PR #22, opened 2019-09-25
rds client: check if db instance not cluster exists

### LINKED ISSUE #21: rds: IsErrorAlreadyExists does not appear to check for all error messages
### What happened?

Somehow my local Crossplane got into a weird state where it is not able to successfully reconcile a mysqlinstance. Here's an example of the status of the object:

```
  status:
    conditions:
    - lastTransitionTime: "2019-09-25T20:05:18Z"
      reason: Managed resource is being created
      status: "False"
      type: Ready
    - lastTransitionTime: "2019-09-25T21:49:28Z"
      message: "DBInstanceAlreadyExists: DB Instance already exists\n\tstatus code:
        400, request id: 910d5710-7374-4e4e-ae3c-37e75a950dbe"
      reason: Encountered an error during managed resource reconciliation
      status: "False"
      type: Synced
```

It looks as though there is an issue with the [IsErrorAlreadyExists function](https://github.com/crossplaneio/stack-aws/blob/1365f37ff9ac847bf75ff28dfa770e30a3006d7c/pkg/clients/aws/rds/rds.go#L112) where it is only checking one of several error codes that could be returned in the case that the resource already exists.

Here is a link to the [specific error message that was being returned](https://github.com/aws/aws-sdk-go-v2/blob/57d74d6ff436076d96a8f1104b53949ff7c8c22e/service/rds/api_errors.go#L127), and the full list of error messages as well. It looks like there is at least one other case.

When looking into this issue, it may make sense to also check for more instances of this class of issue (the class is that only one error code is handled, when several should be handled). On quick inspection, it looks as though the [IsErrorNotFound function](https://github.com/crossplaneio/stack-aws/blob/1365f37ff9ac847bf75ff28dfa770e30a3006d7c/pkg/clients/aws/rds/rds.go#L117) may also have a similar issue. I did not check the logic for other resource types.

### How can we reproduce it?

The gist of reproducing this the way that I ran into it is to create an instance using the [Crossplane Stacks Guide](https://crossplane.io/docs/v0.3/stacks-guide.html) and the AWS option, and then to get the AWS Stack to try to create the instance again. I believe this could be done by running `kubectl edit` on the object and removing the `status.InstanceName`, if it has one.

Crossplane version: helm crossplane/alpha:

```
$ helm list crossplane
NAME            REVISION        UPDATED                         STATUS          CHART                   APP VERSION     NAMESPACE
crossplane      1               Tue Sep 24 16:27:50 2019        DEPLOYED        crossplane-0.3.0        0.3.0           crossplane-system
```

This is running on the Kubernetes from Docker for Mac.

hasheddan:

Updates the RDS `IsAlreadyExistsError` function to check the correct AWS error for DB instance such that "already exists" failures will be ignored on creation.

Fixes #21 

### Checklist

I have:
- [x] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.

[`clusterrole.yaml`]: https://github.com/crossplaneio/crossplane/blob/master/cluster/charts/crossplane/templates/clusterrole.yaml

### DIFF
--- a/pkg/clients/aws/rds/rds.go
+++ b/pkg/clients/aws/rds/rds.go
@@ -107,10 +107,10 @@ func (r *rdsClient) DeleteInstance(name string) (*Instance, error) {
 	return NewInstance(output.DBInstance), nil
 }
 
-// IsErrorAlreadyExists returns true if the supplied error indicates a cluster
-// does already exists.
+// IsErrorAlreadyExists returns true if the supplied error indicates an instance
+// already exists.
 func IsErrorAlreadyExists(err error) bool {
-	return strings.Contains(err.Error(), rds.ErrCodeDBClusterAlreadyExistsFault)
+	return strings.Contains(err.Error(), rds.ErrCodeDBInstanceAlreadyExistsFault)
 }
 
 // IsErrorNotFound helper function to test for ErrCodeDBInstanceNotFoundFault error


negz:
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_CreateDBCluster.html

I wanted to understand what was going wrong here. It sounds like we were checking the error returned when an Aurora DB cluster exists, when in fact we wanted to check the error returned when an RDS instance already exists. Given this is `rds.IsAlreadyExistsError` and not `aws.IsAlreadyExistsError`, I agree that this is the only error we need to test.

PR #29, opened 2019-10-04
Generate crossplane-runtime methodsets

negz:

This PR uses https://github.com/crossplaneio/controller-tools to generate the crossplane-runtime method sets for all Crossplane managed resources and non-portable resource classes, i.e. [`resource.Managed`](https://godoc.org/github.com/crossplaneio/crossplane-runtime/pkg/resource#Managed) and [`resource.NonPortableClass`](https://godoc.org/github.com/crossplaneio/crossplane-runtime/pkg/resource#NonPortableClass). See https://github.com/crossplaneio/crossplane/pull/856 for the equivalent change to Crossplane core.

### Checklist

I have:
- [ ] Updated the RBAC permissions in [`clusterrole.yaml`] to include any new types.

[`clusterrole.yaml`]: https://github.com/crossplaneio/crossplane/blob/master/cluster/charts/crossplane/templates/clusterrole.yaml


apis/cache/v1alpha2/replication_group_types.go
```diff
@@ -316,61 +314,6 @@ type ReplicationGroup struct {
 	Status ReplicationGroupStatus `json:"status,omitempty"`
 }
 
-// SetBindingPhase of this ReplicationGroup.
-func (rg *ReplicationGroup) SetBindingPhase(p runtimev1alpha1.BindingPhase) {
-	rg.Status.SetBindingPhase(p)
-}
-
-// SetConditions of this ReplicationGroup.
-func (rg *ReplicationGroup) SetConditions(c ...runtimev1alpha1.Condition) {
```
soorena776: You have probably addressed this elsewhere, so apologize if I am repeating the same question:

How do we customize any of these methods? I would imagine if all of these methods are similar for different types, one could move them all to crossplane/run-time library and use reflection.

negz: I'm not sure I understand the question. Are you wondering what we'd do if we needed a slightly different implementation of one of these methods? The code generation tool won't replace any method that already exists, so it's safe to implement our own custom variants.

If you're wondering how we update the methods that are generated by the code generation tool, https://github.com/crossplaneio/crossplane-tools/blob/master/internal/method/method.go is the place to look.

PR #42, opened 2019-10-22
Elasticache v1beta1

muvaf:

Refactors AWS ElastiCache to adopt https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md

Part of https://github.com/crossplaneio/stack-aws/issues/30

### Checklist

I have:


apis/cache/v1alpha2/replication_group_types.go
```diff
@@ -63,18 +63,201 @@ var LatestSupportedPatchVersion = map[MinorVersion]PatchVersion{
...
+type ReshardingStatus struct {
+	// Represents the progress of an online resharding operation.
+	SlotMigration SlotMigration `json:"slotMigration"`
+}
+
+// Represents the progress of an online resharding operation.
+// Please also see https://docs.aws.amazon.com/goto/WebAPI/elasticache-2015-02-02/SlotMigration
+type SlotMigration struct {
+	// NOTE(muvaf): Type of ProgressPercentage is float64 in AWS SDK but
+	// float is not supported by controller-runtime.
```
negz: http://giphygifs.s3.amazonaws.com/media/QUaqJRizED5NC/giphy.gif

apis/cache/v1alpha2/zz_generated.managed.go
```diff
@@ -33,6 +33,11 @@ func (mg *ReplicationGroup) GetClaimReference() *corev1.ObjectReference {
 	return mg.Spec.ClaimReference
 }
 
+// GetCondition of this ReplicationGroup.
```
negz: Would be nice to move these code generation updates into a separate commit if it's not too difficult to do so.

pkg/clients/aws.go
```diff
@@ -169,8 +183,52 @@ func Int64Value(v *int64) int {
...
+	}
+	return from
+}
+
+func LateInitializeBool(in *bool, from bool) *bool {
+	if in != nil {
+		return in
+	}
+	return &from
+}
```
negz: https://github.com/crossplaneio/stack-gcp/blob/master/pkg/clients/gcp.go#L247

This looks like it differs logically from the GCP implementation. I believe the intent of all the late init functions are:

1. Return `in` if `in` is not `nil` or if `from` is its zero value.
1. Return `from`.

This implementation would return a pointer to a `false` bool, which I believe will be interpreted as the optional value being set to `false`, not unset.

pkg/clients/elasticache/elasticache.go
```diff
@@ -46,38 +40,6 @@ func NewClient(credentials []byte, region string) (Client, error) {
...
-		Kubernetes resources have a UID for this reason. Crossplane (often) uses
-		this UID to identify the cloud provider resources it manages. UIDs are
-		in practice 36 character V1 UUID strings, but Kubernetes requires we
-		treat them as opaque.
-
-		ElastiCache requires that Replication Groups be identified by a string
-		consisting of no more than 20 characters from the set [-a-z0-9], so we
-		can't use the Kubernetes UID. Instead we hash it, use the 64 bit hash's
-		16 character hex string, and hope we don't get a collision. ¯\_(ツ)_/¯
-	*/
```
negz: RIP Crossplane's largest comment block.

pkg/clients/elasticache/elasticache.go
```diff
@@ -89,209 +51,267 @@ func NewReplicationGroupDescription(g *v1alpha2.ReplicationGroup) string {
...
 // modification input suitable for use with the AWS API.
-func NewModifyReplicationGroupInput(g *v1alpha2.ReplicationGroup) *elasticache.ModifyReplicationGroupInput {
+func NewModifyReplicationGroupInput(g v1alpha2.ReplicationGroupParameters, id string) *elasticache.ModifyReplicationGroupInput {
 	return &elasticache.ModifyReplicationGroupInput{
-		ReplicationGroupId: aws.String(NewReplicationGroupID(g), aws.FieldRequired),
-
-		// TODO(negz): Should this be a configurable part of the replication
-		// group spec? If we did wait until the next maintenance window to apply
-		// changes we'd need some way to account for the pending changes during
-		// our sync logic.
```
negz: I think that removing this TODO makes sense given that deferred apply doesn't really make sense for Crossplane resources, but I wonder if we should document this behaviour somehow.

pkg/controller/cache/managed.go
```diff
@@ -18,9 +18,11 @@ package cache
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
 	"strings"
 
+	elasticache2 "github.com/aws/aws-sdk-go-v2/service/elasticache"
```
negz: I'd prefer a different import name here - something that makes it a little more obvious this is the ElastiCache service and the other `elasticache` is our "client" code. Perhaps name this `elasticacheservice` or just `service`?

pkg/controller/cache/managed.go
```diff
@@ -100,19 +102,19 @@ func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (resource.
 	if c.newClientFn != nil {
 		newClientFn = c.newClientFn
 	}
```
negz: Nitpick - it would be nice to set `c.newClientFn` inside `SetupWithManager` as we did with the GCP updates.

pkg/controller/cache/managed.go
```diff
@@ -121,134 +123,103 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (resource.E
...
+	elasticache.LateInitialize(&cr.Spec.ForProvider, rg)
+
+	o := resource.ExternalObservation{
+		ResourceExists:    true,
+		ConnectionDetails: resource.ConnectionDetails{},
+	}
 
 	// DescribeReplicationGroups can return one or many replication groups. We
 	// ask for one group by name, so we should get either a single element list
 	// or an error.
```
negz: This comment should probably move up to line 126 since it explains why we're writing `rg := rsp.ReplicationGroups[0]`.

pkg/controller/cache/managed.go
```diff
@@ -121,134 +123,103 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (resource.E
...
-	switch g.Status.State {
-	case v1alpha2.StatusAvailable:
-		g.Status.SetConditions(runtimev1alpha1.Available())
-		resource.SetBindable(g)
-	case v1alpha2.StatusCreating:
-		g.Status.SetConditions(runtimev1alpha1.Creating())
-	case v1alpha2.StatusDeleting:
-		g.Status.SetConditions(runtimev1alpha1.Deleting())
+	ccList := make([]elasticache2.CacheCluster, len(cr.Status.AtProvider.MemberClusters))
+	for i, cc := range cr.Status.AtProvider.MemberClusters {
```
negz: Should we update our observation of the replication group before iterating over its member clusters? I don't recall whether it's possible that we might have more or less clusters member clusters than last time we observed the replication group.

pkg/controller/cache/managed.go
```diff
@@ -121,134 +123,103 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (resource.E
...
 	}
+	o.ResourceUpToDate = !elasticache.ReplicationGroupNeedsUpdate(cr.Spec.ForProvider, rg, ccList)
 
-	o := resource.ExternalObservation{
-		ResourceExists:    true,
-		ConnectionDetails: resource.ConnectionDetails{},
+	if cr.Status.AtProvider.ConfigurationEndpoint.Address != "" {
+		o.ConnectionDetails[runtimev1alpha1.ResourceCredentialsSecretEndpointKey] = []byte(cr.Status.AtProvider.ConfigurationEndpoint.Address)
+		b := make([]byte, binary.MaxVarintLen64)
+		binary.LittleEndian.PutUint64(b, uint64(cr.Status.AtProvider.ConfigurationEndpoint.Port))
```
negz: Perhaps we should store ports as a string? i.e:

```go
[]byte(strconv.Itoa(cr.Status.AtProvider.ConfigurationEndpoint.Port))
``` 

I feel like it would be surprising to find an int64 encoded as a byte slice when reading a secret.

negz:
Dropped a few comments - mostly nitpicks. The overall direction look good.

apis/cache/v1alpha2/replication_group_types.go
muvaf: Be careful my friend, knowing too much might get you in trouble..

pkg/clients/aws.go
muvaf: I think ignoring the zero-value may not make sense in all cases. I wonder if we should make them `func(pointer, pointer) pointer` or `func(pointer, pointer) value`. These 2 functions for each basic type would cover the cases where we call late-initialization functions:
```
pointer <-> pointer
pointer <-> value
value <-> pointer
value <-> value
```
The `value` inputs can easily be given as `&value` while the inverse wouldn't make as much sense. If the value really is `false`, we'd take that into account and set the value to false.

BTW, the reason I needed `LateInitializeBool` was the function `automaticFailoverEnabled` that returns `bool`. But taking a second look, it should actually return `*bool` because failover status string could be empty. So, we don't need `LateInitializeBool` anymore.

pkg/clients/elasticache/elasticache.go
muvaf: Having to delete the `¯\_(ツ)_/¯` upset me more.

muvaf:
@negz I was actually still testing this. I'll let you know if a follow-up PR is needed.

PR #45, opened 2019-10-23
Elasticache necessary fixes

muvaf:
The PR https://github.com/crossplaneio/stack-aws/pull/42 is merged but the following changes are needed to make it work.

- fix resourceclass missing forProvider issue.
- rename ApplyImmediately to ApplyModificationsImmediately for better explanation. (not necessary but I thought `ApplyImmediately` may not make much sense for new-comers)
- supply kube client in return value of connector.

### Checklist

I have:


apis/cache/v1beta1/replication_group_types.go
```diff
@@ -244,7 +244,7 @@ type ReplicationGroupParameters struct {
 	//
 	// If false, changes to the nodes in the replication group are applied on the
 	// next maintenance reboot, or the next failure reboot, whichever occurs first.
-	ApplyImmediately bool `json:"applyImmediately"`
+	ApplyModificationsImmediately bool `json:"applyModificationsImmediately"`
```
negz: I guess we have some wiggle room on this naming because it's not really part of the actual ReplicationGroup API object, but rather one of the operations on it. This seems reasonable to me.

https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_ModifyReplicationGroup.html

pkg/controller/cache/claim.go
```diff
@@ -112,11 +112,6 @@ func ConfigureReplicationGroup(_ context.Context, cm resource.Claim, cs resource
 
 	rg.Spec = *spec
 
-	// NOTE(muvaf): ReplicationGroup actually supports memcached as well but the
-	// claim is a _Redis_Cluster, so, we override the value no matter what. Users
-	// should be able to get a memcached through manually creating the ReplicationGroup
-	rg.Spec.ForProvider.Engine = v1beta1.CacheEngineRedis
-
```
negz: What's the context around this removal?

muvaf: `Engine` is a required field and if it's not supplied, the creation of the resource class is rejected immediately anyway. So, for engine type, we rely completely on what resource class gives us and only that, which could be either `redis` or `memcached`.

apis/cache/v1beta1/replication_group_types.go
muvaf: > it's not really part of the actual ReplicationGroup API object, but rather one of the operations on it.

Yeah, actually this goes for some other fields as well. The actual `ReplicationGroup` object hides a lot of properties that you can actually control through operation objects. So, our API object is actually the sum of all of the possible configurations.

muvaf:
@negz I've tested this manually, ready to go!

PR #48, opened 2019-10-24
dbsubnetgroup: move from storage to database group

hasheddan:

Moves `DBSubnetGroup` to `database` group instead of `storage` group. Will follow up with examples update in core Crossplane if merged.

### Checklist

I have:


go.mod
```diff
@@ -3,10 +3,10 @@ module github.com/crossplaneio/stack-aws
 go 1.12
 
 require (
-	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d // indirect
 	github.com/aws/aws-sdk-go-v2 v0.5.0
 	github.com/crossplaneio/crossplane v0.3.1-0.20191023221351-518648b051cd
 	github.com/crossplaneio/crossplane-runtime v0.0.0-20191023215652-0f37bea5496e
+	github.com/crossplaneio/crossplane-tools v0.0.0-20191023215726-61fa1eff2a2e // indirect
```
soorena776: Is this change needed for this fix?

hasheddan: I don't think so. I believe https://github.com/crossplaneio/crossplane/pull/962 is a proposal to fix pinning angryjet, but I just have been doing this for using the latest version. Since it gets removed if someone uses `go mod tidy` I should likely remove it here :)

Issue #30, opened 2019-10-04
AWS RDS and ReplicationGroup v1beta1 Quality


muvaf:
AWS RDS and ReplicationGroup should be refactored to use generic managed reconciler and adopt patterns in https://github.com/crossplaneio/crossplane/pull/840

- [ ] RDS to use generic managed reconciler & API to be v1beta1 quality.
- [ ] ReplicationGroup API to be v1beta1 quality.
- [ ] Update YAMLs in docs that refer to these two resources.

Part of https://github.com/crossplaneio/crossplane/issues/863

PR #49, opened 2019-10-24
RDS v1beta1

muvaf:

Rewrite AWS RDS controller to adopt https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md

Part of #30 

Controller tests and some of the client functions are missing but actual business logic and API is ready to be reviewed.

### Checklist

I have:


muvaf:
@negz @soorena776 all the functionality we need should be in. I'd highly appreciate if you can review. I'm still working on unit tests and manual testing but I don't expect a major change.

apis/database/v1alpha2/rdsinstance_types.go
```diff
@@ -66,52 +66,691 @@ func (v *DBSubnetGroupNameReferencerForRDSInstance) Assign(res resource.CanRefer
...
-	RDSInstanceParameters        `json:",inline"`
+	ForProvider                  RDSInstanceParameters `json:"forProvider,omitempty"`
+
+	// DBSubnetGroupNameRef references to a DBSubnetGroup to retrieve its name
+	// +immutable
+	// +optional
+	DBSubnetGroupNameRef *DBSubnetGroupNameReferencerForRDSInstance `json:"dbsubnetGroupNameRef,omitempty" resource:"attributereferencer"`
+
+	// VPCSecurityGroupIDRefs references to a list of SecurityGroups to retrieve a list of securityGroupIDs
+	// +immutable
```
soorena776: is +immutable feature actually supported by Kubebuilder?
https://book.kubebuilder.io/reference/markers/crd-validation.html

apis/database/v1alpha2/rdsinstance_types.go
```diff
@@ -66,52 +66,691 @@ func (v *DBSubnetGroupNameReferencerForRDSInstance) Assign(res resource.CanRefer
...
+	// A value that specifies that the DB instance class of the DB instance uses
+	// its default processor features.
+	UseDefaultProcessorFeatures *bool `json:"useDefaultProcessorFeatures,omitempty"`
 }
 
 // An RDSInstanceSpec defines the desired state of an RDSInstance.
 type RDSInstanceSpec struct {
 	runtimev1alpha1.ResourceSpec `json:",inline"`
-	RDSInstanceParameters        `json:",inline"`
+	ForProvider                  RDSInstanceParameters `json:"forProvider,omitempty"`
```
soorena776: Nit: I'd keep the `RDSInstanceParameters` inline to keep it consistent with other types in this repository.

apis/database/v1alpha2/rdsinstance_types.go
```diff
@@ -38,19 +38,19 @@ const (
 	PostgresqlEngine = "postgres"
 )
 
-// SecurityGroupIDReferencerForRDSInstance is an attribute referencer that resolves SecurityGroupID from a referenced SecurityGroup
-type SecurityGroupIDReferencerForRDSInstance struct {
+// VPCSecurityGroupIDReferencerForRDSInstance is an attribute referencer that resolves SecurityGroupID from a referenced SecurityGroup
+type VPCSecurityGroupIDReferencerForRDSInstance struct {
```
soorena776: I understand that this renaming is to make it clear that the referenced security group is for VPC, but it breaks the pattern in all the referencers where follow underlyingReferencer+"ForType" naming pattern.

pkg/clients/rds/rds.go
```diff
@@ -124,19 +66,353 @@ func IsErrorNotFound(err error) bool {
...
+func IsUpToDate(in v1alpha2.RDSInstanceParameters, db rds.DBInstance) bool {
+	currentParams := &v1alpha2.RDSInstanceParameters{}
+	LateInitialize(currentParams, db)
+	currentModificationRequest := GenerateModifyDBInstanceInput("", currentParams)
+	candidateModificationRequest := GenerateModifyDBInstanceInput("", &in)
+	return reflect.DeepEqual(currentModificationRequest, candidateModificationRequest)
+}
+
+func GetConnectionDetails(in v1alpha2.RDSInstance) resource.ConnectionDetails {
+	if in.Status.AtProvider.Endpoint.Address == "" {
```
negz: Perhaps we could avoid needing to make this check if we setup a secret propagating reconciler, which would ensure any updates to the managed resource secret were propagated to the resource claim secret. There's an example of this pattern in the EKS controller.

https://godoc.org/github.com/crossplaneio/crossplane-runtime/pkg/resource#NewSecretPropagatingReconciler

pkg/controller/rds/rdsinstance.go
```diff
@@ -67,186 +52,150 @@ type Reconciler struct {
...
-		return reconcile.Result{}, err
+		return resource.ExternalCreation{}, err
 	}
-
-	rdsClient, err := r.connect(instance)
-	if err != nil {
-		return r.fail(instance, err)
+	req := e.client.CreateDBInstanceRequest(rds.GenerateCreateDBInstanceInput(meta.GetExternalName(cr), password, &cr.Spec.ForProvider))
+	req.SetContext(ctx)
+	_, err = req.Send()
```
negz: How does this behave if the RDS instance already exists? Are we intending to fail the reconcile in this case?

Taking a quick skim of the v1beta1 controllers, I think that RDS, ReplicationGroup, and CloudMemorystoreInstance would return an error if trying to create an instance that already exists (which I recall is now our desired behaviour), but it looks like CloudSQL would not. Should I open a PR to update CloudSQL?

pkg/controller/rds/claim.go
```diff
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"strings"
 
+	aws "github.com/crossplaneio/stack-aws/pkg/clients"
+
```
negz:
```go
resource.WithManagedBinder(resource.NewAPIManagedStatusBinder(mgr.GetClient())),
resource.WithManagedFinalizer(resource.NewAPIManagedStatusUnbinder(mgr.GetClient())),
```

It just occurred to me that since we're enabling the status subresource for the RDS controller we'll also need to pass the above arguments to `resource.NewClaimReconcile`. I'm hoping to make this behaviour the default soon given that it should be the default for any new controller we write, per https://github.com/crossplaneio/crossplane-runtime/issues/29.

apis/database/v1alpha2/rdsinstance_types.go
```diff
@@ -129,21 +768,318 @@ const (
...
+	VPCSecurityGroups []VPCSecurityGroupMembership `json:"vpcSecurityGroups,omitempty"`
+}
+
+// An RDSInstanceStatus represents the observed state of an RDSInstance.
+type RDSInstanceStatus struct {
+	runtimev1alpha1.ResourceStatus `json:",inline"`
+	AtProvider                     RDSInstanceObservation `json:"atProvider,omitempty"`
 }
 
 // +kubebuilder:object:root=true
```
negz: I believe you'll need to enable the status subresource comment marker to `RDSInstance` below in order for the managed resource reconciler to work.

pkg/controller/rds/rdsinstance.go
negz: https://github.com/crossplaneio/stack-gcp/pull/62

muvaf: Yes, thanks for the PR. It's a little bit disturbing to have that behavior difference by I think it will be addressed with external policies once and for all.

pkg/controller/rds/rdsinstance.go
```diff
@@ -19,46 +19,31 @@ package rds
...
-
-var (
-	log           = logging.Logger.WithName("controller." + controllerName)
-	ctx           = context.Background()
-	result        = reconcile.Result{}
-	resultRequeue = reconcile.Result{Requeue: true}
+	errNotRDSInstance = "managed resource is not an RDS instance custom resource"
 )
 
 // Reconciler reconciles a Instance object
```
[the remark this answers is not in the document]
muvaf: Questioning how did I not delete that big chunk of code that I worked all they to render useless..

pkg/controller/rds/claim.go
```diff
@@ -275,14 +277,14 @@ func ConfigureMyRDSInstance(_ context.Context, cm resource.Claim, cs resource.Cl
...
 	}
-	spec.Engine = v1alpha2.MysqlEngine
-	v, err := validateEngineVersion(spec.EngineVersion, my.Spec.EngineVersion)
+	spec.ForProvider.Engine = v1alpha2.MysqlEngine
+	v, err := validateEngineVersion(aws.StringValue(spec.ForProvider.EngineVersion), my.Spec.EngineVersion)
 	if err != nil {
 		return err
 	}
-	spec.EngineVersion = v
+	spec.ForProvider.EngineVersion = &v
```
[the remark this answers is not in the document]
muvaf: Yes, makes more sense.

pkg/clients/rds/rds.go
```diff
@@ -124,19 +66,353 @@ func IsErrorNotFound(err error) bool {
...
+	return c
+}
+
+// GenerateModifyDBInstanceInput from RDSInstanceSpec
+func GenerateModifyDBInstanceInput(name string, p *v1alpha2.RDSInstanceParameters) *rds.ModifyDBInstanceInput {
+	// NOTE(muvaf): MasterUserPassword is not used here. So, password is set once
+	// and kept that way.
+	// NOTE(muvaf): Change of DBInstanceIdentifier is supported by AWS but
+	// Crossplane assumes identification info never changes, so, we don't support
+	// it.
```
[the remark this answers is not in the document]
muvaf: You send a `NewDBInstanceIdentifier` in the modify request and after that point on you have to call it with the new identifier. I think that is very imperative, so, a bit hard to support and also really a rare case.

pkg/clients/aws.go
```diff
@@ -120,6 +120,15 @@ func LateInitializeStringPtr(in *string, from *string) *string {
 	return from
 }
 
+// LateInitializeStringPtr returns in if it's non-nil, otherwise returns from
```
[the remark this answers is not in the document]
muvaf: Actually function itself had a bug :( Thanks for pointing!

apis/compute/v1alpha2/types.go
```diff
@@ -177,7 +177,7 @@ type EKSClusterParameters struct {
 	// SecurityGroupIDs of this EKS cluster.
 	SecurityGroupIDs []string `json:"securityGroupIds,omitempty"`
 
-	// SecurityGroupIDRefs is a set of referencers that each retrieve the ID from the referenced SecurityGroup
+	// VPCSecurityGroupIDRefs is a set of referencers that each retrieve the ID from the referenced SecurityGroup
```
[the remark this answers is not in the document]
muvaf: I didn't mean to touch that file actually, auto-refactor must have converted it. Just reverted the line.

apis/database/v1alpha2/rdsinstance_types.go
muvaf: Oh, we're actually going with `forProvider` in spec and `atProvider` in status sub-structs as outlined [here](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#owner-based-struct-tree).

apis/database/v1alpha2/rdsinstance_types.go
muvaf: Hopefully in the future, but we sit tight. https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#immutable-properties

apis/database/v1alpha2/rdsinstance_types.go
muvaf: So, there are two fields in RDS's spec now. One is `VPCSecurityGroupIDs` and the other one is `DBSecurityGroups` and they point to [different type of security groups.](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.RDSSecurityGroups.html)

So, it's actually maps to distinct functionality rather than only naming. Though, I wasn't aware of `underlyingReferencer+"ForType"` pattern. How do you think we should proceed here?

muvaf:
A few cases I ran into during the manual test:
* AWS rejects modification requests if you send a field whose value is same as the current one. So, I've had to generate a patch object using json patch.
* Modifications take time during that we should not retry. `Update` call returns nil if state is not `available`
* Deletion returns error if the object is already in `deleting` state, now that call returns nil if the state is already `deleting`
* `ApplyModificationsImmediately *bool` is false by default but in that case we retry until the next maintenance window since the current state and the desired state don't match. If its value is `nil`, meaning user didn't specify anything, we make it `true`.

Unit tests are broken right now, but they should be easy to fix. I still need a few things to manually test.

negz:
> Modifications take time and during that we should not retry. Update call returns nil if state is not available

This sounds a little like the first case covered in https://github.com/crossplaneio/stack-aws/issues/46. Does it make sense for both resources to behave the same in this case?

pkg/clients/rds/rds.go
```diff
@@ -107,8 +109,6 @@ func GenerateCreateDBInstanceInput(name, password string, p *v1alpha2.RDSInstanc
 		PromotionTier:                      awsclients.Int64Address(p.PromotionTier),
 		PubliclyAccessible:                 p.PubliclyAccessible,
 		StorageEncrypted:                   p.StorageEncrypted,
-		TdeCredentialArn:                   p.TDECredentialArn,
-		TdeCredentialPassword:              p.TDECredentialPassword,
```
negz: Any context on removing these guys? I'm guessing the latter is a credential?

pkg/clients/rds/rds.go
```diff
@@ -135,70 +135,93 @@ func GenerateCreateDBInstanceInput(name, password string, p *v1alpha2.RDSInstanc
...
+			return nil, err
+		}
+		desiredJSON, err := json.Marshal(p)
+		if err != nil {
+			return nil, err
+		}
+		patchJSON, err := jsonpatch.CreateMergePatch(currentJSON, desiredJSON)
+		if err != nil {
+			return nil, err
+		}
```
negz: https://github.com/imdario/mergo

You might also be able to use mergo to do this. It uses a whole bunch of reflection under the hood so it's likely not much better than JSON patching, but would at least avoid all the JSON serialisation.

pkg/clients/rds/rds.go
```diff
@@ -397,17 +423,80 @@ func LateInitialize(in *v1alpha2.RDSInstanceParameters, db rds.DBInstance) {
...
+	// But we detect as if we need to update it all the time. We assign
+	// the actual full version to our spec to avoid unnecessary update signals.
+	if strings.HasPrefix(aws.StringValue(db.EngineVersion), aws.StringValue(in.EngineVersion)) {
+		in.EngineVersion = db.EngineVersion
+	}
 }
 
-func IsUpToDate(in v1alpha2.RDSInstanceParameters, db rds.DBInstance) bool {
-	currentParams := &v1alpha2.RDSInstanceParameters{}
+//func subtract(from, minus *string) *string {
```
negz: Will need to get rid of all this commented out code.

pkg/clients/rds/rds.go
negz: Got it. I thought your intention was to not publish a secret _at all_ until the endpoint is set, but I now see this only contains the endpoint and port.

pkg/controller/database/claim.go
```diff
@@ -96,27 +96,29 @@ type PostgreSQLInstanceClaimController struct{}
...
+		v1beta1.RDSInstanceKind,
+		v1beta1.Group))
 
 	r := resource.NewClaimReconciler(mgr,
 		resource.ClaimKind(databasev1alpha1.PostgreSQLInstanceGroupVersionKind),
-		resource.ClassKind(v1alpha2.RDSInstanceClassGroupVersionKind),
-		resource.ManagedKind(v1alpha2.RDSInstanceGroupVersionKind),
+		resource.ClassKind(v1beta1.RDSInstanceClassGroupVersionKind),
+		resource.ManagedKind(v1beta1.RDSInstanceGroupVersionKind),
+		resource.WithManagedBinder(resource.NewAPIManagedStatusBinder(mgr.GetClient())),
```
negz: This signature changed after my static provisioning work; it's now `(mgr.GetClient(), mgr.GetScheme())`

apis/aws.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	cachev1alpha2 "github.com/crossplaneio/stack-aws/apis/cache/v1beta1"
 	computev1alpha2 "github.com/crossplaneio/stack-aws/apis/compute/v1alpha2"
-	databasev1alpha2 "github.com/crossplaneio/stack-aws/apis/database/v1alpha2"
+	databasev1alpha2 "github.com/crossplaneio/stack-aws/apis/database/v1beta1"
```
hasheddan: I am realizing my `DbSubnetGroup` PR is now complicating this work here :/ @negz @muvaf do you think we should have a `v1beta` and `v1alpha2` (which I will be updating shortly to `v1alpha3`) directory under `database` or also bring `DbSubnetGroup` along to `v1beta1`?

negz: > do you think we should have a v1beta and v1alpha2 (which I will be updating shortly to v1alpha3) directory under database or also bring DbSubnetGroup along to v1beta1?

Let's keep DbSubnetGroup at v1alpha2 - I think it's close but needs a few small touches to reach what we consider v1beta1 readiness.

hasheddan: SGTM

pkg/clients/rds/rds.go
muvaf: I've tried but it doesn't seem to have a mechanism for creating an intermediary _patch_ object to apply to other resources. Now, I looked up what controller-runtime uses for their new `Patch` call and it's the same as we do https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/patch.go#L61

I'll probably want to dig deeper into how to make this glue code more manageable in 0.5, either generation or reflection or json methods. Humans shouldn't be doing this work.

pkg/clients/rds/rds.go
muvaf: >  I'm guessing the latter is a credential?

Right. I've left the following comment in `rdsinstance_types.go`:
```
// TODO(muvaf): get this password as input when we have a way of supplying
// sensitive information as input from the user.

// TDECredentialArn is the ARN from the key store with which to associate the instance for TDE encryption.
// +optional
// TDECredentialArn *string `json:"tdeCredentialArn,omitempty"`
// TDECredentialPassword is the password for the given ARN from the key store in order to access the
// device.
// +optional
// TDECredentialPassword *string `json:"tdeCredentialPassword,omitempty"`
```

As far as I understand, `TDECredentialArn` has no meaning if you don't supply the password. So, I've removed both.

pkg/controller/database/claim.go
muvaf: Thanks for the heads-up! I'll update the crossplane-runtime in a final commit.

muvaf:
> I am realizing my DbSubnetGroup PR is now complicating this work here :/

@hasheddan yeah, took 2 hours to fix the conflicts, separate the packages and try to make angryjet happy 😅  but it should be fine now.

@negz rebased to master, tested a few other cases and put simple fixes and TODOs. I had to squash some commits, so there could be some commits where code doesn't compile. Hope I didn't make reviewing harder!

pkg/controller/database/rdsinstance.go
```diff
@@ -0,0 +1,243 @@
...
+		return resource.ExternalObservation{}, errors.Wrap(resource.Ignore(rds.IsErrorNotFound, err), errDescribeFailed)
+	}
+
+	// Describe requests can be used with filters, which then returns a list.
+	// But we use an explicit identifier, so, if there is no error, there should
+	// be only 1 element in the list.
+	instance := rsp.DBInstances[0]
+	current := cr.Spec.ForProvider.DeepCopy()
+	rds.LateInitialize(&cr.Spec.ForProvider, &instance)
+	if !reflect.DeepEqual(current, &cr.Spec.ForProvider) {
```
negz: No changes needed, but I suspect we can drop this check in future. I'm pretty sure `Update` is a no-op if nothing about the resource has changed.

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -0,0 +1,1160 @@
...
+
+	// DBSubnetGroupNameRef references to a DBSubnetGroup to retrieve its name
+	// +immutable
+	// +optional
+	DBSubnetGroupNameRef *DBSubnetGroupNameReferencerForRDSInstance `json:"dbsubnetGroupNameRef,omitempty" resource:"attributereferencer"`
+
+	// VPCSecurityGroupIDRefs references to a list of SecurityGroups to retrieve a list of securityGroupIDs
+	// +immutable
+	// +optional
+	VPCSecurityGroupIDRefs []*VPCSecurityGroupIDReferencerForRDSInstance `json:"vpcSecurityGroupIDRefs,omitempty" resource:"attributereferencer"`
```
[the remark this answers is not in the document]
muvaf: I've thought about this during[ the one-pager](https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#owner-based-struct-tree) and IMO `forProvider` should include only directly mirrored fields from SDK objects and act as the last representation of what's going to be sent.

But I'm happy to discuss changing that since at that time references were not implemented yet.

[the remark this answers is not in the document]
negz: I feel quite strongly that references belong under `forProvider`.

I can see the user benefit in `forProvider` being literally exactly what is sent to the provider, and moreso see the benefit that provides us when writing translation code. However, from a user's perspective reference fields _are_ for the provider - they're just a different form of specifying the input for the provider. I feel that keeping them immediately next to the underlying fields that they set makes them almost self-documenting. If you see `fooRef` right next to `foo` it's more obvious that the two fields are related than if they're far away from each other in the `spec`. Similarly, it maintains high fidelity for references.

Take the following example:

```yaml
apiVersion: example.org/v1
kind: Example
metadata:
  name: example
spec:
   # Does this field set .spec.forProvider.foo` or  .spec.forProvider.settings.foo?
   # Even outside of this pathological case in which two forProvider fields at
   # different levels have the same name, imagine adding a barRef here for a
   # bar field that was nested two levels deep inside forProvider in a complex
   # managed resource with tens of fields. It would be difficult to correlate them
   # at a glance.
   fooRef:
     name: example
  forProvider:
    foo: example
    settings:
      foo: example
```

Compared to:

```yaml
apiVersion: example.org/v1
kind: Example
metadata:
  name: example
spec:
  forProvider:
    foo: example
    fooRef:
      name: example
    settings:
      foo: example
      fooRef:
        name: example
```

I feel the latter example is much easier to understand.

I think we can avoid like issues with conflicting fields if we establish some rules, for example that such fields are always immediately next to their equivalent field and always named `fooRef` or `fooRefs` for arrays. We should probably also consider a comment marker for documentation purposes like `//+crossplane:reference:populates:foo`.

PR #52, opened 2019-10-26
Elasticache deletion and update calls better UX

### LINKED ISSUE #46: Benign ReplicationGroup reconcilation errors
### What happened?

I noticed the following benign reconcile errors while trying out the new v1beta ReplicationGroup resource.

While creating:
```
    Last Transition Time:  2019-10-24T00:34:21Z
    Message:               cannot modify ElastiCache replication group: InvalidReplicationGroupState: Replication group must be in available state to modify.
                           status code: 400, request id: acf70fbf-1b09-4b26-8916-8b27f4817251
    Reason:                Encountered an error during managed resource reconciliation
    Status:                False
    Type:                  Synced
```

While deleting:
```
    Last Transition Time:  2019-10-24T00:48:22Z
    Message:               cannot delete ElastiCache replication group: InvalidReplicationGroupState: Replication group default-app-redis-dr2rp has status deleting which is not valid for deletion.
                           status code: 400, request id: b4d9f4d2-f278-4f5b-a837-514f23ad1cf5
    Reason:                Encountered an error during managed resource reconciliation
    Status:                False
    Type:                  Synced
```

These errors seem benign - the ReplicationGroup is created and then deleted just fine. We could probably avoid them by having the `ExternalClient` `Update` and `Delete` methods be no-op if the ReplicationGroup under reconciliation is observed to be creating or deleting.

Relates to https://github.com/crossplaneio/stack-aws/issues/30.

### How can we reproduce it?

Create and then delete a ReplicationGroup using the following claim and class:

```yaml
---
apiVersion: cache.aws.crossplane.io/v1beta1
kind: ReplicationGroupClass
metadata:
  name: aws-redis-standard
  labels:
    example: "true"
specTemplate:
  writeConnectionSecretsToNamespace: crossplane-system
  providerRef:
    name: example
  reclaimPolicy: Delete
  forProvider:
    replicationGroupDescription: "An example replication group"
    applyModificationsImmediately: true
    engine: "redis"
    engineVersion: "3.2.4"
    cacheParameterGroupName: default.redis3.2.cluster.on
    cacheNodeType: cache.t2.micro
    automaticFailoverEnabled: true
    numNodeGroups: 2
    replicasPerNodeGroup: 2
---
apiVersion: cache.crossplane.io/v1alpha1
kind: RedisCluster
metadata:
  name: app-redis
spec:
  classSelector:
    matchLabels:
      example: "true"
  writeConnectionSecretToRef:
    name: redissecret
  engineVersion: "3.2"
```

Crossplane version:

muvaf:

Fixes #46 

### Checklist

I have:

### DIFF
--- a/pkg/controller/cache/managed.go
+++ b/pkg/controller/cache/managed.go
@@ -199,6 +199,11 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (resource.Ex
 	if !ok {
 		return resource.ExternalUpdate{}, errors.New(errNotReplicationGroup)
 	}
+	// NOTE(muvaf): AWS API rejects modification requests if the state is not
+	// `available`
+	if cr.Status.AtProvider.Status != v1beta1.StatusAvailable {
+		return resource.ExternalUpdate{}, nil
+	}
 	mr := e.client.ModifyReplicationGroupRequest(elasticache.NewModifyReplicationGroupInput(cr.Spec.ForProvider, meta.GetExternalName(cr)))
 	mr.SetContext(ctx)
 	_, err := mr.Send()
@@ -211,6 +216,9 @@ func (e *external) Delete(ctx context.Context, mg resource.Managed) error {
 		return errors.New(errNotReplicationGroup)
 	}
 	mg.SetConditions(runtimev1alpha1.Deleting())
+	if cr.Status.AtProvider.Status == v1beta1.StatusDeleting {
+		return nil
+	}
 	req := e.client.DeleteReplicationGroupRequest(elasticache.NewDeleteReplicationGroupInput(meta.GetExternalName(cr)))
 	req.SetContext(ctx)
 	_, err := req.Send()
--- a/pkg/controller/cache/managed_test.go
+++ b/pkg/controller/cache/managed_test.go
@@ -492,6 +492,13 @@ func TestObserve(t *testing.T) {
 func TestUpdate(t *testing.T) {
 	cases := []testCase{
 		{
+			name:       "NotInAvailableState",
+			e:          &external{},
+			r:          replicationGroup(withProviderStatus(v1beta1.StatusCreating)),
+			want:       replicationGroup(withProviderStatus(v1beta1.StatusCreating)),
+			returnsErr: false,
+		},
+		{
 			name: "FailedModifyReplicationGroup",
 			e: &external{client: &fake.MockClient{
 				MockDescribeReplicationGroupsRequest: func(_ *elasticache.DescribeReplicationGroupsInput) elasticache.DescribeReplicationGroupsRequest {
@@ -534,11 +541,13 @@ func TestUpdate(t *testing.T) {
 			}},
 			r: replicationGroup(
 				withReplicationGroupID(name),
+				withProviderStatus(v1beta1.StatusAvailable),
 				withConditions(runtimev1alpha1.Available()),
 				withMemberClusters([]string{cacheClusterID}),
 			),
 			want: replicationGroup(
 				withReplicationGroupID(name),
+				withProviderStatus(v1beta1.StatusAvailable),
 				withConditions(runtimev1alpha1.Available()),
 				withMemberClusters([]string{cacheClusterID}),
 			),
@@ -595,10 +604,17 @@ func TestDelete(t *testing.T) {
 					}
 				},
 			}},
-			r: replicationGroup(),
+			r:          replicationGroup(),
+			want:       replicationGroup(withConditions(runtimev1alpha1.Deleting())),
+			returnsErr: false,
+		},
+		{
+			name: "AlreadyDeletingState",
+			e:    &external{},
+			r:    replicationGroup(withProviderStatus(v1beta1.StatusDeleting)),
 			want: replicationGroup(
-				withConditions(runtimev1alpha1.Deleting()),
-			),
+				withProviderStatus(v1beta1.StatusDeleting),
+				withConditions(runtimev1alpha1.Deleting())),
 			returnsErr: false,
 		},
 		{
--- a/pkg/controller/database/rdsinstance.go
+++ b/pkg/controller/database/rdsinstance.go
@@ -165,6 +165,7 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (resource.Ex
 	if !ok {
 		return resource.ExternalCreation{}, errors.New(errNotRDSInstance)
 	}
+	cr.SetConditions(runtimev1alpha1.Creating())
 	if cr.Status.AtProvider.DBInstanceStatus == v1beta1.RDSInstanceStateCreating {
 		return resource.ExternalCreation{}, nil
 	}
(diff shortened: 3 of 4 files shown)


pkg/controller/cache/managed.go
```diff
@@ -164,8 +164,6 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (resource.Ex
 	if !ok {
 		return resource.ExternalCreation{}, errors.New(errNotReplicationGroup)
 	}
-
-	cr.Status.SetConditions(runtimev1alpha1.Creating())
```
negz: Was this deletion intended?

pkg/controller/cache/managed.go
```diff
@@ -199,6 +197,11 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (resource.Ex
 	if !ok {
 		return resource.ExternalUpdate{}, errors.New(errNotReplicationGroup)
 	}
+	// NOTE(muvaf): AWS API rejects modification requests if the state is not
+	// `available`
+	if cr.Status.AtProvider.Status != v1beta1.StatusAvailable {
+		return resource.ExternalUpdate{}, nil
+	}
```
negz: We could also handle this case by setting `UpToDate = true` in `Observe` and never calling update at all, but I like this pattern better because we're not actually up to date, but rather can't update yet.

pkg/controller/cache/managed.go
muvaf: Yes. In other resources, this condition is set in `Observe` right after the object is retrieved, which I think makes sense to decide on the state with the observation we get rather than the action we're about to take. As an example, we can set condition to `Deleting` but the deletion call fails, however we'have marked the object as deleting even though it's perfectly available. This doesn't exactly map to `Creating` since there is no object yet but I like the idea of having the API's response as the only source of truth for the status.

pkg/controller/cache/managed.go
muvaf: > but I like this pattern better because we're not actually up to date, but rather can't update yet.

Me too. Initially wanted the condition to be `status == modifying` but AWS explicitly says it has to be `available`. But I'm wondering if we could at least fire an event or note it down to somewhere that _we're not able to update the resource yet, here are some cat videos you can watch while waiting_

pkg/controller/cache/managed.go
negz: > In other resources, this condition is set in Observe right after the object is retrieved

I did not notice that, but I see now that it's true for all the CloudMemorystore controller, which will also need updating if we want to go this direction for v1beta1. In most (all?) pre-v1beta1 controllers the conditions are set in observe _and_ in create and delete.

That said, I'm not quite sold that we should go this direction. Frequently we describe the `Ready` condition as representing "the lifecycle of the managed resource", but I've always taken that to include capturing our controller's intent, rather than being based purely an observation of the external resource. Put otherwise, we _intentionally_ set the status to `Deleting` even if the delete call fails in order to communicate to the user that we received and are acting on their request to delete the managed resource.

Obviously this is a little confused by the fact that we updated statuses based on _actions_ in some places (create, delete), but on _observations_ in orders (observe), so in this example we'd set `Deleting`, then set `Available` during `Observe`, then set `Deleting` again when we tried to delete the resource again (due to the deletion timestamp being set). 

I do think there's some value in the current pattern. Given that we requeue every 30 seconds when waiting for an operation like create or delete this means when you create a `CloudSQLInstance` you won't see that it's creating until the first reconcile half a minute later.

pkg/controller/cache/managed.go
negz: > at least fire an event or note

We could potentially use a bespoke condition for this, either a variant of `Ready` or an entirely new condition type.

pkg/controller/cache/managed.go
muvaf: > Frequently we describe the Ready condition as representing "the lifecycle of the managed resource", but I've always taken that to include capturing our controller's intent, rather than being based purely an observation of the external resource.

I always think that status is simply the observation of the resource and this wouldn't include controller's intent. But I do see value in emitting the controller's intent.

> I do think there's some value in the current pattern. Given that we requeue every 30 seconds when waiting for an operation like create or delete this change means when you create a CloudSQLInstance you won't see that it's creating until the first reconcile half a minute later.

Good point.

I'll change all resources to set the condition in `Create/Delete`. Maybe we need a condition for `Updating`?

muvaf: @negz How about having managed reconciler do this condition setting before calling the corresponding functions in 0.5?

hasheddan: Definitely a fan of adding this to the managed reconciler for consistency across controllers! 👍 I wonder if it would be useful to have some sort of controller status and provider status that are separate (i.e. "the state of the controller is Deleting but the state of the resource at the provider is Available" in the example above).

muvaf: @hasheddan well, actually `atProvider` has a state field that we calculate the availability from. It's just that the field name is not consistent across resources. In some resources it's `status.atProvider.dbInstanceStatus`, in some of them `status.atProvider.status` etc.

But I'm not sure if the drift between _state intent of the controller for the given resource_ and _state of the resource_ is worth a new condition. Because one converges to the other in max 1 minute in the happy path. For current cases;
* `Create` doesn't really make sense since there is no object if the intent isn't successfully executed,
* `Delete` makes sense to some extent but usually if `Delete` is requested not sure availability info is relevant anymore.

Apart from these, `Update` call is a bit different and that's where I see some value for separation. Because the stack controllers might want to take some actions in case of resource unavailability, like readiness reporting of the application could take managed resources' state into account and intermittent unavailability due to `Update` could be valuable to report. If the `Update` call isn't successful for some reason, stack controller should have the possibility to know the current exact state, which is resource is still safe to use.

Though currently we don't set a condition for `Update` calls, so, it should be fine. But I think it's worth noting that we should try to lean towards keeping the status sub-resource of the CR as close to actual status as possible for the sake of being a reliable lowest level of the Crossplane stacks.

negz: > Maybe we need a condition for Updating?

Yeah I've thought about this before and been on the fence. I agree it would especially make sense in cases where a resource was unavailable or somehow degraded while updating, but I'm less sure it makes sense if resources handle updates quickly and without degradation. Definitely worth opening an issue to think through.

> How about having managed reconciler do this condition setting before calling the corresponding functions in 0.5?

So the historical reason we ask folks to emit these conditions in the `ExternalClient` rather than the managed reconciler is because we didn't want to constrain managed resources to condition reasons like `Creating`, `Deleting`, and `Available`. e.g. Maybe a managed resource wants to name its "Creating" condition `Activating`, and its `Available` reason `Online` or similar. I lean slightly toward maintaining this flexibility, though I can see that we've taken advantage of it in zero cases so far.

> I wonder if it would be useful to have some sort of controller status and provider status that are separate

I tend to agree with @muvaf on this one. With our new `v1beta` resources in almost all cases you can see the exact status of the provider resource in its own language under `.status.atProvider`. In general I think we should be very conservative about adding new condition types.

PR #65, opened 2019-10-31
rds: remove unused  masterpassword field from spec

muvaf:

Remove unused  masterpassword field from spec. It's not used anywhere but somehow exists.

### Checklist

I have:

### DIFF
--- a/apis/database/v1beta1/rdsinstance_types.go
+++ b/apis/database/v1beta1/rdsinstance_types.go
@@ -471,24 +471,6 @@ type RDSInstanceParameters struct {
 	// +optional
 	LicenseModel *string `json:"licenseModel,omitempty"`
 
-	// MasterUserPassword is the password for the master user. The password can include any printable
-	// ASCII character except "/", """, or "@".
-	// Amazon Aurora
-	// Not applicable. The password for the master user is managed by the DB cluster.
-	// For more information, see CreateDBCluster.
-	// MariaDB
-	// Constraints: Must contain from 8 to 41 characters.
-	// Microsoft SQL Server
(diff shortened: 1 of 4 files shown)


negz:
> it might be worth considering if we should allow the ability to create a k8s secret and reference it for the password

@muvaf and I discussed a similar pattern for CloudSQL in https://github.com/crossplaneio/stack-gcp/pull/46. @muvaf did you end up raising an issue to track that we might want this functionality? I agree this is probably something like what we want to do, but there is some interesting interplay between _reading_ credentials from a secret and writing them back to another secret (the connection secret) that we might need to work through.

negz:
I'm thinking perhaps we should hold this for 0.3.0, given it's technically an API change. I could be convinced to include it in 0.2.1 considering it's non-breaking AFAIK (the field will be ignored if submitted now, which will be the same behaviour after).

muvaf:
@negz this issue was raised for that specific case https://github.com/crossplaneio/stack-gcp/issues/54

I think getting a secret as input is a feature that should be implemented per managed resource since the transfer of the values from secret to the SDK api call would happen in the `ExternalClient` implementation. I'm planning to submit some updates to https://github.com/crossplaneio/crossplane/blob/master/design/one-pager-managed-resource-api-design.md and this is one of the things I'll mention.

> I'm thinking perhaps we should hold this for 0.3.0, given it's technically an API change

I am not sure whether it really is an API change from the user's perspective.
Before: User supplies a value and it doesn't work
After: User supplies a value and it disappears, effect on the resource is same.

So, neither their pipeline nor the state of the created object will be affected.

PR #66, opened 2019-11-02
Automatically ignore attribute referencers when comparing

negz:

This method prevents us having to explicitly filter out or unset attribute referencers before comparing objects to determine whether our external resource is up to date. Note the references will exist in the generated patch we use to update our external resource, but should be ignored.

### Checklist

I have:


negz:
This is similar to https://github.com/crossplaneio/stack-gcp/issues/77, except that we were already ignoring the references for RDS manually so there's no current bug in this controller.

PR #74, opened 2019-11-06
Pin code generation tools using Go modules

negz:

This invokes code generation tools that are pinned using go modules. This will allow us to pin angryjet (rather than running it from `master`), and to use `go.mod` `replace` statements to easily work with different tooling builds.

One somewhat significant change here is that `make manifests` becomes part of `make generate`. Previously they both used the same tool (`controller-gen`), but invoked the tool twice, and potentially used different versions. I think this makes sense as long as you mentally model `make generate` as "run generation tools" rather than "generate Go code" specifically.

### Checklist

I have:


negz:
@hasheddan I've made this change for AWS, GCP, and Azure so far. In each case the integration tests seem to break but I can't immediately see why they would. Any ideas?

https://github.com/crossplaneio/stack-gcp/pull/88
https://github.com/crossplaneio/stack-azure/pull/62

hasheddan:
Going back on my previous comment, the issue I was seeing above was because the integration tests were rerun without running the preceding steps. The actual failure is on timeout of the stack-aws installation. Checking why this happens now.

```
[2019-11-06T23:36:06.426Z] >>>>>>> installing stack-aws into "aws" namespace

[2019-11-06T23:36:06.426Z] namespace/aws created

[2019-11-06T23:36:06.426Z] clusterstackinstall.stacks.crossplane.io/stack-aws created

[2019-11-06T23:38:07.278Z] waiting for deployment create in aws...........................TIMEOUT
```

Issue #110, opened 2019-01-18
Utilize IAM Roles for AWS Authentication when running inside of AWS


mfacenet:
**Is this a bug report or feature request?**
* Feature Request

**What should the feature do:**
Utilize IAM Roles to authenticate to the AWS API when available.

**What is use case behind this feature:**
As a engineer, I don't wish to concern myself with the details of securing/rotating security credentials, as such, I want to utilize IAM Roles assigned to the nodes that crossplane is executing on to authenticate to the AWS services.  Optionally I want to utilize kube2iam or kiam to manage which IAM role crossplane has access to.  

**Environment**:
Crossplane running inside of a Kubernetes cluster on AWS, with or without kube2iam/kiam installed

prasek:
@hasheddan is this fixed now with https://github.com/crossplane/stack-aws/pull/126?

hasheddan:
@prasek yep! Thanks for flagging 👍

martin-ducar-gd:
A question here. I see the Service account role support has been added that means you can annotate service account and that will in the end bring ENV variables like AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE to provider-aws pod which it will use to authenticate to AWS API. Which is good but it becomes really hard to do cross account assume. EKS webidentity is somehow fixed to that account where the EKS cluster lives. and assuming role in another account is done as 2 step process. What was I think instead meant here is with tools like kube2iam kiam its much more simplier to cross account assume without POD even knowing it all you do is add annotation to pod: `iam.amazonaws.com/role: "arn:aws:iam::<account_id>:role/<role_name>"` and if your kiam or kube2iam is able to assume that role it will give it to provider-aws via redirecting metadata call to 169.254.169.254... all it needs to do is use aws-sdk with fallback to ec2rolecreds. What I think is missing really is a way  to annotate a provider-aws pod with proper annotation as that is done via `ControllerConfig` and `Provider` abstractions. Is there any way to define that now or really its needed to modify the crossplane logic?

hasheddan:
>  What I think is missing really is a way to annotate a provider-aws pod with proper annotation as that is done via ControllerConfig and Provider abstractions.

[...]

PR #126, opened 2020-02-06
Allow auth using IAM Roles for Service Accounts on EKS

### LINKED ISSUE #109: ServiceAccount integration for AWS

Currently we are unable to create RDS instances using the crossplane. Due to security restrictions we are only allowed or restricted to use ServiceAccounts. 

Ability to support ServiceAccount by using annotations or some other way will avoid using AWS credentials manually.

hasheddan:

This is initial implementation of allowing an AWS `Provider` to authenticate using [IAM roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) when running `crossplane` and `stack-aws` on EKS. A few things to note on this early WIP:

- Auth is only enabled for `RDSInstance` here, but other resources will look essentially the same
- The AWS SDK has been update from v0.5.0 to v0.19.0, which requires `context` to be explicitly passed to `.Send()` methods. I have updated across all controllers and clients, but am just passing `context.TODO()` into those controllers that are not yet using the managed reconciler. I think this should be addressed in a separate PR.
- Because `aws-sdk-go-v2` (which is what we use in `stack-aws`) does not yet have the functionality implemented in v1 [here](https://github.com/aws/aws-sdk-go/pull/2667), I am doing STS auth manually. However, there is [indication](https://github.com/aws/aws-sdk-go-v2/issues/475#issuecomment-583001626) that v2 will have the new features soon. I normally would say we should wait until it is contributed upstream before adding this feature, but given that we have an explicit request here and we can easily update to consume v2 implementation without changing our API, I think it would be reasonable to go ahead and move forward.
- This has been tested successfully on EKS, but does require some manual configuration. We can likely start with just documenting how to set it up (since early users will likely already be familiar), then actually provide it as a deployment option in the future.
- The current implementation will need some clean-up but I would like to open conversation before fixing any stylistic changes.

Fixes #109 & #110 

/cc @phanimullapudi @mfacenet 

### Checklist

I have:


muvaf:
One thing I don't like about this concept is that authentication is done _per pod_, which essentially means you can only use one SA-enabled provider for the whole Crossplane cluster. @hasheddan Is my understanding correct?

Is there a way to overcome this? I image that if we had a controller for `Provider` type whose purpose is:
* whenever an SA-enabled `Provider` is created, create a new `Pod` with that annotated service account attached that will fetch the actual credential data via `AssumeRoleWithWebIdentityRequest`.
* Then we could:
  * either possibly store that into a secret that is referred by the `Provider`, hence stack-aws won't be even changed. But that might be against why user needs to do it in the first place; they don't want to manage the credentials.
  * or in each reconcile stack-aws can send a HTTP query to that `Pod` for credentials and use the retrieved data for requests in that reconcile. So, the `Pod` that was created for SA-enabled provider would essentially act like a credential provider.

Another possibility is that when we see an SA-enabled provider, we could somehow create another stack-aws `Deployment` which has the necessary service account and it'd reconcile only the resources with `spec.providerRef == SA-enabled provider that this stack-aws pod is assigned to` . But this'd complicate the stack stuff.

hasheddan:
@muvaf thanks for your thoughts here! Yes, you do understand correctly. While this is a different pattern, it is less about trying to enable using service accounts in general and more about enabling using service accounts and IAM roles the way that EKS does internally. I agree that most Crossplane users will want to use multiple providers, which you could still do here by managing a "core" provider that uses this service account model, and then letting other providers that reference secrets exist as well (potentially with finer scoped permissions).

In general, I think your points are valuable for a conversation about how auth works across Crossplane. It might be worth exploring if the current secret model is sustainable long-term. However, I do not think enabling the functionality in this PR hurts our authentication story, and it does enable users who prefer this pattern to utilize it.

pkg/clients/aws.go
```diff
@@ -92,6 +98,45 @@ func LoadConfig(data []byte, profile, region string) (*aws.Config, error) {
 	return &config, err
 }
 
+// LoadSAConfig assumes an IAM role configured via a ServiceAccount.
+// https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
+func LoadSAConfig(ctx context.Context, region string) (*aws.Config, error) {
+	cfg, err := external.LoadDefaultAWSConfig()
+	if err != nil {
+		return nil, err
+	}
```
negz: We should wrap these errors now that there's a small handful of branches that return them.

pkg/clients/aws.go
```diff
@@ -92,6 +98,45 @@ func LoadConfig(data []byte, profile, region string) (*aws.Config, error) {
...
+	cfg.Region = region
+	svc := sts.New(cfg)
+
+	b, err := ioutil.ReadFile(os.Getenv("AWS_WEB_IDENTITY_TOKEN_FILE"))
```
negz: It might be worth returning a distinct error if this `os.Getenv` call returns the empty string. My guess is that if this env var is not set we'll return an error like `cannot read file: ""`. Though that said, the logs will include a stack trace which may be sufficient.

pkg/clients/cloudformation/cloudformation.go
```diff
@@ -33,7 +34,7 @@ type Client interface {
 }
 
 type cloudFormationClient struct {
-	cloudformation cfiface.CloudFormationAPI
+	cloudformation cfiface.ClientAPI
```
negz: What's the reason behind this interface change?

pkg/clients/eks/eks.go
```diff
@@ -105,9 +106,9 @@ type AMIClient interface {
 }
 
 type eksClient struct {
-	eks            eksiface.EKSAPI
+	eks            eksiface.ClientAPI
 	amiClient      AMIClient
-	sts            *sts.STS
+	sts            *sts.Client
```
negz: Ditto here - what's the context for this change?

pkg/clients/rds/rds.go
```diff
@@ -44,12 +45,17 @@ type Client interface {
 }
 
 // NewClient creates new RDS RDSClient with provided AWS Configurations/Credentials
-func NewClient(credentials []byte, region string) (Client, error) {
-	cfg, err := awsclients.LoadConfig(credentials, awsclients.DefaultSection, region)
-	if err != nil {
-		return nil, err
+func NewClient(ctx context.Context, credentials []byte, region string, useSA bool) (Client, error) {
```
negz: Style nit: I'd prefer an alternate `NewServiceAccountClient` or a varaidic option like `UsePodServiceAccount` over this boolean argument.

negz:
> I think the fact that authentication is done per pod doesn't really fit how Crossplane works in general.

@muvaf I agree that this is pretty different from how Crossplane is typically configured, but I do think it's a reasonable MVP to solve the immediate problems a few folks have come to us with. It sounds like some orgs aren't permitted to expose their credentials as a Kubernetes `Secret` for security reasons; this matches what the folks at AWS have told us is considered best practice. I like that at least our `Provider` resources have to opt into the functionality, so there's a little gating around whether services get to use the stack pod's credentials.

muvaf:
Alright, but let's make sure this meets their requirements. @hasheddan I assume they use credentials only in this way. Are they OK with using only one IAM role for resource provisioning in a given cluster? They may think that they can deploy more than one Crossplane instance to the same cluster, which is not possible. So, it'd be great if we can confirm what we give is what they want.

Ideally, I'd like us to discuss the final place we'd like to be in and how we would upgrade the clusters working with this MVP to that final place. Not really in a defined and planned way but at least very general path like `when we do X solution, we can migrate the existing setups like that`. But given that even if we mess up with auth and require reconfiguration, no resource will be modified/changed, so, it should be alright to not have an idea about that at the moment.

pkg/clients/cloudformation/cloudformation.go
hasheddan: The interfaces were renamed somewhere in the bump from v0.5 to v0.19 in the AWS SDK

pkg/clients/eks/eks.go
hasheddan: Same as above

hasheddan:
@muvaf that’s a great point! @phanimullapudi do the constraints mentioned above still slow for the functionality you desire?

hasheddan:
Note: PR to implement similar functionality in `aws-sdk-go-v2` has been [opened](https://github.com/aws/aws-sdk-go-v2/pull/488) but is in draft.

apis/v1alpha3/types.go
```diff
@@ -28,6 +28,11 @@ type ProviderSpec struct {
 
 	// Region for managed resources created using this AWS provider.
 	Region string `json:"region"`
+
+	// UseServiceAccount indicates to use an IAM Role associated Kubernetes
+	// ServiceAccount for authentication instead of a credentials Secret.
+	// https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
+	UseServiceAccount *bool `json:"useServiceAccount,omitempty"`
```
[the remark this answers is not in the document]
hasheddan: I was thinking that as well. We technically could leave it and it would be a dummy reference, but I think that is a poor user experience.

phanimullapudi:
> @muvaf that’s a great point! @phanimullapudi do the constraints mentioned above still slow for the functionality you desire?

Yes using one single IAM role should work for us.

apis/v1alpha3/types.go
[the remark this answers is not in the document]
hasheddan: Ref: https://github.com/crossplaneio/crossplane-runtime/pull/115 and https://github.com/crossplaneio/crossplane-tools/pull/15

pkg/controller/database/rdsinstance_test.go
```diff
@@ -166,6 +172,30 @@ func TestConnect(t *testing.T) {
 				cr: instance(),
 			},
 		},
+		"SuccessfulUseSA": {
```
negz: Nitpick; Should we write the full `ServiceAccount` so it's clearer what we're testing in the test output?

pkg/clients/aws.go
```diff
@@ -92,6 +102,45 @@ func LoadConfig(data []byte, profile, region string) (*aws.Config, error) {
 	return &config, err
 }
 
+// UsePodServiceAccount assumes an IAM role configured via a ServiceAccount.
+// https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
```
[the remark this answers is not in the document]
hasheddan: Definitely! Thanks for pointing out!

pkg/controller/cache/managed.go
```diff
@@ -83,15 +86,24 @@ func (c *connecter) Connect(ctx context.Context, mg resource.Managed) (managed.E
 
 	p := &awsv1alpha3.Provider{}
 	if err := c.client.Get(ctx, meta.NamespacedNameOf(g.Spec.ProviderReference), p); err != nil {
-		return nil, errors.Wrap(err, "cannot get provider")
+		return nil, errors.Wrap(err, errGetProvider)
+	}
+
+	if p.Spec.UseServiceAccount != nil && *p.Spec.UseServiceAccount {
```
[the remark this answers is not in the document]
hasheddan: I like that idea 👍

pkg/controller/database/rdsinstance_test.go
hasheddan: For sure!

hasheddan:
@negz thanks for review! Should be ready for merge now :heavy_check_mark:

alemairebe:
Could someone please document how to use this feature ?
Thanks !

hasheddan:
@alemairebe I have added a [doc](https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md) describing how to enable this feature. The `v0.9` release of Crossplane will make this even easier. Please let me know if you have any further questions I can help with :)

alemairebe:
Thank you for this quick answer !
It already helps a lot, my current use-case is to deploy and configure EKS (iam/userdata/asg/...) with TF and use crossplane to abstract access to DB. Could you also document the minimal policy needed by crossplane to handled only RDS ?
Is it planned to enable crossplane to create an IAM role which would be used by the app pod in the stack to access the DB ?

muvaf:
@alemairebe So, you'd like to use Crossplane to create an IAM Role with a few policies attached and then assume it in the app pod to access to the RDS. Is my understanding correct?

You can create the role via [`IAMRole`](https://github.com/crossplane/stack-minimal-aws/blob/master/kustomize/aws/identity/iamrole.yaml) CR and attach the policy via [`IAMRolePolicyAttachment`](https://github.com/crossplane/stack-minimal-aws/blob/master/kustomize/aws/identity/iamrolepolicyattachment.yaml) CR (you can find examples in [stacks guide](https://crossplane.io/docs/v0.8/stacks-guide-aws.html) or official [API reference](https://crossplane.io/docs/v0.8/api/crossplaneio/stack-aws/identity-aws-crossplane-io-v1alpha3.html), too).

How do you plan to enable pod to assume this role? By attaching an annotation on its service account?

alemairebe:
@muvaf Yes, I would like to use Crossplane to : 
1) create an RDS instance with IAMDatabaseAuthentication enabled
2) create a IAM role to access the RDS instance
3) map the IAM role to the masterUsername
4) create a service account with the right annotation to assume the role from a pod

muvaf:
@alemairebe It should be possible to make this scenario work in Crossplane except for the policy creation part that needs to be done for the IAM role created. I'm working on a set of YAMLs to see how it'd unfold e2e. Will let you know when it's ready.

PR #136, opened 2020-02-24
remove check for EKS regions

janwillies:
Remove the check for EKS regions because it's hard to keep the list updated manually

The AWS API will tell us an error anyway (albeit only during runtime). 

Relevant discussion: https://crossplane.slack.com/archives/CEG3T90A1/p1582580141019500?thread_ts=1582578523.016300&cid=CEG3T90A1

### Checklist

I have:

[`app.yaml`]: https://github.com/crossplane/stack-aws/blob/master/config/stack/manifests/app.yaml

### DIFF
--- a/apis/compute/v1alpha3/types.go
+++ b/apis/compute/v1alpha3/types.go
@@ -40,9 +40,6 @@ const (
 	errResourceIsNotEKSCluster = "the managed resource is not an EKSCluster"
 )
 
-// EKSRegion represents an EKS enabled AWS region.
-type EKSRegion string
-
 // VPCIDReferencerForEKSCluster is an attribute referencer that resolves VPCID from a referenced VPC
 type VPCIDReferencerForEKSCluster struct {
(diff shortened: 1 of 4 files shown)


apis/compute/v1alpha3/types.go
```diff
@@ -154,7 +142,6 @@ type EKSClusterParameters struct {
 	// https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html
 
 	// Region for this EKS Cluster.
-	// +kubebuilder:validation:Enum=us-west-2;us-east-1;eu-west-1
 	Region EKSRegion `json:"region"`
```
hasheddan: Since `EKSRegion` is no longer an enum and just a plain string do you want to change `Region` type to `string`?

janwillies: done

negz:
Thanks @janwillies!

PR #141, opened 2020-02-26
move IAMRole and IAMRolePolicyAttachment to v1beta1

sahil-lakhwani:

Add implementations of IAMRole and IAMRolePolicyAttachment that adhere to the managed resources standards

Tested with the following

IAMRole
```
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRole
metadata:
  name: sample-cluster-role
  annotations:
   crossplane.io/external-name: managed-resource-role
spec:
  forProvider:
    description: some description
    assumeRolePolicyDocument: |
      {
        "Version": "2012-10-17",
        "Statement": [
          {
            "Effect": "Allow",
            "Principal": {
              "Service": "eks.amazonaws.com"
            },
            "Action": "sts:AssumeRole"
          }
        ]
      }
  reclaimPolicy: Delete
  providerRef:
    name: example
```

IAMRolePolicyAttament
```
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRolePolicyAttachment
metadata:
  name: sample-role-servicepolicy
spec:
  forProvider:
    roleNameRef:
      name: sample-cluster-role
    # wellknown policy arn
    policyArn: arn:aws:iam::aws:policy/AmazonEKSServicePolicy
  reclaimPolicy: Delete
  providerRef:
    name: example
~                   
```

### Checklist

I have:

- [x]  Run `make reviewable` to ensure this PR is ready for review.

[`app.yaml`]: https://github.com/crossplane/stack-aws/blob/master/config/stack/manifests/app.yaml


sahil-lakhwani:
```
kubectl get iamrole.identity.aws.crossplane.io/metadata-name

NAME            ROLENAME   DESCRIPTION   READY   SYNCED   AGE
metadata-name              update desc   True    True     12s
```

For the `ROLENAME` column, because .spec.rolename has been removed, I tried to pull the value from annotation `crossplane.io/external-name` in the CRD definition. Didn't work for me.
(Pulling values from other annotations with no special characters does seem to work)

pkg/controller/identity/iamrole/controller_test.go
```diff
@@ -57,7 +58,7 @@ func TestMain(m *testing.M) {
 func Test_Connect(t *testing.T) {
```
hasheddan: Can we refactor these tests to match our table-driven pattern in other `v1beta1` resources?

Ref: https://github.com/crossplane/stack-aws/blob/94baed407fc79f3a083c409b42542a00f7f0eb74/pkg/controller/database/rdsinstance_test.go#L106

pkg/controller/identity/iamrolepolicyattachment/controller_test.go
```diff
@@ -57,7 +57,7 @@ func TestMain(m *testing.M) {
 func Test_Connect(t *testing.T) {
```
hasheddan: Same comment here about table-driven tests.

pkg/clients/iam/iamrole.go
```diff
@@ -3,16 +3,54 @@ package iam
...
+
+// GenerateDeleteRoleInput from IAMRoleSpec
+func GenerateDeleteRoleInput(name string) *iam.DeleteRoleInput {
+	m := &iam.DeleteRoleInput{
+		RoleName: aws.String(name),
+	}
+	return m
+}
+
+// UpdateRoleExternalStatus updates the external status object, given the observation
```
hasheddan: Having `External` in the name here as we are actually updating the name of the kubernetes object rather than the external IAM role on AWS. We typically defer this type of functionality to the controller, so this can likely be removed (see comments in controller).

pkg/clients/iam/iamrolepolicyattachment.go
```diff
@@ -16,3 +18,28 @@ type RolePolicyAttachmentClient interface {
...
+// GenerateDetachRolePolicyInput from IAMRolePolicyAttachmentSpec
+func GenerateDetachRolePolicyInput(p *v1beta1.IAMRolePolicyAttachmentParameters) *iam.DetachRolePolicyInput {
+	m := &iam.DetachRolePolicyInput{
+		PolicyArn: aws.String(p.PolicyARN),
+		RoleName:  aws.String(p.RoleName),
+	}
+	return m
+}
+
+// UpdateRolePolicyExternalStatus updates the external status object, given the observation
```
hasheddan: Same comment as in IAM role client.

pkg/controller/identity/iamrole/controller.go
```diff
@@ -107,38 +109,39 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}
 
 	if err != nil {
```
hasheddan: Nitpick: this error check and the one above can be refactored as:

```
if err != nil {
	return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(iam.IsErrorNotFound, err), errGet)
}
```
`ResourceExists: false` is implied and an error will cause requeue :heavy_check_mark:

pkg/controller/identity/iamrole/controller.go
```diff
@@ -107,38 +109,39 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+
+	req := e.client.CreateRoleRequest(iam.GenerateCreateRoleInput(name, &cr.Spec.ForProvider))
 
 	result, err := req.Send(ctx)
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
 
-	cr.UpdateExternalStatus(*result.Role)
+	iam.UpdateRoleExternalStatus(cr, *result.Role)
```
hasheddan: We typically would update the status in the subsequent `Observe`.

Ref: https://github.com/crossplane/stack-aws/blob/94baed407fc79f3a083c409b42542a00f7f0eb74/pkg/controller/database/rdsinstance.go#L140

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -91,13 +91,13 @@ type external struct {
...
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
 
 	req := e.client.ListAttachedRolePoliciesRequest(&awsiam.ListAttachedRolePoliciesInput{
-		RoleName: aws.String(cr.Spec.RoleName),
+		RoleName: aws.String(cr.Spec.ForProvider.RoleName),
 	})
 
 	observed, err := req.Send(ctx)
```
hasheddan: err check here can also be refactored as described in comment on iam role controller

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -127,33 +127,31 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 
 	cr.SetConditions(runtimev1alpha1.Available())
 
-	cr.UpdateExternalStatus(*attachedPolicyObject)
+	iam.UpdateRolePolicyExternalStatus(cr, *attachedPolicyObject)
```
hasheddan: Here we are already updating the status in the `Observe` method (contrary to the iam role controller), but it would be nice to have a `GenerateObservation` function that returns the `AtProvider` observation rather than passing in the object to have its `AtProvider` status set in the client.

pkg/clients/iam/iamrole.go
hasheddan: Instead we would probably want something like a `GenerateObservation` function, ref: https://github.com/crossplane/stack-aws/blob/94baed407fc79f3a083c409b42542a00f7f0eb74/pkg/clients/rds/rds.go#L220

pkg/controller/identity/iamrole/controller_test.go
```diff
@@ -247,38 +248,84 @@ func Test_Create(t *testing.T) {
...
+			ForProvider: v1beta1.IAMRoleParameters{
+				AssumeRolePolicyDocument: "arbitrary role policy doc",
+				Description:              "arbitrary role description",
+			},
+		},
+	}
+	meta.SetExternalName(&mockManaged, "arbitrary role name")
 
-	_, err := mockExternalClient.Update(context.Background(), &mockManaged)
+
```
golangcibot: File is not `goimports`-ed with -local github.com/crossplane/stack-aws (from `goimports`)
```suggestion

```

apis/identity/v1beta1/iamrole_types.go
```diff
@@ -38,13 +34,13 @@ type IAMRoleParameters struct {
 	Description string `json:"description,omitempty"`
 
 	// RoleName presents the name of the IAM role.
-	RoleName string `json:"roleName"`
+	// RoleName string `json:"roleName"`
```
hasheddan: You can go ahead and remove this :)

apis/identity/v1beta1/iamrole_types.go
```diff
@@ -38,13 +34,13 @@ type IAMRoleParameters struct {
 	Description string `json:"description,omitempty"`
```
hasheddan:
```suggestion
	Description *string `json:"description,omitempty"`
```

https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#pointer-types-and-markers

apis/identity/v1beta1/iamrolepolicyattachment_types.go
```diff
@@ -45,7 +42,7 @@ func (v *IAMRoleNameReferencerForIAMRolePolicyAttachment) Assign(res resource.Ca
 		return errors.New(errResourceIsNotIAMRolePolicyAttachment)
 	}
 
-	p.Spec.RoleName = value
+	p.Spec.ForProvider.RoleName = value
 	return nil
 }
```
hasheddan: It wouldn't let me comment directly on these lines, but it looks like `RoleName` and and `RoleNameRef` are both optional fields. Would you mind updating them as follows:

```
        // RoleName presents the name of the IAM role.
        // +optional
	RoleName *string `json:"roleName,omitempty"`

	// RoleNameRef references to an IAMRole to retrieve its Name
        // +optional
	RoleNameRef *IAMRoleNameReferencerForIAMRolePolicyAttachment `json:"roleNameRef,omitempty" resource:"attributereferencer"`
```

apis/identity/v1beta1/iamrole_types.go
```diff
@@ -14,15 +14,11 @@ See the License for the specific language governing permissions and
...
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/aws/aws-sdk-go-v2/service/iam"
-
-	aws "github.com/crossplane/stack-aws/pkg/clients"
-
 	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 )
```
hasheddan: It looks like there are some additional configurable fields for `IAMRole` that can be exposed: https://github.com/aws/aws-sdk-go-v2/blob/9fc62ee75d1acca973ac777e51993fce74f6a27f/service/iam/api_op_CreateRole.go#L13

apis/identity/v1beta1/iamrolepolicyattachment_types.go
hasheddan: @sahil-lakhwani per your message in Slack, I think it is fine to leave `RoleName` as-is (since it is "required" for creation of the external resource but we do not validate it on object creation), but we should add an `// +optional` tag to `RoleNameRef` :) Thank you for noticing this!

Ref: https://github.com/crossplane/stack-gcp/blob/8ece46f91e44e4139ba92f848be687b970acf214/apis/container/v1alpha1/nodepool_types.go#L125

pkg/controller/identity/iamrole/controller.go
```diff
@@ -89,78 +91,77 @@ type external struct {
...
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
 
-	cr.UpdateExternalStatus(*result.Role)
-
 	return managed.ExternalCreation{}, nil
 }
 
 func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
```
hasheddan: The `Update` methods for `IAMRole` and `IAMRolePolicyAttachment` will never be called because we are not checking if `IsUpToDate` in `Observe` method. This can be a tricky thing to get right, but here is a good reference: 

https://github.com/crossplane/stack-gcp/blob/8ece46f91e44e4139ba92f848be687b970acf214/pkg/controller/database/cloudsql.go#L150

https://github.com/crossplane/stack-gcp/blob/8ece46f91e44e4139ba92f848be687b970acf214/pkg/controller/database/cloudsql.go#L156

Let me know if you need more detail :)

infinitecompute:
@sahil-lakhwani What is the latest on updating and merging this PR?

sahil-lakhwani:
> @sahil-lakhwani What is the latest on updating and merging this PR?

Hi @infinitecompute 
There were some confusions, regarding which I had quick discussions with @hasheddan 
And then I had to make some more changes to match the name change (provider-aws)
I have updated the PR.

apis/identity/v1beta1/iamrole_types.go
```diff
@@ -14,37 +14,77 @@ See the License for the specific language governing permissions and
...
 
+// Tag represents user-provided metadata that can be associated
+// with a IAM role. For more information about tagging,
+// see Tagging IAM Identities (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags.html)
+// in the IAM User Guide.
+type Tag struct {
+
+	// The key name that can be used to look up or retrieve the associated value.
+	// For example, Department or Cost Center are common choices.
+	Key string `json:"key,omitempty"`
```
hasheddan: This is required per https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags.html so you can remove `omitempty`:

"You can create a tag with an empty value such as phoneNumber = . You cannot create an empty tag key."

apis/identity/v1beta1/iamrole_types.go
```diff
@@ -14,37 +14,77 @@ See the License for the specific language governing permissions and
...
+	// The value associated with this tag. For example, tags with a key name of
+	// Department could have values such as Human Resources, Accounting, and Support.
+	// Tags with a key name of Cost Center might have values that consist of the
+	// number associated with the different cost centers in your company. Typically,
+	// many resources have tags with the same key name but with different values.
+	//
+	// AWS always interprets the tag Value as a single string. If you need to store
+	// an array, you can store comma-separated values in the string. However, you
+	// must interpret the value in your code.
+	Value string `json:"value,omitempty"`
```
hasheddan: This is optional so you can add `// +optional`

pkg/clients/iam/iamrole.go
```diff
@@ -1,18 +1,138 @@
...
 	GetRoleRequest(*iam.GetRoleInput) iam.GetRoleRequest
 	CreateRoleRequest(*iam.CreateRoleInput) iam.CreateRoleRequest
 	DeleteRoleRequest(*iam.DeleteRoleInput) iam.DeleteRoleRequest
+	UpdateRoleRequest(*iam.UpdateRoleInput) iam.UpdateRoleRequest
 }
 
 // NewRoleClient returns a new client using AWS credentials as JSON encoded data.
 func NewRoleClient(conf *aws.Config) (RoleClient, error) {
 	return iam.New(*conf), nil
 }
```
hasheddan: Ideally we would support multiple auth methods here, as you can see demonstrated here: https://github.com/crossplane/provider-aws/blob/1d51592300403eea476170ac4f1b549cf1ae42a6/pkg/clients/rds/rds.go#L48

pkg/clients/iam/iamrolepolicyattachment.go
```diff
@@ -16,3 +19,45 @@ type RolePolicyAttachmentClient interface {
 func NewRolePolicyAttachmentClient(conf *aws.Config) (RolePolicyAttachmentClient, error) {
 	return iam.New(*conf), nil
 }
```
hasheddan: Same note on auth here as above

pkg/controller/identity/iamrole/controller.go
```diff
@@ -67,7 +73,7 @@ type connector struct {
 }
 
 func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (managed.ExternalClient, error) {
```
hasheddan: It would be preferable to move away from using the `util.RetrieveAwsConfigFromProvider` function here I think, as has been done on other `v1beta1` objects: https://github.com/crossplane/provider-aws/blob/1d51592300403eea476170ac4f1b549cf1ae42a6/pkg/controller/database/rdsinstance.go#L91

However, I do think all of this authentication machinery will be refactored in the not too distant future so I would be fine if this didn't change for now.

pkg/clients/iam/iamrole.go
hasheddan: See comment on the controller for more context. I would be okay with leaving this as-is for now if needed.

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -125,90 +123,61 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
-	dReq := e.client.DetachRolePolicyRequest(&awsiam.DetachRolePolicyInput{
-		PolicyArn: aws.String(cr.Status.AttachedPolicyARN),
-		RoleName:  aws.String(cr.Spec.RoleName),
-	})
-
-	if _, err := dReq.Send(ctx); err != nil {
-		return managed.ExternalUpdate{}, errors.Wrapf(err, errDetach, cr.Status.AttachedPolicyARN, cr.Spec.RoleName)
-	}
-
+	// PolivyARN is the only distinguishing field and on update to that, new policy is attached
```
hasheddan: I think we need to let user know that their `spec` is out of sync with the external resource. However, returning an error here will cause immediate requeue, which I am not sure we want. @negz @muvaf any thoughts here?

pkg/clients/iam/iamrole.go
```diff
@@ -1,18 +1,138 @@
 package iam
```
hasheddan: `IAMRole` client methods should be tested.

pkg/clients/iam/iamrolepolicyattachment.go
```diff
@@ -3,6 +3,9 @@ package iam
 import (
```
hasheddan: `IAMRolePolicyAttachment` methods should be tested.

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -125,90 +123,61 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		}, nil
 	}
 
+	current := cr.Spec.ForProvider.DeepCopy()
+	iam.LateInitializePolicy(&cr.Spec.ForProvider, attachedPolicyObject)
+	if !reflect.DeepEqual(current, &cr.Spec.ForProvider) {
```
muvaf: Same here, `cmp` instead of `reflect`.

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -125,90 +123,61 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+
 	cr.SetConditions(runtimev1alpha1.Available())
 
-	cr.UpdateExternalStatus(*attachedPolicyObject)
+	cr.Status.AtProvider = iam.GenerateRolePolicyObservation(*attachedPolicyObject)
 
 	return managed.ExternalObservation{
-		ResourceExists: true,
+		ResourceExists:   true,
+		ResourceUpToDate: cr.Spec.ForProvider.PolicyARN == cr.Status.AtProvider.AttachedPolicyARN,
```
muvaf: As this is not an update-able object, `ResourceUpToDate` can always return `true`.

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -125,90 +123,61 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
 	}
 
 	cr.SetConditions(runtimev1alpha1.Creating())
 
-	req := e.client.AttachRolePolicyRequest(&awsiam.AttachRolePolicyInput{
-		PolicyArn: aws.String(cr.Spec.PolicyARN),
-		RoleName:  aws.String(cr.Spec.RoleName),
-	})
+	req := e.client.AttachRolePolicyRequest(iam.GenerateAttachRolePolicyInput(&cr.Spec.ForProvider))
+	req.SetContext(ctx)
```
muvaf: As I mentioned in IAMRole controller, you can call `Send` directly on the request and you don't need to set context as `Send` does that when you pass the context in.

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -125,90 +123,61 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
 
 	_, err := req.Send(ctx)
 
-	return managed.ExternalCreation{}, errors.Wrapf(err, errAttach, cr.Spec.PolicyARN, cr.Spec.RoleName)
+	return managed.ExternalCreation{}, errors.Wrapf(err, errAttach, cr.Spec.ForProvider.PolicyARN, cr.Spec.ForProvider.RoleName)
```
muvaf: You don't need to return anything other than the error(`err`) and the error string(`errAttach`) as managed reconciler properly fills the error and logs with necessary identifiers.

pkg/controller/identity/iamrolepolicyattachment/controller.go
muvaf: Is `PolicyARN` actually an immutable field? If so, I think we can skip doing anything in `Update`.

Letting the user know about the drift is challenging in AWS because update requests are constructed only with update-able parameters, hence API server wouldn't return an error like GCP `this field cannot be updated`. Ideally, validation should reject editing an immutable field.

sahil-lakhwani: @hasheddan @muvaf In the current implementation, if the `PolicyARN` in `spec` is changed anyhow, a new policy is being created. And in the `Update` method, we have:
 `if cr.Status.AttachedPolicyARN == "" || cr.Spec.PolicyARN == cr.Status.AttachedPolicyARN`

I think we can check for the above condition in `Observe` itself, and decide either to create a new policy or update the existing.

hasheddan: @sahil-lakhwani I am hesitant to delete and recreate an external object as part of the update logic of a reconciler because it feels like it violates that contract of the lifecycle of a custom resource being tied to the lifecycle of an external resource. @muvaf let me know if you think differently here. It sounds like the path forward for the time being is to just always return `true` for `ResourceUpToDate` if the `IAMRolePolicyAttachment` exists at all (per @muvaf [comment](https://github.com/crossplane/provider-aws/pull/141/files/acb886579b36a352e22e7201e6f3b4f98a8cb786#r390895464)), and keep the current functionality you have here of doing nothing in the `Update` method.

@muvaf I wonder if we should expose something in `crossplane-runtime` that allows consumers to report back that "this resource is not up to date, but we are not trying to fix it". We could do this currently by manually updating status condition here, but I think `crossplane-runtime` would overwrite it after we returned an empty `managed.ExternalUpdate`.

muvaf: Yeah we don't delete anything unless actual deletion request comes in. @sahil-lakhwani this field should be marked as immutable and when we get around implementation of immutability, it will work. But for the time being, that's all we can do.

About the drift in general; there are three kinds of drifts:
* In progress: you changed something and the update process is going on but the property is still the same in external resource, like you increased the size of DB from 20GB to 400GB and `spec` says 400GB immediately while the process is going on. Not much to do here unless provider reports an unavailable state, then we could update the condition. But other than that, status is expected to show that an update is going on if possible.
* Immutable field changed by user. Specifically in AWS, `Modify` requests do not include those fields so we can't have provider return error in those cases. The best we could do is to prevent those fields from being edited.
* Update-able field changed but resulted in error. Provider tells you the error and we show that.

@hasheddan For the second case, we could investigate a mechanism where in `Observe` we compare `spec` with what exists in the provider and update a `condition` saying whether there is a drift without going into details like whether this was because an immutable field or an update is due etc. It's a bit challenging to implement this in a good way though (comparison is hard). Maybe use IsUpToDate check to set a condition? But that'll require full scan of fields opposed to selective scanning that current `IsUpToDate` functions do.

pkg/controller/identity/iamrole/controller.go
```diff
@@ -81,86 +87,101 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
 
 	cr.Status.SetConditions(runtimev1alpha1.Creating())
 
-	req := e.client.CreateRoleRequest(&awsiam.CreateRoleInput{
-		RoleName:                 aws.String(cr.Spec.RoleName),
-		AssumeRolePolicyDocument: aws.String(cr.Spec.AssumeRolePolicyDocument),
-		Description:              aws.String(cr.Spec.Description),
-	})
+	name := meta.GetExternalName(cr)
+	if name == "" {
```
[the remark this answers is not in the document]
sahil-lakhwani: Ok. I thought it is the job of the managed resource controller to set the annotation. But I see it is the annotation is already set before it reaches `Create` of the managed resource controller.

pkg/controller/identity/iamrole/controller.go
```diff
@@ -81,86 +87,101 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
-		return managed.ExternalObservation{
-			ResourceExists: false,
-		}, nil
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(iam.IsErrorNotFound, err), errGet)
 	}
 
-	if err != nil {
-		return managed.ExternalObservation{}, errors.Wrapf(err, errGet, cr.Spec.RoleName)
+	role := *observed.Role
```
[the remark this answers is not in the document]
sahil-lakhwani: @muvaf I think checking for `err` should be enough. If there is no Role, `err` is non-nil.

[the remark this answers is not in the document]
muvaf: That assumes everything works correct in AWS SDK though. Not a big deal, but as a general rule of thumb, I'd go with the fact that if something is pointer then it should be used after we make sure it has a value. Because dereferencing `nil` brings down the whole manager with all of its controllers by `panic`ing.

pkg/clients/iam/iamrole.go
```diff
@@ -1,18 +1,138 @@
...
+	m := &iam.UpdateRoleInput{
+		RoleName:           aws.String(name),
+		Description:        p.Description,
+		MaxSessionDuration: p.MaxSessionDuration,
+	}
+	return m
+}
+
+// GenerateDeleteRoleInput from IAMRoleSpec
+func GenerateDeleteRoleInput(name string) *iam.DeleteRoleInput {
```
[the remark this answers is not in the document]
sahil-lakhwani: @muvaf I remember remember discussing this on sig-services to keep the translation code in separate methods.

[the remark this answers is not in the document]
muvaf: Yeah but I mean the cases where the function may not be considered as translation. For example here, it's basically passing the identifier. So, I think we can get away by directly constructing those structs with 1-2 fields in the place where they are needed to reduce the back-and-forth between the packages. Usually get and create objects are the big ones that require almost-full translation.

pkg/controller/identity/iamrolepolicyattachment/controller.go
sahil-lakhwani: @muvaf `errAttach` has format specifiers, and removing the values for those resulted in this for me (kubectl describe)
```
 Warning  CannotCreateExternalResource  6m59s  managed/iamrolepolicyattachment.identity.aws.crossplane.io  failed to attach the policy with arn %!!(MISSING)v(MISSING) to role %!!(MISSING)v(MISSING): NoSuchEntity: Policy arn:aws:iam::aws:policy/AleaForBusinessDeviceSetup does not exist or is not attachable.
```
The values are not filled and I am not sure how they will. Am I missing something?

sahil-lakhwani:
@hasheddan @muvaf I have added new commits to address the comments.

pkg/clients/iam/iamrole.go
hasheddan: Going to go ahead and resolve this and will see about addressing in a later PR :)

pkg/controller/identity/iamrole/controller.go
hasheddan: As with note on client, we can implement this in a future PR.

pkg/controller/identity/iamrole/controller.go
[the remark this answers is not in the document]
hasheddan: Looks like this still needs to be checked @sahil-lakhwani :)

pkg/controller/identity/iamrole/controller.go
```diff
@@ -81,92 +88,97 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
 	if err != nil {
 		return nil, errors.Wrap(err, errClient)
 	}
-	return &external{c}, nil
+	return &external{c, conn.client}, nil
```
hasheddan: No change needed for now, but I generally prefer to include the names of fields in the struct being returned for readability and new contributors :)

pkg/controller/identity/iamrolepolicyattachment/controller.go
hasheddan: I believe @muvaf just means that you can remove the formatted string and return `errors.Wrap(err, errAttach)` :)

sahil-lakhwani:
> I believe there are a few more of @muvaf comments to address, but going to go ahead and give this the LGTM! @sahil-lakhwani would you mind running a few manual e2e tests (like you did earlier) to verify functionality is operating as expected? Thanks so much for continuing to push through this!

@hasheddan Couldn't directly reply to your comment https://github.com/crossplane/provider-aws/pull/141#discussion_r392948035  but the thing you mentioned has been taken care of:
https://github.com/sahil-lakhwani/provider-aws/blob/54a5aa32696b9434364e038fd0db05a1b5864fc4/pkg/controller/identity/iamrole/controller.go#L113-L115

Also, I saw the test coverage results today, will take a look.

infinitecompute:
Is additional code coverage the only blocker to merging?

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -83,37 +86,30 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
-				ResourceExists: false,
-			}, nil
-		}
-
-		return managed.ExternalObservation{}, errors.Wrapf(err, errGet, cr.Spec.RoleName)
+		return managed.ExternalObservation{}, errors.Wrapf(resource.Ignore(iam.IsErrorNotFound, err), errGet, cr.Spec.ForProvider.RoleName)
 	}
 
 	var attachedPolicyObject *awsiam.AttachedPolicy
 	for i := 0; i < len(observed.AttachedPolicies); i++ {
```
muvaf:
```suggestion
	for _, policy := range observed.AttachedPolicies {
		if cr.Spec.ForProvider.PolicyARN == aws.StringValue(policy.PolicyArn) {
			attachedPolicyObject = &policy
			break
		}
	}
```

Nitpick: using `range` would probably be easier to read.

pkg/controller/identity/iamrolepolicyattachment/controller.go
muvaf:
```suggestion
	return managed.ExternalCreation{}, errors.Wrap(err, errAttach)
```

I'm suggesting this because this. We don't need to repeat what's on the CR as long as the error string is specific enough.

apis/identity/v1beta1/iamrolepolicyattachment_types.go
```diff
@@ -55,20 +52,22 @@ type IAMRolePolicyAttachmentParameters struct {
...
 	// attach.
+	// +immutable
 	PolicyARN string `json:"policyArn"`
 
 	// RoleName presents the name of the IAM role.
 	RoleName string `json:"roleName,omitempty"`
 
 	// RoleNameRef references to an IAMRole to retrieve its Name
+	// +optional
 	RoleNameRef *IAMRoleNameReferencerForIAMRolePolicyAttachment `json:"roleNameRef,omitempty" resource:"attributereferencer"`
```
muvaf: You can delete `resource:"attributereferencer"` as we don't use that anymore.

pkg/controller/identity/iamrole/controller.go
```diff
@@ -81,92 +88,97 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
-	cr.UpdateExternalStatus(*result.Role)
-
-	return managed.ExternalCreation{}, nil
+	_, err := e.client.CreateRoleRequest(iam.GenerateCreateRoleInput(meta.GetExternalName(cr), &cr.Spec.ForProvider)).Send(ctx)
+	return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 }
 
 func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
 	// TODO(soorena776): add more sophisticated Update logic, once we
 	// categorize immutable vs mutable fields (see #727)
```
negz: Is this TODO still relevant?

pkg/controller/identity/iamrolepolicyattachment/controller.go
```diff
@@ -43,18 +44,20 @@ const (
 	errGet              = "failed to get IAMRolePolicyAttachments for role with name: %v"
 	errAttach           = "failed to attach the policy with arn %v to role %v"
 	errDetach           = "failed to detach the policy with arn %v to role %v"
+
+	errKubeUpdateFailed = "cannot update RDS instance custom resource"
```
negz:
```suggestion
	errKubeUpdateFailed = "cannot late initialize IAMRolePolicyAttachment"
```

pkg/controller/identity/iamrole/controller.go
```diff
@@ -81,92 +88,97 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrapf(resource.Ignore(iam.IsErrorNotFound, err), errGet, meta.GetExternalName(cr))
+	}
 
-	if iam.IsErrorNotFound(err) {
-		return managed.ExternalObservation{
-			ResourceExists: false,
-		}, nil
+	if observed.Role == nil {
+		return managed.ExternalObservation{}, errors.New(errSDK)
```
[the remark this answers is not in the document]
sahil-lakhwani: Yeah that's something we rarely expect to occur. We are not depending on the API for the case where the`Role` is still `nil` despite of `err` being `nil`

pkg/controller/identity/iamrolepolicyattachment/controller_test.go
```diff
@@ -19,468 +19,535 @@ package iamrolepolicyattachment
...
+}
 
-	for _, tc := range []struct {
-		description       string
-		managedObj        resource.Managed
-		configErr         error
-		clientErr         error
-		expectedClientNil bool
-		expectedErrNil    bool
+func Test_Connect(t *testing.T) {
```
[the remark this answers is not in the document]
sahil-lakhwani: That seems good. Can you sight an example please?

pkg/controller/identity/iamrolepolicyattachment/controller_test.go
```diff
@@ -19,468 +19,535 @@ package iamrolepolicyattachment
...
+	type args struct {
+		newClientFn func(*aws.Config) (iam.RolePolicyAttachmentClient, error)
+		awsConfigFn func(context.Context, client.Reader, *corev1.ObjectReference) (*aws.Config, error)
+		cr          resource.Managed
+	}
+	type want struct {
+		err error
+	}
+
+	cases := map[string]struct {
```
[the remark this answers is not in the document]
sahil-lakhwani: That seems good. Can you sight an example?

pkg/clients/iam/iamrole.go
```diff
@@ -1,18 +1,120 @@
...
+	}
+	desired, ok := generated.(*iam.Role)
+	if !ok {
+		return true, errors.New(errCheckUpToDate)
+	}
+
+	GenerateIAMRole(in, desired)
+
+	// 'AssumeRolePolicyDocument' is an escaped string in iam.Role and a normal string  in v1beta.IAMRole.
+	// There is no proper way to compare them.
```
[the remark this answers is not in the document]
sahil-lakhwani: So, the problem while escaping was with spaces and new lines. I figured that `json.Compact` did  the work. I've also added a test case for that.

pkg/clients/iam/iamrole.go
```diff
@@ -1,18 +1,120 @@
...
+	in.MaxSessionDuration = awsclients.LateInitializeInt64Ptr(in.MaxSessionDuration, role.MaxSessionDuration)
+	in.Path = awsclients.LateInitializeStringPtr(in.Path, role.Path)
+
+	if in.PermissionsBoundary != nil {
+		in.PermissionsBoundary = awsclients.LateInitializeStringPtr(in.PermissionsBoundary, role.PermissionsBoundary.PermissionsBoundaryArn)
+	}
+
+	for _, tag := range in.Tags {
+		role.Tags = append(role.Tags, iam.Tag{Key: &tag.Key, Value: &tag.Value})
+	}
```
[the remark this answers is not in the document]
sahil-lakhwani: Made a change, not visible here.

pkg/controller/identity/iamrolepolicyattachment/controller_test.go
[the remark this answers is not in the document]
negz: Sure. See https://github.com/crossplane/crossplane-runtime/blob/da385a76/pkg/fieldpath/fieldpath_test.go. Note that I don't think this needs to block this PR.

muvaf:
@negz I was trying to write the YAMLs for the scenario [here](https://github.com/crossplane/provider-aws/pull/126#issuecomment-604302587) and realized that we don't support inline IAM policies. Digging more, I found out `RolePolicy` term is used for inline policy and when we create a `IAMRolePolicyAttachment`, what happens is that the normal policy gets copied to `Role` as managed inline policy. The interesting part is that actual inline policies are added in the same way, as seen [here](https://docs.aws.amazon.com/cli/latest/reference/iam/put-role-policy.html) and there is no way to create a `RolePolicy` without attaching it to a role. So, I think some consolidation could make sense there.

I suggest:
* Rename `IAMRolePolicyAttachment` to `IAMRolePolicy`.
* Add a field [`PolicyDocument`](https://github.com/aws/aws-sdk-go-v2/blob/master/service/iam/api_op_PutRolePolicy.go#L37) to allow specifying the policy inline.
* If the [`PolicyDocument`](https://github.com/aws/aws-sdk-go-v2/blob/master/service/iam/api_op_PutRolePolicy.go#L37) is given, call `PutRolePolicy` and if `PolicyArn` is given call [`AttachRolePolicy`](https://github.com/aws/aws-sdk-go-v2/blob/master/service/iam/api_op_AttachRolePolicy.go#L14).

The idea is that we'd have one managed resource for _inline policy_ operations as opposed to mapping each imperative call to a managed resource. This functionality is additive and not blocking this PR but I wanted to drop it on this PR in case we want to rename the resource, better to do it before bumping to `v1beta1`.

Also, should we drop `IAM` prefix from the name of both IAM resources? We already have `identity` in the group name. Maybe change `identity` group name to `iam` [to match the API](https://github.com/aws/aws-sdk-go-v2/tree/master/service/iam)?

negz:
> Also, should we drop IAM prefix from the name of both IAM resources? We already have identity in the group name. Maybe change identity group name to iam to match the API?

This sounds reasonable to me. We'd have `role.iam.aws.crossplane.io` etc, and the Go types would be imported as `iamv1beta1.Role`.

negz:
> I found out RolePolicy term is used for inline policy

Are you sure this is the case? As far as I can tell there's both `PutRolePolicy` (which specifies a policy to include inline) and `AttachRolePolicy` (which references an existing policy ARN) API calls. Both appear to use the term "role policy" for the policy, whether it's inline or distinct.

> what happens is that the normal policy gets copied to Role as managed inline policy.

This doesn't match my understanding of the API. There seems to be a `ListRolePolicy` API call that lists inline policies, and also a `ListAttachedRolePolicies` that lists attached policies. As far as I can tell when you call `GetRole` it includes neither attached or inline policies.

> we don't support inline IAM policies

@muvaf do we need to support inline IAM policies to support that use case? Under the current model, could we:

1. Create an IAM role that would be used by the pod that will access the RDS instance
1. Create an IAM policy that allows access to the RDS instance
1. Create an IAM role policy attachment to attach the policy from 2 to the role from 1?

> I suggest:

Would this mean we could only attach IAM policies to IAM roles that were managed by Crossplane? It seems to me like there are two ways to associate a policy with a role; either include the policy inline, or add it as a distinct API object and create a reference between the two. The latter seems to be the more granular and flexible model.

muvaf:
> This sounds reasonable to me. We'd have role.iam.aws.crossplane.io etc, and the Go types would be imported as iamv1beta1.Role. I feel okay merging this PR and making that change in a subsequent one though, to make it easier to review.

Sounds good. The rest of the discussions is not a blocker for this PR.

There is [this doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html) that explains the inline and standalone policies. In all cases, you can use standalone policies with any entity (group, role, user etc.). The inline policy has essentially the same semantics except it has a one-to-one relationship with the entity.

Quoting from _Inline Policies_ section of that doc:
> Each policy is an inherent part of the user, group, or role. Notice that two roles include the same policy (the DynamoDB-books-app policy), but they are not sharing a single policy; each role has its own copy of the policy. 

What I meant by copying is the mechanism that is described in the doc: if we used `IAMRolePolicyAttachment` that refers to a standalone policy, even though that standalone policy changes, the policy that is attached to the `Role` won't change. The moment we attach, a copy of the policy is coupled with the role. IMO, that is essentially the same thing with `PutRolePolicy` except that you give the policy to attach instead of refer to an existing one.

> This doesn't match my understanding of the API. There seems to be a ListRolePolicy API call that lists inline policies, and also a ListAttachedRolePolicies that lists attached policies. As far as I can tell when you call GetRole it includes neither attached or inline policies.

Here is the comments from those calls:
> ListAttachedRolePoliciesRequest returns a request value for making API operation for
AWS Identity and Access Management.
Lists all managed policies that are attached to the specified IAM role.

> ListRolePoliciesRequest returns a request value for making API operation for
AWS Identity and Access Management.
Lists the names of the inline policies that are embedded in the specified
IAM role.

I can see your point that they are treated differently. Digging the outputs of those two calls, `ListRolePoliciesRequest` returns the `PolicyNames []string` and `ListAttachedRolePoliciesRequest` returns `[]AttachedPolicy` which is just `PolicyName string` and `PolicyArn string`.

I feel tempted to handle both inline and reference cases in one resource but I can also see your point about high fidelity even though it will increase number of different managed resources.

> Create an IAM policy that allows access to the RDS instance

That's what we are not able to do at the moment and I figured if we were able to expand the attachment for inline policy, then it'd be pretty easy to enable the use-case.

> It seems to me like there are two ways to associate a policy with a role; either include the policy inline, or add it as a distinct API object and create a reference between the two. The latter seems to be the more granular and flexible model.

muvaf:
> This sounds reasonable to me. We'd have role.iam.aws.crossplane.io etc, and the Go types would be imported as iamv1beta1.Role. I feel okay merging this PR and making that change in a subsequent one though, to make it easier to review.

Sounds good. The rest of the discussions is not a blocker for this PR. I can prepare a PR to do that.

There is [this doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_managed-vs-inline.html) that explains the inline and standalone policies. In all cases, you can use standalone managed policies with any entity (group, role, user etc.). The inline policy has essentially the same semantics except it has a one-to-one relationship with the entity.

Quoting from _Inline Policies_ section of that doc:
> Each policy is an inherent part of the user, group, or role. Notice that two roles include the same policy (the DynamoDB-books-app policy), but they are not sharing a single policy; each role has its own copy of the policy. 

What I meant by copying is the mechanism that is described in the doc: if we used `IAMRolePolicyAttachment` that refers to a managed policy, ~~even though that managed policy changes, the policy that is attached to the `Role` won't change. The moment we attach, a copy of the policy is coupled with the role. IMO, that is essentially the same thing with `PutRolePolicy` except that you give the policy to attach instead of referring to an existing one.~~ The CLI equivalent of those actions have more description: [put-role-policy](https://docs.aws.amazon.com/cli/latest/reference/iam/put-role-policy.html) and [attach-role-policy](https://docs.aws.amazon.com/cli/latest/reference/iam/attach-role-policy.html). I think your understanding is closer how it works and attaching is to refer managed policies, not copy it and treat as inline policy.

So, I agree that _managed policy attachment_ and _inline policy addition_ should be different resources.


> Create an IAM policy that allows access to the RDS instance

That's what we are not able to do at the moment and I figured if we were able to expand the attachment for inline policy, then it'd have been pretty easy to enable the use-case.

> It seems to me like there are two ways to associate a policy with a role; either include the policy inline, or add it as a distinct API object and create a reference between the two. The latter seems to be the more granular and flexible model.

I agree. But both ways have their own use cases. In this instance, even though it's possible to create a standalone policy, doing it inline makes more sense since the policy directly refers to an RDS instance and is [the encouraged way](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.IAMPolicy.html). So, at some point we want inline option, too. But I can do with normal managed policy for now.

negz:
> Also, should we drop IAM prefix from the name of both IAM resources? We already have identity in the group name. Maybe change identity group name to iam to match the API?

I mentioned this seemed like a good idea earlier, but now I'm thinking we can probably leave it as is. It seems like https://aws.amazon.com/iam/ is one service of https://aws.amazon.com/identity/, which also include SSO, Cognito, etc. I also suspect including IAM in the CR names will help folks, because you don't usually include the group in `kubectl` commands, and `kubectl get iamrole`, `kubectl get iamrolepolicyattachment` probably makes more sense to AWS users than `kubectl get role`, etc.

muvaf:
> I mentioned this seemed like a good idea earlier, but now I'm thinking we can probably leave it as is. It seems like https://aws.amazon.com/iam/ is one service of https://aws.amazon.com/identity/, which also include SSO, Cognito, etc.

I have a feeling that _identity_ is merely a marketing term. The [grouping in the SDK](https://github.com/aws/aws-sdk-go-v2/tree/master/service) seems like a better candidate to use in our groupings. You can see `cognito` and `iam` on the same level of _services_ there without `identity` encapsulating them. Note that adopting SDK grouping has some nice implications as well such as reuse of some structs in our structs. For example `ec2.Tag` is used by all managed resources in ec2 group or `rds.Tag` is used by all RDS-related stuff.

> I also suspect including IAM in the CR names will help folks, because you don't usually include the group in kubectl commands, and kubectl get iamrole, kubectl get iamrolepolicyattachment probably makes more sense to AWS users than kubectl get role, etc.

Yeah, `kubectl get role` does have an overlap. I'm not sure whether it'd be weird to make group `iam` and still call the role `iamrole`.

FWIW, you don't have to include the full group when there is collision. `kubectl get provider.gcp` and `kubectl get provider.aws` both return the right results. Maybe we could get away with using `kubectl get role.iam` to not have collision with `kubectl get role (kubernetes Role resource)`.

PR #155, opened 2020-03-16
External tagging for RDS, ReplicationGroup and VPC

muvaf:

Implements external tagging for RDS, ReplicationGroup and VPC. I was going to implement it only for RDS and ReplicationGroup but then realized it's really painful to find lost VPCs since their identifiers are generated by AWS, so, I implemented for it as well.

### Checklist

I have:


pkg/controller/cache/managed.go
```diff
@@ -222,6 +223,31 @@ func (e *external) Delete(ctx context.Context, mg resource.Managed) error {
 	return errors.Wrap(resource.Ignore(elasticache.IsNotFound, err), errDeleteReplicationGroup)
 }
 
+type tagger struct {
+	kube client.Client
+}
+
+func (t *tagger) Initialize(ctx context.Context, mg resource.Managed) error {
```
hasheddan: These `Initialize` functions get the tags provided in the spec and the external tags that Crossplane now sets and merges them together then updates the tags in the spec, do I have that right?

muvaf: Yes, that's what they do. Simply appending Crossplane tags to the existing ones.

hasheddan: Awesome, thanks for the clarification :)

muvaf:
@hasheddan yes, I've tested them manually.

pkg/controller/database/rdsinstance.go
```diff
@@ -211,8 +213,23 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 	if err != nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, errPatchCreationFailed)
 	}
-	modify := e.client.ModifyDBInstanceRequest(rds.GenerateModifyDBInstanceInput(meta.GetExternalName(cr), patch))
-	_, err = modify.Send(ctx)
+	_, err = e.client.ModifyDBInstanceRequest(rds.GenerateModifyDBInstanceInput(meta.GetExternalName(cr), patch)).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, errModifyFailed)
+	}
+	if len(patch.Tags) > 0 {
```
negz: Nit: This conditional block is almost half of the method. Should we return early if there are zero tags instead of nesting all of this logic?

PR #156, opened 2020-03-18
S3 bucket: move to v1beta1

sahil-lakhwani:

Move S3Bucket to v1beta1

Tested the claim with:

```
apiVersion: storage.crossplane.io/v1alpha1
kind: Bucket
metadata:
  name: app-bucket
  annotations:
    crossplane.io/external-name: external-bucket
spec:
  classSelector:
    matchLabels:
      example: "true"
      # Azure bucket claims can be satisifed by either storage accounts or
      # containers. Uncomment one of the below if you need one of either.
      # kind: account
      # kind: container
  name: crossplane-example-%s
  writeConnectionSecretToRef:
    name: bucket
```

### Checklist

I have:


pkg/controller/s3/s3bucket.go
```diff
@@ -18,249 +18,216 @@ package s3
...
+	return managed.ExternalCreation{}, errors.Wrap(err, errPolicyAttach)
 }
 
-func (r *Reconciler) _delete(bucket *bucketv1alpha3.S3Bucket, client s3.Service) (reconcile.Result, error) {
-	bucket.Status.SetConditions(runtimev1alpha1.Deleting(), runtimev1alpha1.ReconcileSuccess())
-	if bucket.Spec.ReclaimPolicy == runtimev1alpha1.ReclaimDelete {
-		if err := client.DeleteBucket(bucket); err != nil {
-			return r.fail(bucket, err)
-		}
+func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
```
muvaf: In `Update`, you should make the `Put` call directly. If they do what's expected from a `PUT`, i.e. update or create if not exists, we shouldn't need to worry about patches or differences etc. We'd call `PUT` in any case and move on. Though this needs manual check, it might return `AlreadyExists` error if it's already there.

pkg/controller/s3/s3bucket.go
```diff
@@ -18,249 +18,216 @@ package s3
...
-	if bucket.Status.IAMUsername == "" {
-		return r.fail(bucket, errors.New("username not set, .Status.IAMUsername"))
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mg.(*v1beta1.S3Bucket)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
-	bucketInfo, err := client.GetBucketInfo(bucket.Status.IAMUsername, bucket)
+
+	name := meta.GetExternalName(cr)
```
muvaf: I think you can use `meta.GetExternalName(cr)` directly instead of creating a variable `name`.

pkg/controller/s3/s3bucket.go
```diff
@@ -18,249 +18,216 @@ package s3
...
 	}
 
-	if bucketInfo.Versioning != bucket.Spec.Versioning {
-		err := client.UpdateVersioning(bucket)
-		if err != nil {
-			return r.fail(bucket, err)
-		}
-	}
+	policy := e.getBucketPolicy(meta.GetExternalName(cr))
+
```
muvaf: It's a bit of personal preference, but I don't think empty lines help the reader understand the code better unless they separate steps that each consist of bulk of calls.

apis/storage/v1beta1/bucket_types.go
```diff
@@ -55,41 +46,38 @@ type S3BucketParameters struct {
 	// provisioning.
 	// +kubebuilder:validation:Enum=Read;Write;ReadWrite
 	LocalPermission *storagev1alpha1.LocalPermissionType `json:"localPermission"`
```
muvaf: `LocalPermission` doesn't seem to be included in any call. I'd consider removing it for now until we get to implement those other subresources.

apis/storage/v1beta1/bucket_types.go
```diff
@@ -55,41 +46,38 @@ type S3BucketParameters struct {
 	// provisioning.
```
muvaf: `Versioning` doesn't seem to be included in any call. I'd consider removing it for now until we get to implement those other subresources.

pkg/clients/s3/s3.go
```diff
@@ -18,262 +18,96 @@ package s3
...
-func (c *Client) UpdateVersioning(bucket *v1alpha3.S3Bucket) error {
-	versioningStatus := s3.BucketVersioningStatusSuspended
-	if bucket.Spec.Versioning {
-		versioningStatus = s3.BucketVersioningStatusEnabled
-	}
-	name := bucket.GetBucketName()
-	input := &s3.PutBucketVersioningInput{Bucket: &name, VersioningConfiguration: &s3.VersioningConfiguration{Status: versioningStatus}}
-	_, err := c.s3.PutBucketVersioningRequest(input).Send(context.TODO())
+// IsUpToDate checks if the current spec is in sync with the observed bucket properties.
+func IsUpToDate(p v1beta1.S3BucketParameters, policy string) (bool, error) {
```
muvaf: Normally, what `IsUpToDate` does is that it checks `*Parameters` struct with actual cloud resource struct. However, in this case we have one-to-many relation between `*Parameters` struct and sub-resources. The easiest solution would be to pass in all those resources as arguments to this function, but this would bloat the function signature. So, I think doing the actual `GET` call of each resource and comparing it against `p v1beta1.S3BucketParameters` would make sense for this case.

Right now, what it'd do is to `GET` the policy and compare it with the policy that `p` has and return the result. I'd not use `CreatePatch` as the logic is simple enough.

sahil-lakhwani: Right.
I created the `CreatePatch` method considering there will be more properties to added.

pkg/controller/s3/s3bucket.go
sahil-lakhwani: Okay, I checked that `PutBucketPolicyRequest` doesn't return any error for already existing policy. So we can call it without a check. 
Just that a local change can save an API call. And there will be more API calls for more fields we add, because there is no single consolidated call for S3 bucket properties.

negz:
The current v1alpha1 S3 bucket controller implicitly creates an IAM user and grants that user access to the S3 bucket it creates. This is a legacy pattern that goes against our current design practices. I see that this v1beta1 implementation follows our best practices by not creating an IAM user; the idea being we'd instead model IAM users as a distinct managed resource.

At the time of writing we don't yet have support for an `IAMUser` managed resource in Crossplane. When we do add support we'll need to land https://github.com/crossplane/crossplane/pull/1163 in order to enable the v1alpha1 S3 functionality in which a `Bucket` claim creates both an S3 bucket and an IAM user.

This all means that we will either need to hold off on merging this v1beta1 PR until https://github.com/crossplane/crossplane/pull/1163 is implemented, or merge this PR and tolerate the fact than an IAM user is not automatically provisioned.

prasek:
@negz looks like [S3Bucket](https://doc.crds.dev/github.com/crossplane/provider-aws/storage.aws.crossplane.io/S3Bucket/v1alpha3@v0.11.1) is still `v1alpha3`, so keeping this open for now.

PR #160, opened 2020-03-21
security group: implement missing fields on ingress and egress.

muvaf:

* Implement some of the missing fields.
* Move ingress & egress rule creation to `Update`. The reasoning for this is that if the creation of the security group succeeds but any of the rule creation calls in `Create` fails, we'd not detect it in `Observe` during existence check since it only looks at whether a security group with given id exists. Since we never make those calls again user won't know they failed either. Expanding that existence check to include checking each rule as well but I thought we will move these `AuthorizeSecurityGroupEgressRequest` calls to `Update` anyway when `Update` is implemented.
* Use external name annotation instead of ID in the status.

### Checklist

I have:


pkg/clients/ec2/securitygroup.go
```diff
@@ -34,3 +35,45 @@ func IsSecurityGroupNotFoundErr(err error) bool {
 	}
 	return false
 }
+
+// GenerateEC2Permissions converts object Permissions to ec2 format
+func GenerateEC2Permissions(objectPerms []v1alpha3.IPPermission) []ec2.IpPermission {
+	var permissions []ec2.IpPermission
```
golangcibot: Consider preallocating `permissions` (from `prealloc`)

apis/network/v1alpha3/securitygroup_types.go
```diff
@@ -45,68 +45,174 @@ func (v *VPCIDReferencerForSecurityGroup) Assign(res resource.CanReference, valu
...
 	}
 
-	sg.Spec.VPCID = value
+	sg.Spec.VPCID = aws.String(value)
 	return nil
 }
 
+// SecurityGroupParameters define the desired state of an AWS VPC Security
+// Group.
+type SecurityGroupParameters struct {
```
hasheddan: I had a bit of trouble following the changes in this file due to how github is showing the diff, but it this still additive or should these changes require a bump to `v1alpha4`?

pkg/controller/network/securitygroup/controller.go
```diff
@@ -152,54 +144,54 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
...
 	}
 
-	return managed.ExternalCreation{}, nil
+	// NOTE(muvaf): GroupID is used as external name instead of GroupName because
+	// there are cases where only GroupID is accepted as identifier.
+	meta.SetExternalName(cr, aws.StringValue(result.GroupId))
+	return managed.ExternalCreation{}, errors.Wrap(e.kube.Update(ctx, cr), errSpecUpdate)
 }
 
 func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
```
hasheddan: Just want to verify that if we add `Ingress` and `Egress` rules in `Update` rather than `Create` that the `SecurityGroup` will be created "locked down" by default? Basically I just don't want us to be creating a more permissive security group than the user requests and then updating it. Which I guess could be possible if the user requests something that is more restrictive than the default 🤔 I believe that the default allows inbounf traffic from other instances in the SG, and all outbound traffic https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-security-groups.html#default-security-group

apis/network/v1alpha3/securitygroup_types.go
muvaf: It should be additive. I'm testing it with Wordpress stack that has v1alpha3 `SecurityGroup` resources.

pkg/controller/network/securitygroup/controller.go
muvaf: The section you linked describes the default SG that you get when you create a `VPC`. I interpreted the next section as what we are concerned; [what rules are default when I create a custom security group](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-security-groups.html#creating-your-own-security-groups). It seems like it doesn't allow any inbound traffic but allows all outbound traffic. I feel like this is restrictive enough for a default, what do you think?

muvaf:
@hasheddan Confirmed this is working with Wordpress. I've been going crazy about why this isn't working and apparently the problem was some limit on AWS :( Changed the region and it worked as usual. I couldn't figure out to see actual error anywhere; it's probably CloudFormation created resource that couldn't provision instance. Anyway, it's ready to merge.

pkg/clients/ec2/securitygroup.go
```diff
@@ -34,3 +39,58 @@ func IsSecurityGroupNotFoundErr(err error) bool {
 	}
 	return false
 }
+
+// IsRuleAlreadyExistsErr returns true if the error is because the rule already exists.
+func IsRuleAlreadyExistsErr(err error) bool {
```
negz: Is the `InvalidPermissionDuplicate` code used elsewhere? Should this be `IsInvalidPermissionDuplicateErr` in case we want to use it for other AWS errors with this code?

pkg/controller/network/securitygroup/controller.go
```diff
@@ -86,47 +89,33 @@ func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (manag
...
-	})
-
-	response, err := req.Send(ctx)
-
-	if ec2.IsSecurityGroupNotFoundErr(err) {
-		return managed.ExternalObservation{
-			ResourceExists: false,
-		}, nil
+	if meta.GetExternalName(cr) == "" {
+		return managed.ExternalObservation{}, nil
```
negz: Nit: I preferred the existing pattern of explicitly rather than implicitly returning `ResourceExists: false`, because it documents the intent of the return a little better.

pkg/controller/network/securitygroup/controller.go
negz: I can see @hasheddan's concern, but isn't this inherent to the AWS API? It seems like the API requires us to first create the group then modify its rules (otherwise we'd just do it as one create API call). So the question here is more about _how long_ the security group will be using the default set of rules before we update it. Previously we did both in the same `Create` call, and now we'll do it in `Update`. Keep in mind this probably doesn't delay the calls very much, because we'll be immediately requeued after a creation, notice we need an update, and apply the missing rules.

pkg/controller/network/securitygroup/controller.go
```diff
@@ -152,54 +140,54 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
...
+	}
 	// TODO(soorena776): add more sophisticated Update logic, once we
 	// categorize immutable vs mutable fields (see #727)
 
+	if len(cr.Spec.Ingress) > 0 {
+		_, err := e.sg.AuthorizeSecurityGroupIngressRequest(&awsec2.AuthorizeSecurityGroupIngressInput{
+			GroupId:       aws.String(meta.GetExternalName(cr)),
+			IpPermissions: ec2.GenerateEC2Permissions(cr.Spec.Ingress),
+		}).Send(ctx)
+		if err != nil && !ec2.IsRuleAlreadyExistsErr(err) {
```
negz: Nit: You could probably use `if resource.Ignore(ec2.IsRuleAlreadyExistsErr, err) != nil { }`

negz:
@muvaf Looks like a unit test is failing. Could you take a look before we merge?

pkg/clients/ec2/securitygroup.go
muvaf: It's not used elsewhere but I think having `AlreadyExists` in the name sounds more familiar because actual error changes from one resource to another.

pkg/controller/network/securitygroup/controller.go
muvaf: I went ahead and debugged this using breakpoints. What's happening is that there is no inbound rules and only one outbound rule by default which allows `0.0.0.0`. Even though their API is designed that way, what we do is only **addition**. So, there isn't any insecure->secure switch between `Create`->`Update` for inbound rules. For outbound, it's the same as before; we get default `0.0.0.0` rule and leave it there because we don't know about it and API is additive.

PR #161, opened 2020-03-23
DynamoDB v1alpha1

sahil-lakhwani:

Adds resource DynamoDB Table in alpha1 version.

### Checklist

I have:


apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,259 @@
...
+// StreamSpecification specifies the settings for DynamoDB Streams on the table.
+type StreamSpecification struct {
+
+	// Indicates whether DynamoDB Streams is enabled (true) or disabled (false)
+	// on the table.
+	StreamEnabled *bool `json:"streamEnabled,omitempty"`
+
+	// When an item in the table is modified, StreamViewType determines what information
+	// is written to the stream for this table.
+	StreamViewType string `json:"StreamViewType,omitempty"`
```
hasheddan, line 137: Optional fields should be pointer types (i.e. `*string` here).

muvaf:
I think Crossplane does not necessarily have opinions on the level of the service that the provider supplies. Until now, it's been on machine level, i.e. RDS instance was an _instance_ or EKS was a set of instances. We implemented these resources first because we wanted to enable the core scenarios. However, we don't necessarily have an opinion about their level. For example, we implemented RDS instance, but if you look at the API, we could possibly implement `RDSDatabase`, which is one database in an RDS instance. The former enabled more scenarios but it does not necessarily mean Crossplane provides the lowest level of a service.

In DynamoDB case, if all they allow is to create a table, then the managed resource could do that. It wouldn't be `DynamoDBInstance` but `DynamoDBTable` or directly `DynamoDB` since its users already know that it can only be a table.

IMO, as Crossplane matures to enable more use-cases we'll see wide variety of services implemented. From SaaS providers(like MongoDb) to different levels of service in the same cloud API. For example, I can see a case where we have `RDSDatabase` or `KubernetesNamespace` whose provider is simply a kubeconfig. So, I think as long as we name the resource correctly and the published secret has enough information to consume the resources, it should be doable.

> Creating database tables feels more like an application-level operation, so I am not sure we should support it as a managed resource. @muvaf and I were recently discussing how to handle these types of services. It seems that main thing you would want from an application perspective is for credentials to be provided to be able to connect to Dynamo.

IMO the main thing an application needs is to have the creds mounted but what level of service it needs is up to developer. Wordpress creates a database if it doesn't exist, but, it can also use an existing one. I might want to mount only a database creds for access as opposed to whole MySQL instance creds. So, in general, what I'd advocate for is that let's allow app devs to have options to decide.

sahil-lakhwani:
@hasheddan Just to mention here, DynamoDB accesses work only with IAM. Operations can be authenticated per table (even per column), as per the policy.

negz:
FWIW I agree with @muvaf here - I think we should have this support even if many folks will prefer to create their tables in code.

pkg/clients/dynamodb/dynamodb.go
```diff
@@ -0,0 +1,315 @@
...
+				Projection: &v1alpha1.Projection{
+					NonKeyAttributes: val.Projection.NonKeyAttributes,
+					ProjectionType:   string(val.Projection.ProjectionType),
+				},
+			}
+		}
+		return localSecondaryIndexes
+	}
+	return nil
+}
```
negz, line 315: Nitpick for all of these these `build<T>` functions - it would be nice to 'return early' to move the happy path out of the nested block:

```go
if len(indexes) == 0 {
    return nil
}
localSecondaryIndexes := ...
// etc
return localSecondaryIndexes.
```

apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// Represents the settings used to enable server-side encryption.
+	// +optional
+	SSESpecification *SSESpecification `json:"sseSpecification,omitempty"`
+
+	// The stream settings for DynamoDB Streams on the table. These settings consist of:
+	// +optional
+	StreamSpecification *StreamSpecification `json:"streamSpecification,omitempty"`
+
+	// A list of key-value pairs to label the table.
```
negz, line 195:
```suggestion
	// A list of key-value pairs to label the table.
	// +optional
```

apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,281 @@
...
+// DynamoTableObservation keeps the state for the external resource
+type DynamoTableObservation struct {
+
+	// An array of AttributeDefinition objects. Each of these objects describes
+	// one attribute in the table and index key schema.
+	AttributeDefinitions []AttributeDefinition `json:"attributeDefinitions,omitempty"`
+
+	// // The date and time when the table was created, in UNIX epoch time (http://www.epochconverter.com/)
+	// // format.
+	// CreationDateTime *time.Time `json:"creationDateTime,omitempty"`
```
negz, line 214:
```suggestion
```

apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,281 @@
...
+	TableID string `json:"tableId,omitempty"`
+
+	// The current state of the table:
+	TableStatus string `json:"tableStatus,omitempty"`
+
+	// Unique identifier for the table for which the backup was created.
+	TableName string `json:"tableName,omitempty"`
+}
+
+// A DynamoTableStatus represents the observed state of a DBSubnetGroup.
```
negz, line 250:
```suggestion
// A DynamoTableStatus represents the observed state of a DynamoDB Table.
```

apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,281 @@
...
+type DynamoTableStatus struct {
+	runtimev1alpha1.ResourceStatus `json:",inline"`
+	AtProvider                     DynamoTableObservation `json:"atProvider"`
+}
+
+// +kubebuilder:object:root=true
+
+// A DynamoTable is a managed resource that represents an AWS DynamoDB Table
+// +kubebuilder:printcolumn:name="TABLE_NAME",type="string",JSONPath=".status.atProvider.tableName"
+// +kubebuilder:printcolumn:name="TABLE_STATUS",type="string",JSONPath=".status.atProvider.tableStatus"
```
negz, line 260:
```suggestion
// +kubebuilder:printcolumn:name="TABLE-NAME",type="string",JSONPath=".status.atProvider.tableName"
// +kubebuilder:printcolumn:name="TABLE-STATUS",type="string",JSONPath=".status.atProvider.tableStatus"
```

config/crd/database.aws.crossplane.io_dynamotables.yaml
```diff
@@ -0,0 +1,694 @@
...
+  creationTimestamp: null
+  name: dynamotables.database.aws.crossplane.io
+spec:
+  additionalPrinterColumns:
+  - JSONPath: .status.atProvider.tableName
+    name: TABLE_NAME
+    type: string
+  - JSONPath: .status.atProvider.tableStatus
+    name: TABLE_STATUS
+    type: string
```
negz, line 17:
```suggestion
  - JSONPath: .status.atProvider.tableName
    name: TABLE-NAME
    type: string
  - JSONPath: .status.atProvider.tableStatus
    name: TABLE-STATUS
    type: string
```

apis/database/v1alpha1/dynamodb_types.go
```diff
@@ -0,0 +1,281 @@
...
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
```
negz, line 262: I like this convention of including the READY and SYNCED conditions. 👍

config/crd/database.aws.crossplane.io_dynamotables.yaml
```diff
@@ -0,0 +1,694 @@
...
+              required:
+              - name
+              - namespace
+              type: object
+          required:
+          - forProvider
+          - providerRef
+          type: object
+        status:
+          description: A DynamoTableStatus represents the observed state of a DBSubnetGroup.
```
negz, line 432:
```suggestion
          description: A DynamoTableStatus represents the observed state of a DynamoDB Table.
```

config/crd/database.aws.crossplane.io_dynamotables.yaml
```diff
@@ -0,0 +1,694 @@
...
+              required:
+              - name
+              - namespace
+              type: object
+          required:
+          - forProvider
+          - providerRef
+          type: object
+        status:
+          description: A DynamoTableStatus represents the observed state of a DynamoDB Table.
```
negz, line 432:
```suggestion
          description: A DynamoTableStatus represents the observed state of a DynamoDB
            Table.
```

PR #162, opened 2020-03-23
Move DBSubnetGroup to v1beta1

gauravgahlot:

- Moved DBSubnetGroup implementation from `v1alpha3` to `v1beta1`. 
- Added implementation for DBSubnetGroup update

Tested with the following:

```
kind: DBSubnetGroup
metadata:
  name: sample-dbsubnetgroup
  annotations:
    crossplane.io/external-name: managed-resource-dbsubnetgroup
spec:
  forProvider:
    groupName: my-cool-dbsubnetgroup
    description: EKS vpc to rds
    subnetIdRefs:
      - name: sample-subnet1
      - name: sample-subnet2
      - name: sample-subnet3
    tags:
      - key: name
        value: my-cool-dbsubnetgroup
  reclaimPolicy: Delete
  providerRef:
    name: aws-provider
```

 
### Checklist
I have:
- [x] Ensured this PR contains a neat, self-documenting set of commits.
- [x] Updated any relevant [documentation], examples, or release notes.

[documentation]: https://github.com/crossplane/crossplane/pull/1360


pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -129,21 +130,21 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 }
```
hasheddan: In the `ExternalObservation` above I believe we are going to always be calling for an `Update` because `ResourceUpToDate` will default to `false` (ref: https://github.com/crossplane/provider-aws/blob/a7a3b6ba7d29821a7e84e304f043ec462a1ceb8a/pkg/controller/database/rdsinstance.go#L160). This was fine before because we had a no-op update, but now it looks as though we are performing actual updates, which we won't want to do unless they are needed. We will need to compare the observed DB subnet group with the spec and determine if we need an update, then provide the appropriate value to `ResourceUpToDate`. Let me know if you have any questions :)

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -162,7 +154,7 @@ func (b *DBSubnetGroup) UpdateExternalStatus(observation rds.DBSubnetGroup) {
 		}
```
muvaf: I think the fields in spec and status need some changes. You can take a look at our API patterns doc to get a grasp of our conventions around the fields https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -129,21 +130,21 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
 
 	cr.Status.SetConditions(runtimev1alpha1.Creating())
 
 	input := &awsrds.CreateDBSubnetGroupInput{
-		DBSubnetGroupDescription: aws.String(cr.Spec.DBSubnetGroupDescription),
-		DBSubnetGroupName:        aws.String(cr.Spec.DBSubnetGroupName),
-		SubnetIds:                cr.Spec.SubnetIDs,
+		DBSubnetGroupDescription: aws.String(cr.Spec.ForProvider.DBSubnetGroupDescription),
+		DBSubnetGroupName:        aws.String(cr.Spec.ForProvider.DBSubnetGroupName),
+		SubnetIds:                cr.Spec.ForProvider.SubnetIDs,
```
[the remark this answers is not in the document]
gauravgahlot: Hey @muvaf, the `rds` package already has a declaration for `Tag`. If we want to make Tags part of `DBSubnetGroup`, we will need to declare it with a different name. Please let me know if I'm misunderstanding this.

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -82,6 +88,7 @@ type DBSubnetGroupParameters struct {
 	DBSubnetGroupDescription string `json:"description"`
 
 	// The name for the DB subnet group. This value is stored as a lowercase string.
+	// +immutable
 	DBSubnetGroupName string `json:"groupName"`
```
muvaf, line 92: We use `meta.GetExternalName(cr)` when we need the name, so, this field should be removed now.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,206 @@ package dbsubnetgroup
...
-
-		return managed.ExternalObservation{}, errors.Wrapf(err, errDescribe, cr.Spec.DBSubnetGroupName)
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(dbsg.IsErrorNotFound, err), errDescribe)
 	}
 
 	// in a successful response, there should be one and only one object
-	if len(response.DBSubnetGroups) != 1 {
-		return managed.ExternalObservation{}, errors.Errorf(errMultipleItems, cr.Spec.DBSubnetGroupName)
+	if len(res.DBSubnetGroups) != 1 {
+		return managed.ExternalObservation{}, errors.Errorf(errMultipleItems, cr.Spec.ForProvider.DBSubnetGroupName)
```
muvaf, line 130: if statements holds true when there are any case where the len is not 1, which includes 0 but the error only says there are multiple items. It could be something along the lines `recevied 0 or more than one resources`

pkg/clients/dbsubnetgroup/dbsubnetgroup.go
```diff
@@ -0,0 +1,139 @@
...
+// IsDBSubnetGroupNotFoundErr returns true if the error is because the item doesn't exist
+func IsDBSubnetGroupNotFoundErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), rds.ErrCodeDBSubnetGroupNotFoundFault)
+}
+
+// IsDBSubnetGroupUpToDate checks whether there is a change in any of the modifiable fields.
+func IsDBSubnetGroupUpToDate(p v1beta1.DBSubnetGroupParameters, sg rds.DBSubnetGroup) (bool, error) {
```
muvaf, line 61: There are only 2 fields that might need update since `Modify` call takes only those. So, it might be easier just to check `DBSubnetGroupDescription` and `SubnetIDs` fields and see whether they changed and return.

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -82,6 +88,7 @@ type DBSubnetGroupParameters struct {
 	DBSubnetGroupDescription string `json:"description"`
```
muvaf, line 88: The name of this field could be `Description`

pkg/controller/database/dbsubnetgroup/controller.go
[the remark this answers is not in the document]
muvaf: If I recall correctly, in my comment, I was saying that `DBSubnetGroupParameters` should have a `Tags` field and it seems to be the case now.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,206 @@ package dbsubnetgroup
...
-	for _, t := range cr.Spec.Tags {
-		input.Tags = append(input.Tags, awsrds.Tag{
-			Key:   aws.String(t.Key),
-			Value: aws.String(t.Value),
-		})
+	input := &awsrds.CreateDBSubnetGroupInput{
+		DBSubnetGroupDescription: aws.String(cr.Spec.ForProvider.DBSubnetGroupDescription),
+		DBSubnetGroupName:        aws.String(meta.GetExternalName(cr)),
+		SubnetIds:                cr.Spec.ForProvider.SubnetIDs,
+		Tags:                     []awsrds.Tag{},
```
muvaf, line 182: An empty array of tags is sent here. A `for` similar to the existing one would do the job.

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -151,6 +159,21 @@ type DBSubnetGroupList struct {
 	Items           []DBSubnetGroup `json:"items"`
 }
 
+// DBSubnetGroupObservation is the representation of the current state that is observed
+type DBSubnetGroupObservation struct {
+	// SubnetGroupStatus specifies the current state of this DB subnet group.
+	SubnetGroupStatus string `json:"subnetGroupStatus,omitempty"`
+
+	// DBSubnetGroupArn is the Amazon Resource Name (ARN) for this DB subnet group.
+	DBSubnetGroupArn string `json:"dbSubnetGroupArn,omitempty"`
```
muvaf, line 168: I think `state` and `arn` would be sufficient for names as it's already implied with these fields being under DBSubnetGroup CRD. I see that they are indeed named as `DBSubnetGroupArn` and `SubnetGroupStatus` in cloud API but IMO we should go with `state` and `arn` as it's really unnecessary.

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -112,23 +120,23 @@ type DBSubnetGroupExternalStatus struct {
...
 // +kubebuilder:object:root=true
 
 // A DBSubnetGroup is a managed resource that represents an AWS VPC Database
 // Subnet Group.
-// +kubebuilder:printcolumn:name="GROUPNAME",type="string",JSONPath=".spec.groupName"
-// +kubebuilder:printcolumn:name="DESCRIPTION",type="string",JSONPath=".spec.description"
-// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.groupStatus"
+// +kubebuilder:printcolumn:name="GROUPNAME",type="string",JSONPath=".spec.forProvider.groupName"
+// +kubebuilder:printcolumn:name="DESCRIPTION",type="string",JSONPath=".spec.forProvider.description"
+// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.atProvider.groupStatus"
```
muvaf, line 139:
```suggestion
// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.atProvider.subnetGroupStatus"
```

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -112,23 +120,23 @@ type DBSubnetGroupExternalStatus struct {
...
 }
 
 // +kubebuilder:object:root=true
 
 // A DBSubnetGroup is a managed resource that represents an AWS VPC Database
 // Subnet Group.
-// +kubebuilder:printcolumn:name="GROUPNAME",type="string",JSONPath=".spec.groupName"
-// +kubebuilder:printcolumn:name="DESCRIPTION",type="string",JSONPath=".spec.description"
-// +kubebuilder:printcolumn:name="STATUS",type="string",JSONPath=".status.groupStatus"
+// +kubebuilder:printcolumn:name="GROUPNAME",type="string",JSONPath=".spec.forProvider.groupName"
```
muvaf, line 137:
```suggestion
```

pkg/clients/dbsubnetgroup/dbsubnetgroup.go
```diff
@@ -0,0 +1,139 @@
...
+	patch, err := CreatePatch(&sg, &p)
+	if err != nil {
+		return false, err
+	}
+	return cmp.Equal(&v1beta1.DBSubnetGroupParameters{}, patch, cmpopts.IgnoreInterfaces(struct{ resource.AttributeReferencer }{})), nil
+}
+
+// CreatePatch creates a *v1beta1.DBSubnetGroupParameters that has only the changed
+// values between the target *v1beta1.DBSubnetGroupParameters and the current *rds.DBSubnetGroup
+func CreatePatch(in *rds.DBSubnetGroup, target *v1beta1.DBSubnetGroupParameters) (*v1beta1.DBSubnetGroupParameters, error) {
```
muvaf, line 71: I don't see any usage of `func CreatePatch` other than is-up-to-date check, so, it'd be better if you can remove it after making the is-up-to-date simpler.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,206 @@ package dbsubnetgroup
...
-	// TODO(soorena776): add more sophisticated Update logic, once we
-	// categorize immutable vs mutable fields (see #727)
+	_, err := e.client.ModifyDBSubnetGroupRequest(&awsrds.ModifyDBSubnetGroupInput{
+		DBSubnetGroupName:        aws.String(meta.GetExternalName(cr)),
+		DBSubnetGroupDescription: aws.String(cr.Spec.ForProvider.DBSubnetGroupDescription),
+		SubnetIds:                cr.Spec.ForProvider.SubnetIDs,
+	}).Send(ctx)
 
-	return managed.ExternalUpdate{}, nil
+	return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(dbsg.IsErrorAlreadyExists, err), errUpdate)
```
muvaf, line 205: It's possible to update the tags as well but it requires another type of call `AddTagsToResourceRequest`, you can take a look at RDS controller for an example. In that case, a patch is used but here you can just go with `cr.Spec.ForProvider.Tags` to calculate the list of tags to send. Comments in `AddTagsToResourceRequest` call says it applies to RDS resources but it looks like [DBSubnetGroup is also considered](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Tagging.html) an RDS resource.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,221 @@ package dbsubnetgroup
...
+		DBSubnetGroupDescription: aws.String(cr.Spec.ForProvider.Description),
+		DBSubnetGroupName:        aws.String(meta.GetExternalName(cr)),
+		SubnetIds:                cr.Spec.ForProvider.SubnetIDs,
 	}
 
-	req := e.client.CreateDBSubnetGroupRequest(input)
+	if len(cr.Spec.ForProvider.Tags) != 0 {
+		input.Tags = make([]awsrds.Tag, len(cr.Spec.ForProvider.Tags))
+		for i, val := range cr.Spec.ForProvider.Tags {
+			input.Tags[i] = awsrds.Tag{Key: &val.Key, Value: &val.Value}
```
muvaf, line 182:
```suggestion
			input.Tags[i] = awsrds.Tag{Key: aws.String(val.Key), Value: aws.String(val.Value)}
```

`&val.Key` refers to the address of `val.Key` string. The value of `val.Key` changes at each iteration of the for loop but its address doesn't actually change. What you end up with is that all elements of `input.Tags` pointing to the same address, whose value was assigned in the last iteration of that for loop. So, you get an array full of the same values, which results in `DuplicateKey` error. I've encountered a similar issue [here in RDS](https://github.com/crossplane/provider-aws/pull/178).

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,221 @@ package dbsubnetgroup
...
+		return managed.ExternalUpdate{}, nil
+	}
 
-	return managed.ExternalCreation{}, nil
-}
+	_, err := e.client.ModifyDBSubnetGroupRequest(&awsrds.ModifyDBSubnetGroupInput{
+		DBSubnetGroupName:        aws.String(meta.GetExternalName(cr)),
+		DBSubnetGroupDescription: aws.String(cr.Spec.ForProvider.Description),
+		SubnetIds:                cr.Spec.ForProvider.SubnetIDs,
+	}).Send(ctx)
```
muvaf, line 204: The `err` returned from `ModifyDBSubnetGroupRequest` is not handled here.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -164,22 +165,30 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
...
 
 func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
-	// TODO(soorena776): add more sophisticated Update logic, once we
-	// categorize immutable vs mutable fields (see #727)
+	cr, ok := mgd.(*v1beta1.DBSubnetGroup)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	_, err := e.client.ModifyDBSubnetGroupRequest(&awsrds.ModifyDBSubnetGroupInput{
```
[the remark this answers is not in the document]
muvaf: Oh what I meant here was something similar to `AlreadyExists`, like `NoModificationInModifyRequest` or something like that, not literally `AlreadyExists` error.

You can test creating the resource and making it call this function twice and see whether it returns any error where modification request doesn't actually change anything. If it doesn't return an error, then you don't need to skip anything, including `AlreadyExists`.

pkg/clients/dbsubnetgroup/dbsubnetgroup.go
muvaf: This comment assumed that we were not able to update the tags but that's actually possible later I realized. So, you need to check tags as well.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,221 @@ package dbsubnetgroup
...
+	dbsg.LateInitialize(&cr.Spec.ForProvider, &observed)
+	if !reflect.DeepEqual(current, &cr.Spec.ForProvider) {
+		if err := e.kube.Update(ctx, cr); err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, errKubeUpdateFailed)
+		}
+	}
+	cr.Status.AtProvider = dbsg.GenerateObservation(observed)
+
+	switch cr.Status.AtProvider.State {
+	case v1beta1.DBSubnetGroupStateAvailable:
```
muvaf, line 144: I tested this locally and the status string I got for success was `Complete`, which does not appear in any of the state strings here. These strings look like copy of RDS state strings but every resource has its own set of state strings, so you either need to find them in AWS sdk or test locally to see what's returned.

muvaf:
Please make sure the functionality works as expected by testing manually.

pkg/controller/database/dbsubnetgroup/controller.go
muvaf: It's also fine to just have the success string (`Complete`) and let other states fall into `Unavailable` for now.

gauravgahlot: I noticed that `DBSubnetGroup` does not have any specific status type as it's there for [RDS]. So while I was testing it I kept on receiving `Complete` state. Therefore, I have updated the code to just have the success string (`Complete`) and let other states fall into `Unavailable `as you suggested.

[RDS]: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Status.html

pkg/controller/database/dbsubnetgroup/controller.go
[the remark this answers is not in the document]
gauravgahlot: Here is what I tried:
 - created the resource
 - made a change in CR
 - forced return false from dbsg.IsDBSubnetGroupUpToDate 
 - kubectl apply
As a result, it kept on sending the same parameters in `ModifyDBSubnetGroupRequest`. However, no error was received. So, I have update the code to not skip anything.

pkg/controller/database/dbsubnetgroup/controller.go
```diff
@@ -18,175 +18,215 @@ package dbsubnetgroup
...
+		}
+		_, err = e.client.AddTagsToResourceRequest(&awsrds.AddTagsToResourceInput{
+			ResourceName: aws.String(cr.Status.AtProvider.ARN),
+			Tags:         tags,
+		}).Send(ctx)
+		if err != nil {
+			return managed.ExternalUpdate{}, errors.Wrap(err, errAddTagsFailed)
+		}
+	}
+	return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
```
muvaf, line 218: Nitpick: `err` is from `ModifyDBSubnetGroupRequest`, which is irrelevant at this point. But it's same functionally, since that `err` is `nil` at that point anyway.

apis/database/v1beta1/dbsubnetgroup_referencers_test.go
```diff
@@ -167,7 +166,8 @@ func TestDBSubnetGroupNameReferencerBuild(t *testing.T) {
 			input: input{
 				readerFn: func(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
 					p := obj.(*DBSubnetGroup)
-					p.Spec.DBSubnetGroupName = mockDBSubnetGroupName
+					// p.Spec.ForProvider.DBSubnetGroupName = mockDBSubnetGroupName
```
hasheddan, line 169: I think this can be removed here

apis/database/v1beta1/dbsubnetgroup_types.go
```diff
@@ -151,31 +140,17 @@ type DBSubnetGroupList struct {
...
-func (b *DBSubnetGroup) UpdateExternalStatus(observation rds.DBSubnetGroup) {
-
-	subnets := make([]Subnet, len(observation.Subnets))
-	for i, sn := range observation.Subnets {
-		subnets[i] = Subnet{
-			SubnetID:     aws.StringValue(sn.SubnetIdentifier),
-			SubnetStatus: aws.StringValue(sn.SubnetStatus),
-		}
-	}
+// DBSubnetGroupObservation is the representation of the current state that is observed
```
hasheddan, line 143: nitpick: would you mind moving this up above `DBSubnetGroupStatus` where `DBSubnetGroupExternalStatus` is currently?

pkg/clients/dbsubnetgroup/dbsubnetgroup.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/rds"
+
+	"github.com/crossplane/provider-aws/apis/database/v1beta1"
+	awsclients "github.com/crossplane/provider-aws/pkg/clients"
+)
+
+// Client is the external client used for DBSubnetGroup Custom Resource
+type Client interface { // nolint:gocyclo
```
hasheddan, line 31: nitpick: is `// nolint:gocyclo` needed here?

pkg/clients/rds/fake/dbsubnetgroup.go
```diff
@@ -19,17 +19,20 @@ package fake
 import (
```
hasheddan, line 19: nitpick: might make sense to move this file `fake/dbsubnetgroup.go` from `clients/rds` to `clients/dbsubnetgroup`

PR #164, opened 2020-03-23
add support for custom subnets in EKS

### LINKED ISSUE #163: add support for custom subnets in EKS (secondary cidr ranges)
The current EKS implementation does not allow to create pods in an additional subnet as per https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

I would like crossplane to create EKS clusters with custom network/secondary cidr ranges. Ideally I'd like to configure them like this:

```yaml
apiVersion: compute.aws.crossplane.io/v1alpha3
kind: EKSClusterClass
metadata:
  name: custom-network
  labels:
    aws: "true"
    custom-network: "true"
specTemplate:
  # [...]
  subnetIds:
    - subnet-08a6e42f696140da4
    - subnet-074d2b84fc0fba006
  customSubnetIds:
    - subnet-026c046bb4fbd1468
    - subnet-065d50facdfe89127
```

janwillies:

PoC which adds support for custom subnets/secondary cidr ranges in EKS according to https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

In a nutshell, we need to create an `ENIConfig` CR per AvailabilityZone:
```yaml
apiVersion: crd.k8s.amazonaws.com/v1alpha1
kind: ENIConfig
metadata: 
  name: us-west-2a
spec: 
  securityGroups: 
    - sg-0dff111a1d11c1c11
  subnet: subnet-011b111c1f11fdf11
```
For this we need:
- workergroup securitygroup
- custom subnet & availabilityzone tuple

`func (r *Reconciler) _customnetwork` is the entry point for this.

I'd like some comments on the approach and what's necessary to get this upstream

Fixes #163 

### Checklist

I have:


janwillies:
Thanks for the quick feedback! Challenges with `KubernetesApplication` might be
1. wait until `ENIConfig` was deployed successfully (maybe taint or restart node)
2. add env variables to existing `aws-node` daemon-set

This has to be done before pods are scheduled. I basically followed the same approach as [_awsauth()](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/compute/eks.go#L234)

janwillies:
this is a good writeup with lots of comments: https://github.com/weaveworks/eksctl/issues/1096

muvaf:
I think there could be two paths forward as we get closer to make EKS `v1beta1`, which will be accelerated when [composition](https://github.com/crossplane/crossplane/pull/1163) is implemented, which will be in latest v0.10.0:
* Implement this in EKS controller, then remove it all when we move to `v1beta1` because one of the [high fidelity](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity) goal and then write the necessary `KubernetesApplication` in a composition.
* Create a `KubernetesApplication` in the namespace in the same namespace of the `EKSCluster` connection secret. Mark `EKSCluster` as ready only if the `KubernetesApplication` completed making the patches and creations. Then when we do move to `v1beta1` EKS, the refactor is much less.

I am leaning towards the second scenario where we try to hardcode a _composition_ into `EKSCluster` controller in case CNI is enabled.

I have a feeling that the challenges you listed should be doable. @hasheddan is the one who worked on that resource recently.

> add env variables to existing aws-node daemon-set

I think `KubernetesApplication` patches the remote resource if it already exists. So, it should work in theory though we'd have to test.

> wait until ENIConfig was deployed successfully (before nodes register I believe)

If EKSCluster controller marks the cluster ready only if `KubernetesApplication` is ready in case CNI is enabled, this should be OK.

FWIW, you might need to use `KubernetesApplicationResource` which represents an individual remote cluster resource instead of `KubernetesApplication` so that you get more granular view of the `status` of the remote resource in status of `KubernetesApplicationResource`; which is not the case in `KubernetesApplication` because it includes many resources and only concerned by whether `KubernetesApplicationResource`s are submitted or not.

All this should be testable with current EKSCluster, i.e. after it's ready create the `KubernetesApplication` or set of `KubernetesApplicationResource`s that does what you want and see if it worked.

muvaf:
@janwillies do you think this PR is still relevant?

janwillies:
No, let me close this in favor of a solution on top of v1beta1 eks cluster.

Looking at v1beta1 it seems there is no way to edit Kubernetes resources in the target cluster (like `aws-auth` ConfigMap or the `aws-node` DaemonSet). I wonder what will be a good way to support this.

PR #178, opened 2020-03-28
rds: fix value copy operation of tags and process features

muvaf:

The iteration overrides the same pointer's value. I have encountered this bug during the first implementation of tags and fixed but I don't know how the change got lost :(

Manually tested, it's working.

@jbw976 I think we'll need to do a patch release of AWS provider. Since we tag the resources with default tags, RDS is unusable without this fix. I opened this PR against `release-0.7` because `master` contains some breaking changes like IAMRole being `v1beta1`, so, the patch release should only include this commit.

Fixes https://github.com/crossplane/stack-minimal-aws/issues/12

### Checklist

I have:

### DIFF
--- a/pkg/clients/rds/rds.go
+++ b/pkg/clients/rds/rds.go
@@ -120,8 +120,8 @@ func GenerateCreateDBInstanceInput(name, password string, p *v1beta1.RDSInstance
 		c.ProcessorFeatures = make([]rds.ProcessorFeature, len(p.ProcessorFeatures))
 		for i, val := range p.ProcessorFeatures {
 			c.ProcessorFeatures[i] = rds.ProcessorFeature{
-				Name:  &val.Name,
-				Value: &val.Value,
+				Name:  aws.String(val.Name),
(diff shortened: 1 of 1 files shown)


muvaf:
> @muvaf we can still cherry-pick only the commit in #179 to release-0.7 and it shouldn't cause any issues related to IAM role updates.

@hasheddan If I understood you correctly, that's what this PR does. Two PRs have the same commit but this one is against `release-0.7` and the other one is against `master`.

hasheddan:
@muvaf it is typically best practice to commit to master then git cherry-pick -x to the release branch

muvaf:
@hasheddan oh do you mean the cherry-pick should be done _after_ the commit is merged to master? That makes sense.

hasheddan:
@muvaf yep! Or at least just be explicitly cherry-picked from the `master` commit (i.e. `-x`), which it looks like you force pushed and updated :) Thanks again for fixing this!

PR #182, opened 2020-03-30
move network resources to v1beta1

sahil-lakhwani:

This PR moves all existing network resources to v1beta1

Tested examples in `examples/network`.

### Checklist

I have:


pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
 
 func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
-	cr, ok := mgd.(*v1alpha3.InternetGateway)
+	cr, ok := mgd.(*v1beta1.InternetGateway)
 	if !ok {
 		return errors.New(errUnexpectedObject)
 	}
 
-	if cr.Status.InternetGatewayID == "" {
+	if cr.Status.AtProvider.InternetGatewayID == "" {
```
muvaf, line 250: External name should be checked here and in case it's not saved, it most probably means the resource is not created in the first place. So, I think if the external name is empty, we can return with success.

pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
+	for _, a := range cr.Status.AtProvider.Attachments {
+		_, err := e.client.DetachInternetGatewayRequest(&awsec2.DetachInternetGatewayInput{
+			InternetGatewayId: aws.String(cr.Status.AtProvider.InternetGatewayID),
 			VpcId:             aws.String(a.VPCID),
-		})
+		}).Send(ctx)
 
-		if _, err := dReq.Send(ctx); err != nil {
+		if err != nil {
 			if ec2.IsInternetGatewayNotFoundErr(err) {
```
muvaf, line 264: If there is no internet gateway, we should just return with success.

pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
-			InternetGatewayId: aws.String(cr.Status.InternetGatewayID),
+	for _, a := range cr.Status.AtProvider.Attachments {
+		_, err := e.client.DetachInternetGatewayRequest(&awsec2.DetachInternetGatewayInput{
+			InternetGatewayId: aws.String(cr.Status.AtProvider.InternetGatewayID),
 			VpcId:             aws.String(a.VPCID),
-		})
+		}).Send(ctx)
 
-		if _, err := dReq.Send(ctx); err != nil {
+		if err != nil {
```
muvaf, line 263: Avoiding nested if statements is considered more idiomatic. So, maybe you can check for `NotFound` and then in a separate if statement you can check `if resource.Ignore(ec2.IsInternetGatewayNotFoundErr, err) != nil`

apis/network/v1beta1/internetgateway_types.go
```diff
@@ -43,24 +51,25 @@ func (v *VPCIDReferencerForInternetGateway) Assign(res resource.CanReference, va
...
 	}
 
-	ig.Spec.VPCID = value
+	ig.Spec.ForProvider.VPCID = value
 	return nil
 }
 
 // InternetGatewayParameters define the desired state of an AWS VPC Internet
 // Gateway.
 type InternetGatewayParameters struct {
```
muvaf, line 60: InternetGateway supports tags. You can use `CreateTagsRequest` to add tags.

muvaf:
On another note, please don't hesitate to make changes to the existing logic or decide to take some small divergences to make things simpler as you see fit.

infinitecompute:
@sahil-lakhwani @muvaf What is the checklist to get this closed out?  Can we resolve all conversations that have been addressed please.

apis/network/v1beta1/internetgateway_types.go
```diff
@@ -43,24 +51,28 @@ func (v *VPCIDReferencerForInternetGateway) Assign(res resource.CanReference, va
...
 	// VPCID is the ID of the VPC.
-	VPCID string `json:"vpcId,omitempty"`
+	VPCID string `json:"vpcId"`
```
hasheddan, line 62: I don't think this will work if only `vpcIdRef` is supplied. This API server will reject creation if `vpcId` is not supplied.

apis/network/v1beta1/securitygroup_types.go
```diff
@@ -45,21 +43,13 @@ func (v *VPCIDReferencerForSecurityGroup) Assign(res resource.CanReference, valu
...
-	// VPCID is the ID of the VPC.
-	// +optional
-	VPCID *string `json:"vpcId,omitempty"`
-
-	// VPCIDRef references to a VPC to and retrieves its vpcId
-	// +optional
-	VPCIDRef *VPCIDReferencerForSecurityGroup `json:"vpcIdRef,omitempty"`
-
 	// A description of the security group.
 	Description string `json:"description"`
```
hasheddan, line 55: `GroupName` should be removed here in favor of `ExternalName` annotation.

apis/network/v1beta1/subnet_types.go
```diff
@@ -56,48 +54,74 @@ type SubnetParameters struct {
...
+	// address.
+	// +optional
+	MapPublicIPOnLaunch *bool `json:"mapPublicIPOnLaunch,omitempty"`
+
+	// Tags represents to current ec2 tags.
+	Tags []Tag `json:"tags,omitempty"`
 
 	// VPCID is the ID of the VPC.
-	VPCID string `json:"vpcId,omitempty"`
+	VPCID string `json:"vpcId"`
```
hasheddan, line 82: Same thing here. This needs to remain `omitempty`.

pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
-		For(&v1alpha3.InternetGateway{}).
+		For(&v1beta1.InternetGateway{}).
 		Complete(managed.NewReconciler(mgr,
-			resource.ManagedKind(v1alpha3.InternetGatewayGroupVersionKind),
-			managed.WithExternalConnecter(&connector{client: mgr.GetClient(), newClientFn: ec2.NewInternetGatewayClient, awsConfigFn: utils.RetrieveAwsConfigFromProvider}),
+			resource.ManagedKind(v1beta1.InternetGatewayGroupVersionKind),
+			managed.WithExternalConnecter(&connector{client: mgr.GetClient(), newClientFn: ec2.NewInternetGatewayClient}),
 			managed.WithConnectionPublishers(),
```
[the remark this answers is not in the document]
sahil-lakhwani: @muvaf I have made this change, and taken care of the same for all resources in this PR.

pkg/clients/ec2/internetgateway.go
```diff
@@ -35,3 +44,38 @@ func IsInternetGatewayNotFoundErr(err error) bool {
...
+func IsIgUpToDate(p v1beta1.InternetGatewayParameters, ig ec2.InternetGateway) bool {
+	attachments := ig.Attachments
+
+	// if there are no attachments for obsreved IG and in spec.
+	if len(attachments) == 0 && p.VPCID == "" {
+		return true
+	}
+
+	for _, ig := range attachments {
+		if p.VPCID == aws.StringValue(ig.VpcId) {
```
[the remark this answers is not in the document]
sahil-lakhwani: That is just to find an attachment with `vpcID` which is a string .

apis/network/v1beta1/internetgateway_types.go
```diff
@@ -84,22 +93,25 @@ type InternetGatewayExternalStatus struct {
 	// The ID of the internet gateway.
```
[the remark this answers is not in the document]
sahil-lakhwani: I have made this change for all the resources in this PR.

apis/network/v1beta1/internetgateway_types.go
sahil-lakhwani: Added this.

apis/network/v1beta1/securitygroup_types.go
sahil-lakhwani: `GroupId` has been chosen for that as that is the identifier of a Security Group.

pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
+
+	response, err := e.client.DescribeInternetGatewaysRequest(&awsec2.DescribeInternetGatewaysInput{
+		InternetGatewayIds: []string{cr.Status.AtProvider.InternetGatewayID},
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(ec2.IsInternetGatewayNotFoundErr, err), errDescribe)
+	}
+
+	observed := response.InternetGateways[0]
+
```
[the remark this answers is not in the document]
sahil-lakhwani: I have refactored the code, thanks for this suggestion!

pkg/controller/network/internetgateway/controller.go
```diff
@@ -23,205 +23,255 @@ import (
...
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
 
 	cr.Status.SetConditions(runtimev1alpha1.Creating())
 
-	req := e.client.CreateInternetGatewayRequest(&awsec2.CreateInternetGatewayInput{})
-
-	ig, err := req.Send(ctx)
+	ig, err := e.client.CreateInternetGatewayRequest(&awsec2.CreateInternetGatewayInput{}).Send(ctx)
```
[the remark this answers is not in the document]
sahil-lakhwani: I think the new commit follows this.

sahil-lakhwani:
@muvaf  @hasheddan  I have addressed your comments, and taken care of them across resources. I have replied to comments which are not marked `Outdated` here, they are addressed in the code.

negz:
This PR needs a rebase due to the work we did to migrate all the controllers to support the `crossplane.io/external-name` annotation. It will likely need further rebasing once https://github.com/crossplane/provider-aws/pull/198 lands, so it might be best to rebase on that branch.

pkg/controller/network/routetable/controller.go
```diff
@@ -142,99 +158,134 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
 	cr.Status.SetConditions(runtimev1alpha1.Deleting())
 
-	// in order to delete a route table, all of its dependencies need to be
-	// deleted first
-
-	if err := e.deleteRoutes(ctx, meta.GetExternalName(cr), cr.Status.Routes); err != nil {
-		return err
-	}
-
-	if err := e.deleteAssociations(ctx, cr.Status.Associations); err != nil {
```
negz, line 208: Do we not actually need to delete the routes?

pkg/controller/network/securitygroup/controller.go
```diff
@@ -124,75 +145,104 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+	}
+
+	if patch.Ingress != nil {
+		if _, err := e.sg.AuthorizeSecurityGroupIngressRequest(&awsec2.AuthorizeSecurityGroupIngressInput{
 			GroupId:       aws.String(meta.GetExternalName(cr)),
-			IpPermissions: ec2.GenerateEC2Permissions(cr.Spec.Ingress),
-		}).Send(ctx)
-		if err != nil && !ec2.IsRuleAlreadyExistsErr(err) {
+			IpPermissions: v1beta1.BuildEC2Permissions(cr.Spec.ForProvider.Ingress),
+		}).Send(ctx); err != nil {
```
negz, line 227: It seems that previously we were ignoring errors that satisfied `ec2.IsRuleAlreadyExistsErr`. Do we no longer need to do that?

pkg/controller/network/securitygroup/controller.go
```diff
@@ -124,75 +145,104 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+	if len(patch.Tags) != 0 {
+		if _, err := e.sg.CreateTagsRequest(&awsec2.CreateTagsInput{
+			Resources: []string{meta.GetExternalName(cr)},
+			Tags:      v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags),
+		}).Send(ctx); err != nil {
+			return managed.ExternalUpdate{}, errors.Wrap(err, errCreateTags)
+		}
+	}
+
+	if patch.Ingress != nil {
```
negz, line 223: It looks like we create the security group, but never add any ingress or egress rules until update time. Does this mean there will be a brief window in which the security group will appear to be 'ready' but won't actually be allowing the traffic it is configured to allow?

pkg/controller/network/internetgateway/controller.go
```diff
@@ -97,125 +115,140 @@ type external struct {
...
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
 
-	meta.SetExternalName(cr, aws.StringValue(rsp.InternetGateway.InternetGatewayId))
-	if err := e.kube.Update(ctx, cr); err != nil {
-		return managed.ExternalCreation{}, errors.Wrap(err, errPersistExternalName)
+	meta.SetExternalName(cr, aws.StringValue(ig.InternetGateway.InternetGatewayId))
+
+	return managed.ExternalCreation{}, errors.Wrap(e.kube.Update(ctx, cr), errSpecUpdate)
```
negz, line 170: Here and elsewhere in this PR, it seems like we have a call order of:

1. Set conditions. (Modifies status)
1. Set external name. (Modifies metadata)
1. Call `e.kube.Update()`

This is all followed by a call to `e.kube.Status().Update()` made by the `managed.Reconciler`.

Will calling `Update()` before we call `Status().Update()` undo the status conditions that we set? I can never remember whether `Update()` overwrites status or not.

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
 
 func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.ExternalCreation, error) {
-	cr, ok := mgd.(*v1alpha3.VPC)
+	cr, ok := mgd.(*v1beta1.VPC)
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
 
-	// if VPCID already exists, skip creating the vpc
-	// this happens when an error has occurred when modifying vpc attributes
```
negz, line 150: Is this no longer a concern?

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
+			EnableDnsHostnames: &awsec2.AttributeBooleanValue{Value: cr.Spec.ForProvider.EnableDNSHostNames},
 		},
 	} {
-		attrReq := e.client.ModifyVpcAttributeRequest(input)
-
-		if _, err := attrReq.Send(ctx); err != nil {
+		if _, err := e.client.ModifyVpcAttributeRequest(input).Send(ctx); err != nil {
 			return managed.ExternalCreation{}, errors.Wrap(err, errModifyVPCAttributes)
 		}
 	}
```
negz, line 200: It seems like this PR mostly has a pattern of creating a resource in `Create`, then waiting until `Update` time to set any fields that are modifiable. Should we follow that pattern for VPCs?

pkg/controller/network/internetgateway/controller.go
```diff
@@ -97,125 +115,140 @@ type external struct {
...
-	aReq := e.client.AttachInternetGatewayRequest(&awsec2.AttachInternetGatewayInput{
-		InternetGatewayId: rsp.InternetGateway.InternetGatewayId,
-		VpcId:             aws.String(cr.Spec.VPCID),
-	})
+	response, err := e.client.DescribeInternetGatewaysRequest(&awsec2.DescribeInternetGatewaysInput{
+		InternetGatewayIds: []string{meta.GetExternalName(cr)},
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(ec2.IsInternetGatewayNotFoundErr, err), errDescribe)
+	}
```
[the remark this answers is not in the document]
sahil-lakhwani: It is necessary to detach IG from a VPC in order to attach to another VPC.
I think the main reason of describing IG here is to get the VPCID of the VPC to which the IG is attached to, if any.

pkg/controller/network/routetable/controller.go
sahil-lakhwani: I verified this again. Existing routes do not restrict deletion of the route table. Only associations should be deleted before deleting the route table.

pkg/controller/network/securitygroup/controller.go
sahil-lakhwani: Thanks for the catch!

pkg/controller/network/vpc/controller.go
sahil-lakhwani: After the below comment is addressed, I think this won't be needed.

apis/network/v1beta1/vpc_types.go
```diff
@@ -31,43 +30,58 @@ type VPCParameters struct {
...
+// VPCObservation keeps the state for the external resource
+type VPCObservation struct {
+	// Indicates whether the VPC is the default VPC.
+	IsDefault bool `json:"isDefault,omitempty"`
+
+	// The ID of the AWS account that owns the VPC.
+	OwnerID string `json:"ownerID,omitempty"`
+
 	// VPCState is the current state of the VPC.
 	// +kubebuilder:validation:Enum=pending;available
```
muvaf, line 64:
```suggestion
```

Since this is not a user input, i.e. under status, we can delete the validation.

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
+		InstanceTenancy: awsec2.Tenancy(aws.StringValue(cr.Spec.ForProvider.InstanceTenancy)),
+	}).Send(ctx)
 
-		cr.SetConditions(runtimev1alpha1.Creating())
-		cr.UpdateExternalStatus(*rsp.Vpc)
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
 
+	meta.SetExternalName(cr, aws.StringValue(result.Vpc.VpcId))
```
muvaf, line 184: Right after setting the external name, we should call `e.kube.Update()` so that the annotation is not lost if the next steps fail for some reason.

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
 	// in a successful response, there should be one and only one object
-	if len(rsp.Vpcs) != 1 {
+	if len(response.Vpcs) != 1 {
 		return managed.ExternalObservation{}, errors.New(errMultipleItems)
 	}
 
-	observed := rsp.Vpcs[0]
+	observed := response.Vpcs[0]
 
 	if observed.State == awsec2.VpcStateAvailable {
```
muvaf, line 145: nitpick: we usually go with `switch` instead of else/else-if.

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
+	} else if observed.State == awsec2.VpcStatePending {
+		cr.SetConditions(runtimev1alpha1.Creating())
 	}
 
-	cr.UpdateExternalStatus(observed)
+	// update the CRD spec for any new values from provider
+	current := cr.Spec.ForProvider.DeepCopy()
+	ec2.LateInitializeVPC(&cr.Spec.ForProvider, &observed)
+	if !cmp.Equal(current, &cr.Spec.ForProvider) {
+		if err := e.kube.Update(ctx, cr); err != nil {
```
muvaf, line 155: Once you call `e.kube.Update()`, the `cr` object will be populated with whatever the resulting object is in the api-server and that doesn't include `status`, i.e. all changes you make on `status` before this line gets lost. It's probably called only once or twice as late-init frequency is low but would be better if you just moved the `SetConditions` calls below `e.kube.Update`.

pkg/clients/ec2/vpc.go
```diff
@@ -39,9 +47,35 @@ func IsVPCNotFoundErr(err error) bool {
...
-// IsUpToDate returns true if there is no update-able difference between desired
+// IsVpcUpToDate returns true if there is no update-able difference between desired
 // and observed state of the resource.
-func IsUpToDate(spec v1alpha3.VPCParameters, o ec2.Vpc) bool {
-	actual := v1alpha3.BuildFromEC2Tags(o.Tags)
-	return cmp.Equal(spec.Tags, actual)
+func IsVpcUpToDate(spec v1beta1.VPCParameters, o ec2.Vpc) bool {
+	v1beta1.SortTags(spec.Tags, o.Tags)
+	actual := v1beta1.BuildFromEC2Tags(o.Tags)
+	return cmp.Equal(spec.Tags, actual) && (aws.StringValue(spec.InstanceTenancy) == string(o.InstanceTenancy))
```
muvaf, line 55: `EnableDnsSupport` and `EnableDnsHostnames` can be updated, too.

pkg/clients/ec2/vpc.go
```diff
@@ -39,9 +47,35 @@ func IsVPCNotFoundErr(err error) bool {
...
+}
+
+// GenerateVpcObservation is used to produce v1beta1.VPCObservation from
+// ec2.Vpc.
+func GenerateVpcObservation(vpc ec2.Vpc) v1beta1.VPCObservation {
+	o := v1beta1.VPCObservation{
```
muvaf, line 61: As far as I can tell, the following fields exist in `ec2.Vpc` and can be included in `VPCObservation`:
```
// Information about the IPv4 CIDR blocks associated with the VPC.
CidrBlockAssociationSet []VpcCidrBlockAssociation `locationName:"cidrBlockAssociationSet" locationNameList:"item" type:"list"`

// The ID of the set of DHCP options you've associated with the VPC (or default
// if the default options are associated with the VPC).
DhcpOptionsId *string `locationName:"dhcpOptionsId" type:"string"`

// Information about the IPv6 CIDR blocks associated with the VPC.
Ipv6CidrBlockAssociationSet []VpcIpv6CidrBlockAssociation `locationName:"ipv6CidrBlockAssociationSet" locationNameList:"item" type:"list"`
```

pkg/controller/network/vpc/controller.go
```diff
@@ -97,131 +116,127 @@ type external struct {
...
-		cr.SetConditions(runtimev1alpha1.Creating())
-		cr.UpdateExternalStatus(*rsp.Vpc)
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
 
+	meta.SetExternalName(cr, aws.StringValue(result.Vpc.VpcId))
+
 	// modify vpc attributes
 	for _, input := range []*awsec2.ModifyVpcAttributeInput{
```
muvaf, line 187: These fields can be updated after the creation. I'd suggest moving this call to `Update` method.

pkg/controller/network/subnet/controller.go
```diff
@@ -132,60 +147,120 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		cr.SetConditions(runtimev1alpha1.Creating())
```
muvaf, line 147: Same as above, it's better to make changes on `status` after the late-initialization `Update` call.

pkg/clients/ec2/subnet.go
```diff
@@ -33,3 +48,73 @@ func IsSubnetNotFoundErr(err error) bool {
...
+	patch := &v1beta1.SubnetParameters{}
+	if err := json.Unmarshal(jsonPatch, patch); err != nil {
+		return nil, err
+	}
+	return patch, nil
+}
+
+// IsSubnetUpToDate checks whether there is a change in any of the modifiable fields.
+func IsSubnetUpToDate(p v1beta1.SubnetParameters, s ec2.Subnet) (bool, error) {
+	patch, err := CreateSubnetPatch(s, p)
```
muvaf, line 115: As far as I can see, there are only 2 fields that can be changed. So, I'd just do a `switch` statement for `IsUpToDate` check and get rid of `func CreateSubnetPatch`

muvaf: Same for calling `CreateSubnetPatch` in `Update`. It's barely a few fields, so, handling them manually makes it easier to grasp the code.

pkg/controller/network/securitygroup/controller.go
```diff
@@ -124,75 +145,104 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	// indicates that it's available
 	cr.SetConditions(runtimev1alpha1.Available())
```
muvaf, line 146: Conditions should be set after late-init `Update`.

apis/network/v1beta1/securitygroup_types.go
```diff
@@ -200,28 +202,31 @@ type IPPermission struct {
 // A SecurityGroupSpec defines the desired state of a SecurityGroup.
```
muvaf, line 202: In the two `TODO`s above, it says the jsontags need to be fixed when we bump the version. Could you do that?

pkg/clients/ec2/securitygroup.go
```diff
@@ -94,3 +106,67 @@ func GenerateEC2Permissions(objectPerms []v1alpha3.IPPermission) []ec2.IpPermiss
 	}
 	return permissions
 }
+
+// GenerateSGObservation is used to produce v1beta1.SecurityGroupExternalStatus from
+// ec2.SecurityGroup.
+func GenerateSGObservation(sg ec2.SecurityGroup) v1beta1.SecurityGroupObservation {
+	o := v1beta1.SecurityGroupObservation{
```
muvaf, line 113:
```suggestion
	return v1beta1.SecurityGroupObservation{
```

pkg/controller/network/routetable/controller.go
```diff
@@ -142,99 +158,134 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		cr.SetConditions(runtimev1alpha1.Available())
 	}
```
muvaf, line 160: Why don't we need a late-init for this resource?

pkg/clients/ec2/internetgateway.go
```diff
@@ -35,3 +48,53 @@ func IsInternetGatewayNotFoundErr(err error) bool {
...
+	return v1beta1.InternetGatewayObservation{
+		InternetGatewayID: aws.StringValue(ig.InternetGatewayId),
+		Attachments:       attachments,
+		OwnerID:           aws.StringValue(ig.OwnerId),
+	}
+}
+
+// IsIgUpToDate checks whether there is a change in any of the modifiable fields.
+func IsIgUpToDate(p v1beta1.InternetGatewayParameters, ig ec2.InternetGateway) bool {
+	attachments := ig.Attachments
```
muvaf, line 83:
```suggestion
```

seems fine to just use `ig.Attachments`

pkg/clients/ec2/internetgateway.go
```diff
@@ -35,3 +48,53 @@ func IsInternetGatewayNotFoundErr(err error) bool {
...
+	upToDate := false
```
muvaf, line 84: I think just returning `false` in any comparison statements below should be fine without needing separate `upToDate` variable.

pkg/clients/ec2/routetable.go
```diff
@@ -67,3 +79,127 @@ func IsAssociationIDNotFoundErr(err error) bool {
...
+				associations[i].SubnetIDSelector = target.Routes[i].GatewayIDSelector
+			}
+		}
+		in.Associations = associations
+	}
+
+	if len(in.Tags) == 0 && len(rt.Tags) != 0 {
+		in.Tags = v1beta1.BuildFromEC2Tags(rt.Tags)
+	}
+}
```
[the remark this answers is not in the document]
muvaf: Ideally, there should be a field of the struct and we should late-init only the elements that we don't have locally. But that doesn't handle the fact that user might have wanted to delete that entry. Lately, I've been thinking that maybe we'd require user to prefix `-` in the key field value to let us know that this entry should be removed, similar to kubectl unlabel notation.

The late-init only in case the array is empty is the safe solution but there are cases where there is a constant `default` entry even if you didn't create it, so, we'd be missing that. But I think skipping if the spec array is not empty should be fine for this PR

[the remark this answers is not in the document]
muvaf: FWIW, this function is not used in `Observe`. Only in UpToDate check as far as I can see.

pkg/controller/network/internetgateway/controller.go
muvaf: > Will calling Update() before we call Status().Update() undo the status conditions that we set?

Yes. The operation that undoes the rest of the resource is not `Update` but the call itself actually. What it does when you call either `Status().Update()` or `Update()` is that it issues the respective payload (either only `status` or `metadata+spec`) and then deep copies the resulting object into what you have sent. So, the local object is the most up-to-date version that exists in the api-server, which means the payload that you didn't send is overwritten by whatever is in api-server.

We have synced with @sahil-lakhwani about this and it seems calling `Status().Update()` after the `SetConditions` makes sense because that way if the `CreateInternetGatewayRequest` call fails, the condition will reflect that the error arose during creation.

pkg/controller/network/securitygroup/controller.go
sahil-lakhwani: Have added a check for the ingress/egress rules, and then updating the status.

pkg/clients/ec2/vpc.go
sahil-lakhwani: Have added these fields

sahil-lakhwani:
@muvaf @negz I have addressed the comments. I have resolved some trivial conversations. 
Also, have made sure that the CR Status is updated before it is lost.

apis/network/v1beta1/vpc_types.go
```diff
@@ -0,0 +1,157 @@
...
+	OwnerID string `json:"ownerID,omitempty"`
+
+	// VPCState is the current state of the VPC.
+	VPCState string `json:"vpcState,omitempty"`
+
+	// Tags represents to current ec2 tags.
+	Tags []Tag `json:"tags,omitempty"`
+
+	// The ID of the VPC
+	VPCID string `json:"vpcId,omitempty"`
```
muvaf, line 122: Tags exist in spec and vpcid is the external-name, so, you can remove these from observation.

muvaf:
@hasheddan Is the current group OK for these resources or should it be `ec2.aws.crossplane.io` ?

apis/network/v1beta1/vpc_types.go
```diff
@@ -0,0 +1,157 @@
...
+// A VPCStatus represents the observed state of a VPC.
+type VPCStatus struct {
+	runtimev1alpha1.ResourceStatus `json:",inline"`
+	AtProvider                     VPCObservation `json:"atProvider"`
+}
+
+// +kubebuilder:object:root=true
+
+// A VPC is a managed resource that represents an AWS Virtual Private Cloud.
+// +kubebuilder:printcolumn:name="VPCID",type="string",JSONPath=".status.atProvider.vpcId"
```
muvaf, line 134:
```suggestion
// +kubebuilder:printcolumn:name="VPCID",type="string",JSONPath=".metadata.annotations['crossplane.io/external-name']"
```

I believe this should work but didn't test manually.

muvaf:
@hasheddan we need your approval because of your request for changes.

apis/network/v1beta1/vpc_types.go
muvaf: Here is the working form:
```
// +kubebuilder:printcolumn:name="VPCID",type="string",JSONPath=".metadata.annotations.crossplane\\.io/external-name"
```

sahil-lakhwani: Thanks @muvaf. Added this in a new commit.

muvaf:
Tested with Wordpress and it's working as expected.

Merging this and I will open a follow-up PR to move these network CRDs to `ec2` group per @hasheddan 's comment and then update AWS Sample Stack to use the new specs.

hasheddan:
@muvaf nice, thanks!

Issue #181, opened 2020-03-30
Support backup and restore of all AWS managed resources


negz:
### What problem are you facing?

When a managed resource is backed up and subsequently restored using a tool like https://velero.io the newly restored managed resource should resume reconciliation of the external resource that was created by the backed up resource. This should be true for all AWS managed resources that support the `crossplane.io/external-name` annotation. This annotation set and/or records the actual name of an external resource, causing the newly restored managed resource to use that name.

### How could Crossplane help solve your problem?

Crossplane could ensure all AWS managed resources support the `crossplane.io/external-name` annotation. I believe the following resources need updating:

* [x] `EKSCluster` (in progress https://github.com/crossplane/provider-aws/pull/184)
* [x] `DBSubnetGroup` (in progress as part of v1beta1 work in https://github.com/crossplane/provider-aws/pull/162)
* [x] `S3Bucket` (in progress https://github.com/crossplane/provider-aws/pull/185)
* [x] `InternetGateway` (in progress https://github.com/crossplane/provider-aws/pull/191)
* [x] `RouteTable` (in progress https://github.com/crossplane/provider-aws/pull/191)
* [x] `Subnet` (in progress https://github.com/crossplane/provider-aws/pull/191)
* [x] `VPC` (in progress https://github.com/crossplane/provider-aws/pull/191)

A note on S3 bucket - despite there being a v1beta1 PR in progress, it's possible that PR will block on landing composition so I think we need to hack v1alpha3 to support external names in the meantime.

muvaf:
It seems like all network stuff will be [v1beta1 soon](https://github.com/crossplane/provider-aws/pull/182) and they all use external name in the PR. I'll work on EKSCluster, which is the only one now that is not being worked on.

negz:
Heads up @sahil-lakhwani, @gauravgahlot (and CC @infinitecompute) that #162 and #182 are currently the highest priority controller PRs to get wrapped up and merged, because they'll fix this related issue. Please let me know if they become at risk of not being completed by April 10th.

infinitecompute:
@negz Is CacheSubnet Group [#186](https://github.com/crossplane/provider-aws/pull/186) intentionally excluded from this list?

negz:
@infinitecompute It is. This is a list of existing API types that don't support backup and restore. CacheSubnetGroup doesn't exist today; it's new functionality.

infinitecompute:
[...]

PR #184, opened 2020-03-31
eks: move identification data out of status

muvaf:

For backup support, `status` should be purely observational and its loss should not recur any cost, if possible. This is necessary for [backup and restore](https://github.com/crossplane/provider-aws/issues/181) workflows.

This PR makes EKS controller:
* use external name annotation for cluster name,
* save cloud formation stack id in spec,
* use version retrieved from the server instead of the one saved in status,
* save the version into status continuously instead of one time.

Fix error checking bug for case where the err is nil.

I think we can use external name for both cloud formation stack identification and also cluster name but I didn't want to touch that part; it's working and all will change in v1beta1 anyway.

Manually tested.

### Checklist

I have:

### DIFF
--- a/pkg/controller/compute/eks.go
+++ b/pkg/controller/compute/eks.go
@@ -158,7 +158,7 @@ func (r *Reconciler) _create(instance *awscomputev1alpha3.EKSCluster, client eks
 	instance.Status.SetConditions(runtimev1alpha1.Creating())
 
 	// Create Master
-	createdCluster, err := client.Create(meta.GetExternalName(instance), instance.Spec)
+	_, err := client.Create(meta.GetExternalName(instance), instance.Spec)
 	if err != nil && !eks.IsErrorAlreadyExists(err) {
 		if eks.IsErrorBadRequest(err) {
 			// If this was the first time we encountered this error we'll be
@@ -170,14 +170,8 @@ func (r *Reconciler) _create(instance *awscomputev1alpha3.EKSCluster, client eks
 		}
 		return r.fail(instance, err)
 	}
-
-	// we will need to set State.ClusterVersion it. this is needed to retrieve
-	// the right ami image for the worker nodes
-	instance.Status.ClusterVersion = createdCluster.Version
-
 	// Update status
 	instance.Status.State = awscomputev1alpha3.ClusterStatusCreating
-
 	instance.Status.SetConditions(runtimev1alpha1.ReconcileSuccess())
 
 	// We'll be requeued immediately the first time we update our status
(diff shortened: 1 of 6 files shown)


pkg/controller/compute/eks.go
```diff
@@ -428,11 +419,13 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
 
-	// Create cluster instance
-	if instance.Status.ClusterName == "" {
+	cluster, err := eksClient.Get(meta.GetExternalName(instance))
+	switch {
+	case resource.Ignore(eks.IsErrorNotFound, err) != nil:
+		return r.fail(instance, err)
+	case eks.IsErrorNotFound(err):
 		return r.create(instance, eksClient)
```
negz, line 427: Switches are processed in order, so this could probably just be:

```go
switch {
case eks.IsErrorNotFound(err):
    return r.create(instance, eksClient)
case err != nil:
    return r.fail(instance, err)
}
```

pkg/controller/compute/eks.go
```diff
@@ -273,26 +266,20 @@ func (r *Reconciler) _awsauth(cluster *eks.Cluster, instance *awscomputev1alpha3
 	return err
 }
 
-func (r *Reconciler) _sync(instance *awscomputev1alpha3.EKSCluster, client eks.Client) (reconcile.Result, error) {
-	cluster, err := client.Get(instance.Status.ClusterName)
-	if err != nil {
-		return r.fail(instance, err)
-	}
-
+func (r *Reconciler) _sync(instance *awscomputev1alpha3.EKSCluster, cluster *eks.Cluster, client eks.Client) (reconcile.Result, error) {
```
negz, line 269: Would it make sense to just pass the version here rather than the whole `*eks.Cluster`, since that's the only thing we use?

muvaf: Not sure if I'm looking at the place you're referring to. In `_sync` function, `cluster` is used by [`r.secret`](https://github.com/crossplane/provider-aws/pull/184/files#diff-cd0d23278d543d908fd906517f19a18dR310) and [`r.auth`](https://github.com/crossplane/provider-aws/pull/184/files#diff-cd0d23278d543d908fd906517f19a18dR306) calls as well as for its other properties like [`endpoint`](https://github.com/crossplane/provider-aws/pull/184/files#diff-cd0d23278d543d908fd906517f19a18dR315) and state.

muvaf:
@negz I fixed the broken test. It should be ready to merge.

PR #186, opened 2020-04-01
add cache subnet group

sahil-lakhwani:

This PR adds CacheSubnetGroup resource.
Example can be seen in `example/cache/subnet.yaml`

### Checklist

I have:


pkg/controller/cache/cachesubnetgroup/controller.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+	cr, ok := mg.(*v1alpha1.CacheSubnetGroup)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errNotSubnetGroup)
+	}
+
+	// If the list of spec.subnetIds is different from the resource subnetIds,
+	// this call will return non-nil error
```
negz, line 167: Does this mean we can effectively only update the description field? It seems odd that they allow the caller to supply a list of subnet IDs in the update call, but then return an error if they differ from the current list of subnet IDs.

pkg/clients/elasticache/elasticache.go
```diff
@@ -368,3 +375,23 @@ func isErrorCodeEqual(errorCode string, err error) bool {
...
+					break
+				} else {
+					r = false
+				}
+			}
+		}
+		return r
+	})()
+
+	return subnetEqual && p.Description == aws.StringValue(sg.CacheSubnetGroupDescription)
```
negz, line 396: It seems like the goal here is to determine whether the list of subnets is equal, irrespective of order. Performance probably doesn't matter that much given that there's not likely to be more than a handful of subnets, but we could avoid the `O(n^m)` nested loop by turning one of the inputs into a set, for example:

```go
func IsSubnetGroupUpToDate(p cachev1alpha1.CacheSubnetGroupParameters, sg elasticache.CacheSubnetGroup) bool {
	if p.Description != aws.StringValue(sg.CacheSubnetGroupDescription) {
		return false
	}

	if len(p.SubnetIds) != len(sg.Subnets) {
		return false
	}

	exists := make(map[string]bool)
	for _, id := range sg.Subnets {
		exists[id] = true
	}
	for _, id := range p.SubnetIds {
		if !exists[id] {
			return false
		}
	}

	return true
}
```

pkg/controller/cache/cachesubnetgroup/controller.go
sahil-lakhwani: @negz  I think the comment was misleading. I was able to update a CacheSubnetGroup with new `subnetIds`.

apis/cache/v1alpha1/cache_subnet_group_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// A list of  Subnet IDs for the cache subnet group.
+	// +optional
+	SubnetIDs []string `json:"subnetIds,omitempty"`
+
+	// SubnetIDRefs references to a Subnet to and retrieves its SubnetID
+	// +optional
+	SubnetIDRefs []runtimev1alpha1.Reference `json:"subnetIdRefs,omitempty"`
+
+	// SubnetIDSelector selects a set of references that each retrieve the subnetID from the referenced Subnet
```
negz, line 37:
```suggestion
	// SubnetIDSelector selects a set of references that each retrieve the subnetID from the referenced Subnet
	// +optional
```

PR #194, opened 2020-04-13
IAMUser

sahil-lakhwani:

Adds IAM User resource.

As per discussions on https://github.com/crossplane/provider-aws/pull/141, the resource is named `User` instead of `IAMUser`. The API type name still remains identity.aws.crossplane.io . This and other other resource name (`IAM*`) can be renamed in a separate PR.

Example: `examples/iam/user.yaml`

### Checklist

I have:


apis/identity/v1alpha1/register.go
```diff
@@ -0,0 +1,53 @@
...
+
+var (
+	// SchemeGroupVersion is group version used to register these objects
+	SchemeGroupVersion = schema.GroupVersion{Group: Group, Version: Version}
+
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
+	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+)
+
+// USER type metadata.
```
negz, line 43:
```suggestion
// User type metadata.
```

apis/identity/v1alpha1/register.go
```diff
@@ -0,0 +1,53 @@
...
+var (
+	UserKind             = reflect.TypeOf(User{}).Name()
+	UserGroupKind        = schema.GroupKind{Group: Group, Kind: UserKind}.String()
+	UserKindAPIVersion   = UserKind + "." + SchemeGroupVersion.String()
+	UserGroupVersionKind = SchemeGroupVersion.WithKind(UserKind)
+)
+
+func init() {
+	SchemeBuilder.Register(&User{}, &UserList{})
```
negz, line 52: Would you mind renaming these to `IAMUser` for now? I know we discussed renaming, but I think for the time being we should be consistent with the existing `IAM<T>` types.

pkg/controller/identity/iamuser/controller.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	"github.com/crossplane/provider-aws/apis/identity/v1alpha1"
+	awsv1alpha3 "github.com/crossplane/provider-aws/apis/v1alpha3"
+	awsclients "github.com/crossplane/provider-aws/pkg/clients"
+	"github.com/crossplane/provider-aws/pkg/clients/iam"
+)
+
+const (
+	errNotUserInstance = "managed resource is not an User instance custom resource"
```
negz, line 45: Nit: I think just "User" would be preferable to "User instance".

pkg/controller/identity/iamuser/controller.go
```diff
@@ -0,0 +1,251 @@
...
+
+	groupsResp, err := e.client.ListGroupsForUserRequest(&awsiam.ListGroupsForUserInput{
+		UserName: aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
+
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
+	}
+
+	if !iam.CompareGroups(cr.Spec.ForProvider, groupsResp.Groups) {
```
negz, line 198: I wonder whether we should model IAM group membership with a distinct CR, the same way we model role and user policy attachments?

Our rationale with policy attachments was that we get maximum flexibility when the user, policy, and relationship between the two are all distinct CRs. This way you can model only a user, or only a policy in Crossplane, and still create relationships with users or policys that are not modelled with Crossplane. Presumably if we model group membership as an aspect of an IAM user and later added an IAM group CR, we'd have no way to add users to that group unless those users were modelled in Crossplane, right?

sahil-lakhwani: > Presumably if we model group membership as an aspect of an IAM user and later added an IAM group CR, we'd have no way to add users to that group unless those users were modelled in Crossplane, right?

@negz  I am not sure if I get this completely. I mean if we model the group membership as an aspect of IAMUser now and later add an IAMGroup CR, how would that be a problem?

negz: @sahil-lakhwani If we have:

* An `IAMGroup` CR
* An `IAMUser` that may specify the groups the user should be in.

This allows us to:

* Add an `IAMUser` managed by Crossplane to an `IAMGroup` managed by Crossplane.
* Add an `IAMUser` managed by Crossplane to an IAM group not managed by Crossplane.

However it does not allow us to add an IAM user that is not managed by Crossplane to an `IAMGroup` that is managed by Crossplane; the `IAMUser` must be modelled by Crossplane.

If we instead had:

* An `IAMGroup` CR.
* An `IAMUser` CR`.
* An `IAMGroupMembership` CR.

We would be able to:

* Add an `IAMUser` managed by Crossplane to an `IAMGroup` managed by Crossplane.
* Add an `IAMUser` managed by Crossplane to an IAM group not managed by Crossplane.
* Add an IAM user not managed by Crossplane to an `IAMGroup` managed by Crossplane.
* Add an IAM user not managed by Crossplane to an IAM group not managed by Crossplane.

pkg/clients/iam/iamuser.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	in.Path = awsclients.LateInitializeStringPtr(in.Path, user.Path)
+	if user.PermissionsBoundary != nil {
+		in.PermissionsBoundary = awsclients.LateInitializeStringPtr(in.PermissionsBoundary, user.PermissionsBoundary.PermissionsBoundaryArn)
+	}
+
+	for _, tag := range user.Tags {
+		in.Tags = append(in.Tags, v1alpha1.Tag{Key: *tag.Key, Value: *tag.Value})
+	}
```
negz, line 48: I think this will constantly append the same set of tags on every reconcile. For example presuming the external resource is tagged `cool: true`, each time we observe it call `LateInitializeUser` and append `cool: true` to our slice of tags, even if that tag is already there, thus growing the slice of tags indefinitely.

I believe this bug exists in the `IAMRole` controller too: https://github.com/crossplane/provider-aws/pull/141/files#diff-f4cf5668c6e624de813a694175ff2b28R111

pkg/controller/identity/iamuser/controller.go
```diff
@@ -0,0 +1,251 @@
...
+
+		// Remove user from current groups
+		for _, val := range groupsResp.Groups {
+			if _, err = e.client.RemoveUserFromGroupRequest(&awsiam.RemoveUserFromGroupInput{
+				GroupName: val.GroupName,
+				UserName:  aws.String(meta.GetExternalName(cr)),
+			}).Send(ctx); err != nil {
+				return managed.ExternalUpdate{}, errors.Wrap(err, errRemoveFromGroup)
+			}
+		}
```
negz, line 208: Does this mean that there will be a brief window in which this user is removed from all of the groups we want them to be in, before we re-add them immediately below? It seems like it would be safer to calculate the diff and only remove them from the groups that we know we no longer want them in.

negz: @sahil-lakhwani I think this needs to be addressed before we merge. I'd be happy with either only removing the groups that the user should actually be removed from, or just removing group membership from this CR altogether.

apis/identity/v1alpha1/register.go
```diff
@@ -0,0 +1,53 @@
...
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
+	SchemeBuilder = &scheme.Builder{GroupVersion: SchemeGroupVersion}
+)
+
+// User type metadata.
+var (
+	UserKind             = reflect.TypeOf(IAMUser{}).Name()
+	UserGroupKind        = schema.GroupKind{Group: Group, Kind: UserKind}.String()
+	UserKindAPIVersion   = UserKind + "." + SchemeGroupVersion.String()
+	UserGroupVersionKind = SchemeGroupVersion.WithKind(UserKind)
```
negz, line 48: These should all be `IAMUser<X>` too, to match the name of the struct.

pkg/controller/identity/iamuser/controller.go
sahil-lakhwani: Thanks @negz . I have removed the group membership for now.

apis/identity/v1alpha1/register.go
sahil-lakhwani: Update these. (The comment `UserPolicyAttachment type metadata.` is updated in the IAMUserPolicyAttachment PR)

pkg/controller/identity/iamuser/controller.go
```diff
@@ -0,0 +1,200 @@
...
+func SetupIAMUser(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(v1alpha1.IAMUserGroupKind)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.IAMUser{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(v1alpha1.IAMUserGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewUserClient}),
+			managed.WithConnectionPublishers(),
```
hasheddan, line 71: Could also implement `tagger` here as it appears `IAMUser` supports tags.

pkg/controller/identity/iamuser/controller.go
hasheddan: @negz I just wanted to note here that I was not able to track down how this looked previously, so not sure if comment was addressed sufficiently or not

PR #196, opened 2020-04-14
IAMUserPolicyAttachment

sahil-lakhwani:
### Description of your changes

Adds `UserPolicyAttachment` resource, which helps to attach IAM managed policy to an IAM user.

This depends on (and is a checkout from)  https://github.com/crossplane/provider-aws/pull/194 and should not be merged before that

Example: `examples/iam/userpolicyattachment.yaml` 

### Checklist

I have:
- [x] Run `make reviewable` to ensure this PR is ready for review.
- [x] Ensured this PR contains a neat, self documenting set of commits.
- [ ] Updated any relevant [documentation], [examples], or [release notes].
- [ ] Updated the dependencies in [`app.yaml`] to include any new role permissions.

[documentation]: https://github.com/crossplane/crossplane/tree/master/docs
[examples]: https://github.com/crossplane/crossplane/tree/master/cluster/examples
[release notes]: https://github.com/crossplane/crossplane/tree/master/PendingReleaseNotes.md
[`app.yaml`]: https://github.com/crossplane/provider-aws/blob/master/config/stack/manifests/app.yaml

PR #197, opened 2020-04-17
IAMPolicy

sahil-lakhwani:

Adds IAM Policy as a managed resource.

Example: `example/iam/policy.yaml`

This PR depends on #194 & #196, and should not be merged before those

For the update case:
An update to the policy is actually creating a new version of that policy. The approach in this PR sets the latest version as the default version. Also, because only maximum of 5 versions are allowed, the oldest version is deleted for an update request when there are already 5 versions.

### Checklist

I have:


pkg/clients/iam/iampolicy.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, err
+	}
+	return iam.New(*cfg), nil
+}
+
+// IsPolicyUpToDate checks whether there is a change in any of the modifiable fields in policy.
+func IsPolicyUpToDate(in v1alpha1.PolicyParameters, policy iam.PolicyVersion) (bool, error) {
+	// The AWS API reutrns Policy Document as an escaped string.
+	// Due to differences in the methods to escape a string, the comparison result between
+	// the spec.Documet and policy.Document can sometimes be false negative (due to spaces, line feeds).
```
negz, line 39:
```suggestion
	// the spec.Document and policy.Document can sometimes be false negative (due to spaces, line feeds).
```

negz: Also, great comment! This helps understand what's going on here.

apis/identity/v1alpha1/iampolicy_types.go
```diff
@@ -0,0 +1,98 @@
...
+// An IAMPolicySpec defines the desired state of an IAMPolicy.
+type IAMPolicySpec struct {
+	runtimev1alpha1.ResourceSpec `json:",inline"`
+	ForProvider                  IAMPolicyParameters `json:"forProvider"`
+}
+
+// IAMPolicyObservation keeps the state for the external resource
+type IAMPolicyObservation struct {
+	// The Amazon PolicyObservation Name (ARN) of the policy
+	Arn string `json:"arn,omitempty"`
```
negz, line 48:
```suggestion
	ARN string `json:"arn,omitempty"`
```

apis/identity/v1alpha1/register.go
```diff
@@ -0,0 +1,71 @@
...
+	UserGroupVersionKind = SchemeGroupVersion.WithKind(UserKind)
+)
+
+// Policy type metadata.
+var (
+	PolicyKind             = reflect.TypeOf(IAMPolicy{}).Name()
+	PolicyGroupKind        = schema.GroupKind{Group: Group, Kind: PolicyKind}.String()
+	PolicyKindAPIVersion   = PolicyKind + "." + SchemeGroupVersion.String()
+	PolicyGroupVersionKind = SchemeGroupVersion.WithKind(PolicyKind)
+)
```
negz, line 57: These should be `IAMPolicy<X>` to match the name of the struct.

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+	kube   client.Client
+}
+
+func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mgd.(*v1alpha1.IAMPolicy)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
+	}
+
+	if cr.Status.AtProvider.Arn == "" {
```
negz, line 121:
```suggestion
	if cr.Status.AtProvider.ARN == "" {
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+		return managed.ExternalObservation{}, nil
+	}
+
+	policyResp, err := e.client.GetPolicyRequest(&awsiam.GetPolicyInput{
+		PolicyArn: aws.String(cr.Status.AtProvider.Arn),
```
negz, line 126:
```suggestion
		PolicyArn: aws.String(cr.Status.AtProvider.ARN),
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+
+	if policyResp.Policy == nil {
+		return managed.ExternalObservation{}, errors.New(errEmptyPolicy)
+	}
+	policy := policyResp.Policy
+
+	cr.SetConditions(runtimev1alpha1.Available())
+
+	cr.Status.AtProvider = v1alpha1.IAMPolicyObservation{
+		Arn:                           aws.StringValue(policy.Arn),
```
negz, line 141:
```suggestion
		ARN:                           aws.StringValue(policy.Arn),
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+		AttachmentCount:               aws.Int64Value(policy.AttachmentCount),
+		DefaultVersionID:              aws.StringValue(policy.DefaultVersionId),
+		IsAttachable:                  aws.BoolValue(policy.IsAttachable),
+		PermissionsBoundaryUsageCount: aws.Int64Value(policy.PermissionsBoundaryUsageCount),
+		PolicyID:                      aws.StringValue(policy.PolicyId),
+	}
+
+	versionRsp, err := e.client.GetPolicyVersionRequest(&awsiam.GetPolicyVersionInput{
+		PolicyArn: aws.String(cr.Status.AtProvider.Arn),
```
negz, line 150:
```suggestion
		PolicyArn: aws.String(cr.Status.AtProvider.ARN),
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+		Path:           cr.Spec.ForProvider.Path,
+		PolicyDocument: aws.String(cr.Spec.ForProvider.Document),
+		PolicyName:     aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
+
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
+	}
+
+	cr.Status.AtProvider.Arn = aws.StringValue(createResp.CreatePolicyOutput.Policy.Arn)
```
negz, line 188:
```suggestion
	cr.Status.AtProvider.ARN = aws.StringValue(createResp.CreatePolicyOutput.Policy.Arn)
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	// An update to AWS Policy is a new version of that policy.
+	// A maximum of 5 versions are allowed. Below, the oldest version is deleted
+	// for an update request when 5 versions already exist.
+	// The new version is set as default.
+
+	if err := e.deleteOldestVersion(ctx, cr.Status.AtProvider.Arn); err != nil {
```
negz, line 204:
```suggestion
	if err := e.deleteOldestVersion(ctx, cr.Status.AtProvider.ARN); err != nil {
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+		return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
+	}
+
+	_, err := e.client.CreatePolicyVersionRequest(&awsiam.CreatePolicyVersionInput{
+		PolicyArn:      aws.String(cr.Status.AtProvider.Arn),
```
negz, line 209:
```suggestion
		PolicyArn:      aws.String(cr.Status.AtProvider.ARN),
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+	return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
+}
+
+func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
+	cr, ok := mgd.(*v1alpha1.IAMPolicy)
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+
+	if err := e.deleteNonDefaultVersions(ctx, cr.Status.AtProvider.Arn); err != nil {
```
negz, line 223:
```suggestion
	if err := e.deleteNonDefaultVersions(ctx, cr.Status.AtProvider.ARN); err != nil {
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+		return errors.Wrap(err, errDelete)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Deleting())
+
+	_, err := e.client.DeletePolicyRequest(&awsiam.DeletePolicyInput{
+		PolicyArn: aws.String(cr.Status.AtProvider.Arn),
```
negz, line 230:
```suggestion
		PolicyArn: aws.String(cr.Status.AtProvider.ARN),
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Creating())
+
+	createResp, err := e.client.CreatePolicyRequest(&awsiam.CreatePolicyInput{
+		Description:    cr.Spec.ForProvider.Description,
+		Path:           cr.Spec.ForProvider.Path,
+		PolicyDocument: aws.String(cr.Spec.ForProvider.Document),
+		PolicyName:     aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
```
negz, line 182: This seems like a tricky API in that the create call takes a user-specified name, but returns a fully qualified ARN that must be stored and provided when observing, updating, or deleting the policy. Currently we're storing that ARN in `cr.Status.AtProvider.ARN`, but I think we should store in in the `crossplane.io/external-name` annotation too.

Kubernetes requires us to be able to recover if we lose object status for some reason, so we can't rely on the ARN being stored there. For example if you back up this CR using https://velero.io/ the status will be removed when the CR is restored to the cluster.

```go
meta.SetExternalName(cr, aws.StringValue(createResp.CreatePolicyOutput.Policy.Arn))
if err := e.kube.Update(ctx, cr); err != nil {
   // ... handle error
}

// We must update status _after_ the above Update call, because the Update call overwrites
// cr.Status with whatever is currently in the API server.
cr.Status.AtProvider.Arn = aws.StringValue(createResp.CreatePolicyOutput.Policy.Arn)
return managed.ExternalCreation{}, errors.Wrap(e.kube.Status().Update(ctx, cr), errCreate)
```

I believe we can add something like the above at line 188, then just use `meta.GetExternalName` in all API calls. It will return only the name for the first observe and create, then return the ARN on subsequent observes, updates, and deletes.

sahil-lakhwani: @negz As per our conversation on Slack https://crossplane.slack.com/archives/CEF5N8X08/p1588019841267300 , I am just concerned about the case where the external-name is reset by a second `kubectl apply`

negz: Good point! As discussed on our call earlier it looks like we address a similar case in the SecurityGroup resource by storing its (unqualified) name in `spec`, then storing the unique identifier that the AWS API returns as our external name. In this case the user would be expected to never set the external name annotation.


https://github.com/crossplane/provider-aws/blob/211791/pkg/controller/network/securitygroup/controller.go#L142

apis/identity/v1alpha1/register.go
sahil-lakhwani: I have made the change

pkg/controller/identity/iampolicy/controller.go
sahil-lakhwani: Thanks @negz I've made the required change.

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -0,0 +1,300 @@
...
+// SetupIAMPolicy adds a controller that reconciles IAM Policy.
+func SetupIAMPolicy(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(v1alpha1.IAMPolicyGroupKind)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.IAMPolicy{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(v1alpha1.IAMPolicyGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewPolicyClient}),
```
negz, line 70:
```suggestion
			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewPolicyClient}),
			managed.WithInitializers(),
```

I think we should add the above option to disable [the default behaviour of setting the external name to the managed resource's name](https://github.com/crossplane/crossplane-runtime/blob/290de73499493eae198345e75cf79e1608f25d56/pkg/reconciler/managed/api.go#L49).

I suspect this is working at the moment, but a little misleading, because:

1. When the managed resource is first reconciled we set the external name to the managed resource name.
1. During Observe we notice that the external name is not an ARN, so we report that the resource doesn't exist.
1. During Create we overwrite the external name with the created ARN.

sahil-lakhwani: @negz Thanks for the catch, I have updated this.

PR #215, opened 2020-05-11
Add support for AWS Route53 Hosted Zone and Resource Record Set

### LINKED ISSUE #172: Add Route53 DNS Support
DNS provisioning from Route53.  

Enables creation of Route53 DNS resources.  This is a subset of Route53 features (to be listed)

### References
*

arush-sal:
### Checklist

I have:


apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,133 @@
...
+	// other than Route 53, change the name servers for your domain to the set of
+	// NameServers that CreateHostedZone returns in DelegationSet.
+	// +immutable
+	Name *string `json:"name"`
+
+	// (Private hosted zones only) The ID of an Amazon VPC that you're
+	// associating with this hosted zone. You can specify only one Amazon VPC
+	// when you create a private hosted zone.
+	// +immutable
+	VPCId *string `json:"VPCId,omitempty"`
```
negz, line 87: It seems like this is a reference to a VPC. Should it have a corresponding `VPCIdRef` and `VPCIdSelector`?

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,133 @@
...
+	if op.HostedZone.Id != nil {
+		status.ID = *op.HostedZone.Id
+	}
+	if op.HostedZone.ResourceRecordSetCount != nil {
+		status.ResourceRecordCount = *op.HostedZone.ResourceRecordSetCount
+	}
+	if op.Location != nil {
+		status.Location = *op.Location
+	}
+}
```
negz, line 133: We typically prefer to keep these "translation" functions in `pkg/client` rather than as methods on the types.

pkg/clients/zone/zone_test.go
```diff
@@ -0,0 +1,140 @@
...
+import (
+	"errors"
+	"reflect"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws/awserr"
+	"github.com/aws/aws-sdk-go-v2/service/route53"
+)
+
+func Test_getRegion(t *testing.T) {
```
negz, line 12: I'd prefer to stick to camel case [per Go convention](https://golang.org/doc/effective_go.html#mixed-caps) even in test names. I'd go with `TestGetRegion` here.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+		if aws.StringValue(rrSet.Name) == *p.Name {
+			return rrSet, nil
+		}
+	}
+
+	return route53.ResourceRecordSet{}, &NotFound{}
+}
+
+// IsRRSetNotFoundErr returns true if the error is because the item doesn't exist
+func IsRRSetNotFoundErr(err error) bool {
```
negz, line 170: https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

The above article makes a good case for assertion that an error satisfies a particular interface, rather than that it is of a particular concrete type.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+	patch := &v1alpha3.ResourceRecordSetParameters{}
+	if err := json.Unmarshal(jsonPatch, patch); err != nil {
+		return nil, err
+	}
+	patch.ZoneID = patch.Name
+	return patch, nil
+}
+
+// GetResourceRecordSetOrErr returns recordSet if present or err
+func GetResourceRecordSetOrErr(ctx context.Context,
```
negz, line 150: I haven't seen this `OrErr` convention in Go before - I think just `GetResourceRecordSet` would be appropriate. The signature indicates that an error may be returned.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+			in.Records[i] = aws.StringValue(val.Value)
+		}
+	}
+}
+
+// CreatePatch creates a *v1beta1.ResourceRecordSetParameters that has only the changed
+// values between the target *v1beta1.ResourceRecordSetParameters and the current
+// *route53.ResourceRecordSet
+func CreatePatch(in *route53.ResourceRecordSet,
+	target *v1alpha3.ResourceRecordSetParameters) (*v1alpha3.ResourceRecordSetParameters, error) {
```
negz, line 132: This is subjective, but I prefer to avoid wrapping these longer function declarations. It's hard to follow where the params end and the function body starts since they're at the same level of indentation. These lines are a little long, but I usually don't worry about that too much unless the line is getting to well over the 120 char mark.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+		ResourceRecords: resourceRecords,
+	}
+	changeBatch := &route53.ChangeBatch{
+		Changes: []route53.Change{
+			{
+				Action:            action,
+				ResourceRecordSet: &resourceRecordSet,
+			},
+		},
+	}
```
negz, line 71: Nit: Is there much value in creating `changeBatch` and `resourceRecordSet` as distinct variables vs just declaring them inline below?

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+func GenerateChangeResourceRecordSetsInput(p *v1alpha3.ResourceRecordSetParameters, action route53.ChangeAction) *route53.ChangeResourceRecordSetsInput {
+	var ttl *int64
+	if p.TTL == nil {
+		num := int64(300)
+		ttl = &num
+	} else {
+		ttl = p.TTL
+	}
+
+	resourceRecords := make([]route53.ResourceRecord, 0)
```
negz, line 50:
```suggestion
	resourceRecords := make([]route53.ResourceRecord, 0, len(p.Records))
```

It seems like we always know how many records there will be, so we can be a little more efficient by pre-allocating a backing array of that length for our slice.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+	for _, r := range p.Records {
+		record := r
+		resourceRecords = append(resourceRecords, route53.ResourceRecord{
+			Value: &record,
+		})
+	}
+
+	resourceRecordSet := route53.ResourceRecordSet{
```
negz, line 58: Nit: Go has an unconventional convention of _recommending_ shorter variable names, especially in cases like this where the variable is shortlived (it's only used close to where it's declared) is of a descriptive type.

https://talks.golang.org/2014/names.slide#4

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+	rrType := string(rrSet.Type)
+	in.Type = awsclients.LateInitializeStringPtr(in.Type, &rrType)
+
+	in.TTL = awsclients.LateInitializeInt64Ptr(in.TTL, rrSet.TTL)
+	if len(in.Records) == 0 && len(rrSet.ResourceRecords) != 0 {
+		in.Records = make([]string, len(rrSet.ResourceRecords))
+		for i, val := range rrSet.ResourceRecords {
+			in.Records[i] = aws.StringValue(val.Value)
+		}
+	}
```
negz, line 125: Do we need to handle the case in which we have a partial view of the records in our spec? i.e. Perhaps `spec.records` has two items, but there are actually three items according to the AWS API?

pkg/clients/zone/zone.go
```diff
@@ -0,0 +1,112 @@
...
+		"ap-southeast-2": route53.VPCRegionApSoutheast2,
+		"ap-south-1":     route53.VPCRegionApSouth1,
+		"ap-northeast-1": route53.VPCRegionApNortheast1,
+		"ap-northeast-2": route53.VPCRegionApNortheast2,
+		"ap-northeast-3": route53.VPCRegionApNortheast3,
+		"eu-north-1":     route53.VPCRegionEuNorth1,
+		"sa-east-1":      route53.VPCRegionSaEast1,
+		"ca-central-1":   route53.VPCRegionCaCentral1,
+		"cn-north-1":     route53.VPCRegionCnNorth1,
+	}
```
negz, line 100: https://godoc.org/github.com/aws/aws-sdk-go-v2/service/route53#VPCRegion

VPCRegion appears to be a subtype of `string`, so I think we can just use [type conversion](https://tour.golang.org/basics/13) here. For example:

```go
var string region = "us-east-1"

// r will be of type route53.VPCRegion
r := route53.VPCRegion(region)
```

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,158 @@
...
+func (conn *connector) Connect(ctx context.Context, mgd resource.Managed) (managed.ExternalClient, error) {
+	cr, ok := mgd.(*v1alpha3.ResourceRecordSet)
+	if !ok {
+		return nil, errors.New(errUnexpectedObject)
+	}
+
+	awsconfig, err := conn.awsConfigFn(ctx, conn.client, cr.Spec.ProviderReference)
+	if err != nil {
+		return nil, err
+	}
```
negz, line 64: https://github.com/crossplane/provider-aws/blob/2117914/pkg/controller/cache/cachesubnetgroup/controller.go#L90

Do we handle the case where we want to authenticate with a service account, like the above example?

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,177 @@
...
+	if err := json.Unmarshal(jsonPatch, patch); err != nil {
+		return nil, err
+	}
+	patch.ZoneID = patch.Name
+	return patch, nil
+}
+
+// GetResourceRecordSetOrErr returns recordSet if present or err
+func GetResourceRecordSetOrErr(ctx context.Context,
+	c Client, p v1alpha3.ResourceRecordSetParameters) (route53.ResourceRecordSet, error) {
```
negz, line 151: Perhaps this should just take the ZoneID and record set name as arguments, rather than the whole params?

apis/network/v1alpha3/resourcerecordset_types.go
```diff
@@ -0,0 +1,149 @@
...
+
+	// ZoneIDSelector selects a reference to a Zone to and retrieves its ZoneID
+	// +optional
+	ZoneIDSelector *runtimev1alpha1.Selector `json:"zoneIdSelector,omitempty"`
+}
+
+// ResourceRecordSetObservation keeps the state for the external resource.
+type ResourceRecordSetObservation struct {
+	// Name of the record that you want to create, update, or delete.
+	Name *string `json:"Name"`
```
negz, line 130: I don't think we need to repeat the name in the observation - per my comment above I think it should be the `ExternalName` annotation.

apis/network/v1alpha3/resourcerecordset_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// The current or new DNS record value, not to exceed 4,000 characters. In the
+	// case of a DELETE action, if the current value does not match the actual value,
+	// an error is returned.
+	Value *string `json:"Value"`
+}
+
+// ResourceRecordSetParameters define the desired state of an AWS Route53 Resource Record.
+type ResourceRecordSetParameters struct {
+	// Name of the record that you want to create, update, or delete.
+	Name *string `json:"name"`
```
negz, line 104: It seems like this name is the unique identifier of the record set? If so I think we should remove this field and derive the name from the CR's external name annotation using `meta.GetExternalName`.

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,158 @@
...
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Deleting())
+
+	input := resourcerecordset.GenerateChangeResourceRecordSetsInput(&cr.Spec.ForProvider, route53.ChangeActionDelete)
+	//TODO: Create an ErrorIs function to check the error returned
+	//Currently we aren't checking the error returned
+	//This is just to satisfy the linter
```
negz, line 154: Do we get an error if we try to remove a record set that doesn't exist?

pkg/controller/network/resourcerecordset/controller_test.go
```diff
@@ -0,0 +1,365 @@
...
+	}{
+		{
+			"upexpected managed resource should return error",
+			unexpectedItem,
+			nil,
+			nil,
+			false,
+			false,
+			false,
+		},
```
negz, line 127: I think these tests would be much easier to follow if we included field names in the structs.

pkg/controller/network/resourcerecordset/controller_test.go
```diff
@@ -0,0 +1,365 @@
...
+	} {
+		mockClientErr = tc.clientErr
+		itemsList = tc.itemsReturned
+		res, err := mockExternalClient.Observe(context.Background(), tc.managedObj)
+
+		g.Expect(err == nil).To(gomega.Equal(tc.expectedErrNil), tc.description)
+		g.Expect(res.ResourceExists).To(gomega.Equal(tc.expectedResourceExist), tc.description)
```
negz, line 161: Similar to above, I'd like to avoid using gomega in favour of our table driven testing convention.

pkg/controller/network/zone/controller.go
```diff
@@ -0,0 +1,192 @@
...
+	return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
+}
+
+func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
+	cr, ok := mgd.(*v1alpha3.Zone)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	_, err := e.client.UpdateZoneRequest(&cr.Status.AtProvider.ID, cr.Spec.ForProvider.Comment).Send(ctx)
```
negz, line 164: We can't rely on storing things in our status, unfortunately. I think this `ID` should be stored in the external name annotation via `GetExternalName` and `SetExternalName`.

pkg/controller/network/resourcerecordset/controller.go
arush-sal: Yes. Since we are using the [utils.RetrieveAwsConfigFromProvider](https://github.com/crossplane/provider-aws/pkg/controller/utils/utils.go#L38) function to get our `*aws.Config` the same has been taken care of in the mentioned function itself.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: As far as I know, not in particular. I decided to keep it that way just for the sake of increased readability. But have made the changed that to inline declaration now.

arush-sal:
@negz 

I wanted some clarification around the implementation of the `Alias` feature of Resource Records. The aforementioned feature requires a `Hosted Zone ID` of the resource that we are trying create the alias record with i.e if we are trying to create an `A` type resource record for an ELB we would need the Hosted Zone ID for that ELB that would be specific to the region it is in.

Typically when we try to do the same via AWS console the Hosted Zone ID field gets auto populated but in our case we will have to come up with a sort of resolver for the same. The resolver would be used to get the Hosted Zone ID of the respective service for the corresponding AWS region.
As an example, the list of the Hosted Zone IDs for the ELB/ALB, and NLB can be found [here.](https://docs.aws.amazon.com/general/latest/gr/elb.html)

The same can be done or implemented in the following two ways:

* Create a unified resolver for each service as part of the Resource Record client.
* Create a resolver function or method individually for each service that we implemented.

Since in either of the approach mentioned above it could take some significant amount of time for the implementation, therefore please let me know if I should continue with the feature implementation and if yes, which of the above two listed approaches should I be taking.

apis/network/v1alpha3/network_test.go
```diff
@@ -90,3 +91,24 @@ func Test_Common_BuildFromEC2Tags(t *testing.T) {
 
 	g.Expect(len(res)).To(gomega.Equal(2))
 }
+
+func Test_Zone_ZoneObservationUpdate(t *testing.T) {
+	g := gomega.NewGomegaWithT(t)
```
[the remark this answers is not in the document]
arush-sal: The requested test has been removed in favor of the suggested table driven style.

pkg/controller/network/resourcerecordset/controller_test.go
arush-sal: The requested test has been rewritten with the suggested table driven style.

apis/network/v1alpha3/resourcerecordset_types.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// DNSRecordType defines the valid DNS Record Types that can be used.
+type DNSRecordType string
+
+const (
```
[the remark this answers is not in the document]
arush-sal: The acronyms have been capitalized as suggested.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,133 @@
...
+	AtProvider                     ZoneObservation `json:"atProvider"`
+}
+
+// ZoneParameters define the desired state of an AWS Route53 Hosted Zone.
+type ZoneParameters struct {
+	// CallerReference is an unique string that identifies the request and that
+	// allows failed Zone create requests to be retried without the risk of
+	// executing the operation twice.
+	// +immutable
+	CallerReference *string `json:"callerref"`
```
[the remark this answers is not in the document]
arush-sal: We have implemented auto generation and updation of the `CallerRefrence` via `github.com/google/uuid` package.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Any comments that you want to include about the hosted zone.
+	Comment *string `json:"comment,omitempty"`
+
+	// A value that indicates whether this is a private hosted zone.
+	// +immutable
+	PrivateZone *bool `json:"privatezone,omitempty"`
```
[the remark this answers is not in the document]
arush-sal: The JSON tags have been renamed as suggested.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,133 @@
...
+// ZoneParameters define the desired state of an AWS Route53 Hosted Zone.
+type ZoneParameters struct {
+	// CallerReference is an unique string that identifies the request and that
+	// allows failed Zone create requests to be retried without the risk of
+	// executing the operation twice.
+	// +immutable
+	CallerReference *string `json:"callerref"`
+
+	// Any comments that you want to include about the hosted zone.
+	Comment *string `json:"comment,omitempty"`
```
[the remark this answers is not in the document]
arush-sal: `// +optional` annotation has been added to all the fields that had `omitempty`

apis/network/v1alpha3/zone_types.go
arush-sal: At the moment when this PR was raised the VPC referencer wasn't implemented but has been implemented now.

apis/network/v1alpha3/zone_types.go
arush-sal: This method has been moved from types file to the client package

pkg/clients/zone/zone_test.go
arush-sal: My bad here. The mentioned test has been renamed accordingly.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: The requested change to the way `IsRRSetNotFound` error was handled has been changed as per the recommendation from above article.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: Yes, the respective function has been renamed to `GetResourceRecordSet`

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: The respective and subsequent functions have been unwrapped as suggested.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: That makes sense, I have added `len(p.Records)` to the make call.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: Since we merged it with a single struct declaration therefore this has been resolved.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: For the resource records, we are maintaining our Object spec as the final source of truth. Therefore in such a case, I believe the additional record will be updated with what's available in the spec.

pkg/clients/zone/zone.go
arush-sal: The requested change has been implemented

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// www.example.com. The trailing dot is optional; Amazon Route 53 assumes that
+	// the domain name is fully qualified. This means that Route 53 treats www.example.com
+	// (without a trailing dot) and www.example.com. (with a trailing dot) as identical.
+	//
+	// If you're creating a public hosted zone, this is the name you have registered
+	// with your DNS registrar. If your domain name is registered with a registrar
+	// other than Route 53, change the name servers for your domain to the set of
+	// NameServers that CreateHostedZone returns in DelegationSet.
+	// +immutable
+	// +kubebuilder:validation:Required
```
muvaf, line 67: We don't use that kubebuilder validation for required fields. You can see how the types should look like for various cases here https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// +optional
+	VPCID *string `json:"VPCId,omitempty"`
+
+	// (Private hosted zones only) The region that an Amazon VPC was created in.
+	// +immutable
+	// +optional
+	VPCRegion *string `json:"VPCRegion,omitempty"`
+
+	// (Private hosted zones only) VPCIdRef references a VPC to retrieves its VPC Id
+	// +optional
```
muvaf, line 99: I believe VPCIDRef should also be `+immutable` if VPCID is immmutable.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// +immutable
+	// +optional
+	PrivateZone *bool `json:"privateZone,omitempty"`
+
+	// (Private hosted zones only) The ID of an Amazon VPC that you're
+	// associating with this hosted zone. You can specify only one Amazon VPC
+	// when you create a private hosted zone.
+	// +immutable
+	// +optional
+	VPCID *string `json:"VPCId,omitempty"`
```
muvaf, line 91:
```suggestion
	VPCID *string `json:"vpcId,omitempty"`
```

In jsontags we use camel case.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,173 @@
+/*
+Copyright 2019 The Crossplane Authors.
```
muvaf, line 2:
```suggestion
Copyright 2020 The Crossplane Authors.
```

The date here is about when the file is created.

pkg/clients/zone/zone.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws/awserr"
+	"github.com/aws/aws-sdk-go-v2/service/route53"
+	"github.com/aws/aws-sdk-go-v2/service/route53/route53iface"
+	"github.com/google/uuid"
+
+	"github.com/crossplane/provider-aws/apis/network/v1alpha3"
+)
+
+// Client defines Route53 Client operations
+type Client interface {
```
muvaf, line 14: I don't think we need a level of indirection here. We usually implement only a fraction of the interface that AWS already provides and then in controllers we use methods to produce input for those functions. See RDS client interface here https://github.com/crossplane/provider-aws/blob/118e08d/pkg/clients/rds/rds.go#L39

pkg/clients/zone/zone.go
```diff
@@ -0,0 +1,101 @@
...
+// Hosted Zone can be fetched and checked for existence.
+func (c *zoneClient) GetZoneRequest(id string) route53.GetHostedZoneRequest {
+	return c.zone.GetHostedZoneRequest(&route53.GetHostedZoneInput{
+		Id: &id,
+	})
+}
+
+// CreateZoneRequest returns a route53 CreateHostedZoneRequest using which a route53
+// Hosted Zone can be created.
+func (c *zoneClient) CreateZoneRequest(cr *v1alpha3.Zone) route53.CreateHostedZoneRequest {
```
muvaf, line 40: Same as above comment. The preference has been that the actual AWS API call happens in controller code. Client code is usually used for producing inputs for those calls and converting their output to be used in our structs.

pkg/clients/zone/zone.go
```diff
@@ -0,0 +1,101 @@
...
+	if cr.Spec.ForProvider.CallerReference == nil {
+		cr.Spec.ForProvider.CallerReference = getCallerRef()
```
muvaf, line 42:
```
	// A unique string that identifies the request and that allows failed CreateHostedZone
	// requests to be retried without the risk of executing the operation twice.
	// You must use a unique CallerReference string every time you submit a CreateHostedZone
	// request. CallerReference can be any unique string, for example, a date/time
	// stamp.
	//
	// CallerReference is a required field
```

I believe we can use `metadata.resourceVersion` for this field so that we don't call `Create` twice with exactly same CR.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// ZoneStatus represents the observed state of a Zone.
+type ZoneStatus struct {
+	runtimev1alpha1.ResourceStatus `json:",inline"`
+	AtProvider                     ZoneObservation `json:"atProvider"`
+}
+
+// ZoneParameters define the desired state of an AWS Route53 Hosted Zone.
+type ZoneParameters struct {
```
muvaf, line 56: Looking at `CreateHostedZone`, `HostedZoneConfig` seems to be a sub-struct of the main object but here the properties of `HostedZoneConfig` like `Comment` and `PrivateZone` are at the top level. As per [managed resource design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity), we're trying to stick to whatever cloud provider exposes as API, so, it'd be good to follow that.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+	VPCRegion *string `json:"VPCRegion,omitempty"`
+
+	// (Private hosted zones only) VPCIdRef references a VPC to retrieves its VPC Id
+	// +optional
+	VPCIDRef *runtimev1alpha1.Reference `json:"VPCIdRef,omitempty"`
+
+	// VPCIdSelector selects a reference to a VPC to retrieves its VPC Id
+	// +optional
+	VPCIDSelector *runtimev1alpha1.Selector `json:"VPCIdSelector,omitempty"`
+}
```
muvaf, line 105: I think there is also `DelegationSetId` property that should be included even though we don't yet have the `DelegationSet` managed resource.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+// ZoneObservation keeps the state for the external resource.
+type ZoneObservation struct {
+	// ID is the unique URL representing the new hosted zone.
+	ID string `json:"Id"`
```
muvaf, line 110: This is already saved as external-name, so, we don't need to expose it here again.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// The number of resource record sets in the hosted zone.
+	ResourceRecordCount int64 `json:"ResourceRecordSetCount"`
+}
+
+// DelegationSet describes the name servers for this hosted zone.
+type DelegationSet struct {
```
muvaf, line 117: This struct doesn't seem to be used anywhere.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+)
+
+// +kubebuilder:object:root=true
+
+// Zone is a managed resource that represents an AWS Route53 Hosted Zone.
+// +kubebuilder:printcolumn:name="ID",type="string",JSONPath=".status.atProvider.Id"
```
muvaf, line 28:
```suggestion
// +kubebuilder:printcolumn:name="ID",type="string",JSONPath=".metadata.annotations['crossplane.io/external-name']"
```

should work but didn't test manually

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+// +kubebuilder:printcolumn:name="RRs",type="integer",JSONPath=".status.atProvider.ResourceRecordSetCount"
```
muvaf, line 29: Rather than count, I believe `spec.forProvider.name` could be more useful for a glance view.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+	// +immutable
+	// +kubebuilder:validation:Required
+	Name *string `json:"name"`
+
+	// CallerReference is an unique string that identifies the request and that
+	// allows failed Zone create requests to be retried without the risk of
+	// executing the operation twice.
+	// +immutable
+	// +optional
+	CallerReference *string `json:"callerref,omitempty"`
```
muvaf, line 75:
```suggestion
	CallerReference *string `json:"callerReference,omitempty"`
```

apis/network/v1alpha3/resourcerecordset_types.go
```diff
@@ -0,0 +1,137 @@
...
+// +kubebuilder:object:root=true
+
+// ResourceRecordSet is a managed resource that represents an AWS Route53 Resource Record.
+// +kubebuilder:printcolumn:name="TYPE",type="string",JSONPath=".spec.forProvider.type"
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+type ResourceRecordSet struct {
```
muvaf, line 72: I believe the name should be `ResourceRecord`. The `List` and `Change` calls use `ResourceRecordSet` name but that's because they work as batch operations. Here, we expose individual resource record as CR.

pkg/controller/network/zone/controller.go
```diff
@@ -0,0 +1,202 @@
...
+	client zone.Client
+}
+
+func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mgd.(*v1alpha3.Zone)
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
+	}
+
+	err := e.kube.Update(ctx, configSanitize(cr))
```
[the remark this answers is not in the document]
arush-sal: Since the `comment` is the only field that can be updated, therefore I decided that if none is provided it best to set it to a default value so that it later can be checked against for the updation event. That is the `configSanitize` is supposed to check the current value and set it to its type default if nil. Which we then update the Kubernetes object with.

muvaf:
> Typically when we try to do the same via AWS console the Hosted Zone ID field gets auto populated but in our case we will have to come up with a sort of resolver for the same. The resolver would be used to get the Hosted Zone ID of the respective service for the corresponding AWS region.
As an example, the list of the Hosted Zone IDs for the ELB/ALB, and NLB can be found here.

The auto-fill feature of AWS console is certainly nice but I'm not sure if it's really necessary, especially at v1alpha1 level. I believe it should be fine just to treat those fields as like any other field for now. The actual values can be found in AWS documentation by the user

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,173 @@
...
+// NewClient creates new AWS client with provided AWS Configuration/Credentials
+func NewClient(config *aws.Config) Client {
+	return route53.New(*config)
+}
+
+// GenerateChangeResourceRecordSetsInput prepares input for a ChangeResourceRecordSetsInput
+func GenerateChangeResourceRecordSetsInput(p *v1alpha3.ResourceRecordSetParameters, action route53.ChangeAction) *route53.ChangeResourceRecordSetsInput {
+	var ttl *int64
+	if p.TTL == nil {
+		num := int64(300)
```
muvaf, line 60: How did you come up with `300`?

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,173 @@
...
+			return rrSet, nil
+		}
+	}
+
+	return route53.ResourceRecordSet{}, &NotFound{}
+}
+
+// StringValueFromRRType returns the string value of aws ResourceRecordType
+// passed in or "" if nil
+func StringValueFromRRType(rrType route53.RRType) string {
```
muvaf, line 171: This function doesn't seem to be used anywhere.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+// IsUpToDate checks if object is up to date
+func IsUpToDate(p v1alpha3.ResourceRecordSetParameters, rrset route53.ResourceRecordSet) (bool, error) {
+	// check for the root "." found in DNS entries and add if not found
+	if !strings.HasSuffix(*p.Name, ".") {
+		p.Name = aws.String(fmt.Sprintf("%s.", *p.Name))
+	}
+	patch, err := CreatePatch(&rrset, &p)
```
muvaf, line 98: This resource is so simple that I don't think we need to use JSON patch stuff for comparisons. You can just compare all the properties one by one to be on the safe side.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,173 @@
...
+	req := c.ListResourceRecordSetsRequest(&route53.ListResourceRecordSetsInput{
+		HostedZoneId: p.ZoneID,
+	})
+	res, err := req.Send(ctx)
+	if err != nil {
+		return route53.ResourceRecordSet{}, err
+	}
+
+	for _, rrSet := range res.ResourceRecordSets {
+		if aws.StringValue(rrSet.Name) == aws.StringValue(p.Name) && string(rrSet.Type) == aws.StringValue(p.Type) {
```
muvaf, line 161: I think you can have multiple records for `*.google.com` with same type like `CNAME` or others. So, you can't really use name+type as unique identifier. Am I wrong?

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mgd.(*v1alpha3.ResourceRecordSet)
+
+	if !ok {
+		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
+	}
+
+	rrset, err := resourcerecordset.GetResourceRecordSet(ctx, e.client, cr.Spec.ForProvider)
+	if _, ok := err.(*resourcerecordset.NotFound); ok {
```
muvaf, line 102: We just implement a `func IsErrorNotFound(err error) bool` in clients folder for each resource if we can't use other checks and use that to check error type instead of doing the casting here.

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
+	if err != nil {
+		// Either there is err and retry. Or Resource does not exist.
+		return managed.ExternalObservation{
+			ResourceExists:    false,
+			ConnectionDetails: managed.ConnectionDetails{},
+		}, errors.Wrap(resource.Ignore(zone.IsErrorNoSuchHostedZone, err), errList)
```
muvaf, line 110: I don't think the error you need to skip is `NoSuchHostedZone`. It should be the error that is returned when you call `GET` on the resource.

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+
+	upToDate, err := resourcerecordset.IsUpToDate(cr.Spec.ForProvider, rrset)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, errState)
+	}
+
+	return managed.ExternalObservation{
+		ResourceExists:    true,
+		ResourceUpToDate:  upToDate,
+		ConnectionDetails: managed.ConnectionDetails{},
```
muvaf, line 123:
```suggestion
```

You can just leave it as empty.

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	cr.Status.SetConditions(runtimev1alpha1.Creating())
+
+	input := resourcerecordset.GenerateChangeResourceRecordSetsInput(&cr.Spec.ForProvider, route53.ChangeActionCreate)
+	_, err := e.client.ChangeResourceRecordSetsRequest(input).Send(ctx)
+
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
+	}
+
+	return managed.ExternalCreation{}, errors.Wrap(nil, errCreate)
```
muvaf, line 142:
```suggestion
	return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
```

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Creating())
+
+	input := resourcerecordset.GenerateChangeResourceRecordSetsInput(&cr.Spec.ForProvider, route53.ChangeActionCreate)
+	_, err := e.client.ChangeResourceRecordSetsRequest(input).Send(ctx)
+
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
+	}
```
muvaf, line 140:
```suggestion
```

pkg/controller/network/resourcerecordset/controller.go
```diff
@@ -0,0 +1,179 @@
...
+func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.ExternalCreation, error) {
+	cr, ok := mgd.(*v1alpha3.ResourceRecordSet)
+	if !ok {
+		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Creating())
+
+	input := resourcerecordset.GenerateChangeResourceRecordSetsInput(&cr.Spec.ForProvider, route53.ChangeActionCreate)
+	_, err := e.client.ChangeResourceRecordSetsRequest(input).Send(ctx)
```
muvaf, line 136: It's interesting that AWS doesn't really have a unique identifier for this resource. I am wondering whether it should just be an array in `HostedZone` CR. I know Terraform implements it as [distinct resource](https://www.terraform.io/docs/providers/aws/r/route53_record.html) but I can't easily see how we can make it really work. @negz @hasheddan @kasey any thoughts?

apis/network/v1alpha3/resourcerecordset_types.go
```diff
@@ -0,0 +1,137 @@
...
+// ResourceRecord holds the DNS value to be used for the record.
+type ResourceRecord struct {
+	// The current or new DNS record value, not to exceed 4,000 characters. In the
+	// case of a DELETE action, if the current value does not match the actual value,
+	// an error is returned.
+	Value *string `json:"Value"`
+}
+
+// ResourceRecordSetParameters define the desired state of an AWS Route53 Resource Record.
+type ResourceRecordSetParameters struct {
```
muvaf, line 100: Some fields are missing such as `Weight`. Also I'd suggest looking at [types section here](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#pointer-types-and-markers) to see how their types and jsontags should look like.

apis/network/v1alpha3/zone_types.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Zone is a managed resource that represents an AWS Route53 Hosted Zone.
+// +kubebuilder:printcolumn:name="ID",type="string",JSONPath=".status.atProvider.Id"
+// +kubebuilder:printcolumn:name="RRs",type="integer",JSONPath=".status.atProvider.ResourceRecordSetCount"
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster
+type Zone struct {
```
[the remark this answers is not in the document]
arush-sal: The kind has been renamed to `HostedZone` as mentioned.

apis/network/v1alpha3/zone_types.go
arush-sal: This has been removed.

apis/network/v1alpha3/zone_types.go
arush-sal: `+immutable` has been added to `vpcIdRef` and `vpcIdSelector`

apis/network/v1alpha3/zone_types.go
arush-sal: all the JSON tags with `VPC` has been changed to `vpc`

```go
	// (Private hosted Hostedzones only) The ID of an Amazon VPC that you're
	// associating with this hosted Hostedzone. You can specify only one Amazon VPC
	// when you create a private hosted Hostedzone.
	// +immutable
	// +optional
	VPCId *string `json:"vpcId,omitempty"`

	// (Private hosted Hostedzones only) The region that an Amazon VPC was created in.
	// +immutable
	// +optional
	VPCRegion *string `json:"vpcRegion,omitempty"`

	// (Private hosted Hostedzones only) VPCIdRef references a VPC to retrieves its VPC Id
	// +immutable
	// +optional
	VPCIdRef *runtimev1alpha1.Reference `json:"vpcIdRef,omitempty"`

	// VPCIdSelector selects a reference to a VPC to retrieves its VPC Id
	// +immutable
	// +optional
	VPCIdSelector *runtimev1alpha1.Selector `json:"vpcIdSelector,omitempty"`
```

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: The proper date and license has been added to all the files.

apis/network/v1alpha3/zone_types.go
arush-sal: I was told to keep the structs fields around the concrete types as much as possible and hence I choose to do this. I have added `HostedZoneConfig` as suggested.

apis/network/v1alpha3/zone_types.go
arush-sal: As far as I have seen in all the tests done, the `spec.forProvider.name` is anyway included. And if we explicitly specify that here then it appears twice in the `get` columns.

apis/network/v1alpha3/zone_types.go
arush-sal: this has been renamed to `callerReference` as suggested.

apis/network/v1alpha3/zone_types.go
arush-sal: The same has been added now.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: `300` is the default TTL value by AWS

apis/network/v1alpha3/zone_types.go
arush-sal: We decided to include it as the final source of truth in case of any discrepancy.

pkg/clients/zone/zone.go
arush-sal: This has been changed accordingly, as suggested.

pkg/clients/zone/zone.go
arush-sal: This has been changed accordingly, as suggested.

pkg/clients/zone/zone.go
arush-sal: `metadata.resourceVersion` won't work here as it keeps on changes with every update to the object hence I have used `metadata.UID` instead.

apis/network/v1alpha3/zone_types.go
arush-sal: With the latest changes, it is being used now.

apis/network/v1alpha3/zone_types.go
arush-sal: I tested it and it doesn't seem to be working.

apis/network/v1alpha3/resourcerecordset_types.go
arush-sal: The name is `ResourceRecordSet` because we aren't exposing individual Resource Record but rather a bunch/group  of those together as ```
Records []string `json:"records"`
``` in the `ResourceRecordParameters`

File: apis/network/v1alpha3/resourcerecordset_types.go (hunk 51 above)
Comment (arush-sal) [reply apis/network/v1alpha3/resourcerecordset_types.go]: All of the fields that are available with AWS SDK's Resource Record Set has been added.

**Comment by arush-sal** [comment]
Review comments have been addressed, ready to go ahead.

File: apis/network/v1alpha3/resourcerecordset_types.go (hunk 15 above)
Comment (arush-sal) [reply apis/network/v1alpha3/resourcerecordset_types.go]: I believe that from instead of asking the user to provide the name under 

```
metadata:
  annotation: crossplane.io/external-name:
```
it would be easier from a UX perspective to let them specify that in 

```
spec:
  forProvider:
    name:
```

and then set the annotation internally, as that way it also follows the struct design as specified by the AWS SDK.

File: pkg/clients/resourcerecordset/resourcerecordset.go (hunk 42 above)
Comment (arush-sal) [reply pkg/clients/resourcerecordset/resourcerecordset.go:172]: That is correct and hence the function has been removed.

File: pkg/clients/resourcerecordset/resourcerecordset.go (hunk 44 above)
Comment (arush-sal) [reply pkg/clients/resourcerecordset/resourcerecordset.go]: Here we are checking for `ResourceRecordSet` not an individual `ResourceRecord` and we cannot have multiple `ResourceRecordSet` with the same name+type and hence that works as the unique identifier.

File: pkg/controller/network/resourcerecordset/controller.go (hunk 47 above)
Comment (arush-sal) [reply pkg/controller/network/resourcerecordset/controller.go:123]: Suggest change has been incorporated.

File: pkg/controller/network/resourcerecordset/controller.go (hunk 49 above)
Comment (arush-sal) [reply pkg/controller/network/resourcerecordset/controller.go:140]: Suggest change has been incorporated.

File: pkg/controller/network/resourcerecordset/controller.go (hunk 48 above)
Comment (arush-sal) [reply pkg/controller/network/resourcerecordset/controller.go:142]: Suggest change has been incorporated.

File: pkg/controller/network/resourcerecordset/controller.go (hunk 45 above)
Comment (arush-sal) [reply pkg/controller/network/resourcerecordset/controller.go]: Suggest change has been incorporated.

File: pkg/controller/network/resourcerecordset/controller.go (hunk 46 above)
Comment (arush-sal) [reply pkg/controller/network/resourcerecordset/controller.go]: Suggest change has been incorporated.

**Comment by arush-sal** [comment]
Review comments have been addressed, ready to go ahead.

please note that the branch name suggests it to be `v1alpha3` but it actually is `v1alpha1`

**Comment by muvaf** [comment]
Let's discuss `ResourceRecordSet` today in the community meeting. Following is how the API is shaped:

* `CreateHostedZoneInput` doesn't have any field for resource records.
* `ChangeResourceRecordSetsInput` is the only call that is used for `ResourceRecordSet` creation and deletion. It does batch operation, i.e. you pass an array of `ResourceRecordSet` objects. It determines the operation via a string field: `CREATE`, `DELETE`, `UPSERT`.
* `ListResourceRecordSetsInput` returns an array of `ResourceRecordSet`s that belong to the given Hosted Zone.
* There is no unique identifier for a `ResourceRecordSet`. To do a filtering on the `ListResourceRecordSetsInput` call, one has to use `name + type + setIdentifier` (`setIdentifier` is just like a tag; it's used only if the record set has non-simple routing policy which can have `name+type` collisions).
* All calls have to include Hosted Zone ID.

With this information, my inclination is that it should just be a part of `HostedZone` as an field whose type is array rather than a separate CR. This has the advantage of late-init of default resource records but the tradeoff is that you cannot independently manage a `ResourceRecordSet` without having its `HostedZone` in the Crossplane. Although I am not sure how common that would be, I'm not very familiar with `HostedZone` in general so I'd love to hear some real use cases. Also a resource having no real unique identifier brings some questions about its external-name annotation, too. What would be its value? A combination of several fields?

Looking at corresponding Terraform [resource](https://www.terraform.io/docs/providers/aws/r/route53_record.html), it is a distinct type and they seem to go with `ZONEID_RECORDNAME_TYPE_SET-IDENTIFIER` as identifier for import. CloudFormation treats it as [separate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html) as well.

Both approaches seem valid. I'm a bit leaning towards having it as part of `HostedZone` partly because when you create a `HostedZone`, you get some default resource record sets and those would be late-inited and I feel a bit uncomfortable having a CR that doesn't have an actual unique identifier. But as I said, I don't have much experience with this API so I don't feel strongly either way, though Terraform/CloudFormation treating it as distinct is definitely something to consider.

File: pkg/clients/resourcerecordset/resourcerecordset.go [hunk 53]
```diff
@@ -0,0 +1,220 @@
...
+func IsUpToDate(p v1alpha1.ResourceRecordSetParameters, rrset route53.ResourceRecordSet) (bool, error) {
+	// check for the root "." found in DNS entries and add if not found
+	if !strings.HasSuffix(*p.Name, ".") {
+		p.Name = aws.String(fmt.Sprintf("%s.", *p.Name))
+	}
+	patch, err := CreatePatch(&rrset, &p)
+	if err != nil {
+		return false, err
+	}
+	return cmp.Equal(&v1alpha1.ResourceRecordSetParameters{}, patch), nil
```
muvaf, line 145: It needs to skip reference and selector fields.

pkg/controller/route53/hostedzone/controller.go
```diff
@@ -0,0 +1,175 @@
...
+	if res != nil {
+		cr.Status.AtProvider = hostedzone.GenerateObservation(res.CreateHostedZoneOutput)
+		meta.SetExternalName(cr, aws.StringValue(res.CreateHostedZoneOutput.HostedZone.Id))
+	}
+
+	if err := e.kube.Update(ctx, cr); err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errKubeUpdate)
+	}
+
+	return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
```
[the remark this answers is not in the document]
arush-sal: Suggested change has been incorporated.

pkg/controller/route53/hostedzone/controller.go
```diff
@@ -0,0 +1,175 @@
...
+	res, err := e.client.GetHostedZoneRequest(&route53.GetHostedZoneInput{
+		Id: aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
+
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(hostedzone.IsErrorNoSuchHostedZone, err), errGet)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Available())
+	cr.Status.AtProvider.HostedZone.ResourceRecordSetCount = aws.Int64Value(res.HostedZone.ResourceRecordSetCount)
```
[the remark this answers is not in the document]
arush-sal: Suggested change has been incorporated.

pkg/clients/hostedzone/hostedzone.go
```diff
@@ -0,0 +1,148 @@
...
+			CallerReference:        aws.StringValue(op.HostedZone.CallerReference),
+			ID:                     aws.StringValue(op.HostedZone.Id),
+			Name:                   aws.StringValue(op.HostedZone.Name),
+			ResourceRecordSetCount: aws.Int64Value(op.HostedZone.ResourceRecordSetCount),
+		}
+
+		if op.HostedZone.Config != nil {
+			o.HostedZone.Config = v1alpha1.HostedZoneConfig{
+				PrivateZone: awsclients.LateInitializeBoolPtr(op.HostedZone.Config.PrivateZone, aws.Bool(false)),
+				Comment:     awsclients.LateInitializeStringPtr(op.HostedZone.Config.Comment, aws.String("")),
```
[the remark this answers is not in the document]
arush-sal: My bad, suggested change has been incorporated.

pkg/clients/hostedzone/hostedzone.go
```diff
@@ -0,0 +1,148 @@
...
+func GenerateUpdateHostedZoneCommentInput(cfg *v1alpha1.HostedZoneConfig, id *string) *route53.UpdateHostedZoneCommentInput {
+	var c *string
+	c = aws.String("")
+	if cfg != nil && cfg.Comment != nil {
+		c = cfg.Comment
+	}
+	return &route53.UpdateHostedZoneCommentInput{
+		Comment: c,
+		Id:      id,
+	}
```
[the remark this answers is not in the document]
arush-sal: I have already tried that but if `cfg` is nil this returns `null pointer dereference` that is why I added extra check for `cfg != nil` but yeah I certainly can remove `cfg.Comment != nil` as that was there just for added safety.

pkg/clients/hostedzone/hostedzone.go
```diff
@@ -0,0 +1,148 @@
...
+
+// IsUpToDate check whether the comment in Spec and Response are same or not
+func IsUpToDate(cfg *v1alpha1.HostedZoneConfig, res *route53.HostedZoneConfig) bool {
+	if cfg == nil && res.Comment != nil {
+		return false
+	}
+	if cfg != nil && !cmp.Equal(aws.StringValue(cfg.Comment), aws.StringValue(res.Comment)) {
+		return false
+	}
+	return true
```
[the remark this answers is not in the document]
arush-sal: Suggested change has been incorporated.

apis/route53/v1alpha1/hostedzone_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// NameServers that CreateHostedHostedZone returns in DelegationSet.
+	// +immutable
+	Name *string `json:"name"`
+
+	// CallerReference is an unique string that identifies the request and that
+	// allows failed HostedZone create requests to be retried without the risk of
+	// executing the operation twice.
+	// +immutable
+	// +optional
+	CallerReference *string `json:"callerReference,omitempty"`
```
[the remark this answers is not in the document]
arush-sal: I had added that field with the idea to let the user choose the CallerReference if they want. But have removed the same as requested.

pkg/clients/hostedzone/hostedzone.go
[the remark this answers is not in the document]
arush-sal: I tried the suggested change but encountered the `null pointer dereference` error.

apis/route53/v1alpha1/hostedzone_types.go
```diff
@@ -0,0 +1,199 @@
...
+	// Comment that you want to include about the hosted zone.
+	Comment *string `json:"comment,omitempty"`
+
+	// PrivateZone indicates whether this is a private hosted zone.
+	// +immutable
+	PrivateZone *bool `json:"privateZone,omitempty"`
+}
+
+// HostedZoneObservation keeps the state for the external resource.
+type HostedZoneObservation struct {
```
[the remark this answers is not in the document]
arush-sal: I initially had it like that but changed the same to match the AWS representation of the output. I have removed the common fields from observation now i.e `Name` and `Config` from the `HostedZoneResponse` struct.

pkg/clients/hostedzone/hostedzone.go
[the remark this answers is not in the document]
arush-sal: But this doesn't makes much sense to me. Because `res.PrivateZone` is auto populated to `false` as the default by AWS. So technically `res` will never be `nil`. Also, if we have not provided either of the `cfg.Comment` or `cfg.PrivateZone` it will be `nil` and will return `false` which is not the intended behavior.

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -0,0 +1,220 @@
...
+	req := c.ListResourceRecordSetsRequest(&route53.ListResourceRecordSetsInput{
+		HostedZoneId: zoneID,
+	})
+	res, err := req.Send(ctx)
+	if err != nil {
+		return route53.ResourceRecordSet{}, err
+	}
+
+	for _, rrSet := range res.ResourceRecordSets {
+		if aws.StringValue(rrSet.Name) == aws.StringValue(name) && string(rrSet.Type) == aws.StringValue(rrType) {
```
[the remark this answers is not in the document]
arush-sal: Suggested change has been incorporated.

pkg/clients/resourcerecordset/resourcerecordset.go
arush-sal: Suggested change has been incorporated.

arush-sal:
@muvaf yeah sure. I'll squash the commits, fix the conflicts, and will let you know.

muvaf:
I have discovered a few issues but I'll fix them in a separate PR. Thanks @arush-sal !

PR #217, opened 2020-05-12
added aws-provider to provider ref so manifest can be deployed with n…

jbrawdy:
…o modifications by the user.

I have:
- [ ] Run `make reviewable` to ensure this PR is ready for review.
- [x ] Ensured this PR contains a neat, self documenting set of commits.

### DIFF
--- a/examples/cache/rediscluster/resource-class.yaml
+++ b/examples/cache/rediscluster/resource-class.yaml
@@ -8,7 +8,7 @@ metadata:
 specTemplate:
   writeConnectionSecretsToNamespace: crossplane-system
   providerRef:
-    name: example
+    name: aws-provider
   reclaimPolicy: Delete
   forProvider:
     replicationGroupDescription: "An example replication group"


negz:
Thanks @jbrawdy! I think part of the problem here is that we have some example providers named `provider-aws` and others named `example`. :( I can see the one named `example` below. Where did you come across the one named `provider-aws`?

https://github.com/crossplane/provider-aws/blob/2117914c48d853890e6f75f055c88436eaf24443/examples/aws-provider.yaml#L16

jbrawdy:
> Thanks @jbrawdy! I think part of the problem here is that we have some example providers named `aws-provider` and others named `example`. :( I can see the one named `example` below. Where did you come across the one named `aws-provider`?
> 
> https://github.com/crossplane/provider-aws/blob/2117914c48d853890e6f75f055c88436eaf24443/examples/aws-provider.yaml#L16

Thanks for looking at it! When I was working through the example[ here](https://crossplane.io/docs/v0.10/getting-started/dynamic.html) `gcp-provider` was used for `providerRef`. When I changed the `providerRef` to `aws-provider` this example file worked. When I left the `providerRef` as `example`, it would not provision anything on AWS.

jbrawdy:
> I think the confusion here likely arises because the docs use the `Provider` named `aws-provider` [here](https://crossplane.io/docs/v0.10/cloud-providers/aws/aws-provider.html). We also use the name `example` in the [GCP examples](https://github.com/crossplane/provider-gcp/blob/0e59cad47ebf855fd3f505f3118dc939d4d6f136/examples/gcp-provider.yaml#L16) in the repo. It might be good to standardize on a common name across docs / examples as it is natural to move from one to the other. That being said, this should also be less of an issue once we have a quick start for each cloud in v0.11 docs :)

Sounds good. That makes sense. Thanks.

PR #222, opened 2020-05-17
Add SQS resource

sahil-lakhwani:

This PR adds `Queue` resource offered by the AWS Simple Queue Service.

An example to create a `Queue` resource can be found at `examples/applicationintegration/queue.yaml`

### Checklist

I have:


apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// +optional
+	RedrivePolicy RedrivePolicy `json:"redrivePolicy,omitempty"`
+
+	// The visibility timeout for the queue, in seconds.
+	// +optional
+	VisibilityTimeout *int64 `json:"visibilityTimeout,omitempty"`
+
+	// The ID of an AWS-managed customer master key (CMK) for Amazon SQS or a custom CMK.
+	// +optional
+	KmsMasterKeyID *string `json:"kmsMasterKeyID,omitempty"`
```
hasheddan, line 111:
```suggestion
	KmsMasterKeyID *string `json:"kmsMasterKeyId,omitempty"`
```

pkg/controller/applicationintegration/sqs/contoller.go
```diff
@@ -0,0 +1,244 @@
...
+
+func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.ExternalCreation, error) {
+	cr, ok := mg.(*v1alpha1.Queue)
+	if !ok {
+		return managed.ExternalCreation{}, errors.New(errNotQueue)
+	}
+
+	cr.SetConditions(runtimev1alpha1.Creating())
+
+	// FIFO queues names should start with ".fifo"
```
hasheddan, line 166:
```suggestion
	// FIFO queues names should end with ".fifo"
```

pkg/clients/sqs/sqs.go
```diff
@@ -0,0 +1,240 @@
...
+
+// IsUpToDate checks whether there is a change in any of the modifiable fields.
+func IsUpToDate(p v1alpha1.QueueParameters, attributes map[string]string, tags map[string]string) bool { // nolint:gocyclo
+	if len(p.Tags) != len(tags) {
+		return false
+	}
+	pTags := make(map[string]string, len(p.Tags))
+	for _, tag := range p.Tags {
+		pTags[tag.Key] = tag.Value
+	}
```
hasheddan, line 167: Does `p.Tags` need to be copied into `pTags` for comparison?

pkg/controller/applicationintegration/sqs/contoller.go
```diff
@@ -0,0 +1,244 @@
...
+func SetupQueue(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(v1alpha1.QueueGroupKind)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&v1alpha1.Queue{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(v1alpha1.QueueGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: sqs.NewClient}),
+			managed.WithInitializers(),
```
hasheddan, line 78: It appears you are overriding initializers, but also using `ExternalName` in making calls. I think you will likely want remove this line or include `NewNameAsExternalName` and any other initializers you would need, ref: https://github.com/crossplane/provider-aws/blob/276140071bc332832593952092bcb2507af56f23/pkg/controller/database/rdsinstance.go#L73

pkg/controller/applicationintegration/sqs/contoller.go
```diff
@@ -0,0 +1,244 @@
...
+
+	cr.SetConditions(runtimev1alpha1.Creating())
+
+	// FIFO queues names should start with ".fifo"
+	if aws.BoolValue(cr.Spec.ForProvider.FifoQueue) && !strings.HasSuffix(cr.Spec.ForProvider.Name, fifoQueueSuffix) {
+		return managed.ExternalCreation{}, errors.New(errInvalidNameForFifoQueue)
+	}
+
+	createResp, err := e.client.CreateQueueRequest(&awssqs.CreateQueueInput{
+		Attributes: sqs.GenerateCreateAttributes(&cr.Spec.ForProvider),
```
hasheddan, line 172: Representing these attributes as a map is an unfortunate design choice in the AWS SDK :/ Thanks for making the most of this! :)

pkg/controller/applicationintegration/sqs/contoller.go
sahil-lakhwani: The `ExternalName` is being set in the `Create` method, with the value of `QueueURL`. That is being done to not depend on Status to maintain `QueueURL`, which is the identifier for the queue.

apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+// Tag is a key value pairs attached to a Amazon SQS queue.
+type Tag struct {
+
+	// The key name that can be used to look up or retrieve the associated value.
+	Key string `json:"key"`
+
+	// The value associated with a key in a tag.
+	Value string `json:"value,omitempty"`
```
negz, line 56: This has an `omitempty` tag, but isn't a pointer or marked `// +optional`. Is it an optional field?

apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+type QueueParameters struct {
+	// The length of time, in seconds, for which the delivery
+	// of all messages in the queue is delayed.
+	// +optional
+	DelaySeconds *int64 `json:"delaySeconds,omitempty"`
+
+	// Designates a queue as FIFO.
+	// +immutable
+	// +optional
+	FifoQueue *bool `json:"fifoQueue,omitempty"`
```
negz, line 82:
```suggestion
	FIFOQueue *bool `json:"fifoQueue,omitempty"`
```

We want to try maintain the Go practice of using uppercase acrnonyms. Note this does apply to struct tags, which should always mirror the AWS API.

apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The length of time, in seconds, for which a ReceiveMessage
+	// action waits for a message to arrive.
+	// +optional
+	ReceiveMessageWaitTimeSeconds *int64 `json:"receiveMessageWaitTimeSeconds,omitempty"`
+
+	// RedrivePolicy includes the parameters for the dead-letter queue
+	// functionality of the source queue.
+	// +optional
+	RedrivePolicy RedrivePolicy `json:"redrivePolicy,omitempty"`
```
negz, line 103: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

I believe this should be a pointer.

apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The visibility timeout for the queue, in seconds.
+	// +optional
+	VisibilityTimeout *int64 `json:"visibilityTimeout,omitempty"`
+
+	// The ID of an AWS-managed customer master key (CMK) for Amazon SQS or a custom CMK.
+	// +optional
+	KmsMasterKeyID *string `json:"kmsMasterKeyId,omitempty"`
```
negz, line 111:
```suggestion
	KMSMasterKeyID *string `json:"kmsMasterKeyId,omitempty"`
```

apis/applicationintegration/v1alpha1/sqs_types.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// The length of time, in seconds, for which
+	// Amazon SQS can reuse a data key to encrypt or decrypt messages before calling AWS KMS again.
+	// +optional
+	KmsDataKeyReusePeriodSeconds *int64 `json:"kmsDataKeyReusePeriodSeconds,omitempty"`
```
negz, line 116:
```suggestion
	KMSDataKeyReusePeriodSeconds *int64 `json:"kmsDataKeyReusePeriodSeconds,omitempty"`
```

pkg/clients/sqs/sqs.go
```diff
@@ -0,0 +1,237 @@
...
+		for _, val := range tags {
+			m[val.Key] = val.Value
+		}
+		return m
+	}
+	return nil
+}
+
+// IsErrorNotFound checks if the error returned by AWS API says that the queue being probed doesn't exist
+func IsErrorNotFound(err error) bool {
```
negz, line 121: I _think_ other instances of this pattern mostly omit `Error` from the name - i.e. `IsNotFound`.

pkg/controller/applicationintegration/sqs/controller.go
```diff
@@ -0,0 +1,244 @@
...
+	createResp, err := e.client.CreateQueueRequest(&awssqs.CreateQueueInput{
+		Attributes: sqs.GenerateCreateAttributes(&cr.Spec.ForProvider),
+		QueueName:  aws.String(cr.Spec.ForProvider.Name),
+		Tags:       sqs.GenerateQueueTags(cr.Spec.ForProvider.Tags),
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
+	}
+
+	meta.SetExternalName(cr, aws.StringValue(createResp.QueueUrl))
```
negz, line 180: I see that the queue URL is the unique identifier for a queue, so we need to use that when observing, updating, and deleting the queue. The docs for the below API imply that it's possible to look up a queue URL using only a queue name. Could we store just the queue name in our external name annotation, then use the API below to lookup the URL based on that name when observing, updating, or deleting? This would let us get rid of the `spec.forProvider.name` field and just use the external name annotation, which would be ideal.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueUrl.html

pkg/controller/applicationintegration/sqs/controller.go
```diff
@@ -0,0 +1,244 @@
...
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(sqs.IsErrorNotFound, err), errGetQueueAttributesFailed)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Available())
+
+	cr.Status.AtProvider = v1alpha1.QueueObservation{
+		ARN: resAttributes.Attributes[v1alpha1.AttributeQueueArn],
+	}
```
negz, line 135: Does the `URL` field of `QueueObservation` get set anywhere?

sahil-lakhwani: I have updated this this code section, was some reason it is not being reflected here.

apis/applicationintegration/v1alpha1/sqs_types.go
sahil-lakhwani: I have updated this this code section, was some reason it is not being reflected here.

pkg/controller/applicationintegration/sqs/controller.go
sahil-lakhwani: I have made the required changes after setting the external-name to the queue name and removing the `Name` field from `ForProvider`

sahil-lakhwani:
@negz  I have addressed your review comments.

apis/applicationintegration/v1alpha1/sqs_types.go
negz: @sahil-lakhwani I can see the new `// +optional` tag but it still seems that this is a `string` and not a `*string`. Are you saying you've made that change but it's not showing up in GitHub for some reason?

negz:
@hasheddan Can you please approve this PR if you feel your comments were addressed in order to clear out your requested changes?

apis/applicationintegration/v1alpha1/sqs_types.go
sahil-lakhwani: Okay I had missed changing that field to pointer type. I have changed it in the new commit. Thanks for pointing that!

sahil-lakhwani:
> @sahil-lakhwani I'm seeing a failing test - could you please look into this.

@negz I have fixed this. 
(This was probably due to the upgrade to the AWS SDK after the rebase.)

negz:
Thanks @sahil-lakhwani!

PR #242, opened 2020-06-03
Pass fsGroup in install.yaml

### LINKED ISSUE #211: Re-enable support for IAM Roles for Service Accounts
### What happened?

To conform with Crossplane's new default security context for stacks (https://github.com/crossplane/crossplane/pull/1444), the `provider-aws` container now runs as non-root (https://github.com/crossplane/provider-aws/pull/202). Because we set the non-root user in the Dockerfile, users must rebuild the container to run as root user in order to be able to read the AWS credentials that are injected from the service account into `/var/run/secrets/` in the container.

### How can we fix it?

There are a few immediate options I could see here:

1. Create a separate release for running in "insecure mode". This could use a workaround like https://github.com/kubernetes-sigs/external-dns/pull/1185#issuecomment-530439786.
2. Add fields on `ClusterStackInstall` (or the next iteration of the installation unit). This seems like the most sustainable long-term solution as there is desire to move away from including an `install.yaml` in provider packages (https://github.com/crossplane/crossplane/issues/1441).
3. Document how a user may build their own package from source in [AUTHENTICATION.md](https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md). This is a good short-term solution.

hasheddan:

This re-enables support for IAM Roles for Service Accounts by passing an fsGroup,
meaning that the projected service account token volume will be owned by the GID and the GID will be added as a supplemental group to the container process.

Fixes #211 
Depends on https://github.com/crossplane/crossplane/pull/1577

### Checklist

I have:

### DIFF
--- a/config/package/manifests/install.yaml
+++ b/config/package/manifests/install.yaml
@@ -15,6 +15,8 @@ spec:
       labels:
         core.crossplane.io/name: "provider-aws"
     spec:
+      securityContext:
+        fsGroup: 65534
       containers:
       - name: "provider-aws-controller"
         env:
--- a/config/package/manifests/install.yaml
+++ b/config/package/manifests/install.yaml
@@ -16,6 +16,10 @@ spec:
         core.crossplane.io/name: "provider-aws"
     spec:
       securityContext:
+        # TODO(hasheddan): this can be removed when there is no desire
+        # to support pre-v1.19 Kubernetes clusters
+        # https://github.com/kubernetes/enhancements/pull/1598
+        # https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8
         fsGroup: 65534
       containers:
       - name: "provider-aws-controller"


config/package/manifests/install.yaml
```diff
@@ -15,6 +15,12 @@ spec:
       labels:
         core.crossplane.io/name: "provider-aws"
     spec:
+      securityContext:
+        # TODO(hasheddan): this can be removed when there is no desire
+        # to support pre-v1.19 Kubernetes clusters
+        # https://github.com/kubernetes/enhancements/pull/1598
+        # https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8
+        fsGroup: 65534
```
jbw976, line 23: is this fully compatible with the last official release of crossplane (v0.11)? i.e. if someone installs this latest provider-aws but hasn't updated crossplane to a newer master version, will it work OK?

hasheddan: @jbw976 this will not break `master` of `provider-aws` with Crossplane `v0.11`/`v0.11.1`, but this will be stripped out meaning that the IAM Role for Service Accounts will not work until someone updates their Crossplane to `master`. So in short, this will not introduce any regressions, but will also not be a fix without updating Crossplane. (One could manually edit this as well I think to get the functionality with older Crossplane versions though).

jbw976: Yes that sounds perfect.  Totally fine with it not being "fixed" until their crossplane version is also updated. The important part is that they won't experience any crashes or complete breaks by using this new provider-aws update without also updating crossplane.

PR #244, opened 2020-06-09
Added AWS SNS Notification API

### LINKED ISSUE #175: Simple Notification Service support
SNS support 

Add SNS managed resource

### References
*

rahulchheda:

Fixes #175

    - Added SNS Topic CRUD methods + tests
    - Added SNS Subscription CRUD methods + tests

### Checklist

I have:


rahulchheda:
TODO: 
    - fix CI Build Issue
    - fix Cyclometric Complexity a Observe func SNSSubscription (make reviewable)
    - Add resource in app.yaml

rahulchheda:
**FORCED PUSHED ON THIS BRANCH WITHOUT ANY CHANGES**
Had some issues with rebasing.

apis/notification/v1alpha1/snssubscription_types.go
```diff
@@ -0,0 +1,102 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SNSSubscriptionParameters define the desired state of a AWS SNS Topic
+type SNSSubscriptionParameters struct {
+
+	// TopicArn is the Arn of the SNS Topic
+	// +immutable
+	TopicArn string `json:"topicArn,omitempty"`
```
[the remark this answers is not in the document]
negz: It should also be a string pointer (`*string`), and ideally be named `TopicARN` to comply with Go rules about acronyms. Note that the struct tag should stay identical with the AWS API, which means it should still be `topicArn`. The acronym comment applies to some other field names - e.g. `KMS` and not `Kms`

apis/notification/v1alpha1/snssubscription_types.go
```diff
@@ -0,0 +1,143 @@
...
+	// +optional
+	TopicArnSelector *runtimev1alpha1.Selector `json:"topicArnSelector,omitempty"`
+
+	// The subscription's protocol.
+	// +immutable
+	Protocol *string `json:"protocol"`
+
+	// The subscription's endpoint
+	// +immutable
+	Endpoint *string `json:"endpoint"`
```
negz, line 47: It appears that Endpoint and Protocol are not optional. If so, they should be strings and not pointers to string.

apis/notification/v1alpha1/snssubscription_types.go
```diff
@@ -0,0 +1,143 @@
...
+	ConfirmationPending ConfirmationStatus = "ConfirmationPending"
+	// ConfirmationSuccessful represents confirmed SNS Subscription
+	ConfirmationSuccessful ConfirmationStatus = "Confirmed"
+)
+
+// SNSSubscriptionObservation represents the observed state of a AWS SNS Topic
+type SNSSubscriptionObservation struct {
+
+	// The subscription's owner.
+	// +optional
```
negz, line 97: I don't think we usually worry too much about the `// +optional` convention for status messages (that are never written by humans). That being said, it doesn't hurt anything to have them there.

apis/notification/v1alpha1/snstopic_types.go
```diff
@@ -0,0 +1,147 @@
...
+	// Department could have values such as Human Resources, Accounting, and Support.
+	// Tags with a key name of Cost Center might have values that consist of the
+	// number associated with the different cost centers in your company. Typically,
+	// many resources have tags with the same key name but with different values.
+	//
+	// AWS always interprets the tag Value as a single string. If you need to store
+	// an array, you can store comma-separated values in the string. However, you
+	// must interpret the value in your code.
+	// +optional
+	Value string `json:"value,omitempty"`
```
negz, line 45: I think this should be a string pointer since it's optional.

apis/notification/v1alpha1/snstopic_types.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// SNSTopicParameters define the desired state of a AWS SNS Topic
+type SNSTopicParameters struct {
+	// Name refers to the name of the AWS SNS Topic
+	// +immutable
+	Name string `json:"name"`
```
negz, line 52: Could we use the external name annotation for this?

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,184 @@
...
+	//SNSSubscriptionNotFound is the error code that is returned if SNS Subscription is not present
+	SNSSubscriptionNotFound = "InvalidSNSSubscription.NotFound"
+
+	// SNSSubscriptionInPendingConfirmation will be raise when SNS Subscription is found
+	// but in "pending confirmation" state.
+	SNSSubscriptionInPendingConfirmation = "InvalidSNSSubscription.PendingConfirmation"
+)
+
+// IsErrorSubscriptionNotFound returns true if the error code indicates that the item was not found
+func IsErrorSubscriptionNotFound(err error) bool {
```
negz, line 41: Nit: I don't think we usually include `Error` in these kind of checks - so this would just be `IsSubscriptionNotFound`. Ditto for other similar checks in this package.

apis/notification/v1alpha1/snstopic_types.go
rahulchheda: So, this Name field represents the name of SNS Topic, and the external name is updated with the ARN of this SNS Topic. If topicARN is not needed as external-name, we can shift it to status.
Not fixing this as of now, if you agree with the changes, I can go ahead.

rahulchheda:
@negz I have made all the appropriate changes.

negz:
@rahulchheda It looks like there's an in-progress conflict resolution checked into this PR. Please work with @sahil-lakhwani to get this PR to a point where he's satisfied with it, then let me know when it is ready for me to review again.

sahil-lakhwani:
> @rahulchheda It looks like there's an in-progress conflict resolution checked into this PR. Please work with @sahil-lakhwani to get this PR to a point where he's satisfied with it, then let me know when it is ready for me to review again.

@negz  -  @rahulchheda has addressed all the changes that I could request.

pkg/controller/notification/snssubscription/controller.go
```diff
@@ -0,0 +1,243 @@
...
+		return managed.ExternalObservation{},
+			errors.Wrap(resource.Ignore(sns.IsSubscriptionNotFound, err), errList)
+	}
+
+	// Filters the list of subscription with matching values in CR
+	sub, err := snsclient.GetSNSSubscription(subscriptionList, cr)
+	if err != nil {
+		return managed.ExternalObservation{},
+			errors.Wrap(resource.Ignore(sns.IsSubscriptionNotFound, err), errGetSubscription)
+	}
```
negz, line 136: It seems like this observe function does roughly two things:

1. Lists all subscriptions to the topic ARN of our subscription in order to determine whether our subscription exists.
1. Looks up the attributes of our subscription by its ARN in order to propagate its status.

According to https://docs.aws.amazon.com/sns/latest/api/API_GetSubscriptionAttributes.html the GetSubscriptionAttributesRequest will return HTTP 404 if the subscription does not exist. Could we therefore skip listing the subscriptions, and simply infer that the subscription did not exist if we called GetSubscriptionAttributesRequest and got a 404?

pkg/controller/notification/snssubscription/controller.go
```diff
@@ -0,0 +1,243 @@
...
+}
+
+func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
+	cr, ok := mgd.(*v1alpha1.SNSSubscription)
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+
+	if !awsarn.IsARN(meta.GetExternalName(cr)) && *cr.Status.AtProvider.Status == v1alpha1.ConfirmationPending {
+		return errors.New("Can't delete a subscription in PendingConfirmation state")
```
negz, line 233: Should this error string be a constant like the others? Also, I don't think it should start with a capital per https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,183 @@
...
+			changedAttrs[k] = v
+		}
+	}
+
+	return changedAttrs
+}
+
+// IsSNSSubscriptionUpToDate checks if object is up to date
+func IsSNSSubscriptionUpToDate(p v1alpha1.SNSSubscriptionParameters, sub *sns.Subscription, subAttributes map[string]string) bool {
+	return p.Endpoint == aws.StringValue(sub.Endpoint) && p.Protocol == aws.StringValue(sub.Protocol) && isSNSSubscriptionAttributesUpToDate(p, subAttributes)
```
negz, line 173: Is there much point comparing endpoint and protocol here? I believe they're immutable, so we wouldn't actually update them even if they weren't up to date. In fact, we'd probably enter a tight reconcile loop because we'd continually consider them to need updating, but not update them.

pkg/controller/notification/snssubscription/controller.go
negz: Ah - I see that we use the returned `sub` to pass to `IsSNSSubscriptionUpToDate`. I don't think we need to do this though - see my comment on that function.

pkg/clients/sns/snstopic.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+
+	return input
+}
+
+// LateInitializeTopic fills the empty fields in *v1alpha1.SNSTopicParameters with the
+// values seen in sns.Topic.
+func LateInitializeTopic(in *v1alpha1.SNSTopicParameters, topic sns.Topic, attrs map[string]string) {
+	in.Name = *awsclients.LateInitializeStringPtr(&in.Name, topic.TopicArn)
```
negz, line 100: Do we really want to late initialise the name to the ARN? Aren't they different things? Is there ever a case that the name would be empty in our CR? It seems like a required and immutable field.

pkg/controller/notification/snstopic/controller.go
```diff
@@ -0,0 +1,221 @@
...
+	// their ARN.
+	if !awsarn.IsARN(meta.GetExternalName(cr)) {
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
+
+	// Fetch the list of Topics
+	topicList, err := e.client.ListTopicsRequest(&awssns.ListTopicsInput{}).Send(ctx)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(sns.IsTopicNotFound, err), errList)
+	}
```
negz, line 124: Similar to my comment about subscriptions, is there a value in listing topics? it seems like we could just go straight to calling `GetTopicAttributesRequest` and determine whether the topic exists based on whether that API call returns a HTTP 404.

I see that we select a listed topic and use that for `LateInitializeTopic`, but per my comment on that function I don't think we need to.

apis/notification/v1alpha1/snstopic_types.go
negz: Got it - upon deeper review the pattern you have now makes sense. Thanks!

negz:
Thanks @rahulchheda and @sahil-lakhwani!

negz:
Hi @rahulchheda,

Apologies if I wasn't explicit with my last review pass, but I'm waiting on you to address my last round of review feedback. Please comment on this PR once you've addressed the feedback and are waiting on another review.

rahulchheda:
Got it, will make it reviewable ASAP, I think I missed those comments

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,183 @@
...
+
+// getSubAttributes returns map of SNS Sunscription Attributes
+func getSubAttributes(p v1alpha1.SNSSubscriptionParameters) map[string]string {
+
+	attr := make(map[string]string)
+
+	attr["DeliveryPolicy"] = aws.StringValue(p.DeliveryPolicy)
+	attr["FilterPolicy"] = aws.StringValue(p.FilterPolicy)
+	attr["RawMessageDelivery"] = aws.StringValue(p.RawMessageDelivery)
+	attr["RedrivePolicy"] = aws.StringValue(p.RedrivePolicy)
```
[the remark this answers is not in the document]
rahulchheda: Yes, these exists in SNSTopicAttributes.
ref: https://docs.aws.amazon.com/cli/latest/reference/sns/get-subscription-attributes.html#output

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,183 @@
...
+	in.RedrivePolicy = awsclients.LateInitializeStringPtr(in.RedrivePolicy, aws.String(subAttributes["RedrivePolicy"]))
+}
+
+// GetSNSSubscription returns SNSSubscription from List of SNSSubscription
+func GetSNSSubscription(res *sns.ListSubscriptionsByTopicResponse, cr *v1alpha1.SNSSubscription) (sns.Subscription, error) {
+
+	for _, sub := range res.Subscriptions {
+		if cmp.Equal(sub.TopicArn, cr.Spec.ForProvider.TopicARN) && cmp.Equal(sub.Endpoint, &cr.Spec.ForProvider.Endpoint) && cmp.Equal(sub.Protocol, &cr.Spec.ForProvider.Protocol) {
+			return sub, nil
+		}
```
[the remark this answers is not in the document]
rahulchheda: So, the Sub ARN, is included in it's attributes, not when we list the Subscriptions. That's why validating all the info gathered from the items in the list of Sub ARN (when filtering it through TopicARN). The only redundant guy here is TopicARN (as we are checking it again after using it a filter).

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,183 @@
...
+	return p.Endpoint == aws.StringValue(sub.Endpoint) && p.Protocol == aws.StringValue(sub.Protocol) && isSNSSubscriptionAttributesUpToDate(p, subAttributes)
+}
+
+// isSNSSubscriptionAttributesUpToDate checks if attributes are up to date
+func isSNSSubscriptionAttributesUpToDate(p v1alpha1.SNSSubscriptionParameters, subAttributes map[string]string) bool {
+	return *p.DeliveryPolicy == subAttributes["DeliveryPolicy"] &&
+		*p.FilterPolicy == subAttributes["FilterPolicy"] &&
+		*p.RawMessageDelivery == subAttributes["RawMessageDelivery"] &&
+		*p.RedrivePolicy == subAttributes["RedrivePolicy"] &&
+		*p.TopicARN == subAttributes["TopicArn"]
```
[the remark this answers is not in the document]
rahulchheda: In my testing rounds, it comes up as empty, rather than non-nil, so yeah this code won't panic. I can add validation of this, if necessary.

pkg/clients/sns/snssubscription.go
```diff
@@ -0,0 +1,183 @@
...
+	SNSSubscriptionNotFound = "InvalidSNSSubscription.NotFound"
+
+	// SNSSubscriptionInPendingConfirmation will be raise when SNS Subscription is found
+	// but in "pending confirmation" state.
+	SNSSubscriptionInPendingConfirmation = "InvalidSNSSubscription.PendingConfirmation"
+)
+
+// IsSubscriptionNotFound returns true if the error code indicates that the item was not found
+func IsSubscriptionNotFound(err error) bool {
+	if _, ok := err.(*SubscriptionNotFound); ok {
```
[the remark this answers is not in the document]
rahulchheda: Removed this.

rahulchheda:
Apologies @negz for delaying this PR Review. I have tried to resolve most of your comments, added comments on a few. Let me if some improvement needed.

Issue #148, opened 2020-03-03
Replace Web Identity Token logic on next release of aws-sdk-go-v2


hasheddan:
### What problem are you facing?

We currently assume IAM roles for authentication to the AWS API using our own custom logic to get Web Identity Tokens from the file-system of the controller container. The v1 version of `aws-sdk-go` supported this use-case as an identity provider, but the v2 version did not, so we were forced to use our own implementation in the short-term.

### How could Crossplane help solve your problem?

This logic has now been included in `aws-sdk-go-v2` as part of https://github.com/aws/aws-sdk-go-v2/pull/488. It should be available in the next release (`v0.20.0`), and we should move to using it instead of our own implementation.

hasheddan:
As of [v0.20.0](https://github.com/aws/aws-sdk-go-v2/releases/tag/v0.20.0), the Web Identity Credential Provider has been added. We should bump our dependency on the AWS SDK and use this functionality instead of our own implementation: https://github.com/crossplane/provider-aws/blob/62cca4c8941462a65d6499d6bba43b7cd84cdc38/pkg/clients/aws.go#L108

haarchri:
@hasheddan is this still relevant ?

PR #245, opened 2020-06-09
EKS v1beta1 Cluster

hasheddan:

This still needs some work, but it is functioning for Create / Update / Delete. I have not deleted the existing `v1alpha3` `EKSCluster` for now because I plan on implementing the `Cluster` and `NodeGroup` types separately. 

~Unit tests are still needed across the board for client / controller.~

~Manual testing is in progress.~

Manually tested by creating the YAML manifest added in `examples/eks` and was able to successfully Create / Update / Delete. Also verified that `kubeconfig` field in connection `Secret` allows connection to the cluster when Public access is enabled :+1: 

@janwillies @Henni let me know if yall would be interested in helping out with manual testing here :+1: 

### Checklist

I have:


hasheddan:
Commits will need to be tidied up before merge

config/crd/eks.aws.crossplane.io_clusters.yaml
```diff
@@ -0,0 +1,558 @@
...
+                resourceVersion:
+                  description: 'Specific resourceVersion to which this reference is
+                    made, if any. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency'
+                  type: string
+                uid:
+                  description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
+                  type: string
+              type: object
+            forProvider:
+              description: ClusterParameters define the desired state of an AWS Elastic
```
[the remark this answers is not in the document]
hasheddan: Thanks for the review! Most comments have been addressed :) Could you expound on what you mean here?

apis/eks/v1beta1/types.go
```diff
@@ -0,0 +1,313 @@
...
+// ClusterObservation is the observed state of a cluster.
+type ClusterObservation struct {
+	// The Amazon Resource Name (ARN) of the cluster.
+	Arn string `json:"arn,omitempty"`
+
+	// The Unix epoch timestamp in seconds for when the cluster was created.
+	CreatedAt *metav1.Time `json:"createdAt,omitempty"`
+
+	// The endpoint for your Kubernetes API server.
+	Endpoint string `json:"endpoint,omitempty"`
```
hasheddan, line 230: This is included in the status of many of our managed resources, but I debated whether we should keep it here or not. While it is not _super_ sensitive information, it is included in the connection `Secret` and I am not sure there is a use-case where someone would want the `Endpoint` without wanting to connect.

apis/eks/v1beta1/types.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// The identity provider information for the cluster.
+	Identity Identity `json:"identity,omitempty"`
```
hasheddan, line 233: Another field that I debated whether including in the status or not. It is not in our connection `Secret` currently so I left it.

pkg/clients/aws.go
```diff
@@ -67,12 +67,12 @@ func CredentialsIDSecret(data []byte, profile string) (string, string, error) {
 		return "", "", err
 	}
 
-	id, err := iniProfile.GetKey(external.AWSAccessKeyIDEnvVar)
+	id, err := iniProfile.GetKey("AWS_ACCESS_KEY_ID")
```
hasheddan, line 70: These variables are no longer exported in the v0.23 AWS SDK.

hasheddan: I think we can update how this is done as the SDK now supports credentials chain, but I would prefer to defer that to a later PR.

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,335 @@ package eks
...
-}
-
-func getMostRecentImage(images []*ec2.Image) *ec2.Image {
-	var result *ec2.Image
-	for i := 0; i < len(images); i++ {
-		if result == nil || aws.StringValue(result.CreationDate) < aws.StringValue(images[i].CreationDate) {
-			result = images[i]
+	// NOTE(hasheddan): AWS removes insignificant bits from CIDRs, so we must
+	// compare by converting user-supplied CIDRs to network blocks. We only skip
+	// comparison if both external and local have no CIDR blocks defined.
```
hasheddan, line 275: Taking a look at `TestIsUpToDate` should paint a better picture of what is happening here.

pkg/controller/cache/cachesubnetgroup/controller_test.go
```diff
@@ -270,7 +270,7 @@ func TestObserve(t *testing.T) {
 				cache: &fake.MockClient{
 					MockDescribeCacheSubnetGroupsRequest: func(input *awscache.DescribeCacheSubnetGroupsInput) awscache.DescribeCacheSubnetGroupsRequest {
 						return awscache.DescribeCacheSubnetGroupsRequest{
-							Request: &aws.Request{HTTPRequest: &http.Request{}, Data: &awscache.DescribeCacheSubnetGroupsOutput{
+							Request: &aws.Request{HTTPRequest: &http.Request{}, Retryer: aws.NoOpRetryer{}, Data: &awscache.DescribeCacheSubnetGroupsOutput{
```
hasheddan, line 273: This `aws.NoOpRetryer{}` is the only change in many of these test files. It is simply an artifact of bumping the SDK version. Feel free to skip over these changes.

pkg/controller/eks/cluster.go
```diff
@@ -0,0 +1,240 @@
...
+	rsp, err := e.client.DescribeClusterRequest(&awseks.DescribeClusterInput{Name: aws.String(meta.GetExternalName(cr))}).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, errDescribeFailed)
+	}
+	patch, err := eks.CreatePatch(rsp.Cluster, &cr.Spec.ForProvider)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, errPatchCreationFailed)
+	}
+	if _, err = e.client.UpdateClusterConfigRequest(eks.GenerateUpdateClusterConfigInput(meta.GetExternalName(cr), patch)).Send(ctx); err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(eks.IsErrorInUse, err), errUpdateConfigFailed)
```
hasheddan, line 196: Note that I am ignoring `IsErrorInUse` so that we do a full wait if another Update operation is still underway. This means that when this error occurs we will report in logs / status that a successful Update request was made. I think this is fine as the only time this error will occur is when we are already updating some other field, and we will eventually finish that update and then try this one again.

pkg/controller/eks/secret.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/secret"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/crossplane/crossplane/apis/compute/v1alpha1"
+	workloadv1alpha1 "github.com/crossplane/crossplane/apis/workload/v1alpha1"
+
+	"github.com/crossplane/provider-aws/apis/eks/v1beta1"
+)
+
+// SetupClusterSecret adds a controller that propagates EKS Cluster connection
```
hasheddan, line 34: This and the target controller are essentially directly copied from the old EKS implementation.

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,335 @@ package eks
...
 }
 
-func getImageWithID(imgName string, images []*ec2.Image) (*ec2.Image, error) {
-	for _, img := range images {
-		if imgName == aws.StringValue(img.ImageId) {
-			return img, nil
-		}
+// GetConnectionDetails extracts managed.ConnectionDetails out of eks.Cluster.
+func GetConnectionDetails(cluster *eks.Cluster, stsClient STSClient) managed.ConnectionDetails {
+	if cluster == nil || cluster.Name == nil || cluster.Endpoint == nil || cluster.CertificateAuthority == nil || cluster.CertificateAuthority.Data == nil {
```
hasheddan, line 308: This is a pretty verbose check here, and while working on this controller I notice that we make assumptions in a lot of places that we will not get `nil` values for subfields when the parent is not `nil`. In practice this works fine because AWS returns these values or we check an error before dereferencing all the time. However, it makes me a little uncomfortable that we don't always check every `nil` pointer. In this case, I do this check because I don't want to write a connection details until I have everything we need for a valid `kubeconfig`.

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,338 @@ package eks
...
+		if len(in.ResourcesVpcConfig.SecurityGroupIDs) == 0 && len(cluster.ResourcesVpcConfig.SecurityGroupIds) > 0 {
+			in.ResourcesVpcConfig.SecurityGroupIDs = cluster.ResourcesVpcConfig.SecurityGroupIds
+		}
+		if len(in.ResourcesVpcConfig.SubnetIDs) == 0 && len(cluster.ResourcesVpcConfig.SubnetIds) > 0 {
+			in.ResourcesVpcConfig.SubnetIDs = cluster.ResourcesVpcConfig.SubnetIds
+		}
 	}
+	in.RoleArn = awsclients.LateInitializeStringPtr(in.RoleArn, cluster.RoleArn)
+	in.Version = awsclients.LateInitializeStringPtr(in.Version, cluster.Version)
+}
```
muvaf, line 267: Looks like `Tags` are not late-initialized.

pkg/controller/eks/cluster.go
```diff
@@ -0,0 +1,240 @@
...
+	cr, ok := mg.(*v1beta1.Cluster)
+	if !ok {
+		return managed.ExternalCreation{}, errors.New(errNotEKSCluster)
+	}
+	cr.SetConditions(runtimev1alpha1.Creating())
+	if cr.Status.AtProvider.Status == v1beta1.ClusterStatusCreating {
+		return managed.ExternalCreation{}, nil
+	}
+	_, err := e.client.CreateClusterRequest(eks.GenerateCreateClusterInput(meta.GetExternalName(cr), &cr.Spec.ForProvider)).Send(ctx)
+	return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
```
muvaf, line 172: It could be worth skipping `AlreadyExists` error if the `Create` request is not idempotent.

apis/eks/v1beta1/types.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// The metadata to apply to the cluster to assist with categorization and organization.
+	// Each tag consists of a key and an optional value, both of which you define.
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// The desired Kubernetes version for your cluster. If you don't specify a value
+	// here, the latest version available in Amazon EKS is used.
+	// +optional
+	Version *string `json:"version,omitempty"`
```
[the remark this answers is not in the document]
muvaf: Is there a format specification for this field? It'd be easier for people to know what's expected as version.

pkg/clients/eks/eks.go
muvaf: We should always check for `nil` if the type is a pointer type. Are there places in other parts of the codebase that you realized we don't check for `nil` even though we should?

pkg/controller/eks/cluster.go
muvaf: If these requests cause the state to go into `Updating` state, then it seems that should be fine after the next reconcile. However, I'm wondering whether an empty `UpdateClusterConfigRequest` will cause that as well, which might render the rest of the logic useless.

pkg/clients/aws.go
muvaf: Is there an issue tracking this?

apis/eks/v1beta1/types.go
hasheddan: @muvaf What I am saying here is that I don't know if it should be considered "non-sensitive" data. There is a spectrum of sensitivity, and I would say the API server endpoint is not the most sensitive field, but I also wouldn't make it visible for no reason. I think it is likely fine to leave for now, but just wanted to start the conversation :+1:

pkg/clients/eks/eks.go
```diff
@@ -1,11 +1,11 @@
 /*
-Copyright 2019 The Crossplane Authors.
+Copyright 2020 The Crossplane Authors.
```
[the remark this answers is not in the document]
hasheddan: @muvaf yes, what happened here is that the old file was renamed, then this one got created so it appears as a diff rather than a new file. I am happy to change it back to `2019` if you think necessary.

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
hasheddan: @muvaf this actually just takes `1.15`, `1.16`, etc. and the EKS platform version is a separate field in the status. I have tested providing this manually and updating it to new version. I am not sure we should impose a restriction here as AWS does not specifically give one and it would increase maintenance burden if we had to keep a list of valid values here.

pkg/clients/aws.go
hasheddan: @muvaf I will tackle it as part of https://github.com/crossplane/provider-aws/issues/148

pkg/controller/eks/cluster.go
hasheddan: @muvaf I can test again, but given that I was able to update the cluster version, which is a separate call after this one, I am guessing that an empty update is just ignored

pkg/clients/eks/eks.go
hasheddan: Yes, and I believe we had a discussion about this on a previous issue. This isn't exactly checking for `nil`, but we have a few areas like this where we might, for example, not explicitly check that we don't have an array out of bounds  https://github.com/crossplane/provider-aws/blob/3238aeef3ccc0dab5672482577749432d2d60082/pkg/controller/database/rdsinstance.go#L136

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,338 @@ package eks
...
+				return false, nil
+			}
 		}
 	}
 
-	return result
+	return cmp.Equal(&v1beta1.ClusterParameters{}, patch, cmpopts.EquateEmpty(),
+		cmpopts.IgnoreTypes(&v1alpha1.Reference{}, &v1alpha1.Selector{}),
+		cmpopts.IgnoreFields(v1beta1.ClusterParameters{}, "Tags"),
+		cmpopts.IgnoreFields(v1beta1.VpcConfigRequest{}, "SecurityGroupIDRefs", "SubnetIDRefs", "PublicAccessCidrs")), nil
```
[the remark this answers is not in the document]
hasheddan: @muvaf yes this explicit ignore is needed here because the type is actually `[]v1alpha1.Reference`

pkg/clients/eks/eks.go
hasheddan: I was looking at `RDSInstance` and noticed that we don't late initialize tags there either, and I think that may be intentional. When we do the tagging in the `Initializer`, we add the tags directly to the managed resource, then they get propagated to the external infrastructure. Therefore, we will never have an empty tag map, so if we late initialize tags how are we supposed to reason about what should be there or not? Let's say a tag gets added in the AWS UI, it gets added to the managed resource by late initialization, then we want to remove it. If we explicitly remove it from the managed resource, we will hit `Observe` again, and late initialize the tag right back onto the resource, then see no diff and never make the update.

pkg/controller/eks/cluster.go
hasheddan: @muvaf I think this will be handled by checking if the resource is in `CREATING` state in the line above right? On this topic though, I dropped a comment on another PR about the ability to provide an `idempotencyToken` on some AWS resources. I opted not to use it here because it seems that these requests are already idempotent, and reasoning about when to change the token actually adds a pretty significant layer of complexity that may require further investigation about addressing idempotency across all Crossplane. However, wanted to get your take on it: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.23.0/service/eks?tab=doc#CreateClusterInput

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,338 @@ package eks
...
+			if !netMap[pc] {
+				return false, nil
+			}
 		}
 	}
 
-	return result
+	return cmp.Equal(&v1beta1.ClusterParameters{}, patch, cmpopts.EquateEmpty(),
+		cmpopts.IgnoreTypes(&v1alpha1.Reference{}, &v1alpha1.Selector{}),
+		cmpopts.IgnoreFields(v1beta1.ClusterParameters{}, "Tags"),
```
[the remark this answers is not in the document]
hasheddan: This was another place I was looking at `RDSInstance`.
https://github.com/crossplane/provider-aws/blob/3238aeef3ccc0dab5672482577749432d2d60082/pkg/clients/rds/rds.go#L448
The result here is that `Tags` never get updated unless something else changes in the `Spec`, which is not desirable. However, I believe the reason this is happening is because there will always be a diff with tags because of the way late initialization is ignoring them (per my earlier comment). We could probably fix this in both places by just doing a totally separate check on the tags. What do you think?

apis/eks/v1beta1/types.go
```diff
@@ -0,0 +1,313 @@
...
+	// The cluster control plane logging configuration for your cluster.
+	ClusterLogging []LogSetup `json:"clusterLogging"`
+}
+
+// LogSetup specifies the logging types that are enabled.
+type LogSetup struct {
+	// If a log type is enabled, that log type exports its control plane logs to
+	// CloudWatch Logs. If a log type isn't enabled, that log type doesn't export
+	// its control plane logs. Each individual log type can be enabled or disabled
+	// independently.
```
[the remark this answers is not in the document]
hasheddan: I believe it should be `// +optional` so i will update. Though it does feel kind of weird to just provide a list of logging types that you want disabled without explicitly saying `Enabled: false`. That being said, I don't make the rules here, AWS does :sweat_smile:

apis/eks/v1beta1/types.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// RoleArnRef is a reference to an IAMRole used to set
+	// the RoleArn.
+	// +immutable
+	// +optional
+	RoleArnRef *runtimev1alpha1.Reference `json:"roleArnRef,omitempty"`
+
+	// RoleArnSelector selects references to IAMRole used
+	// to set the RoleArn.
+	// +immutable
```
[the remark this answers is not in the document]
hasheddan: @muvaf agreed, should the reference also be immutable though? What if you reference something that doesn't exist and you want to change it? This actually brings up a larger discussion on these fields. Most of them are not `immutable` after creation of the managed resources, they are `immutable` after creation of the external infrastructure. If you have an invalid parameter for an `immutable` field we should probably let you update it (I realize that right now these immutable tags aren't really doing anything, but in the future they would theoretically represent an immutable field and be checked by API server. Maybe we are okay just saying you have to totally recreate the object if you provide an invalid value for an immutable field?). cc @negz @kasey

pkg/clients/eks/eks.go
```diff
@@ -19,702 +19,338 @@ package eks
...
-		"ClusterName":                      name,
-		"VpcId":                            spec.VPCID,
-		"Subnets":                          subnetIDs,
-		"NodeImageId":                      aws.StringValue(ami.ImageId),
-		"NodeInstanceType":                 spec.WorkerNodes.NodeInstanceType,
-		"BootstrapArguments":               spec.WorkerNodes.BootstrapArguments,
-		"NodeGroupName":                    spec.WorkerNodes.NodeGroupName,
-		"ClusterControlPlaneSecurityGroup": spec.WorkerNodes.ClusterControlPlaneSecurityGroup,
+	if len(p.EncryptionConfig) > 0 {
+		config := make([]eks.EncryptionConfig, len(p.EncryptionConfig))
```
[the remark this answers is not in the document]
hasheddan: Good catch! Updated here and a few other similar places.

hasheddan:
Rebasing this since `v1beta1` network resources PR has now been merged

pkg/clients/eks/eks.go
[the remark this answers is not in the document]
hasheddan: @muvaf I have updated how this is handled in this commit: https://github.com/crossplane/provider-aws/pull/245/commits/6359cf52b4f23762e6dc37e9d40d9d1121889f0c

pkg/clients/eks/eks.go
muvaf: The reason we don't late-init there was that the received object didn't have the tags, you had to make another request, as far as I remember. Late-init right now does its operation as long as the map is empty and as you pointed out this is kind of pointless since it'll never be an empty map because of the tagging initializer. But I don't really feel comfortable when one part of the codebase makes an assumption about the other on that level, i.e. they should be complete on what they do separately from each other. For example, late-init function could be used in another context or tagger could be removed etc.

pkg/controller/eks/cluster.go
muvaf: > I think this will be handled by checking if the resource is in CREATING state in the line above right?

Yes, it's fine either way. I personally just skip if I know that in 100% of the cases I'd like to skip that error even though it's unlikely to occur. But it can be either way.

>  I opted not to use it here because it seems that these requests are already idempotent, and reasoning about when to change the token actually adds a pretty significant layer of complexity that may require further investigation about addressing idempotency across all Crossplane. However, wanted to get your take on it

Yeah, I noticed that in the API and I agree with you that it adds some complexity that may not be really necessary. Maybe `resourceVersion` could be used as token there but it's probably just fine as is.

pkg/clients/eks/eks.go
hasheddan: Hmm it seems like we are making the decision that Crossplane will _always_ add tags to a resource if it is an option. Are you suggesting that we check for nil tags here even though we know tags will never be nil? I get that it would be more consistent with the general purpose of late initialization, but it seems like we will always be using late initialization in this context which makes it feel extraneous.

hasheddan: ^ that being said, I am happy to add it if you feel strongly

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
muvaf: > agreed, should the reference also be immutable though? What if you reference something that doesn't exist and you want to change it? This actually brings up a larger discussion on these fields.

I think we need to define what's actually immutable here: the actual role that will be attached or vpc of the RDS instance that will be created in. In explicit value and reference cases, user's input is directly that; the actual role whether it's on AWS level (actual value) or Crossplane level (reference). But that's not true for selectors; the input is not the config itself, it's a set of instructions to find one. So, the final value that is set by selector on the reference field should be immutable but there is no much reason for instructions themselves to be immutable. If it finds one then the ref will be populated and that's it, but if it couldn't find one then new labels can be added and the result of that selection should be effective because we'd not be changing the actual config; it was never set.

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
muvaf: I was just wondering whether we should include a format in the comment like it needs to be `x.yy` format rather than `x.yy.zz` or actual eks version like `x.yy.zz.eks99`. If it's just plain `x.yy`, maybe add a word like `Example: 1.15`?

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
muvaf: Yeah, that does sound weird 😅

pkg/clients/eks/eks.go
muvaf: Yeah, this looks like something to update some time. It's not critical probably because it's meaninglessly an array; if `NotFound` it returns error anyway and it can never be `>1` elements since we give exact name but yeah, why not check if there is a possibility.

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
hasheddan: I see your point here, updated to only have the `Selector` not be `// +immutable`.

pkg/clients/eks/eks.go
hasheddan: Since we can remove the tag check in `IsUpToDate` it is much cleaner if we follow your suggestion here. Thanks for catching this! :)

apis/eks/v1beta1/types.go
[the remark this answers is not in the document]
hasheddan: I like the idea of the example, added :+1:

PR #253, opened 2020-06-19
Add support for AWS Redshift v1alpha1 

### LINKED ISSUE #230: Add Support for AWS Redshift

The support for AWS Redshift is missing from Crossplane right now.

We should add support for AWS Redshift through which we can create, manage, and update an AWS Redshift cluster with various clusters config parameters.

arush-sal:
### Checklist

I have:


pkg/controller/database/redshift/controller.go
```diff
@@ -0,0 +1,223 @@
...
+		For(&v1alpha1.Redshift{}).
+		Complete(managed.NewReconciler(
+			mgr, resource.ManagedKind(v1alpha1.RedshiftGroupVersionKind),
+			managed.WithExternalConnecter(
+				&connector{kube: mgr.GetClient(),
+					newClientFn: redshift.NewClient,
+					awsConfigFn: utils.RetrieveAwsConfigFromProvider},
+			),
+			managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),
+			managed.WithConnectionPublishers(),
```
muvaf, line 68: Related to our meeting about the secret; an empty set of connection publishers are sent. That's why the secret is never published.

arush-sal: The suggested change has been incorporated.

negz:
Hi @arush-sal, can you please get someone from your team to review this PR, then comment here when that review is addressed and you're ready for a Crossplane maintainer to review? Thanks

arush-sal:
@negz yes sure. I have requested @sahil-lakhwani for the same, awaiting his review.

pkg/clients/redshift/redshift.go
```diff
@@ -0,0 +1,504 @@
...
+		return o
+	}
+	if patch.ClusterVersion != nil {
+		o.ClusterVersion = p.ClusterVersion
+	}
+	// The request to modify publicly accessible options for the cluster cannot be made in the same request as other modifications
+	if patch.ElasticIP != nil || patch.PubliclyAccessible != nil {
+		o.ElasticIp = p.ElasticIP
+		o.PubliclyAccessible = p.PubliclyAccessible
+		return o
```
sahil-lakhwani, line 264: As per the comment the `ModifyClusterInput` should just have `ElasticIp` and `PubliclyAccessible` populated at this state, but I think `AllowVersionUpgrade` and `AutomatedSnapshotRetentionPeriod` may also be set from the above lines.

@arush-sal Can this be the case ever?

pkg/clients/redshift/redshift.go
```diff
@@ -0,0 +1,504 @@
...
+	}
+	if patch.Encrypted != nil {
+		o.Encrypted = p.Encrypted
+	}
+	// Enhanced VPC routing changes must be made in a different request than other changes to the cluster
+	if patch.EnhancedVPCRouting != nil {
+		o.EnhancedVpcRouting = p.EnhancedVPCRouting
+		return o
```
sahil-lakhwani, line 272: Same comment here.

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+	HSMConfigurationIdentifier *string `json:"hsmConfigurationIdentifier,omitempty"`
+
+	// IAMRoles is a list of AWS Identity and Access Management (IAM) roles that can be used
+	// by the cluster to access other AWS services. You must supply the IAM roles
+	// in their Amazon Resource Name (ARN) format. You can supply up to 10 IAM roles
+	// in a single request.
+	// A cluster can have up to 10 IAM roles associated with it at any time.
+	// kubebuilder:validation:MaxItems=10
+	// +optional
+	IAMRoles []string `json:"iamRoles,omitempty"`
```
sahil-lakhwani, line 202: I think you can add a reference to the IAMRole custom resource here.

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+	// The specified Availability Zone must be in the same AWS Region as the current endpoint.
+	// +optional
+	AvailabilityZone *string `json:"availabilityZone,omitempty"`
+
+	// ClusterParameterGroupName is the name of the DB cluster parameter group to use for the DB cluster.
+	// Default: The default Amazon Redshift cluster parameter group. For information
+	// about the default parameter group, go to Working with Amazon Redshift Parameter
+	// Groups (https://docs.aws.amazon.com/redshift/latest/mgmt/working-with-parameter-groups.html)
+	// +optional
+	ClusterParameterGroupName *string `json:"clusterParameterGroupName,omitempty"`
```
sahil-lakhwani, line 100: Would it make sense to have  ParameterGroup as a separate custom resource which can then by attached to clusters?

pkg/controller/database/redshift/controller.go
```diff
@@ -0,0 +1,218 @@
...
+func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
+	cr, ok := mgd.(*v1alpha1.Redshift)
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+	cr.SetConditions(runtimev1alpha1.Deleting())
+	if cr.Status.AtProvider.ClusterStatus == v1alpha1.StateDeleting {
+		return nil
+	}
+	_, err := e.Update(ctx, cr)
```
sahil-lakhwani, line 211: @arush-sal  Can you add some comment here describing why there's a call to `Update` her?

negz:
Thanks for reviewing @sahil-lakhwani! I'll wait for you to approve this PR before we have one of our maintainers review it.

pkg/clients/redshift/redshift.go
arush-sal: @sahil-lakhwani I have tested this and as the comment mention `"The request to modify publicly accessible options for the cluster cannot be made in the same request as other modifications"`. So either it can be a change to `ElasticIp` or `PubliclyAccessible` field and no other change or else there will be an error returned.

pkg/clients/redshift/redshift.go
arush-sal: Same as above. No other request/field or change can be sent in a request where we are also returning value to `EnhancedVpcRouting`

pkg/controller/database/redshift/controller.go
arush-sal: Yeah, I can't think of a reason why I did that. It certainly doesn't make sense, must be a typo. Sorry about that.

apis/database/v1alpha1/redshift_types.go
arush-sal: Yes certainly. But given the time and sheer number parameters it can accept, I decided to go ahead with this barebones structure for the `v1alpha1`. Leaving a little room for improvement around there for the next version iteration.

apis/database/v1alpha1/redshift_types.go
arush-sal: Indeed.

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+	//    * First character must be a letter.
+	//    * Cannot be a reserved word. A list of reserved words can be found in
+	//    Reserved Words (https://docs.aws.amazon.com/redshift/latest/dg/r_pg_keywords.html)
+	//    in the Amazon Redshift Database Developer Guide.
+	// +immutable
+	MasterUsername *string `json:"masterUsername"`
+
+	// NewClusterIdentifier is the new identifier you want to use for the cluster.
+	// +optional
+	NewClusterIdentifier *string `json:"newClusterIdentifier,omitempty"`
```
hasheddan, line 249: I think this should be the `external-name` annotation, so it should be omitted in body. https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#custom-resource-instance-name

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+	// for the cluster that is being created.
+	// Constraints:
+	//    * Must be between 8 and 64 characters in length.
+	//    * Must contain at least one uppercase letter.
+	//    * Must contain at least one lowercase letter.
+	//    * Must contain one number.
+	//    * Can be any printable ASCII character (ASCII code 33 to 126) except '
+	//    (single quote), " (double quote), \, /, @, or space.
+	// +optional
+	NewMasterUserPassword *string `json:"masterUserPassword,omitempty"`
```
hasheddan, line 234: Might be nice to add some notes about default behavior here if this field is not supplied.

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+	//
+	// Values: active, applying
+	HSMStatus HSMStatus `json:"hsmStatus,omitempty"`
+
+	// IAMRoles is a list of AWS Identity and Access Management (IAM) roles that can be used
+	// by the cluster to access other AWS services. You must supply the IAM roles
+	// in their Amazon Resource Name (ARN) format. You can supply up to 10 IAM roles
+	// in a single request.
+	// A cluster can have up to 10 IAM roles associated with it at any time.
+	IAMRoles []ClusterIAMRole `json:"iamRoles,omitempty"`
```
hasheddan, line 427: This and a few other fields that are in the `spec` appear to be duplicated in the `status`. They should be sufficient in the `spec` where they can be late initialized

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,713 @@
...
+// +kubebuilder:object:root=true
+
+// An Redshift is a managed resource that represents an AWS Redshift cluster.
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="STATE",type="string",JSONPath=".status.atProvider.clusterStatus"
+// +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,aws}
+type Redshift struct {
```
hasheddan, line 51: I think this resource should actually be `Cluster` and should be in the `redshift.aws.crossplane.io` group

pkg/clients/redshift/redshift.go
```diff
@@ -0,0 +1,504 @@
...
+func IsUpToDate(p v1alpha1.RedshiftParameters, cl redshift.Cluster) (bool, error) { // nolint:gocyclo
+	// We need to check it explicitly as redshift.Cluster can have multiple ClusterParameterGroups
+	found := isClusterParameterGroupNameUpdated(p.ClusterParameterGroupName, cl.ClusterParameterGroups)
+
+	// Check if it is a cluster rename request
+	if p.NewClusterIdentifier != nil && (aws.StringValue(p.NewClusterIdentifier) != aws.StringValue(cl.ClusterIdentifier)) {
+		return false, nil
+	}
+
+	// Since redshift.Cluster doesn't have a ClusterType field therefore determine its value basecd upon number of nodes.
```
hasheddan, line 137:
```suggestion
	// Since redshift.Cluster doesn't have a ClusterType field therefore determine its value based upon number of nodes.
```

pkg/clients/redshift/redshift.go
```diff
@@ -0,0 +1,504 @@
...
+	if aws.Int64Value(cl.NumberOfNodes) > 1 && aws.StringValue(p.ClusterType) != "multi-node" {
```
hasheddan, line 138: Do we need to check these explicitly if they are also being late initialized correctly in the `Patch` operation? https://github.com/crossplane/provider-aws/pull/253/files#diff-b96ea1b7a25a962d50f42ee655cc5283R78

pkg/clients/redshift/redshift_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+package redshift
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/redshift"
+	"github.com/google/go-cmp/cmp"
+
+	// runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
```
hasheddan, line 25: Is this intended to be removed?

pkg/clients/redshift/redshift_test.go
```diff
@@ -0,0 +1,161 @@
...
+	aws "github.com/crossplane/provider-aws/pkg/clients"
+)
+
+var (
+	dbName     = "example-name"
+	nodeType   = "dc1.large"
+	singleNode = "single-node"
+)
+
+func TestCreatePatch(t *testing.T) {
```
hasheddan, line 37: Looks like a few functions are implicitly tested, but would like to see some explicit tests for something like `LateInitialize`

pkg/controller/database/redshift/controller.go
```diff
@@ -0,0 +1,218 @@
...
+		ClusterIdentifier: aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(redshift.IsErrorNotFound, err), errDescribeFailed)
+	}
+
+	// Describe requests can be used with filters, which then returns a list.
+	// But we use an explicit identifier, so, if there is no error, there should
+	// be only 1 element in the list.
+	instance := rsp.Clusters[0]
```
hasheddan, line 112: I know this has been done elsewhere, but I would prefer that we check the length of the cluster slice instead of assuming that the API will always behave correctly.

pkg/controller/database/redshift/controller.go
hasheddan: This comment was marked as resolved but I still see `Update` here. Is it needed?

apis/database/v1alpha1/redshift_types.go
arush-sal: @hasheddan What you are referring to is `Cluster Identifier` and that is implemented through the `external-name` annotation. `NewClusterIdentifier` is for when the user wants to rename the cluster.

apis/database/v1alpha1/redshift_types.go
arush-sal: `NewMasterUserPassword` is for when the user wants to update or change the password generated by the controller at time of the cluster creation. So there is no default behavior to this. If you want to update the existing password you supply the value to it, else the password remains the same to what it was at creation.

apis/database/v1alpha1/redshift_types.go
arush-sal: I have removed `IAMRoles`, `EnhancedVPCRouting`, `KMSKeyID`, `PreferredMaintenanceWindow`, and,  `SnapshotScheduleIdentifier`.  I decided to keep some other fields that are also present in Parameters because AWS creates and sets a default for them if none is provided. So the idea was to be able to reflect those defaults. Let me know if I should go ahead and remove those other fields also(namely, `ClusterSubnetGroupName`, `ClusterSecurityGroups`, and `VPCSecurityGroups`).

pkg/clients/redshift/redshift.go
arush-sal: Sorry for the typo.

pkg/controller/database/redshift/controller.go
arush-sal: No, it is not. and the requested changes will be available in the latest push.

pkg/clients/redshift/redshift_test.go
arush-sal: Yes. Am sorry about that.

pkg/clients/redshift/redshift_test.go
arush-sal: Sure. Will add tests for those also.

pkg/controller/database/redshift/controller.go
arush-sal: I have replaced that with the following snippet:

```go
if len(rsp.Clusters) > 1 {
		return managed.ExternalObservation{}, errors.Wrap(errors.New(errMultipleCluster), errMultipleCluster)
	}
instance := rsp.Clusters[0]
```

apis/database/v1alpha1/redshift_types.go
arush-sal: Sure. I'll make that change.

pkg/clients/redshift/redshift.go
arush-sal: The idea is to identify the change when the user modifies `ClusterType`  to a valid value. As in `LateInitialize`, we just set it to a sane value, not actual observed value.

apis/database/v1alpha1/redshift_types.go
hasheddan: @arush-sal typically we would want to populate the spec with the defaults for these values that AWS sets if no values are provided by the user

pkg/controller/database/redshift/controller.go
hasheddan: @arush-sal my primary concern was actually with making sure the slice is not `nil`. So maybe we should change that check to `if len(rsp.Clusters) != 1` and update the error message appropriately

apis/database/v1alpha1/redshift_types.go
hasheddan: Ah I see. This is a very imperative operation by the AWS API here. Typically we would want to make this more declarative by just allowing the user to change the field and then reconciling the external state to match. This could be done for something like `MasterUserPassword`, but since the identifier is used to actually get the external state before we modify it, we clearly cannot do the same. I wonder if we should just omit this functionality for now. Any thoughts @muvaf?

pkg/controller/database/redshift/controller.go
arush-sal: @hasheddan We would get an empty slice only if there isn't a cluster with the name. In which we will also receive an err which will be caught by the above line. But still, as you suggested. I'll make the respective changes.

hasheddan:
@arush-sal could you ping me when this is ready for another review? :)

apis/database/v1alpha1/redshift_types.go
```diff
@@ -0,0 +1,652 @@
...
+	// Default: The default Amazon Redshift cluster parameter group. For information
+	// about the default parameter group, go to Working with Amazon Redshift Parameter
+	// Groups (https://docs.aws.amazon.com/redshift/latest/mgmt/working-with-parameter-groups.html)
+	// +optional
+	ClusterParameterGroupName *string `json:"clusterParameterGroupName,omitempty"`
+
+	// SecurityGroups is a list of security groups to associate with this cluster.
+	// Default: The default cluster security group for Amazon Redshift.
+	// +optional
+	ClusterSecurityGroups []string `json:"clusterSecurityGroups,omitempty"`
```
hasheddan, line 101: Can we add referencers for https://doc.crds.dev/github.com/crossplane/provider-aws/ec2.aws.crossplane.io/SecurityGroup/v1beta1 type here?

pkg/controller/database/redshift/controller_test.go
```diff
@@ -0,0 +1,656 @@
...
+)
+
+const (
+	providerName = "aws-creds"
+	// secretNamespace = "crossplane-system"
+	testRegion = "us-east-1"
+
+	// connectionSecretName = "my-little-secret"
+	// secretKey            = "credentials"
+	// credData             = "confidential!"
```
hasheddan, line 48: Looks like this needs to be removed

arush-sal: This has been addressed.

apis/database/v1alpha1/redshift_types.go
arush-sal: This has been added.

arush-sal:
@hasheddan All changes have been done and pushed.

apis/redshift/v1alpha1/doc.go
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1alpha1 contains managed resources for AWS Redshift services such as
+// DynamoDB.
```
hasheddan, line 18:
```suggestion
// Cluster.
```

arush-sal:
@hasheddan Yes all the changes before the last commit have been tested manually for `Create`, `Update`, and `Delete` events.

hasheddan:
Nice! Thanks for your hard work here @arush-sal :)

Issue #251, opened 2020-06-18
v1beta1 EKS Cluster cannot have tags removed


hasheddan:
### What happened?

The v1beta1 EKS `Cluster` resource cannot have tags removed because when a diff in tags is observed, only the [TagResourceRequest](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.23.0/service/eks?tab=doc#Client.TagResourceRequest) call is made. If a tag is to be removed, we should be calling the [UntagResourceRequest](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.23.0/service/eks?tab=doc#Client.UntagResourceRequest) method.

### How can we reproduce it?

Create an EKS `Cluster` with tags. Attempt to remove a tag.

### What environment did it happen in?
Crossplane version: `v0.11.1`
provider-aws version: https://github.com/crossplane/provider-aws/commit/1d7f738aa6ead6aebe2fb6cb13155c23b5231399

muvaf:
Fixed by https://github.com/crossplane/provider-aws/pull/288

PR #254, opened 2020-06-19
EKS v1alpha1 NodeGroup

hasheddan:

Adds support for EKS v1alpha1 `NodeGroup`.

TODO:
- [x] Unit tests for client and controller (including adding mocks) (still adding some client unit tests)
- [x] Manual testing
- [x] Handling labels (see TODO)
- [x] Handling tag diffs (see https://github.com/crossplane/provider-aws/issues/251 for same issue in `Cluster`)

### Checklist

I have:


apis/eks/v1alpha1/types.go
```diff
@@ -0,0 +1,383 @@
...
+const (
+	AMITypesAl2X8664    AMITypes = "AL2_x86_64"
+	AMITypesAl2X8664Gpu AMITypes = "AL2_x86_64_GPU"
+)
+
+// NodeGroupIssueCode is a code for an issue in a NodeGroup.
+type NodeGroupIssueCode string
+
+// Types of NodeGroup issue codes.
+const (
```
muvaf, line 38: I think having the set of possible inputs or outputs is useful when the set does not change an we get a validation that'll improve the UX. While `NodeGroupIssueCode` values probably won't change, I can't see how useful it is to have all possibilities in the codebase. So, I'd prefer removing them and just having `string.`

apis/eks/v1alpha1/types.go
```diff
@@ -0,0 +1,383 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+)
+
+// AMITypes is a type of AMI.
+type AMITypes string
```
muvaf, line 26: The same as below, however, in this case where this type is used is an input parameter. I think this is useful only if we get a validation. We are bitten by enum validation before as the set of values changed, in this case it doesn't seem to be that `AMITypes` will be changed but no one can say it cannot. So, I think we should either just go with `string` and have no validation  or use this type but also provide enum validation via kubebuilder tag.

pkg/clients/eks/nodegroup.go
```diff
@@ -0,0 +1,202 @@
+/*
```
muvaf, line 1: In `clients` package, we have always had 1 package per 1 managed resource but in this case it seems `nodegroup` functions are in the `eks` package together with `cluster`. What do you think about separating them and having `nodegroup` and `cluster` separate?

pkg/clients/eks/nodegroup.go
```diff
@@ -0,0 +1,202 @@
...
+			MinSize:     ng.ScalingConfig.MinSize,
+			MaxSize:     ng.ScalingConfig.MaxSize,
+		}
+	}
+	in.ReleaseVersion = awsclients.LateInitializeStringPtr(in.ReleaseVersion, ng.ReleaseVersion)
+	in.Version = awsclients.LateInitializeStringPtr(in.Version, ng.Version)
+	// NOTE(hasheddan): we always will set the default Crossplane tags in
+	// practice during initialization in the controller, but we check if no tags
+	// exist for consistency with expected late initialization behavior.
+	if in.Tags == nil {
```
muvaf, line 170:
```suggestion
	if len(in.Tags) == 0 {
```

provides a more guaranteed result.

apis/eks/v1alpha1/types.go
```diff
@@ -0,0 +1,383 @@
...
+	// SourceSecurityGroupSelector selects references to SecurityGroups used
+	// to set the SourceSecurityGroups.
+	// +optional
+	SourceSecurityGroupSelector *runtimev1alpha1.Selector `json:"sourceSecurityGroupSelector,omitempty"`
+}
+
+// NodeGroupScalingConfig is the configuration for scaling a node group.
+type NodeGroupScalingConfig struct {
+	// The current number of worker nodes that the managed node group should maintain.
+	DesiredSize *int64 `json:"desiredSize,omitempty"`
```
muvaf, line 225: Are these three fields optional? If so, can we mark them as such?

apis/eks/v1alpha1/types.go
```diff
@@ -0,0 +1,383 @@
...
+	AtProvider                     NodeGroupObservation `json:"atProvider,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// A NodeGroup is a managed resource that represents an AWS Elastic Kubernetes
+// Service NodeGroup.
+// +kubebuilder:printcolumn:name="READY",type="string",JSONPath=".status.conditions[?(@.type=='Ready')].status"
+// +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
+// +kubebuilder:printcolumn:name="RECLAIM-POLICY",type="string",JSONPath=".spec.reclaimPolicy"
```
muvaf, line 364: Similar to `NodePool` in GKE, it could be useful to show the cluster name as printer column.

pkg/clients/eks/nodegroup.go
hasheddan: They are using the same client so it seems redundant. Do you see any advantage to separating them?

apis/eks/v1alpha1/types.go
hasheddan: SGTM 👍

apis/eks/v1alpha1/types.go
hasheddan: Sounds reasonable to me

apis/eks/v1alpha1/types.go
hasheddan: hmm yeah I don't know which to do here. I imagine that these will expand over time (potentially over a very long time), which makes me want to not restrict the allowed values. However, they are not necessarily intuitive values so that make me want to optimize for user experience. I think I will opt for just using `string` here since the values are denoted in the field description, and so folks will be able to look up allowed values on https://doc.crds.dev 👍

hasheddan:
@muvaf I believe all comments are addressed. Still adding some client pkg tests but I don't believe that should block your review

pkg/clients/eks/nodegroup.go
muvaf: No, not much advantage given they are using the exact same client.

pkg/controller/eks/nodegroup/controller.go
```diff
@@ -0,0 +1,239 @@
...
+	cr.SetConditions(runtimev1alpha1.Creating())
+	if cr.Status.AtProvider.Status == v1alpha1.NodeGroupStatusCreating {
+		return managed.ExternalCreation{}, nil
+	}
+	_, err := e.client.CreateNodegroupRequest(eks.GenerateCreateNodeGroupInput(meta.GetExternalName(cr), &cr.Spec.ForProvider)).Send(ctx)
+	// NOTE(hasheddan): we are ignoring invalid request errors here so that we
+	// do not thrash while waiting for a referenced cluster to become ready.
+	// Node group will report that a successful creation request was made so
+	// this can be somewhat misleading.
+	return managed.ExternalCreation{}, errors.Wrap(resource.Ignore(eks.IsErrorInvalidRequest, err), errCreateFailed)
```
muvaf, line 169:
```suggestion
	return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
```

I think my comment got lost between reviews but that was the reason I hadn't approved the PR in the first pass.

The reconciliation doesn't start until all referenced resources are ready so I wouldn't expect this to result in error in that case. If the user supplied a raw cluster name and it's not ready, I think it should be fine to return error until it works. However, if we ignore all invalid requests then this will cause cases where the creation doesn't work but there is also no error. So, I don't think we should ignore the invalid request errors here since that encapsulates many errors that we don't know.

hasheddan: I see your point and believe that the comment I have there reflects the same sentiment. However, in practice, I think it will be common to create a node group at the same time as the cluster, and returning an error here can cause repeated calls to AWS over an extended period of time while the cluster becomes ready. This can lead to hitting rate limits, which could disrupt reconciliation of other resources as well. Since this is an alpha resource and until we have back off built into Crossplane runtime I think doing this is the right call.

muvaf: Maybe I’m missing something here. Is there anything different here than other resources? For example, Subnet does not call create until the vpc it references is ready and if raw vpc id is given instead of reference, it just calls Create every 30 seconds and returns error if it’s not ready. Or NodePool of GKE.

hasheddan: @muvaf nope! You are totally correct, I misinterpreted the first comment and also forgot that we do a `shortWait` instead of returning the error in the reconcile requeue here: https://github.com/crossplane/crossplane-runtime/blob/9a9a434f732111c340d5bdd7ecccb486a025ec2c/pkg/reconciler/managed/reconciler.go#L631

Thanks for pointing this out and apologies for me missing the point originally! Will update now :)

hasheddan: @muvaf fixed, I will make sure to clean up commit history before merge here :+1:

PR #255, opened 2020-06-19
Fix panic in RouteTable client

hasheddan:

The LateInitialize function for RouteTable is implemented incorrectly,
leading to the possibility of hitting array out-of-bounds panics. This
was likely missed because insufficient unit tests were provided for this
package. This PR is a simple fix to make us not panic, but in updating I
observed that there are fields missing from the RouteTable object and
it is also currently taking the place of multiple resource types in the
AWS API, such as Route and Association. These should likely be split out
into separate resources, but that fix is not provided in this PR. It
should also be noted that this PR does not fix the continual updates
that are made to a RouteTable despite the external resource being up to
date. If we do not break RouteTable into multiple resources then this
should be fixed.

### Checklist

I have:


hasheddan:
@negz I think that would be fine. If we want it to be `v1beta1` by the next release it probably would be fine to leave as is and just fix in `master`, but if we aren't sure then I think we should bump it back down.

hasheddan:
@negz I think there may be some conversation to be had about what a beta version of `RouteTable` will look like so I have bumped it back down to `v1alpha4`. Would you mind giving a +1 to indicate that this PR still has your approval with these fairly significant additional changes?

negz:
@hasheddan Consider this comment my +1.

muvaf:
> but in updating I observed that there are fields missing from the RouteTable object

This is the creation struct of RouteTable:
```
type CreateRouteTableInput struct {
	_ struct{} `type:"structure"`

	// Checks whether you have the required permissions for the action, without
	// actually making the request, and provides an error response. If you have
	// the required permissions, the error response is DryRunOperation. Otherwise,
	// it is UnauthorizedOperation.
	DryRun *bool `locationName:"dryRun" type:"boolean"`

	// The ID of the VPC.
	//
	// VpcId is a required field
	VpcId *string `locationName:"vpcId" type:"string" required:"true"`
}
```

So, it has only VPC ID and we expose that. Are you referring some fields from other API calls that should be in `RouteTable` object? (I just realized that `Route` and `Association` calls do have more fields that we currently expose, if that's what you mean)

> it is also currently taking the place of multiple resource types in the AWS API, such as Route and Association. These should likely be split out into separate resources, but that fix is not provided in this PR.

For a resource to be a managed resource, we need unique identifier to identify it in the cloud provider. In case of `Route` resource,  it doesn't have any kind of identifier except the ID of `RouteTable` that it belongs to (create call output returns only a bool). In `Association` case, its creation call does return an identifier to be used for deletion but it has neither `Describe` nor `Modify/Update` calls. However, both are included in the output of `DescribeRouteTable` call. So, it's not as clear cut but I think making them part of `RouteTable` was the right call.

hasheddan:
@muvaf it may be the best option to include these resource types in the `RouteTable` type, but if that is the case then all of their fields should be supported before we call this resource `v1beta1`. I personally do not think a type not having Update / Modify calls is justification for it not being a separate type. I also do not think an object being included in the Describe of another means that it should be part of that type.

There are many examples of this: subnets (and other network resource types) are frequently included in Describe of the parent network but are separate types. Node pools are frequently included in the Describe of a Kubernetes cluster but are separate types.

I am not sure what qualifies as a “unique identifier”, but a `Route` can be deleted by specifying the `RouteTableId` and the appropriate `CidrBlock`. Do these fields not serve to uniquely identify it in the cloud provider?

muvaf:
> all of their fields should be supported before we call this resource v1beta1.

Yes, I agree. Thanks for lowering the version before we make that contract. We need a follow-up PR to implement the missing fields on `Route` and `Association`.

> I personally do not think a type not having Update / Modify calls is justification for it not being a separate type. I also do not think an object being included in the Describe of another means that it should be part of that type.

It doesn't have `Describe` call either but yeah, there are several examples as you pointed out: it's not clear cut. For example, `Tag`s in every group could also be considered as distinct resource if `Create`/`Delete`/`Update` methods being present was enough. So, we can come up with arguments for both approaches. That's why it's harder to come up with a general rule for AWS resources compared to other cloud providers.

I believe, in such cases, it depends on the use cases we'd like to support. When I look at the API, I see that `RouteTable` is merely a logical grouping for routes and associations. They are like configurations of `RouteTable`, similar to how `Bucket` API is constructed; they've got many different calls for setting each configuration. From that perspective, I think that the majority of people do not treat them distinctly from their parent `RouteTable`.

On the other hand, [Terraform](https://www.terraform.io/docs/providers/aws/r/route_table_association.html) seems to be treating them as distinct and [CloudFormation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-route.html) treats only `Route` as a distinct resource.

IMO, as a general approach, I'd say starting with consolidated CRs and then fanning out to distinct CRDs if users request is simpler and relatively safer. It's simpler because users write less YAML and also devs write one less controller; safer because users can open issues to make it distinct but if they are unnecessarily distinct (assumption) in the first place, we wouldn't see people openning issue to consolidate them because they'd already be enabled but with supposedly worse UX. In case of migration from one to other (if happens), I can see our webhook mechanisms being able to create separate CRs for those resources by cut-paste them from the main `RouteTable` CR.

PR #261, opened 2020-06-25
Ignore namespace when getting provider from reference

hasheddan:

Though the Provider type is cluster-scoped, we use
corev1.ObjectReference as the field so a user may also include a
namespace. When getting the Provider, we never want to include an
extraneous namespace because it may cause the operation to fail.

Temporary fix that will be further enhanced by https://github.com/crossplane/crossplane-runtime/pull/182

Slack thread: https://crossplane.slack.com/archives/CEG3T90A1/p1593112864471900

### Checklist

I have:

### DIFF
--- a/pkg/controller/cache/cachesubnetgroup/controller.go
+++ b/pkg/controller/cache/cachesubnetgroup/controller.go
@@ -83,7 +83,7 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 	}
 
 	p := &awsv1alpha3.Provider{}
-	if err := c.client.Get(ctx, meta.NamespacedNameOf(g.Spec.ProviderReference), p); err != nil {
+	if err := c.client.Get(ctx, types.NamespacedName{Name: g.Spec.ProviderReference.Name}, p); err != nil {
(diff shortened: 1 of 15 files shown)


negz:
Are you sure including a namespace when getting a cluster scoped resource causes the operation to fail? I thought it would just be ignored.

hasheddan:
@negz I thought it would also, but slack thread indicates it does cause it to fail. Verifying manually now

hasheddan:
@negz verified by creating `Provider` and referencing it from `Cluster` managed resource:
```
conditions:
  - lastTransitionTime: "2020-06-25T21:32:54Z"
    message: 'connect failed: cannot get provider: Provider.aws.crossplane.io "aws-provider"
      not found'
    reason: Encountered an error during resource reconciliation
    status: "False"
    type: Synced
```

I then removed namespace and it worked as expected. I wonder if this is a recent change in `controller-runtime` :thinking:

hasheddan:
I'll open symmetrical PR's for other providers once this is merged :+1:

hasheddan:
Closing this in favor of longer-term solution in https://github.com/crossplane/crossplane-runtime/pull/182

PR #265, opened 2020-06-29
Adding RDS Client tests

### LINKED ISSUE #247: Add additional unit tests for RDS client package

The `pkg/clients/rds` package only has ~35% unit test coverage. This is well below the desired level.

Implement additional unit tests to achieve coverage similar to other client packages (ideally 85%+).

enderv:
Signed-off-by: Chris <chris.vodak@gmail.com>

Adding more unit tests to RDS Client
Fixes #247

### Checklist

I have:

### DIFF
--- a/pkg/clients/rds/rds_test.go
+++ b/pkg/clients/rds/rds_test.go
@@ -33,8 +33,40 @@ import (
 )
 
 var (
-	dbName           = "example-name"
-	characterSetName = "utf8"
+	address           = "address"
+	arn               = "my:arn"
+	az                = "az"
+	characterSetName  = "utf8"
+	cloudwatchExports = []string{"test"}
+	clusterName       = "testCluster"
+	dbName            = "example-name"
+	description       = "testDescription"
+	domain            = "domain"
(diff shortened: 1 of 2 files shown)


pkg/clients/rds/rds_test.go
```diff
@@ -162,3 +166,689 @@ func TestIsUpToDate(t *testing.T) {
...
+	storageType := "storageType"
+	instanceClass := "class"
+	engine := "5.6.41"
+	truncEngine := "5.6"
+	multiAZ := false
+	trueFlag := true
+	kmsID := "kms"
+	username := "username"
+	window := "window"
+	existingName := "existing"
```
hasheddan, line 483: nitpick: looks like these values are used across a number of the tests, how would you feel about moving them out to a global `var` block so they do not have to be reproduced in each?

enderv: Works for me. Will get that up when I can

hasheddan: Sounds good :) Ping me when ready for another review :+1:

PR #270, opened 2020-07-02
Remove unused dbClusterParamterGroup field from RDS

muvaf:

Removes `dbClusterParamterGroup` as it seems it's not used in any API call and backup retention call was given as `0` for all cases for some reason, that's also fixed.

### Checklist

I have:

### DIFF
--- a/apis/database/v1beta1/rdsinstance_types.go
+++ b/apis/database/v1beta1/rdsinstance_types.go
@@ -207,11 +207,6 @@ type RDSInstanceParameters struct {
 	// +optional
 	DBClusterIdentifier *string `json:"dbClusterIdentifier,omitempty"`
 
-	// DBClusterParameterGroupName is the name of the DB cluster parameter group to use for the DB cluster.
-	// +immutable
-	// +optional
-	DBClusterParameterGroupName *string `json:"dbClusterParameterGroupName,omitempty"`
-
 	// DBInstanceClass is the compute and memory capacity of the DB instance, for example, db.m4.large.
 	// Not all DB instance classes are available in all AWS Regions, or for all
 	// database engines. For the full list of DB instance classes, and availability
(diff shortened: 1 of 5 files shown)


apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -207,11 +207,6 @@ type RDSInstanceParameters struct {
 	// +optional
 	DBClusterIdentifier *string `json:"dbClusterIdentifier,omitempty"`
 
-	// DBClusterParameterGroupName is the name of the DB cluster parameter group to use for the DB cluster.
-	// +immutable
-	// +optional
-	DBClusterParameterGroupName *string `json:"dbClusterParameterGroupName,omitempty"`
```
hasheddan, line 213: Just want to check my understanding here:
- It looks like there is a separate `DBCluster` type that we will want to support
- The `DBCluster` type can have a parameter group called `DBClusterParameterGroup`
- The `DBClusterParameterGroup` itself should likely be its own resource
- It looks like you can create Aurora databases using `RDSInstance` or `DBCluster`?

Does all of that sound right to you? I was just trying to parse [the pkg docs](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.23.0/service/rds?tab=doc) and this was my understanding.

muvaf: Yes, I think that summary is accurate. Though the goal of this PR is less related to cluster/instance specifics; this field literally wasn't used anywhere in the codebase.

hasheddan: Gotcha! Yeah I just wanted to make sure the fix shouldn't be that we _should_ be using this, rather than removing it, but it sounds like this is the right move 👍

Issue #267, opened 2020-07-01
Creating an aurora-mysql RDSInstance fails with allocatedStorage, dbInstanceClass missing bug


kelvinwijaya:
### What happened?
Crossplane throwing an error when spinning aurora based DB, basically I would like to spin up DB that has read replicas and that is why I am chossing engine type aurora and aurora-mysql. 

According to the provider-aws RDSInstance api doc, those field such as allocateStorage and dbInstanceClass is not required for aurora based DB.

So i did try without putting any of these 2 fields and i am hitting into parameters validation issue mentioning that those fields are required

And the i put in those variables to test and i am running into another issue:

```
Warning  CannotCreateExternalResource  27s (x15 over 34s)  managed/rdsinstance.database.aws.crossplane.io  (combined from similar eve
nts): cannot create RDS instance: InvalidParameterCombination: Invalid storage size for engine name oscar and storage type aurora: 20
           status code: 400, request id: af92dcd3-3d74-42ac-985c-606cb8976c33
```

**Wierd thing is the log showing engine name transform to oscar, and storage type is aurora which is not defined anywhere in my manifests**

### How can we reproduce it?

- Just a snippet of what are the Composition file:

```
apiVersion: apiextensions.crossplane.io/v1alpha1
kind: Composition
metadata:
  name: mysqlinstances.crd.aws.mydomain.com
  labels:
    provider: aws
    tier: default
    vpc: dev
spec:
  writeConnectionSecretsToNamespace: crossplane-system
  reclaimPolicy: Delete
  from:
    apiVersion: crd.aws.mydomain.com/v1alpha1
    kind: MySQLInstance
  to:
    - base:
        apiVersion: database.aws.crossplane.io/v1beta1
        kind: RDSInstance
        spec:
          forProvider:
            masterUsername: masteruser
            engine: aurora
            allocatedStorage: 20
            skipFinalSnapshotBeforeDeletion: true
            publiclyAccessible: true
            deletionProtection: false
            dbClusterParameterGroupName: "mydbclusterparametergroup" 
          writeConnectionSecretToRef:
            namespace: crossplane-system
          providerRef:
            name: aws-provider
          reclaimPolicy: Delete
      patches:
        - fromFieldPath: "metadata.uid"
          toFieldPath: "spec.writeConnectionSecretToRef.name"
          transforms:
            - type: string
              string:
                fmt: "%s-mysql"
        - fromFieldPath: "spec.parameters.size"
[...]

PR #273, opened 2020-07-03
Add MasterPasswordSecretRef to allow specifying a password for RDS

muvaf:

We already had plans to use `Secret`s as input for sensitive fields [per managed resource design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#sensitive-input-fields) but I think there hasn't been actual implementation of it yet.

I have implemented it for RDS mainly to address the use case @kelvinwijaya described in https://github.com/crossplane/provider-aws/issues/267 ; it's not possible to create a read replica instance for aurora to join an existing DB cluster since the password is required to be empty.

I have tested creating an Aurora MySQL cluster with no IAM auth and used the following YAML to join that cluster:
```yaml
kind: RDSInstance
metadata:
  name: mvf-test-replica
spec:
  forProvider:
    engine: aurora
    dbClusterIdentifier: mvf-cl-1
    dbInstanceClass: db.t2.small
    availabilityZone: us-east-2a
    masterPasswordSecretRef:
      name: mvf-secret
      namespace: crossplane-system
      key: password
  writeConnectionSecretToRef:
    name: instance-secret
    namespace: crossplane-system
  providerRef:
    name: aws-provider
  reclaimPolicy: Delete
---
apiVersion: v1
kind: Secret
metadata:
  name: mvf-secret
  namespace: crossplane-system
type: Opaque
```

As you can see, in order to send an empty password, we need to reference an empty `Secret`, i.e. existence of a reference means opt-out of password generation. The opt-in for generation could be implemented but that'd be a breaking change for the current users in v1beta1, so, we can consider it in later API versions of this resource.

Fixes https://github.com/crossplane/provider-aws/issues/267

### Checklist

I have:

### DIFF
--- a/apis/database/v1beta1/rdsinstance_types.go
+++ b/apis/database/v1beta1/rdsinstance_types.go
@@ -430,6 +430,13 @@ type RDSInstanceParameters struct {
 	// +optional
 	MasterUsername *string `json:"masterUsername,omitempty"`
 
+	// MasterPasswordSecretRef references the secret that contains the password used
+	// in the creation of this RDS instance. If no reference is given, a password
+	// will be auto-generated.
+	// +optional
+	// +immutable
+	MasterPasswordSecretRef *runtimev1alpha1.SecretKeySelector `json:"masterPasswordSecretRef,omitempty"`
+
 	// MonitoringInterval is the interval, in seconds, between points when Enhanced Monitoring metrics
 	// are collected for the DB instance. To disable collecting Enhanced Monitoring
 	// metrics, specify 0. The default is 0.
--- a/apis/database/v1beta1/zz_generated.deepcopy.go
+++ b/apis/database/v1beta1/zz_generated.deepcopy.go
@@ -697,6 +697,11 @@ func (in *RDSInstanceParameters) DeepCopyInto(out *RDSInstanceParameters) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.MasterPasswordSecretRef != nil {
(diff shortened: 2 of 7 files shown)


muvaf:
> By future "opt-in for generation" do you mean that a user would supply a secret reference and then the controller would populate that secret with the password?

Not really. I meant that there would be a field like `autoGeneratePassword` that you can toggle. For example, in order to not supply a password, which is a valid case as shown, we still have to supply an empty `Secret`. It is a correct approach because empty password is also a valid password to supply but maybe we can consider making this default depending on the community use cases so that you explicitly have to say `autoGeneratePassword: true` if you'd like Crossplane to generate one. Otherwise, if there is a secret ref then use it but if that's also not given then the password is `nil`.

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -430,6 +430,15 @@ type RDSInstanceParameters struct {
 	// +optional
 	MasterUsername *string `json:"masterUsername,omitempty"`
 
+	// MasterPasswordSecretRef references the secret that contains the password used
+	// in the creation of this RDS instance. If a reference is not given, a password
+	// will be auto-generated.
+	//
+	// The key in the secret should be "password"
```
negz, line 437: Should this be configurable?

muvaf: Yeah, thanks for catching! Changed to `SecretKeySelector`.

PR #275, opened 2020-07-07
ResourceRecordSet Follow-up 

muvaf:

This PR adds a few v1beta1 features to `ResourceRecordSet` managed resource. Tested manually.

### Checklist

I have:


pkg/controller/route53/resourcerecordset/controller.go
```diff
@@ -55,59 +63,78 @@ func SetupResourceRecordSet(mgr ctrl.Manager, l logging.Logger) error {
...
-	c := conn.newClientFn(awsconfig)
+	s := &corev1.Secret{}
+	n := types.NamespacedName{Namespace: p.Spec.CredentialsSecretRef.Namespace, Name: p.Spec.CredentialsSecretRef.Name}
+	if err := c.kube.Get(ctx, n, s); err != nil {
+		return nil, errors.Wrap(err, errGetProviderSecret)
+	}
 
-	return &external{kube: conn.client, client: c}, nil
+	rdsClient, err := c.newClientFn(ctx, s.Data[p.Spec.CredentialsSecretRef.Key], p.Spec.Region, awsclients.UseProviderSecret)
+	return &external{client: rdsClient, kube: c.kube}, errors.Wrap(err, errCreateR53Client)
```
hasheddan, line 106:
```suggestion
	r53Client, err := c.newClientFn(ctx, s.Data[p.Spec.CredentialsSecretRef.Key], p.Spec.Region, awsclients.UseProviderSecret)
	return &external{client: r53Client, kube: c.kube}, errors.Wrap(err, errCreateR53Client)
```

pkg/controller/route53/resourcerecordset/controller.go
```diff
@@ -55,59 +63,78 @@ func SetupResourceRecordSet(mgr ctrl.Manager, l logging.Logger) error {
...
-	cr, ok := mgd.(*v1alpha1.ResourceRecordSet)
-
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+	cr, ok := mg.(*v1alpha1.ResourceRecordSet)
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
 
-	rrset, err := resourcerecordset.GetResourceRecordSet(ctx, e.client, cr.Spec.ForProvider.Name, cr.Spec.ForProvider.ZoneID, cr.Spec.ForProvider.Type, cr.Spec.ForProvider.SetIdentifier)
+	rrs, err := resourcerecordset.GetResourceRecordSet(ctx, meta.GetExternalName(cr), cr.Spec.ForProvider, e.client)
```
hasheddan, line 120: Any reason why we don't use client directly here or was this just how it was implemented?

apis/route53/v1alpha1/resourcerecordset_types.go
```diff
@@ -17,167 +17,411 @@ limitations under the License.
 package v1alpha1
 
 import (
-	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
-// DNSRecordType defines the valid DNS Record Types that can be used.
-type DNSRecordType string
```
hasheddan, line 25: Was this being used anywhere?

pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -44,84 +43,92 @@ type Client interface {
 	ListResourceRecordSetsRequest(input *route53.ListResourceRecordSetsInput) route53.ListResourceRecordSetsRequest
 }
 
-// RRsetNotFound will be raised when there is no ResourceRecordSet
-type RRsetNotFound struct{}
+// NotFoundError will be raised when there is no ResourceRecordSet
+type NotFoundError struct{}
```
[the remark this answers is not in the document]
muvaf: Yeah, there is no not found error for resource record sets because they only have `LIST` method, which would just return 0 elements if nothing was found. The workaround implemented here is to return a handmade error in case there is no `ResourceRecordSet` in the list that matches our identifier set.

pkg/controller/route53/resourcerecordset/controller.go
muvaf: It's because the client has only `LIST`(used for get) and `CHANGE`(used for create, update and delete) methods and we need to do some filtering/special case handling on the list response, so, another function that handles this logic made sense.

apis/route53/v1alpha1/resourcerecordset_types.go
muvaf: No, it was stale.

muvaf:
@hasheddan could you take a second look?

PR #288, opened 2020-07-14
Separate add and remove tag operations for v1beta1 EKS cluster

### LINKED ISSUE #276: v1beta1 EKS Cluster cannot change the API server endpoint access from public to private
### What happened?
The v1beta1 EKS Cluster resource cannot change the API server endpoint access from public to private. 

### How can we reproduce it?
Create an EKS Cluster with public access. Attempt to change it to private.
```
      Endpoint Private Access:  true
      Endpoint Public Access:   false
```
Nothing happens.

Crossplane version: v0.11.1
provider-aws version: 1d7f738

hasheddan:

The AWS SDK has separate methods for adding and removing tags from an
EKS cluster. This updates the update method of its controller to make
separate calls for adding and removing tags.

Fixes #276 #251 

cc @janwillies 

### Checklist

I have:

### DIFF
--- a/pkg/controller/eks/cluster.go
+++ b/pkg/controller/eks/cluster.go
@@ -172,7 +172,7 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 	return managed.ExternalCreation{}, errors.Wrap(err, errCreateFailed)
 }
 
-func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) { // nolint:gocyclo
 	cr, ok := mg.(*v1beta1.Cluster)
 	if !ok {
 		return managed.ExternalUpdate{}, errors.New(errNotEKSCluster)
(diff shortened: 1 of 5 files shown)


pkg/clients/eks/eks.go
```diff
@@ -302,8 +302,13 @@ func IsUpToDate(p *v1beta1.ClusterParameters, cluster *eks.Cluster) (bool, error
 		}
 	}
 
+	// if a, r := awsclients.DiffTags(p.Tags, cluster.Tags); len(a) != 0 || len(r) != 0 {
+	// 	return false, nil
+	// }
+
 	return cmp.Equal(&v1beta1.ClusterParameters{}, patch, cmpopts.EquateEmpty(),
 		cmpopts.IgnoreTypes(&v1alpha1.Reference{}, &v1alpha1.Selector{}),
+		// cmpopts.IgnoreFields(v1beta1.ClusterParameters{}, "Tags"),
```
muvaf, line 311: Seems like these additions are non-functional.

pkg/controller/eks/cluster.go
```diff
@@ -188,14 +188,17 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 	if err != nil || rsp.Cluster == nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, errDescribeFailed)
 	}
+	if add, remove := awsclients.DiffTags(cr.Spec.ForProvider.Tags, rsp.Cluster.Tags); len(remove) != 0 {
+		_, err := e.client.UntagResourceRequest(&awseks.UntagResourceInput{ResourceArn: rsp.Cluster.Arn, TagKeys: remove}).Send(ctx)
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(eks.IsErrorInUse, err), errAddTagsFailed)
+	} else if len(add) != 0 {
```
muvaf, line 194: Is there a specific reason for not calling Tag and Untag consecutively if needed?

pkg/clients/eks/eks.go
hasheddan: Thanks for catching this! Forgot to remove them from earlier testing.

pkg/controller/eks/cluster.go
hasheddan: Mostly because I wasn't sure if these were eventually consistent API calls, in which case I was trying to avoid trying to remove a tag and then trying to add it before it was removed. That being said, if we failed we would just try again so we could likely go ahead and call in succession. For the subsequent update calls, I think it may make sense to issue them on a separate reconcile because if the tags differ that may be the only change required. Alternatively, we could check if the tags were actually changed explicitly in `IsUpToDate`. Do you have a preference?

pkg/clients/eks/eks.go
hasheddan: Fixed :+1:

pkg/controller/eks/cluster.go
muvaf: > Mostly because I wasn't sure if these were eventually consistent API calls, in which case I was trying to avoid trying to remove a tag and then trying to add it before it was removed.

I think they block and return only if the tagging is completed. I'm not 100% sure but I haven't seen an AWS call that is not completed immediately after return of the call unless it reports that it's in progress.


I would actually prefer that at the end of the `Update` call, we are on the desired state if possible. In that case, it seems like:
```
add, remove := ...
if err := untag... {}
if err := tag ... {}
```

would work, right? The diff function populates both maps correctly from what I can tell.

hasheddan: @muvaf yep! That is correct. Do you think we should also make the other update calls on this reconcile as well?

muvaf: Do you mean not returning after these two calls are made? If so, yes, Update input generators do not include tags so it should be safe but I'm not sure what the API will return if there is no change because the sole reason of `Update` could be the tags. I don't know if that's easily possible but we can either:
* Ignore `no modification requests were made` error which says the update is already the observed state or
* check whether the resulting update input object is different than observation for each call. Something like `EnsureClusterVersion(...)` call where it first checks the diff for the inputs it's looking for and makes the actual `UpdateClusterVersionRequest` call only if there is a diff.

The latter seems safer but I feel like former one has better ROI in terms of complexity because the case where `Update` called but only one of those update calls are necessary happens but happens only once; in the next reconcile `IsUpToDate` returns `true` anyway. But if neither of the approaches are not easy to achieve right now, I'd not persist since even if `no modification requests were made` error is received, the next reconcile will clear that immediately.

hasheddan: > But if neither of the approaches are not easy to achieve right now, I'd not persist

Got a little lost in the negation here :sweat_smile: I could modify `IsUpToDate` to return multiple values to say:
- Are tags up to date?
- Is cluster version up to date?
- Is config up to date?

We would trigger `Update` on any of those being `false`, but then only make the individual update requests for the one that is not up to date. Is that aligned with your thinking?

hasheddan: @muvaf I opted for always making the cluster config update for now. tag / untag / version update will only happen when needed.

pkg/controller/eks/cluster.go
```diff
@@ -188,17 +188,25 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
 		return managed.ExternalUpdate{}, errors.Wrap(err, errPatchCreationFailed)
 	}
-	if len(patch.Tags) > 0 {
-		_, err := e.client.TagResourceRequest(&awseks.TagResourceInput{ResourceArn: rsp.Cluster.Arn, Tags: cr.Spec.ForProvider.Tags}).Send(ctx)
-		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(eks.IsErrorInUse, err), errAddTagsFailed)
-	}
 	if patch.Version != nil {
-		_, err := e.client.UpdateClusterVersionRequest(&awseks.UpdateClusterVersionInput{Name: awsclients.String(meta.GetExternalName(cr)), Version: patch.Version}).Send(ctx)
-		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(eks.IsErrorInUse, err), errUpdateVersionFailed)
+		if _, err := e.client.UpdateClusterVersionRequest(&awseks.UpdateClusterVersionInput{Name: awsclients.String(meta.GetExternalName(cr)), Version: patch.Version}).Send(ctx); err != nil {
```
muvaf, line 207: If this succeeds, will the next `UpdateClusterConfigRequest` work since it'll be in `modifying` state?

hasheddan: @muvaf I am not sure. That is why I was originally returning after each call. Either way, if the update for version is still being made on the subsequent reconcile then the `UpdateClusterConfigRequest` will fail. We could just skip making any update call if the cluster is in `Modifying` state.

PR #292, opened 2020-07-22
fix make run

janwillies:

Fixes `make run` which was expecting the binary to be named `provider` instead of `provider-aws`

### Checklist

I have:


hasheddan:
@janwillies this is a really helpful fix 👍 Would you want to check and see if it is also needed for other providers?

PR #298, opened 2020-07-24
security group: remove the default egress and fix comparison issue

### LINKED ISSUE #296: SecurityGroups never finish creating
### What happened?
SecurityGroups never finish. The status Ready is always false. I have waited for quite some time and it does not seem to change. 

### How can we reproduce it?
Try this out:
```yaml
kind: SecurityGroup
metadata:
  name: foo-tester
spec:
  forProvider:
    description: A test sg to see if it can finish
    egress:
    - fromPort: -1
      ipProtocol: "-1"
      ipRanges:
      - cidrIp: 0.0.0.0/0
        description: Any IP address.
      toPort: -1
    groupName: foo-tester
    ingress:
    - fromPort: 6379
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 10.101.0.0/16
        description: Just the cidr from the vpc only.
      toPort: 6379
    vpcIdRef:
      name: my-vpc
  providerRef:
    name: aws-provider
  reclaimPolicy: Delete
```

Crossplane version: 0.12.0
Provider-aws: 0.11.0

muvaf:

`SecurityGroup` creates an inital egress rule and if user's egress rule corresponds to this existing one, no change is propagated such as description so it always returned that the resource is not up-to-date. Additionally, `-1` value for `fromPort` and `toPort` is special for AWS and they return `nil` for values of those fields if you had submitted `-1` and that'd cause not up-to-date error since spec has `-1` but observed value seems to be `nil`.

Manually tested.

Fixes #296 

### Checklist

I have:

### DIFF
--- a/pkg/clients/ec2/securitygroup.go
+++ b/pkg/clients/ec2/securitygroup.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 
+	awsgo "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
@@ -30,6 +31,7 @@ type SecurityGroupClient interface {
 	DescribeSecurityGroupsRequest(input *ec2.DescribeSecurityGroupsInput) ec2.DescribeSecurityGroupsRequest
 	AuthorizeSecurityGroupIngressRequest(input *ec2.AuthorizeSecurityGroupIngressInput) ec2.AuthorizeSecurityGroupIngressRequest
 	AuthorizeSecurityGroupEgressRequest(input *ec2.AuthorizeSecurityGroupEgressInput) ec2.AuthorizeSecurityGroupEgressRequest
+	RevokeSecurityGroupEgressRequest(input *ec2.RevokeSecurityGroupEgressInput) ec2.RevokeSecurityGroupEgressRequest
 	CreateTagsRequest(input *ec2.CreateTagsInput) ec2.CreateTagsRequest
 }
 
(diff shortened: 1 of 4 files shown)


pkg/clients/ec2/securitygroup.go
```diff
@@ -149,6 +151,24 @@ func CreateSGPatch(in ec2.SecurityGroup, target v1beta1.SecurityGroupParameters)
 	v1beta1.SortTags(target.Tags, in.Tags)
 	LateInitializeSG(currentParams, &in)
 
+	// NOTE(muvaf): Sending -1 is sent as FromPort or ToPort is valid but the
```
hasheddan, line 154:
```suggestion
	// NOTE(muvaf): Sending -1 as FromPort or ToPort is valid but the
```

pkg/controller/ec2/securitygroup/controller.go
```diff
@@ -185,18 +186,29 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
...
 	}
-
 	meta.SetExternalName(cr, aws.StringValue(result.GroupId))
 
-	return managed.ExternalCreation{}, errors.Wrap(e.kube.Update(ctx, cr), errSpecUpdate)
+	if err := e.kube.Update(ctx, cr); err != nil {
+		return managed.ExternalCreation{}, errors.Wrap(err, errSpecUpdate)
+	}
+	// NOTE(muvaf): AWS creates an initial egress rule and there is no way to
+	// disable it with the create call. So, we revoke it right after the creation.
```
hasheddan, line 201: Should we always revoke? If the user supplies no egress rules should we assume they want the default?

muvaf: I think so. The CR is supposed to serve as the source of truth. In other cases, we'd cover that in late-init but when user does give an entry in that array the default one is not late-inited so we get into a loop of up-to-date.

Also in AWS console, that rule is populated in creation options and when you remove it and click create, you see revoke call being made in the progress bar.

hasheddan: Nice! This sounds like a reasonable solution here, thanks for explaining :)

PR #303, opened 2020-07-27
Fix nil pointer dereference in IAM referencers

hasheddan:

This fixes a potential panic with IAM referencers by making the required fields non-pointer types. Also updates required EKS fields to be non-pointer types as well.

### Checklist

I have:


apis/identity/v1alpha1/referencers.go
```diff
@@ -83,12 +82,12 @@ func (mg *IAMGroupUserMembership) ResolveReferences(ctx context.Context, c clien
 	if err != nil {
 		return err
 	}
-	mg.Spec.ForProvider.UserName = &user.ResolvedValue
+	mg.Spec.ForProvider.UserName = user.ResolvedValue
 	mg.Spec.ForProvider.UserNameRef = user.ResolvedReference
 
 	// Resolve spec.forProvider.groupName
 	group, err := r.Resolve(ctx, reference.ResolutionRequest{
-		CurrentValue: *mg.Spec.ForProvider.GroupName,
```
hasheddan, line 91: For reviewers: this (and same situation below) is the primary need for this fix. I initially thought to just use `reference.FromPtrValue` but then realized these fields should not be optional anyway.

apis/eks/v1alpha1/types.go
```diff
@@ -51,7 +51,7 @@ type NodeGroupParameters struct {
 	//
 	// ClusterName is a required field
 	// +immutable
-	ClusterName *string `json:"clusterName,omitempty"`
```
hasheddan, line 54: The EKS changes are not at risk of panic and do not actually change the generated CRD, but went ahead and fixed them while I was here.

Issue #297, opened 2020-07-24
Egress and Ingress rules should be separate from SecurityGroup


muvaf:
### What happened?

SecurityGroup object has arrays for both ingress and egress rules but according to our [managed resource API design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity), they should be separate CRDs.

muvaf:
I am not 100% sure about this as these rules do not have anything to identify them at all. It's literally the set of all properties that make up an identifier and this argument could go for any array element of all managed resources.

muvaf:
FWIW, deletion of `SecurityGroup` also deletes all `Ingress` and `Egress` rules. So, in case we did separate them into their own types, we'd see leftover rules in the cluster whose `SecurityGroup` doesn't exist anymore.

haarchri:
We have SecurityGroupRule since Release v0.30.0

PR #306, opened 2020-07-28
security group: implement full-fledged conversion functions for ip permissions

### LINKED ISSUE #304: SecurityGroup does not allow to include ingress source from another SG
### What happened?

Trying to add ingress source from another security group but it returns error:

```
message: "update failed: failed to authorize ingress rules: InvalidParameterValue:
CIDR block sg-xxxxxx is malformed\n\tstatus code: 400, request id:
6733d531-97d7-43d1-b23c-5df88444fe4d"
```

Tested from AWS UI console to entry ingress record as security group is ok

example:

```
kind: SecurityGroup
metadata:
  name: mysql.secgrp.abc.com
  labels:
    provider: aws
    tier: default
    vpc: dev
spec:
  forProvider:
    description: Created by crossplane
    groupName: crossplane-mysql-sg
    ingress:
      - fromPort: 3306
        toPort: 3306
        ipProtocol: TCP
        ipRanges:
          - cidrIp: "sg-xxxxxxxxx"
            description: "Access from EKS nodes"
    vpcId: vpc-xxxxxxxxx
  providerRef:
    name: aws-provider
```

### How can we reproduce it?
See the above sample manifest for SG

Crossplane version: 0.11.1
provider-aws: master (most recent)

muvaf:

There is a `GenerateEC2Permissions` that covers all fields but it's not used; another one with less fields is used. So, I made a change to make it use the full one and implemented its missing correspondent conversion.

Fixes #304 

### Checklist

I have:

### DIFF
--- a/config/crd/ec2.aws.crossplane.io_securitygroups.yaml
+++ b/config/crd/ec2.aws.crossplane.io_securitygroups.yaml
@@ -283,7 +283,6 @@ spec:
                         type: array
                     required:
                     - ipProtocol
-                    - ipRanges
                     type: object
                   type: array
                 groupName:
@@ -431,7 +430,6 @@ spec:
                         type: array
                     required:
                     - ipProtocol
-                    - ipRanges
                     type: object
                   type: array
                 tags:
--- a/pkg/clients/ec2/securitygroup.go
+++ b/pkg/clients/ec2/securitygroup.go
@@ -267,15 +267,15 @@ func LateInitializeIPPermissions(spec []v1beta1.IPPermission, o []ec2.IpPermissi
 // values between the target *v1beta1.SecurityGroupParameters and the current
 // *ec2.SecurityGroup
 func CreateSGPatch(in ec2.SecurityGroup, target v1beta1.SecurityGroupParameters) (*v1beta1.SecurityGroupParameters, error) { // nolint:gocyclo
-	// We initialize these arrays with the correct length so that late-init
-	// iteration works; workaround for using empty SecurityGroupParameters with
-	// late-init.
-	currentParams := &v1beta1.SecurityGroupParameters{
-		Ingress: make([]v1beta1.IPPermission, len(in.IpPermissions)),
-		Egress:  make([]v1beta1.IPPermission, len(in.IpPermissionsEgress)),
-	}
 	v1beta1.SortTags(target.Tags, in.Tags)
-	LateInitializeSG(currentParams, &in)
+	currentParams := &v1beta1.SecurityGroupParameters{
+		Description: awsclients.StringValue(in.Description),
+		GroupName:   awsclients.StringValue(in.GroupName),
+		VPCID:       in.VpcId,
+	}
+	currentParams.Tags = v1beta1.BuildFromEC2Tags(in.Tags)
+	currentParams.Ingress = GenerateIPPermissions(in.IpPermissions)
+	currentParams.Egress = GenerateIPPermissions(in.IpPermissionsEgress)
 	// NOTE(muvaf): Sending -1 as FromPort or ToPort is valid but the returned
 	// object does not have that value. So, in case we have sent -1, we assume
 	// that the returned value is also -1 in case if it's nil.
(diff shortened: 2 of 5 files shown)


muvaf:
I just tested this manually and it seems like some problems were buried because we didn't use some of the fields. Put some workarounds but I didn't want to invest too much as both Ingress and Egress arrays are going to be separate https://github.com/crossplane/provider-aws/issues/297 . So, I think we should merge this PR to address the urgent problems but next release should have those types as separate CRDs, deprecate `v1beta1` SecurityGroup and publish `v1beta2` SecurityGroup with instructions to manually migrate existing ones since the change will be a breaking one.

muvaf:
@hasheddan I'll be offline. Feel free to merge the PR if you'd like to.

config/crd/ec2.aws.crossplane.io_securitygroups.yaml
```diff
@@ -283,7 +283,6 @@ spec:
                         type: array
                     required:
                     - ipProtocol
-                    - ipRanges
```
hasheddan, line 286: Just a note, I feel like we should formalize our CRD versioning strategy. I know we generally follow the k8s conventions, but we have a few different scenarios that I feel like we encounter. For instance, this is technically a change here but obviously will not break any instances of this CRD that already exist so there is no need for a version bump. Maybe this is obvious, but may be nice for end users to be able to see exactly what we adhere to.

pkg/clients/ec2/securitygroup.go
```diff
@@ -129,28 +175,107 @@ func LateInitializeSG(in *v1beta1.SecurityGroupParameters, sg *ec2.SecurityGroup
...
 	if len(in.Tags) == 0 && len(sg.Tags) != 0 {
 		in.Tags = v1beta1.BuildFromEC2Tags(sg.Tags)
 	}
 }
 
+// LateInitializeIPPermissions returns an array of []v1beta1.IPPermission whose
+// empty optional fields are filled with what's observed in []ec2.IpPermission.
+//
+// Note that since there is no unique identifier for each IPPermission, its order
+// is assumed to be stable and used indexes are used as identifier.
```
hasheddan, line 196: nitpick:
```suggestion
// is assumed to be stable and indexes are used as identifier.
```

hasheddan: Any idea if AWS sorts rules in a consistent order? Seems like if we don't mirror then it will look like we always need update.

config/crd/ec2.aws.crossplane.io_securitygroups.yaml
muvaf: I believe we explicitly refer to Kubernetes policy for versioning: https://crossplane.io/docs/master/introduction/managed-resources.html Not sure what additions we would need to make to that though.

hasheddan: Yeah basically all I would add is that in a case like this we don't need to bump to `v1beta2` because this does not impact user experience, but maybe that is implied

pkg/clients/ec2/securitygroup.go
muvaf: No one knows, really. This is the problem with resources without any identifiers. Though manually sorting both of the arrays could be a good idea, it will be meaningless when we separate those rules into their [own types](https://github.com/crossplane/provider-aws/issues/297). Let me see if I can write something to address this.

hasheddan: Fair point, I just don't want to give the illusion to users that we support this inline for now if in reality it is going to thrash because that could cause issues with other requests.

hasheddan: In regards to sorting, it seems like we should establish uniqueness of an entry on the fromPort/toPort/IpProtocol combination.

muvaf: > it seems like we should establish uniqueness of an entry on the fromPort/toPort/IpProtocol combination.

I'm not sure. Looking at comments of those fields, even `IpProtocol` can get `-1` value so they don't really match their actual configurations. For example, I've tested that there could be a case where the port range is `0 - 3306`, protocol `TCP` but one referring to a security group as source and another one referring to a IpV4 CIDR block as source.

muvaf: Alright, I've tried a few things to sort correctly and it gets really ugly. I feel like I spent way more time than necessary on this now. I think it's OK to not sort it and rely on AWS and this is still improvement over the existing behavior.

hasheddan: Fair enough. I guess worst case scenario we only send a successful update every so often. Feel free to merge when ready then I can start release proceedings :+1:

PR #318, opened 2020-08-13
adds sorting for security group ingress/egress permissions

### LINKED ISSUE #317: SecurityGroup Inaccurate Invalid JSON Comparison
### What happened?

The AWS Security group accepts lists of IPPermission structs to store the ingress and egress rules. However, if multiple rules are provided, it is possible the CR will never resolve. The controller compares the JSON from AWS to a marshaled version of the SecurityGroup, since we cannot make any guarantees about the order of elements in the JSON list sent back by the API, we cannot just compare the two marshaled bodies. 

### How can we reproduce it?

In my example, I was creating a security group with the following ingress/egress values

```          ingress:
            - fromPort: 6379
              toPort: 6379
              ipProtocol: tcp
              ipRanges:
              - cidrIp:  x.x.x.x/x
                description: Everywhere
            - fromPort: 5432
              toPort: 5432
              ipProtocol: tcp
              ipRanges:
              - cidrIp:  x.x.x.x/x
                description: Everywhere
            egress:
            - fromPort: 6379
              toPort: 6379
              ipProtocol: tcp
              ipRanges:
              - cidrIp: x.x.x.x/x
                description: Everywhere
            - fromPort: 5432
              toPort: 5432
              ipProtocol: tcp
              ipRanges:
              - cidrIp:  x.x.x.x/x
                description: Everywhere
```

But since it's not sorted/ordered, the list AWS sent back was in a different order, which means the comparison would always yield a response that stated it needed to be updated.

Crossplane version: crossplane-alpha
AWS Provider: 0187f39ff12e995dbab3e1a9d4ee0b70f506ec0b

### Proposed Fix

I will be opening a PR in a moment with a working fix for this issue. I am currently sorting the Ingress/Egress lists from AWS and for the CR itself. I would imagine this is a common issue, is there a more general fix (or more lenient JSON merge function?)

krishchow:

Fixes #317 

### Checklist

I have:
- [:heavy_check_mark:] Run `make reviewable` to ensure this PR is ready for review.
- [:heavy_check_mark:] Ensured this PR contains a neat, self documenting set of commits.
- [:heavy_check_mark:] Updated any relevant [documentation], [examples], or [release notes].
- [:heavy_check_mark:] Updated the dependencies in [`app.yaml`] to include any new role permissions.

### Suggestions

Although this proposed fix works, I am sure there are other ways of handling this same issue. If anyone has any suggestions, I would be open to those ideas :)


krishchow:
Thanks for sharing the previous conversation @hasheddan! It seems like @muvaf ran into a similar situation I did, I am sure the way I tackled this is not the cleanest. I would probably more optimal to completely revamp the existing comparison. This was essentially the same issue I ran into when I created the s3bucketpolicy. There, instead of comparing the JSON bodies I ended up unmarshalling both into the same struct and just comparing the structs.

pkg/clients/ec2/securitygroup.go
```diff
@@ -152,6 +153,9 @@ func GenerateIPPermissions(objectPerms []ec2.IpPermission) []v1beta1.IPPermissio
 		}
 		permissions[i] = ipPerm
 	}
+	sort.Slice(permissions, func(i, j int) bool {
```
jeffvance, line 156: Nit: this works find but perhaps consider a new func that sorts permissions and returns it. Eg:

155      }
159      return sortBtPort(*permissions)
160 }

Issue #321, opened 2020-08-25
S3 BucketPolicy creation failed


kelvinwijaya:
Hi @hasheddan @krishchow ,

Understand that S3BucketPolicy has just been merged to master branch, i have tested the S3 Bucket Policy Support to be attached to my S3 bucket

**First scenario** is to use matchLabels to link to my S3 bucket and it failed to find the resource

Errors:

```
message: 'cannot resolve references: no managed resources matched selector'
```

```
apiVersion: storage.aws.crossplane.io/v1alpha1
kind: S3BucketPolicy
metadata:
  name: mys3bucketpolicy-label
spec:
  forProvider:
    version: "2012-10-17"
    bucketNameSelector:
      matchControllerRef: false
      matchLabels:
        creator : crossplane
    userName: crossplane-bucket-mybucket-noext-thxxb-hzm6b
    statement:
      - action:
        - "s3:GetObject"
        - "s3:GetObjectAcl"
        - "s3:GetObjectVersion"
        - "s3:PutObject"
        - "s3:PutObjectAcl"
        - "s3:DeleteObject"
        - "s3:DeleteObjectVersion"
        effect: Allow
        principal:
          allowAnon: true
        resource: 
          - "arn:aws:s3:::*"
        sid: DownloadandUpload
  providerRef:
    name: myaws-provider
  reclaimPolicy: Delete
```

**Second scenario** is to use map to the exact bucketname and username. This seem to have no issue to tie to the bucket but i got malform policy

Errors:
```
message: "create failed: failed to attach the policy to role: MalformedPolicy:
      Policy has invalid resource\n\tstatus code: 400, request id: 124589344580AC47,
      host id: Jtb5QA7NvnOcnMil0Ait77X9+EG9pwc7TQbyPT8IrgFKjx1nO3Oq7aCyjx5xNZkQlB3/QG74eEM="
```

```
apiVersion: storage.aws.crossplane.io/v1alpha1
kind: S3BucketPolicy
metadata:
  name: mys3bucketpolicy-direct
spec:
  forProvider:
    version: "2012-10-17"
    bucketName: mybucket-noext-thxxb-hzm6b
    userName: crossplane-bucket-mybucket-noext-thxxb-hzm6b
    statement:
      - sid: DownloadandUpload
        action:
          - "s3:GetObject"
          - "s3:GetObjectAcl"
          - "s3:GetObjectVersion"
          - "s3:PutObject"
          - "s3:PutObjectAcl"
[...]

PR #322, opened 2020-08-26
Adds examples for S3BucketPolicy

krishchow:
This PR aims to add more examples for the S3BucketPolicy, this should hopefully provide more context for some of the features provided by the resource. There is also a minor bugfix/optimization that I squshed into this PR.

Partial Fix for #321 

### Checklist

I have:
- [:heavy_check_mark:] Run `make reviewable` to ensure this PR is ready for review.
- [:heavy_check_mark:] Ensured this PR contains a neat, self documenting set of commits.
- [:heavy_check_mark:] Updated any relevant [documentation], [examples], or [release notes].
- [:heavy_check_mark:] Updated the dependencies in [`app.yaml`] to include any new role permissions.

### DIFF
--- a/examples/storage/policy/bucket-policy-bucket.yaml
+++ b/examples/storage/policy/bucket-policy-bucket.yaml
@@ -1,0 +1,16 @@
+apiVersion: storage.aws.crossplane.io/v1alpha3
+kind: S3Bucket
+metadata:
+  name: test.s3bucket.aws.crossplane.io
+  labels:
+    type: test
+spec:
+  localPermission: ReadWrite
+  region: us-east-2
+  cannedACL: private
+  writeConnectionSecretToRef:
+    namespace: crossplane-system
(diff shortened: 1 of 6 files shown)


krishchow:
@kelvinwijaya - Hopefully the examples here are helpful

kelvinwijaya:
Indeed the examples will be very helpful, thanks @krishchow 👍

krishchow:
@hasheddan Done! The first commit adds 4 examples for policies, along with the sample bucket for each of those policies. If the bucket shouldn't be there, I can remove it, but I put it there since the two resources are closely related.

The second commit just allows the Controller to handle the case where the "effectIAMUser" flag is set to true, and the Principal was not passed in by the user. Previously, this case was a little ambiguous, so I changed the code to be more explicit about creating the Principal struct when needed.

PR #328, opened 2020-09-02
Add S3 Bucket Tag in S3 Bucket v1alpha3 api

kelvinwijaya:
Description of your changes
This PR aims to add Tagging feature support for the S3Bucket v1alpha3, this feature should help in terms of add tags for AWS billing purposes. Also the changes might be useful and be propagated into the next version v1beta1 release

Checklist
I have:

[✔️] Run make reviewable to ensure this PR is ready for review.
[✔️] Ensured this PR contains a neat, self documenting set of commits.
[✔️] Updated any relevant documentation, examples, or release notes.
[ ] Updated the dependencies in app.yaml to include any new role permissions


pkg/controller/s3/s3bucket.go
```diff
@@ -183,6 +186,28 @@ func (r *Reconciler) _sync(bucket *bucketv1alpha3.S3Bucket, client s3.Service) (
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// helper function to perform S3 Bucket Tags update
+func updateTagging(bucket *bucketv1alpha3.S3Bucket, client s3.Service) error {
+	if bucket.Spec.Tags != nil {
```
muvaf, line 202: I see that we're not checking whether we need an update or not. Is this because `PutBucketTagging` is idempotent?

muvaf (approved):
LGTM! From what I can tell, it makes the `PutBucketTaggingRequest` in each reconcile even if they don't need to be updated and there isn't a mechanism to delete a tag. But since v1beta1 is on its way, I think it's fine to have these missing pieces for now to unblock key use cases with tags.

pkg/controller/s3/s3bucket.go
hasheddan: Would probably be good to still not make the call even if idempotent

examples/storage/bucket/resource-class.yaml
```diff
@@ -14,3 +14,25 @@ specTemplate:
   providerConfigRef:
     name: example
   reclaimPolicy: Delete
+---
+apiVersion: storage.aws.crossplane.io/v1alpha3
```
hasheddan, line 18: You are welcome to just replace the existing one with the one with tags if you like :+1:

hasheddan (commented):
CI failure was a flake, just retriggered :+1:

pkg/controller/s3/s3bucket.go
kelvinwijaya: Yes it should be idempotent, the condition is just to cater CR that don't required tagging so it doesn't make unnecessary API call

kelvinwijaya:
> CI failure was a flake, just retriggered 👍

kelvinwijaya:
Sorry that i have close and re-open the PR again to trigger the CI as i have no permission in Jenkins for replay the CI tasks

kelvinwijaya:
Hi @hasheddan @muvaf ,

Thanks for reviewing, all the checks have passed already, please help to merge this PR as i am not authorized

pkg/controller/s3/s3bucket.go
muvaf: I think what @hasheddan means is that the call is being made every time as long as there are tags on the resource and that may not be good even if it doesn't return an error. One reason is that unnecessary API calls could fill the API rate limit of AWS and another reason is that some resources get in an `updating` state even if the update call doesn't change anything. I think the latter does not apply here since tags are considered as separate resource in AWS and do not affect the state of the main resouce, the former argument still holds but to prevent that we'd need to make a Get call to check the change. @hasheddan do you think it's worth adding that call given that this code piece will be replaced by v1beta1 by the release anyway?

hasheddan: I think it is fine to skip for now :)

PR #330, opened 2020-09-03
Fix RDS Instance Password Reset Loop

### LINKED ISSUE #329: RDS instance gets caught in a password reset loop
### What happened?

What we expected: When creating an RDS instance with the `masterPasswordSecretRef` field set, I'd expect for Crossplane to reset the master credentials for the RDS instance once (to ensure they're set properly) and then consider the RDS instance as synced and ready for use.

What happened: When creating an RDS instance with the `masterPasswordSecretRef` field set, the RDS instance is stuck in a loop, continually entering the `resetting-master-credentials` state. In order to resolve the issue, you need to alter the CRD and remove the `masterPasswordSecretRef` field.

### How can we reproduce it?

Create an RDS instance with the `masterPasswordSecretRef` and it will always enter the reset loop.

Crossplane version: 0.12
Cloud provider: aws (0.11)
Kubernetes version: v1.18.2 (using KIND)

PocketMobsters:

Fixes #329 

RDSInstances will now only update the master password if there's a difference between the exported secret (if exists) and the referenced input secret from `masterPasswordSecretRef`.
### Checklist

I have:

### DIFF
--- a/pkg/clients/rds/rds.go
+++ b/pkg/clients/rds/rds.go
@@ -440,9 +440,9 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
 	//  <Modify/Create/Delete>DBInstanceInput objects but not in DBInstance
 	//  object are not late-inited. So, this func always returns true when
 	//  those configurations are changed by the user.
-
-	// TODO(muvaf): If a secret is provided for password, this logic should check
-	// whether it's changed by comparing it to the password in the published secret.
+	
+	// TODO(pocketmobsters): AllowMajorVersionUpgrade is another field we've
+	// identified that will cause this func to always return true.
 	patch, err := CreatePatch(&db, &p)
 	if err != nil {
 		return false, err
@@ -452,6 +452,7 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "Tags"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "SkipFinalSnapshotBeforeDeletion"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "FinalDBSnapshotIdentifier"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef")
 	), nil
 }
 
--- a/pkg/clients/rds/rds.go
+++ b/pkg/clients/rds/rds.go
@@ -467,27 +467,25 @@ func IsUpToDate(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance,
 
 // GetPassword fetches the referenced input password for an RDSInstance CRD and determines whether it has changed or not
 func GetPassword(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance) (newPwd string, changed bool, err error) {
-	if r.Spec.ForProvider.MasterPasswordSecretRef == nil && r.Spec.WriteConnectionSecretToReference == nil {
+	if r.Spec.ForProvider.MasterPasswordSecretRef == nil {
 		return "", false, nil
 	}
-	if r.Spec.ForProvider.MasterPasswordSecretRef != nil {
-		nn := types.NamespacedName{
-			Name:      r.Spec.ForProvider.MasterPasswordSecretRef.Name,
-			Namespace: r.Spec.ForProvider.MasterPasswordSecretRef.Namespace,
-		}
-		s := &corev1.Secret{}
-		if err := kube.Get(ctx, nn, s); err != nil {
-			return "", false, errors.Wrap(err, errGetPasswordSecretFailed)
-		}
-		newPwd = string(s.Data[r.Spec.ForProvider.MasterPasswordSecretRef.Key])
+	nn := types.NamespacedName{
+		Name:      r.Spec.ForProvider.MasterPasswordSecretRef.Name,
+		Namespace: r.Spec.ForProvider.MasterPasswordSecretRef.Namespace,
 	}
+	s := &corev1.Secret{}
+	if err := kube.Get(ctx, nn, s); err != nil {
+		return "", false, errors.Wrap(err, errGetPasswordSecretFailed)
+	}
+	newPwd = string(s.Data[r.Spec.ForProvider.MasterPasswordSecretRef.Key])
 
 	if r.Spec.WriteConnectionSecretToReference != nil {
-		nn := types.NamespacedName{
+		nn = types.NamespacedName{
 			Name:      r.Spec.WriteConnectionSecretToReference.Name,
 			Namespace: r.Spec.WriteConnectionSecretToReference.Namespace,
 		}
-		s := &corev1.Secret{}
+		s = &corev1.Secret{}
 		//the output secret may not exist yet, so we can skip returning an error
 		//if the error is NotFound
 		if err := kube.Get(ctx, nn, s); resource.IgnoreNotFound(err) != nil {
--- a/pkg/clients/rds/rds.go
+++ b/pkg/clients/rds/rds.go
@@ -17,6 +17,7 @@ limitations under the License.
 package rds
 
 import (
+	"context"
 	"encoding/json"
 	"strconv"
 	"strings"
@@ -25,13 +26,21 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/rds"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 
 	"github.com/crossplane/provider-aws/apis/database/v1beta1"
 	awsclients "github.com/crossplane/provider-aws/pkg/clients"
+)
+
+const (
+	errGetPasswordSecretFailed = "cannot get password secret"
 )
 
 // Client defines RDS RDSClient operations
@@ -435,7 +444,11 @@ func LateInitialize(in *v1beta1.RDSInstanceParameters, db *rds.DBInstance) { // 
 }
 
 // IsUpToDate checks whether there is a change in any of the modifiable fields.
-func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error) {
+func IsUpToDate(ctx context.Context, kube *client.Client, p v1beta1.RDSInstanceParameters, db rds.DBInstance, outCreds *v1alpha1.SecretReference) (bool, error) {
+	_, pwdChanged, err := GetPassword(ctx, kube, p.MasterPasswordSecretRef, outCreds)
+	if err != nil {
+		return false, err
+	}
 	patch, err := CreatePatch(&db, &p)
 	if err != nil {
 		return false, err
@@ -445,10 +458,59 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "Tags"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "SkipFinalSnapshotBeforeDeletion"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "FinalDBSnapshotIdentifier"),
-		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "ApplyModificationsImmediately"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "AllowMajorVersionUpgrade"),
-	), nil
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
+	) && !pwdChanged, nil
+}
+
+// GetPassword fetches the referenced input password for an RDSInstance CRD and determines whether it has changed or not
+func GetPassword(ctx context.Context, kube *client.Client, inputCreds *v1alpha1.SecretKeySelector, outCreds *v1alpha1.SecretReference) (string, bool, error) {
+	if kube == nil {
+		return "", false, nil
+	}
+	if inputCreds == nil && outCreds == nil {
+		return "", false, nil
+	}
+
+	var err error
+	var newPwd string
+	var pwdChanged bool
+	if inputCreds != nil {
+		nn := types.NamespacedName{
+			Name:      inputCreds.Name,
+			Namespace: inputCreds.Namespace,
+		}
+		newPwd, err = getPasswordFromSecret(ctx, kube, nn, inputCreds.Key)
+		if err != nil {
+			return "", false, errors.Wrap(err, errGetPasswordSecretFailed)
+		}
+	}
+
+	if outCreds != nil {
+		nn := types.NamespacedName{
+			Name:      outCreds.Name,
+			Namespace: outCreds.Namespace,
+		}
+		outPwd, _ := getPasswordFromSecret(ctx, kube, nn, v1alpha1.ResourceCredentialsSecretPasswordKey)
+		if outPwd == newPwd {
+			return newPwd, false, nil
+		}
+	}
+
+	return newPwd, pwdChanged, nil
+}
+
+func getPasswordFromSecret(ctx context.Context, kube *client.Client, namespacedName types.NamespacedName, key string) (string, error) {
+	if kube == nil {
+		return "", nil
+	}
+	k := *kube
+	s := &corev1.Secret{}
+	if err := k.Get(ctx, namespacedName, s); err != nil {
+		return "", err
+	}
+	return string(s.Data[key]), nil
 }
 
 // GetConnectionDetails extracts managed.ConnectionDetails out of v1beta1.RDSInstance.
--- a/pkg/clients/rds/rds.go
+++ b/pkg/clients/rds/rds.go
@@ -493,9 +493,11 @@ func GetPassword(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance
 			Namespace: r.Spec.WriteConnectionSecretToReference.Namespace,
 		}
 		s := &corev1.Secret{}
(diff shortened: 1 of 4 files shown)


PocketMobsters:
Tacking on some additional fields to ignore in IsUpToDate. 

ApplyModificationsImmediately will cause this function to always return true because it is not returned by AWS. If this is set, it WILL be included in any subsequent modify call so it won't break any behavior as any other field that should be sent out will appropriately return true from this function.

Similarly with AllowMajorVersionUpgrade it is not returned by AWS but their [docs](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Upgrading.html#USER_UpgradeDBInstance.Upgrading.Manual) mention that it should be included with a call that changes EngineVersion. By itself, it should not trigger a modify call to be sent. 

These two fields will ALSO cause the operator to go into a continual loop of processing these instances which is only really noticeable if you have debug logging enabled. Each time this happens, it is reporting that there were no modifications included in the request.

pkg/controller/database/rdsinstance.go
```diff
@@ -89,11 +89,29 @@ type external struct {
...
 }
 
-func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) {
+func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.ExternalObservation, error) { // nolint:gocyclo
 	cr, ok := mg.(*v1beta1.RDSInstance)
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errNotRDSInstance)
 	}
+
+	pwdChanged := false
```
muvaf, line 98: I believe this check belongs to `IsUpToDate` function. As you said, we don't have another resource whose `IsUpToDate` func has a client but we don't have another resource in AWS that takes a secret ref as input either, so this could be example to future implementations.

I would suggest implementing this functionality in a way to be called like `pw, changed, err := rds.CheckPassword(...)` so that we can call it in `IsUpToDate`, `Update` and `Create`, essentially wherever we need the password. `pkg/clients/rds/rds.go` could be a good fit for this utility.

pkg/clients/rds/rds.go
```diff
@@ -452,6 +445,9 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "Tags"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "SkipFinalSnapshotBeforeDeletion"),
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "FinalDBSnapshotIdentifier"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
```
muvaf, line 448: Thanks for catching these!

muvaf (commented):
Hi @PocketMobsters ! Your solution looks solid, I just left a comment about where it could go. Thank you for catching this bug and opening the PR!

kelvinwijaya:
Thanks @PocketMobsters , i think the fix would help my existing use case for updating RDSInstances password by secret object reference (As per my current understanding the password is auto-generated), i did some workaround to run another controller for updating the RDSInstances password, but seem i don't have to anymore

PocketMobsters:
@muvaf I rebased my changes so everything should be up to date with master as of the time of this comment.

I moved the password change detection logic to the RDS client and thus I'm passing some additional information (context and kubernetes client) into `IsUpToDate` so it can be used in `GetPassword`. I really liked your suggestion of the return signature being `pw, changed, err`. 

After I made my changes I realized you would have named this function `CheckPassword` and I'm totally down to rename it, just want to get a bit of a review on the new approach and whether its something you'd like in the code base.

I was having some trouble coming up with a good way to test the new functionality, as I didn't really notice anything existing that mocks external calls and the contribution guidelines specifically prohibit adding a third party testing library. As it stands now, my code is untested programmatically - I've just put it through a bunch of manual integration tests. I looked at the existing E2E integration test and it appears to revolve around making sure the provider itself can be installed and brings up the appropriate deployments and pods so I don't think its appropriate for this either. If you have any ideas or existing examples of integration tests like that I'd be happy to revise my changes accordingly!

pkg/clients/rds/rds.go
```diff
@@ -435,14 +444,11 @@ func LateInitialize(in *v1beta1.RDSInstanceParameters, db *rds.DBInstance) { //
...
 // IsUpToDate checks whether there is a change in any of the modifiable fields.
-func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error) {
-	// TODO(muvaf): ApplyImmediately and other configurations that exist in
-	//  <Modify/Create/Delete>DBInstanceInput objects but not in DBInstance
-	//  object are not late-inited. So, this func always returns true when
-	//  those configurations are changed by the user.
-
-	// TODO(muvaf): If a secret is provided for password, this logic should check
-	// whether it's changed by comparing it to the password in the published secret.
+func IsUpToDate(ctx context.Context, kube *client.Client, p v1beta1.RDSInstanceParameters, db rds.DBInstance, outCreds *v1alpha1.SecretReference) (bool, error) {
```
muvaf, line 447:
```suggestion
func IsUpToDate(ctx context.Context, kube *client.Client, p v1beta1.RDSInstance, db rds.DBInstance) (bool, error) {
```

What do you think about supplying the whole `RDSInstance` and get the necessart secret refs in the function?

muvaf:
```suggestion
func IsUpToDate(ctx context.Context, kube client.Client, p v1beta1.RDSInstanceParameters, db rds.DBInstance, outCreds *v1alpha1.SecretReference) (bool, error) {
```

Since `client.Client` is an interface, we don't need to send it via pointer.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,59 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "FinalDBSnapshotIdentifier"),
-	), nil
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "ApplyModificationsImmediately"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "AllowMajorVersionUpgrade"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
+	) && !pwdChanged, nil
+}
+
+// GetPassword fetches the referenced input password for an RDSInstance CRD and determines whether it has changed or not
+func GetPassword(ctx context.Context, kube *client.Client, inputCreds *v1alpha1.SecretKeySelector, outCreds *v1alpha1.SecretReference) (string, bool, error) {
```
muvaf, line 468:
```suggestion
func GetPassword(ctx context.Context, kube client.Client, rds *v1beta1.RDSInstance) (string, bool, error) {
```

Similar suggestions as above.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,59 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
+		return "", false, nil
+	}
+	if inputCreds == nil && outCreds == nil {
+		return "", false, nil
+	}
+
+	var err error
+	var newPwd string
+	var pwdChanged bool
+	if inputCreds != nil {
```
muvaf, line 479: I believe once we accept `*v1beta1.RDSInstance` and do the necessary checks like `if rds.Spec.ForProvider.MasterPasswordSecretRef == nil` in the beginning, we don't need to check them anymore. Once they are out of the way, it could be a simple flow like `do the checks -> get the input secret -> get the output secret -> return newPwd, newPwd != oldPwd, nil`

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,59 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
+		outPwd, _ := getPasswordFromSecret(ctx, kube, nn, v1alpha1.ResourceCredentialsSecretPasswordKey)
+		if outPwd == newPwd {
+			return newPwd, false, nil
+		}
+	}
+
+	return newPwd, pwdChanged, nil
+}
+
+func getPasswordFromSecret(ctx context.Context, kube *client.Client, namespacedName types.NamespacedName, key string) (string, error) {
```
muvaf, line 504: From what I can tell, once we use `client.Client`, this function would not be really necessary and we can do the same operation in place in `GetPassword`.

pkg/controller/database/rdsinstance.go
```diff
@@ -204,18 +205,16 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
 	modify := rds.GenerateModifyDBInstanceInput(meta.GetExternalName(cr), patch)
 	var conn managed.ConnectionDetails
 	if cr.Spec.ForProvider.MasterPasswordSecretRef != nil {
```
muvaf, line 207: I think we can skip the `if cr.Spec.ForProvider.MasterPasswordSecretRef != nil` here since we do it in `GetPassword` anyway. The `if changed` condition covers the case where there is no change.

pkg/controller/database/rdsinstance.go
```diff
@@ -204,18 +205,16 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
+		if changed {
+			conn = managed.ConnectionDetails{
+				runtimev1alpha1.ResourceCredentialsSecretPasswordKey: []byte(pwd),
+			}
+			modify.MasterUserPassword = aws.String(pwd)
 		}
-		modify.MasterUserPassword = aws.String(string(s.Data[cr.Spec.ForProvider.MasterPasswordSecretRef.Key]))
 	}
 	if _, err = e.client.ModifyDBInstanceRequest(modify).Send(ctx); err != nil {
 		return managed.ExternalUpdate{}, errors.Wrap(err, errModifyFailed)
```
muvaf, line 220: Would you be use `GetPassword` in `Create`, too? We wouldn't really check whether password changed but we'd need the desired password and generate a new one only if the returned password is `""`.

muvaf (commented):
Thank you for the update! It's looking pretty good! I've left a few simplification suggestions.

Regarding tests, I see that `pkg/client/rds/rds_test.go` file assumes that all functions are bare conversions, hence, doesn't have any test with clients. But we do have such tests and you can find some examples in [`pkg/controller/database/rdsinstance_test.go`](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/database/rdsinstance_test.go#L232) and in other controller tests.

pkg/controller/database/rdsinstance.go
PocketMobsters: Wonderful comments, I'll make the according changes and look at the tests you referenced to see what I can do with the clients!

pkg/clients/rds/rds.go
PocketMobsters: I basically just exit early if neither are set. It's kind of a redundant conditional upon further inspection.

I need to check each one specifically, because otherwise accessing the name, namespace and key (for input) will throw an error. If you take a look at the updated function and can figure out a less verbose way to write it and cover the cases it does I'll happily make changes!

PocketMobsters:
I've successfully forgotten to sign off my commits at least 3 times here. One day I'll get it done without DCO reminding me.

@muvaf I believe I've addressed most of your comments, with the exception being the flow change in `GetPassword` because I'm not 100% sure how to change it to work properly without the individual checks. I think some of the confusion comes from the early exit if neither are set. 

I'd like to apologize for the delay on getting these changes in, I've been pulling on a couple different threads at work in the meantime.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,50 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "AllowMajorVersionUpgrade"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
+	) && !pwdChanged, nil
+}
+
+// GetPassword fetches the referenced input password for an RDSInstance CRD and determines whether it has changed or not
+func GetPassword(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance) (string, bool, error) {
+	if kube == nil {
+		return "", false, nil
+	}
```
muvaf, line 471:
```suggestion
```

I don't think we need to check this because if that's `nil`, probably many of the calls before this function will fail anyway.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,50 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
+	if r.Spec.ForProvider.MasterPasswordSecretRef == nil && r.Spec.WriteConnectionSecretToReference == nil {
```
muvaf, line 472:
```suggestion
	if r.Spec.ForProvider.MasterPasswordSecretRef == nil || r.Spec.WriteConnectionSecretToReference == nil {
```

In order for this function to make its job, both fields have to be set. So, if any of them isn't there, then we should return early. As written here, that's not an error case, it's just saying that we don't need the utility of `GetPassword` for those cases.

muvaf: After you make this change, in the rest of the function you can be sure that they are not `nil` so you can get rid of two `if` conditions below.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,50 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
+			Name:      r.Spec.WriteConnectionSecretToReference.Name,
+			Namespace: r.Spec.WriteConnectionSecretToReference.Namespace,
+		}
+		s := &corev1.Secret{}
+		if err := kube.Get(ctx, nn, s); err != nil {
+			pwdChanged = newPwd != ""
+		} else {
+			outPwd := string(s.Data[v1alpha1.ResourceCredentialsSecretPasswordKey])
+			pwdChanged = newPwd != "" && outPwd != newPwd
+		}
```
muvaf, line 501:
```suggestion
		if err := kube.Get(ctx, nn, s); resource.IgnoreNotFound(err) != nil {
			return "", false, err
		}
		pwdChanged = newPwd != string(s.Data[v1alpha1.ResourceCredentialsSecretPasswordKey])
		err = nil
		return
```

On `NotFound` case is OK to skip. For other errors, we need to return them. Also we don't need to check specifically against `""` because if it's `NotFound`, then `string(s.Data[v1alpha1.ResourceCredentialsSecretPasswordKey])` will be `""`.

pkg/clients/rds/rds.go
```diff
@@ -452,7 +458,50 @@ func IsUpToDate(p v1beta1.RDSInstanceParameters, db rds.DBInstance) (bool, error
...
 		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "FinalDBSnapshotIdentifier"),
-	), nil
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "ApplyModificationsImmediately"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "AllowMajorVersionUpgrade"),
+		cmpopts.IgnoreFields(v1beta1.RDSInstanceParameters{}, "MasterPasswordSecretRef"),
+	) && !pwdChanged, nil
+}
+
+// GetPassword fetches the referenced input password for an RDSInstance CRD and determines whether it has changed or not
+func GetPassword(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance) (string, bool, error) {
```
muvaf, line 468:
```suggestion
func GetPassword(ctx context.Context, kube client.Client, r *v1beta1.RDSInstance) (newPwd string, pwdChanged bool, err error) {
```

I think naming the returned variables if there are more than two would serve as a nice documentation to understand what the `bool` is for.

pkg/clients/rds/rds.go
muvaf: I left some suggestions about this.

muvaf:
> I'd like to apologize for the delay on getting these changes in, I've been pulling on a couple different threads at work in the meantime.

No need to apologize, it's perfectly fine. We wouldn't even know about this problem. Thank you for driving this, it's pretty close to be merged!

pkg/clients/rds/rds.go
PocketMobsters: Great call

pkg/clients/rds/rds.go
PocketMobsters: Neither `masterPasswordSecretRef` (input) nor `writeConnectionSecretToRef` (output) are marked as required, so I figure we should try and account for situations where either are set. I figure the best way to dissect this would be to walk through the contexts in which this function may be called.

-----

### Create

When creating an instance, we only really check the input. If input is set, we'll use that value for the password we provide to AWS. If the input field is not set, or the value in the referenced secret is set to `""` we'll autogenerate a password. I've created a small diagram for [it here](https://i.imgur.com/Qn7BJTL.png).

If we change the `GetPassword` function to return if either input or output references are nil, we wouldn't report back what the input password is so I'd revert the change in `Create` and have it pull the input secret directly.

### Update

I created [another diagram](https://i.imgur.com/kEfYaRR.png) to help convey my thought process. As you can see, we don't really do anything unless both fields are set, so I can see how making this change would clean up the code a bit.

### Observe

I created [a diagram](https://i.imgur.com/pPM4rCz.png) to visually represent this flow as well. Again we only really do work if both fields are set.

-----

If we have `Create` pull the password directly as it was doing before, I can 100% support cleaning up the code a bit. I'd also feel like a rename from `GetPassword` to `CheckPassword` as suggested in your original round of reviews would completely describe what it is doing. Otherwise, I feel we should at least continue supporting fields being partially set, to account for the `Create` case.

pkg/clients/rds/rds.go
PocketMobsters: I kept the `newPwd != ""` because I think we want to make sure that at least some value was pulled from the input secret, otherwise we may report that the password changed (a value in output but no value in input) but also send back an empty string for `newPwd`. This wouldn't matter in `Create` as we only care about the value of the password, but I'm unsure what the implications are of providing an empty password in `Update` and what having that included in the AWS request body would do

pkg/clients/rds/rds.go
muvaf: Thanks for all the details! I think I understand what you're describing. Seems like we have only one case that's tricky; input is given but output is not given during `Create`. I couldn't really think of many scenarios where you give an input secret (meaning you use pwd auth) but not use the output. However, for the implementation to be _correct_, i.e. user supplied a password and that should be enough for us to use it, I agree with you that this should be taken into account.

So, in all cases, if input is empty we should return early but output being empty is not of a concern so that we can take `Create` into account. So, I'd suggest:
```suggestion
	if r.Spec.ForProvider.MasterPasswordSecretRef == nil {
```

and assume that `r.Spec.ForProvider.MasterPasswordSecretRef` is not `nil` in the rest of the function. The check for `if r.Spec.WriteConnectionSecretToReference != nil {` in the later parts of the function seems necessary.

PocketMobsters: Pushed this up

muvaf (approved):
LGTM! Thank you @PocketMobsters for working on this!

PR #335, opened 2020-09-10
fixes #332 subnet reconciliatiion

### LINKED ISSUE #332: aws subnet reconciliation fails
### What happened?

When an AWS subnet managed by Crossplane is deleted and when the controller reconciles it fails with an error.

### How can we reproduce it?

* Apply the below YAML spec

YAML spec:

```yaml
---
kind: VPC
metadata:
  name: frezbo-vpc
spec:
  forProvider:
    cidrBlock: 10.16.0.0/16
    enableDnsSupport: true
    enableDnsHostNames: true
    instanceTenancy: default
    tags:
      - key: Name
        value: frezbo-vpc
  reclaimPolicy: Delete
  providerRef:
    name: aws-provider
---
kind: Subnet
metadata:
  name: frezbo-pub
spec:
  forProvider:
    cidrBlock: 10.16.1.0/24
    tags:
      - key: Name
        value: frezbo-pub
    vpcIdRef:
      name: frezbo-vpc
    mapPublicIPOnLaunch: true
  reclaimPolicy: Delete
  providerRef:
    name: aws-provider
```

* Wait for the subnet to be created
* Delete the subnet from the AWS console

Error:

```bash
❯ k describe subnet frezbo-pub
Name:         frezbo-pub
Namespace:
Labels:       <none>
Annotations:  crossplane.io/external-name: subnet-04632929ad60075a9
API Version:  ec2.aws.crossplane.io/v1beta1
Kind:         Subnet
Metadata:
  Creation Timestamp:  2020-09-09T16:11:48Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generation:  3
  Managed Fields:
    API Version:  ec2.aws.crossplane.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:kubectl.kubernetes.io/last-applied-configuration:
      f:spec:
        .:
        f:forProvider:
          .:
          f:cidrBlock:
          f:mapPublicIPOnLaunch:
          f:vpcIdRef:
            .:
            f:name:
        f:providerRef:
          .:
          f:name:
        f:reclaimPolicy:
    Manager:      kubectl-client-side-apply
    Operation:    Update
    Time:         2020-09-09T16:11:48Z
    API Version:  ec2.aws.crossplane.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:crossplane.io/external-name:
        f:finalizers:
          .:
          v:"finalizer.managedresource.crossplane.io":
      f:spec:
        f:forProvider:
          f:assignIpv6AddressOnCreation:
          f:availabilityZone:
          f:availabilityZoneId:
          f:tags:
          f:vpcId:
      f:status:
        .:
        f:atProvider:
          .:
          f:availableIpAddressCount:
          f:subnetId:
          f:subnetState:
        f:conditions:
    Manager:         crossplane-aws-provider
    Operation:       Update
    Time:            2020-09-09T16:15:39Z
  Resource Version:  65529
  Self Link:         /apis/ec2.aws.crossplane.io/v1beta1/subnets/frezbo-pub
  UID:               543e5d76-0a20-47b9-94d5-1ce15e209f52
Spec:
  For Provider:
    assignIpv6AddressOnCreation:  false
    Availability Zone:            us-east-1e
    Availability Zone Id:         use1-az3
    Cidr Block:                   10.16.1.0/24
    Map Public IP On Launch:      true
    Tags:
      Key:    Name
      Value:  frezbo-pub
    Vpc Id:   vpc-03218d0c7d6f9e6c7
    Vpc Id Ref:
      Name:  frezbo-vpc
  Provider Ref:
    Name:          aws-provider
  Reclaim Policy:  Delete
Status:
  At Provider:
    Available Ip Address Count:  251
    Subnet Id:                   subnet-04632929ad60075a9
    Subnet State:                available
  Conditions:
    Last Transition Time:  2020-09-09T16:15:39Z
    Message:               create failed: failed to create the Subnet resource: InvalidParameterCombination: The parameter availabilityZone cannot be used with the parameter availabilityZoneId
                           status code: 400, request id: 96acc599-75fa-4092-8141-b90d0e54ac31
    Reason:                Encountered an error during resource reconciliation
    Status:                False
    Type:                  Synced
    Last Transition Time:  2020-09-09T16:14:42Z
    Reason:                Resource is being created
    Status:                False
    Type:                  Ready
```

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

When the subnet resource has been externally deleted, the current CR fields has `availabilityZone` and `availabilityZoneID` set. To work around the issue of not passing in both fields `availabilityZoneID` is updated to `nil`

Fixes #332 
-->

### Checklist
I have:
- [ x ] Run `make reviewable` to ensure this PR is ready for review.
- [ x ] Ensured this PR contains a neat, self documenting set of commits.
- [ x ] Updated any relevant [documentation], [examples], or [release notes].
- [ x ] Updated the dependencies in [`app.yaml`] to include any new role permissions.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -76,6 +76,7 @@ github.com/creack/pty v1.1.7/go.mod h1:lj5s0c3V2DBrqTV7llrYr5NG6My20zk30Fl46Y7Do
 github.com/crossplane/crossplane v0.11.0 h1:uye4oDt8Pz7JoGBI0qtPuUHENDxLu0LIxhX+gyTUySM=
 github.com/crossplane/crossplane v0.11.0/go.mod h1:+AH83k737iABb8jUKyo8zML73JZZioeybuShGyAtbkg=
 github.com/crossplane/crossplane-runtime v0.7.1-0.20200421211018-be37c50cc2ab/go.mod h1:e12p4X6dbtqd9GOnph78Epd6aezyvmcMM1+6aQy3VzY=
+github.com/crossplane/crossplane-runtime v0.9.0 h1:K6/tLhXKzhsEUUddTvEWWnQLLrawWyw1ptNK7NBDpDU=
 github.com/crossplane/crossplane-runtime v0.9.0/go.mod h1:gNY/21MLBaz5KNP7hmfXbBXp8reYRbwY5B/97Kp4tgM=
 github.com/crossplane/crossplane-runtime v0.9.1-0.20200827134951-29150890d4c1 h1:WLHvlKo69ii3ezcCEXK0XVHZRoBMQex2ylQEcCUmxO4=
 github.com/crossplane/crossplane-runtime v0.9.1-0.20200827134951-29150890d4c1/go.mod h1:gNY/21MLBaz5KNP7hmfXbBXp8reYRbwY5B/97Kp4tgM=
--- a/pkg/controller/ec2/subnet/controller.go
+++ b/pkg/controller/ec2/subnet/controller.go
@@ -103,6 +103,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}).Send(ctx)
 
 	if err != nil {
+		// when the subnet is deleted externally and we sent a create request with the cr data in k8s
+		// either of AvailabilityZone or AvailabilityZoneID needs to be set
+		if cr.Spec.ForProvider.AvailabilityZone != nil && cr.Spec.ForProvider.AvailabilityZoneID != nil {
+			cr.Spec.ForProvider.AvailabilityZoneID = nil
+		}
+		if err := e.kube.Update(ctx, cr); err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, errKubeUpdateFailed)
+		}
 		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsSubnetNotFoundErr, err), errDescribe)
 	}
 
@@ -141,6 +149,10 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
 	cr, ok := mgd.(*v1beta1.Subnet)
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
+	}
+
+	if cr.Spec.ForProvider.AvailabilityZone != nil && cr.Spec.ForProvider.AvailabilityZoneID != nil {
+		return managed.ExternalCreation{}, errors.New("Both AvailabilityZone and AvailabilityZoneID cannot be passed")
 	}
 
 	cr.Status.SetConditions(runtimev1alpha1.Creating())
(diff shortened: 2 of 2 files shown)


frezbo:
This is a little hack to get around the issue, let me know if there's a better way.

muvaf:
@frezbo Do you encounter #332 only in the cases where you have your `Subnet` CR in your cluster but something deleted the actual subnet in AWS console?

frezbo:
@muvaf yes you are right, it only happens if the subnet is deleted from the console and the `Subnet` CR is already present.

muvaf:
Got it. In Kubernetes conventions, `spec` is the utmost truth for the desired state and the controllers try really hard to not override a value there because then they would risk destroying user's intent. So, none of Crossplane controllers actually changes a field in spec unless it's empty, i.e. during [late-init](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#late-initialization) we fill only the empty values, not change the existing ones.

So, I believe returning an error and having the user manually fix it would be the safer option compared to removing a value in the spec. Given that it's an edge case where Crossplane will try recreate the resource even if we don't return an error, it should be fair to ask user to fix the problem so that Crossplane keeps to be dummy and respects the user's input no matter what.

frezbo:
@muvaf  I agree to that, but in this case, I did not pass in those fields, the `late-init` added those. For example this is my subnet spec:

```yaml
kind: Subnet
metadata:
  name: frezbo-pub
spec:
  forProvider:
    cidrBlock: 10.64.1.0/24
    tags:
      - key: Name
        value: frezbo-pub
    vpcId: "vpc-096af621e9899c2a7"
    mapPublicIPOnLaunch: true
  reclaimPolicy: Delete
  providerConfigRef:
    name: aws-provider
```

in this case the user provided spec did not have those additional fields. As a user I would expect it to re-create it with the spec I provided, not the spec that got updated. Consider a large infrastrucuture definition where the subnet got deleted. it would break a lot of downstream dependencies and it would be a lot of work tracking down what failed. 

Does this UX make sense?

frezbo:
I think this issue would be present for many other Cloud resources. It makes perfect sense for a k8s construct like a `pod ` or `deployment` to have the `spec` be the source of truth. But when it managing a cloud resource a user would expect it to be recreated with the `spec` the user provided, not the `spec` in k8s

muvaf:
The main thing here is that different reconciliation loops always look up to spec for desired state and since late-init fills the empty values, it's the desired state now. There are pros and cons of doing late-init but in our discussions during the [API design](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md) we have reached the same conclusion with Kubernetes and started doing it in all controllers.

One of the costs of doing late-init, as you experienced, the spec is the superset of what users provides during the initial creation. While this causes some confusion, the advantages of being able to see everything about the resource via `kubectl` over weighs it. For example, `RDSInstance` has over 50 fields but you can create it with only 2-3 fields. Once it's created, all those fields that are filled by AWS start to show up on your CR, hence later on during Day 2 operations when you want to make a change, you don't really need to go to console to see what's the existing value. It's all there for you to see and change.

Also keep in mind that the recreation is only a by-product here. Crossplane aims to do what users tell it to do. For example, the initial creation of CR does mean that user wants to recreate the resource or change of a field means that user wants a different value for that field. However, in this case, without any user action in the cluster, Crossplane recreates the resource if it's deleted in the AWS console. That's not really something we aim for but we have it, in fact, it's not really useless besides stateless resources like network. For example, if your RDS instance is deleted in the Console, recreation by Crossplane isn't really meaningful - the data is lost.

To conclude, for the reasons above, I don't think the value we get here justifies making a change in an existing value of the `spec`.

frezbo:
@muvaf Awesome. Thanks for the explanation. I was not aware of the design ideas. 👍 . I'll close this out.

Issue #307, opened 2020-07-29
Support for AWS Elastic Container Registry (ECR)


muvaf:
### What problem are you facing?

There is no managed resource for ECR.

### How could Crossplane help solve your problem?

Implement a managed resource for ECR.

enderv:
Taking a stab at this as I have time

PR #337, opened 2020-09-14
Adding ECR support

enderv:

Adding support for Elastic Container Repository. Fixes https://github.com/crossplane/provider-aws/issues/307
### Checklist

I have:


apis/ecr/v1alpha1/repository_types.go
```diff
@@ -0,0 +1,122 @@
...
+
+// ImageScanningConfiguration Scanning Configuration
+type ImageScanningConfiguration struct {
+
+	// The setting that determines whether images are scanned after being pushed
+	// to a repository. If set to true, images will be scanned after being pushed.
+	// If this parameter is not specified, it will default to false and images will
+	// not be scanned unless a scan is manually started with the StartImageScan
+	// API.
+	ScanOnPush bool `json:"scanOnPush,omitempty"`
```
hasheddan, line 89: I will typically make a single field in an embedded struct required (i.e. no `omitempty` here) if its parent is optional so that supplying the parent necessitates supplying the single child.

apis/ecr/v1alpha1/repository_types.go
```diff
@@ -0,0 +1,122 @@
...
+	ForProvider RepositoryParameters `json:"forProvider"`
+}
+
+// RepositoryObservation keeps the state for the external resource
+type RepositoryObservation struct {
+	// The date and time, in JavaScript date format, when the repository was created.
+	CreatedAt *metav1.Time `json:"createdAt,omitempty"`
+
+	// The image scanning configuration for a repository.
+	ImageScanningConfiguration ImageScanningConfiguration `json:"imageScanningConfiguration,omitempty"`
```
hasheddan, line 59: Can we omit and late initialize in the spec?

pkg/controller/ecr/repository/controller.go
```diff
@@ -0,0 +1,281 @@
...
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+type connector struct {
+	kube client.Client
+}
+
+func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
+	cfg, err := awsclients.GetConfig(ctx, c.kube, mg, "")
```
hasheddan, line 82: Is `Repository` global or scoped to a region? If the former we should pass `"aws-global"` here, if the latter we should add `region` field to spec and pass it here 👍 (this is brand new, see https://github.com/crossplane/provider-aws/pull/339 for more info)

hasheddan (commented):
This is looking really good @enderv! Thanks for adding this, a few minor comments below 👍

apis/ecr/v1alpha1/repository_types.go
enderv: Not sure what you mean on this one. Have it being set in late initialize already if left out of parameters. Thought all the properties should be in the observation?

hasheddan: I just meant that I don't think we need it present in the observation if it will always match the spec exactly

enderv: Oh gotcha. I was looking at it as this could actually be different in the spec vs what is actually in aws as you can update this field in the spec. So there is that brief period where they could be different before update runs. But if thats not a concern then yeah I can pull it.

apis/ecr/v1alpha1/repository_types.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+)
+
+// RepositoryParameters define the desired state of an AWS Elastic Container Repository
+type RepositoryParameters struct {
+
+	// Region is the region you'd like your Certificate to be created in.
```
hasheddan, line 28:
```suggestion
	// Region is the region you'd like your Repository to be created in.
```

I am guessing this is the intended?

apis/ecr/v1alpha1/repository_types.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Metadata tagging key value pairs
+	// +optional
+	Tags []Tag `json:"tags,omitempty"`
+}
+
+// A RepositorySpec defines the desired state of a Elastic Container Repository.
+type RepositorySpec struct {
+	runtimev1alpha1.ResourceSpec `json:",inline"`
+	// +optional
```
hasheddan, line 52:
```suggestion
```

apis/ecr/v1alpha1/repository_types.go
hasheddan: Hmm yeah I think generally we would say that if the field in the spec is not accurate then the resource should indicate that it is updating to match that state, but not necessarily that the current state of the field is reflected in the status. I think we could also remove `ImageTagMutability` from the status below. More info [here](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity).

pkg/controller/ecr/repository/controller.go
```diff
@@ -0,0 +1,285 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	"github.com/crossplane/provider-aws/apis/ecr/v1alpha1"
+	awsclients "github.com/crossplane/provider-aws/pkg/clients"
+	ecr "github.com/crossplane/provider-aws/pkg/clients/ecr"
+)
+
+const (
+	errUnexpectedObject = "The managed resource is not an repository resource"
```
hasheddan, line 43:
```suggestion
	errUnexpectedObject = "managed resource is not an repository resource"
```

hasheddan (commented):
@enderv apologies for the delay here, a few minor updates then I think we are good to go :+1:

apis/ecr/v1alpha1/repository_types.go
enderv: yepp good ole copy paste

apis/ecr/v1alpha1/repository_types.go
enderv: Thanks for the link! Went ahead and removed them both.

enderv:
@hasheddan No worries and should be good to go now. Think I got everything.

hasheddan (approved):
LGTM! Thanks for the great work here @enderv!

Issue #336, opened 2020-09-11
ProviderConfig does not have region but this is required by some services 


muvaf:
### What happened?

When you try to create a `VPC` with `master` branch, you'll get the following error:
```yaml
  - lastTransitionTime: "2020-09-10T13:25:59Z"
    message: 'create failed: failed to create the VPC resource: unknown endpoint,
      could not resolve endpoint, partition: "aws", service: "ec2", region: "", known:
      [af-south-1 ap-northeast-2 eu-central-1 eu-west-1 fips-us-east-2 fips-us-west-1
      me-south-1 us-west-2 ap-southeast-2 eu-north-1 fips-ca-central-1 sa-east-1 us-east-1
      ap-south-1 ap-southeast-1 ca-central-1 eu-west-2 us-east-2 ap-east-1 ap-northeast-1
      eu-south-1 eu-west-3 fips-us-east-1 fips-us-west-2 us-west-1]'
    reason: Encountered an error during resource reconciliation
    status: "False"
    type: Synced
```

The reason is that client has to be configured with region so that it can construct the endpoint. However, a random endpoint doesn't suffice - it determines in which region the resource should be created. That's pretty different than other providers in the sense that a configuration parameter is included in client config instead of creation request.

### How can we reproduce it?

Create `VPC`

### What environment did it happen in?
Crossplane version: master
Provider-aws:  master

Issue #38, opened 2018-11-25
The AWS cloud providers should not include the region


bassam:
The region used by a given resource should be late bound and not part of the AWS Provider object.

PR #339, opened 2020-09-15
Fix missing region in all resources

muvaf:

Adds region field, does some fixes for CRDs and updates the examples. Note that `region` field is kept optional in `beta` resources. So, people who use `providerRef` will not be broken in this version.

I'm trying to test each and every resource and plan to do a revamp of the examples folder(fix many of them, add missing ones) which will include removal of all claim/class CRs. I want that folder to work with `kubectl apply -f` so maybe in the future we can have some kind of integration test for them.

Fixes https://github.com/crossplane/provider-aws/issues/336

### Checklist

I have:


muvaf:
It's all well and working now. I had to test each and every managed resource we have and do fixes :( `examples` folder is now accurate and we have an example for all managed resources.

hasheddan (commented):
Thanks for tackling this @muvaf! Mostly looks good to me with a few small comments :) Feel free to ping me for approval after fixes 👍

muvaf:
@hasheddan Thanks for the review! Addressed all of your comments.

hasheddan (approved):
Awesome! Thanks again @muvaf :)

kasey:
Adding some metadata: this PR is a fix for a breaking change introduced in https://github.com/crossplane/provider-aws/pull/320 which addressed https://github.com/crossplane/provider-aws/issues/38

PR #344, opened 2020-09-22
Remove claims and classes

muvaf:

This PR also removes the compute/v1alpha3 EKSCluster, which we have been keeping around because the v1beta1 EKSCluster is compatible only with composition and not resource claims.

See https://github.com/crossplane/crossplane/issues/1670 for context.
 

### Checklist

I have:


muvaf:
Updated with `master` branches and manually tested.

hasheddan (changes requested):
@muvaf I believe you'll need to remove package install from integration tests here: https://github.com/crossplane/provider-aws/blob/eab4250ee41843b8f4e317facae82fe213747179/cluster/local/integration_tests.sh#L236

pkg/controller/s3/s3bucket.go
```diff
@@ -210,7 +208,7 @@ func updateTagging(bucket *bucketv1alpha3.S3Bucket, client s3.Service) error {
 
 func (r *Reconciler) _delete(bucket *bucketv1alpha3.S3Bucket, client s3.Service) (reconcile.Result, error) {
 	bucket.Status.SetConditions(runtimev1alpha1.Deleting(), runtimev1alpha1.ReconcileSuccess())
-	if bucket.Spec.ReclaimPolicy == runtimev1alpha1.ReclaimDelete {
+	if bucket.Spec.DeletionPolicy == runtimev1alpha1.DeletionDelete {
```
negz, line 211: iirc Delete is now the default. Do we need to handle the case where the policy not set?

negz (approved):
I left one comment, and I think the integration tests will need to be disabled, but otherwise LGTM. For the integration tests I've taken an approach of removing them completely in a separate commit so that we can easily revert and update when the new packages are ready.

pkg/controller/s3/s3bucket.go
muvaf: This controller will be [gone](https://github.com/crossplane/provider-aws/pull/331) in the next days, so I don't think we need to handle it. I just made changes to get the controller in a state that can be compiled.

hasheddan (approved):
I'll add back integration tests when new packaging is ready :+1:

Issue #368, opened 2020-10-06
S3 AccelerateConfig is never deleted once created


muvaf:
### What happened?

If `spec.forProvider.accelerateConfig` is filled once, it's never deleted since at each reconcile it's being [late-initialized](https://github.com/crossplane/provider-aws/blob/b935c3e/pkg/controller/s3/bucketresources/accelerateConfig.go#L52). We need to know whether we should make `DeleteAccelerateConfig` call or late-init the existing accelerate config.

### How can we reproduce it?

Create a `Bucket` with `spec.forProvider.accelerateConfig` filled. Then try to delete `spec.forProvider.accelerateConfig`. It'll come back.

### What environment did it happen in?
AWS Provider: https://github.com/crossplane/provider-aws/commit/bac07bf21a97dc11d00b73903e992f1fe3f1fdc6

krishchow:
This was actually one of the cases I spent some time thinking about - for most of the resources where late initialize is implemented, there isn't actually an explicit delete function - e.g., on the Go SDK, there is no delete acceleration config call. Instead, by default, the value for these fields is _""_ (or sometimes _nil_), and once it has been set once, to "delete" it, you need to switch the value to suspended.

krishchow:
1. AccelerationConfiguration - not much we can do for this one, users will need to manually set the value to Suspended to "delete" it. 
2. LoggingConfiguration - we can wrap the internal struct with a LoggingEnabled struct, and just reflect the behavior of AWS
3. NotificationConfiguration - not sure how best to handle this currently.
4. RequestPaymentConfiguration - this cannot be deleted, so the late initialize here is okay.
5. VersioningConfiguration - this works similar to the AccelerationConfiguration, where the user will need to set the values to reflect the default values (suspended and disabled).

muvaf:
For resources that have `delete` calls, it seems like we either won't _delete_ them or not late-init them and I think your call makes sense; don't late-init if they have `delete` calls. Because I can see cases where the `delete` is actually only thing that'll get user where they would like to end up. For example, even if you change some settings in `WebsiteConfiguration`, a bucket whose website config doesn't exist could very well be a state user desires.

For the ones without delete calls, I think we should just late init and leave `Delete` func empty. Whatever config switch we think makes sense may not correspond to what user thinks as _deletion_. So, it'd be safer to let them decide what they'd like to do with the resource once they created it. Will open a PR and request your review

PR #369, opened 2020-10-06
Cleanup for S3

muvaf:

https://github.com/crossplane/provider-aws/pull/331 has recently been merged. This PR does some small renaming changes.

@krishchow Though one thing I notices when I was testing is that subresources are actually never deleted if late-init is implemented. Because late-init fills the spec with what exists and `Observe` returns `NeedsUpdate` only if the spec of that resource is `nil`, for example `Spec.ForProvider.AccelerateConfiguration`. I don't have a very good solution that keeps both late-init and subresource deletion work as expected. I opened https://github.com/crossplane/provider-aws/issues/368 so that we can circle back to think about it. 

Fixes #

I have:

Manually with `examples/s3` folder.


examples/s3/bucketpolicy.yaml
```diff
@@ -21,7 +20,7 @@ spec:
         principal:
           allowAnon: true # this boolean flag makes the bucket public, this takes priority over the aws principal.
         resource:
-          - test.s3bucket.aws.crossplane.io
+          - test-bucket-mvf
```
hasheddan, line 23: Just wanted to confirm this is a valid value here?

pkg/controller/s3/bucket/CORSConfig.go
```diff
@@ -14,14 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package bucketresources
+package bucket
```
hasheddan, line 17: nitpick: I might have missed conversation on original PR, but was there a specific reason for adding these subresource clients to `controller` rather than `client` dir?

examples/s3/bucketpolicy.yaml
muvaf: Nope, thanks for catching!

pkg/controller/s3/bucket/CORSConfig.go
muvaf: I guess they could go to either one of them. They are very similar to `ExternalClient` with their interface which makes them closer to be in `controller` package but at the same time they are called by the main controller which is similar to other things in `clients` folder. So, I'm on the fence.

On another note, fix for https://github.com/crossplane/provider-aws/issues/368 will likely change the interface and a few things so we can discuss this there as well since the shape of the interface would affect where they go.

krishchow: It was previously in the client dir, and I moved it under the controller since each struct is responsible for maintaining the lifecycle of a specific sub-resource similar to the controller. From my perspective, it doesn't make that much sense to have it under the client, since the structs are responsible for reconciling in the state of their resource.

pkg/controller/s3/bucket.go
```diff
@@ -142,11 +141,8 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
-	_, err := e.s3client.CreateBucketRequest(s3.GenerateCreateBucketInput(meta.GetExternalName(cr), cr.Spec.ForProvider)).Send(ctx)
-	if s3.IsAlreadyExists(err) {
-		return managed.ExternalCreation{}, nil
-	}
```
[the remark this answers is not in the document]
muvaf: > If this scenario is not valid (i.e., brownfield buckets)

@krishchow What do you mean by brownfield buckets?

My reasoning to remove this statement was that in `Observe`, we report back whether resource exists or not. Then managed reconciler [calls](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reconciler/managed/reconciler.go#L622) `Create` only if it doesn't exist. So, the only case we'd get an `AlreadyExists` error is if the bucket is created between those two calls (nanosecond level diff maybe?), which is negligible.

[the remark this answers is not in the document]
krishchow: Ah, I am specifically referring to the case where the user wants to add an S3 bucket that have previously created into the cluster as a Bucket. Currently, that error is ignored intentionally so we can support this use case. If you don't think this is in-scope for the AWS provider at this time, we can ignore this use case for now and potentially return to this discussion in the future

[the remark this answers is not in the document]
muvaf: > I am specifically referring to the case where the user wants to add an S3 bucket that have previously created into the cluster as a Bucket. Currently, that error is ignored intentionally so we can support this use case.

I believe this use case is [covered](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/s3/bucket.go#L99) by `Observe`. If it doesn't exist, then `ResourceExists: false` (not an error case) and `Create` will be called. If it does, then it'll move on as usual as if it'd been created by provider-aws.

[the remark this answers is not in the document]
krishchow: Ah yeah, you are correct, I guess I am focusing too much on that specific case - this PR LGTM :)

examples/s3/bucketpolicy.yaml
hasheddan: @muvaf looks like this may not have been updated yet?

muvaf: @hasheddan fixed now

hasheddan (approved):
LGTM! Thanks @muvaf!

Issue #372, opened 2020-10-08
Route53 ResourceRecordSet with AilasTarget is not working


YasukeXXX:
### What happened?

I tried to manage record set with crossplane, but I got an error.

error messsage:  
```
   Warning  CannotCreateExternalResource  4s (x3 over 6s)  managed/resourcerecordset.route53.aws.crossplane.io  failed to create the ResourceRecordSet resource: InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, ChangeResourceRecordSetsInput.ChangeBatch.Changes[0].ResourceRecordSet.ResourceRecords.
```

manifest:
```yaml
apiVersion: route53.aws.crossplane.io/v1alpha1
kind: ResourceRecordSet
metadata:
  name: test
  annotations:
    crossplane.io/external-name: test.xxxxxxx.com.
spec:
  forProvider:
    type: A
    zoneId: XXXXXXXXXXXXX
    aliasTarget:
      dnsName: xxxxx.region.elb.amazonaws.com
      evaluateTargetHealth: true
      hostedZoneId: XXXXXXXXXXXXX
  reclaimPolicy: Orphan
  providerConfigRef:
    name: aws-provider

```
Is there something wrong?

### How can we reproduce it?

Fill `xxx` values and apply above manifest.

### What environment did it happen in?
Crossplane version: 0.12.0

muvaf:
It seems like AWS has an unnecessary validation in its Go SDK. Opened a bug here https://github.com/aws/aws-sdk-go-v2/issues/803

PR #376, opened 2020-10-12
Adopt AWS SDK behavior for resource record array

muvaf:
## Description of your changes

@skmcgrail [informed](https://github.com/aws/aws-sdk-go-v2/issues/803#issuecomment-706436385) that the array here is expected to be `nil` if it's empty. The issue doesn't exist in the newer versions of AWS SDK but until we update to the new versions, this fix is necessary.

Fixes https://github.com/crossplane/provider-aws/issues/372

Fixes #

I have:

Manually tested with an alias target.


hasheddan (approved):
Nice! Thanks @muvaf 🎉

PR #377, opened 2020-10-12
s3: subresources that cannot be deleted should not have deletion implemented

muvaf:

Removes the `Delete` implementations of S3 subresources that do not have corresponding deletion call and fixes some bugs related to versioning and tagging.

Fixes https://github.com/crossplane/provider-aws/issues/368
Fixes https://github.com/crossplane/provider-aws/issues/379

Fixes #

I have:

Manually with `examples/s3/bucket.yaml`

### DIFF
--- a/pkg/controller/s3/bucket/versioningConfig.go
+++ b/pkg/controller/s3/bucket/versioningConfig.go
@@ -53,12 +53,8 @@ func (in *VersioningConfigurationClient) LateInitialize(ctx context.Context, buc
 		bucket.Spec.ForProvider.VersioningConfiguration = &v1beta1.VersioningConfiguration{}
 		config = bucket.Spec.ForProvider.VersioningConfiguration
 	}
-	if len(external.Status) != 0 { // By default Status is the string ""
-		config.Status = aws.String(string(external.Status))
-	}
-	if len(external.MFADelete) != 0 { // By default MFADelete is the string ""
-		config.MFADelete = aws.String(string(external.MFADelete))
-	}
+	config.Status = aws.LateInitializeStringPtr(config.Status, aws.String(string(external.Status)))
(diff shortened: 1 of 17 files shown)


hasheddan (approved):
LGTM, but would also like @krishchow to sign off 👍

pkg/controller/s3/bucket/versioningConfig.go
```diff
@@ -54,12 +53,8 @@ func (in *VersioningConfigurationClient) LateInitialize(ctx context.Context, buc
...
 	}
-	if len(external.Status) != 0 { // By default Status is the string ""
-		config.Status = aws.String(string(external.Status))
-	}
-	if len(external.MFADelete) != 0 { // By default MFADelete is the string ""
-		config.MFADelete = aws.String(string(external.MFADelete))
-	}
+	config.Status = aws.LateInitializeStringPtr(config.Status, aws.String(string(external.Status)))
+	config.MFADelete = aws.LateInitializeStringPtr(config.MFADelete, aws.String(string(external.MFADelete)))
 	return nil
```
krishchow, line 58: Minor nitpick: I believe the default values of external.Status and external.MFADelete is empty strings `""`, so with this implementation, we are always late-initializing the config to be a struct with those values, even though that's not what the user provided, or what the external representation is. I think the current implementation works, but it just leaves some ambiguous values.

krishchow (commented):
Aside from the one-note, this PR LGTM, I think it does a nice job of making everything more consistent externally and therefore easier to reason about what late-initializes and what doesn't.

pkg/controller/s3/bucket/versioningConfig.go
muvaf: I checked the implementation and `aws.String(string(external.Status))` returns `nil` if the given string is `""`. So, we don't give any value if the returned string is `""`

PR #382, opened 2020-10-13
Custom Service Endpoint Support

### LINKED ISSUE #199: Custom Endpoints
Deploying a Crossplane enabled Kubernetes in Private AWS VPCs that has custom service endpoints.

The AWS Golang SDK supports resolution for custom endpoints:

https://github.com/aws/aws-sdk-go/blob/9c9db07028aeb7d607e54b2b5aa97bbbbebcbbfd/service/s3/endpoint.go#L102
https://github.com/aws/aws-sdk-go/blob/4b8b821e1bcd8466b496d1856ea11b735e549e4c/models/endpoints/endpoints.json
https://github.com/aws/aws-sdk-go/blob/f26268b77ea9619d95b1d04b6f3e3c5a03b61110/example/aws/endpoints/customEndpoint/customEndpoint.go

Does provider-aws already support custom endpoints for services? If not, where would be a good entry-point for implementation?

Promaethius:

Adds annotation parsing in the form of
```
aws.crossplane.io/endpointURL:
aws.crossplane.io/endpointSigningRegion:
```

When a new service is created, these annotations are used to add custom urls and signing regions to a service configuration. The service name is matched against the `apiVersion` so that a resolver function is generated specific to that service.

Fixes #199 

I have:

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -25,11 +25,13 @@ import (
 	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/session"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
@@ -96,7 +98,8 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 
 	switch s := pc.Spec.Credentials.Source; s { //nolint:exhaustive
 	case runtimev1alpha1.CredentialsSourceInjectedIdentity:
-		return UsePodServiceAccount(ctx, []byte{}, DefaultSection, region)
+		cfg, err := UsePodServiceAccount(ctx, []byte{}, DefaultSection, region)
+		return SetResolver(ctx, mg, cfg), err
 	case runtimev1alpha1.CredentialsSourceSecret:
 		csr := pc.Spec.Credentials.SecretRef
 		if csr == nil {
@@ -106,10 +109,33 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 		if err := c.Get(ctx, types.NamespacedName{Namespace: csr.Namespace, Name: csr.Name}, s); err != nil {
 			return nil, errors.Wrap(err, "cannot get credentials secret")
 		}
-		return UseProviderSecret(ctx, s.Data[csr.Key], DefaultSection, region)
+		cfg, err := UseProviderSecret(ctx, s.Data[csr.Key], DefaultSection, region)
+		return SetResolver(ctx, mg, cfg), err
 	default:
 		return nil, errors.Errorf("credentials source %s is not currently supported", s)
 	}
+}
+
+// SetResolver parses annotations from the managed resource
+// and returns a configuration accordingly.
+func SetResolver(ctx context.Context, mg resource.Managed, cfg *aws.Config) *aws.Config {
+	if URL, ok := mg.GetAnnotations()["aws.alpha.crossplane.io/endpointURL"]; ok {
+		if SigningRegion, ok := mg.GetAnnotations()["aws.alpha.crossplane.io/endpointSigningRegion"]; ok {
+			defaultResolver := endpoints.NewDefaultResolver()
+			endpointResolver := func(service, region string) (aws.Endpoint, error) {
+				if strings.Contains(mg.GetObjectKind().GroupVersionKind().Version, service) {
+					return aws.Endpoint{
+						URL:           URL,
+						SigningRegion: SigningRegion,
+					}, nil
+				}
+
+				return defaultResolver.ResolveEndpoint(service, region)
+			}
+			cfg.EndpointResolver = aws.EndpointResolverFunc(endpointResolver)
+		}
+	}
+	return cfg
 }
 
 // UseProvider to produce a config that can be used to authenticate to AWS.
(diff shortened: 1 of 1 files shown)


Promaethius:
```
annotations:
  aws.crossplane.io/endpointURL: http://localhost
  aws.crossplane.io/endpointSigningRegion: localhost
```
-- or --
```
annotations:
  aws.crossplane.io/customendpoint:
    URL: http://localhost
    SigningRegion: localhost
```

pkg/clients/aws.go
```diff
@@ -102,12 +105,35 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
...
+		return SetResolver(ctx, mg, cfg), err
 	default:
 		return nil, errors.Errorf("credentials source %s is not currently supported", s)
 	}
 }
 
+// SetResolver parses annotations from the managed resource
+// and returns a configuration accordingly.
+func SetResolver(ctx context.Context, mg resource.Managed, cfg *aws.Config) *aws.Config {
+	if URL, ok := mg.GetAnnotations()["aws.crossplane.io/endpointURL"]; ok {
```
negz, line 118: Do these endpoints change on a service-to-service basis? I'm wondering if we should put the annotations on the `ProviderConfig` instead (or even make them a first class field). If we overrode the endpoint on the `ProviderConfig`, would the same endpoint apply to both RDS instances and S3 buckets (for example)?

Promaethius: it changes from service to service unfortunately

pkg/clients/aws.go
```diff
@@ -102,12 +105,35 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
...
+		if SigningRegion, ok := mg.GetAnnotations()["aws.crossplane.io/endpointSigningRegion"]; ok {
+			defaultResolver := endpoints.NewDefaultResolver()
+			endpointResolver := func(service, region string) (aws.Endpoint, error) {
+				if strings.Contains(mg.GetResourceVersion(), service) {
```
negz, line 122: I believe `GetResourceVersion` will return the `metadata.resourceVersion` of the managed resource, which is like:

```yaml
metadata:
  resourceVersion: "147443"
```

Is using this field intentional? It doesn't seem like it would ever contain the `service` name.

Promaethius: Ah I'm trying to get the `apiVersion` field.

pkg/clients/aws.go
```diff
@@ -102,12 +105,35 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
...
 	default:
 		return nil, errors.Errorf("credentials source %s is not currently supported", s)
 	}
 }
 
+// SetResolver parses annotations from the managed resource
+// and returns a configuration accordingly.
+func SetResolver(ctx context.Context, mg resource.Managed, cfg *aws.Config) *aws.Config {
+	if URL, ok := mg.GetAnnotations()["aws.crossplane.io/endpointURL"]; ok {
+		if SigningRegion, ok := mg.GetAnnotations()["aws.crossplane.io/endpointSigningRegion"]; ok {
```
hasheddan, line 119: What would you think about making these annotations `aws.alpha.crossplane.io` since this is a new feature?

Promaethius: Works for me. What do you think about using an annotation object rather than several annotations?

hasheddan: I think it is _generally_ recommended to use separate annotations

pkg/clients/aws.go
Promaethius: Is it `mg.GetObjectKind().GroupVersionKind().String()`

pkg/clients/aws.go
Promaethius: Ok then I'll leave it as is

Promaethius:
@negz @hasheddan changes made.

hasheddan (approved):
Looking good to me here! Thanks for this @Promaethius! Have you done any manual testing with this change?

Promaethius:
@hasheddan did manual testing confirming that a random URL would make the client complain in the object description and that switching to aws given endpoints would work. Added a little more functionality. Crossplane aws objects don't stick to the ServiceID requirements and there's no way for the client to pass it into the config if the client is initialized WITH the config so it has to be an extra annotation :/

Promaethius:
@muvaf could I get your eyes on this?

hasheddan:
@Promaethius apologies for the long delay here! Could you rebase on `master`?

pkg/clients/aws.go
```diff
@@ -92,7 +94,8 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 
 	switch s := pc.Spec.Credentials.Source; s { //nolint:exhaustive
 	case runtimev1alpha1.CredentialsSourceInjectedIdentity:
-		return UsePodServiceAccount(ctx, []byte{}, DefaultSection, region)
+		cfg, err := UsePodServiceAccount(ctx, []byte{}, DefaultSection, region)
+		return SetResolver(ctx, mg, cfg), err
```
muvaf, line 98: We have an alternative `UseProviderConfigV1` in `master`. Could you add that statement there as well?

muvaf (approved):
I have one addition I'd like to be made but otherwise looks good to go!

pkg/clients/aws.go
Promaethius: @muvaf I can't find `UseProviderConfigV1`. What is it?

muvaf: Sorry the wrong name, it's [`GetConfigV1`](https://github.com/crossplane/provider-aws/blob/master/pkg/clients/aws.go#L246). It's an alternative method for controllers that use AWS SDK v1, which are mostly the ones generated by AWS ACK. I can make the addition later if you'd like this to get merged now.

Promaethius: I had to implement it as a v1 function cause the versions diverge now on how to handle endpoints but this should do it

Issue #383, opened 2020-10-14
Update AWS SDK v2 version


muvaf:
### What problem are you facing?

I've been getting `The XML you provided was not well-formed or did not validate against our published schema` in some S3 calls and I suspect it might be related to the AWS SDK version (`github.com/aws/aws-sdk-go-v2 v0.23.0`) we are using.

### How could Crossplane help solve your problem?

Update to [v0.26.0](https://github.com/aws/aws-sdk-go-v2/releases/tag/v0.26.0) which has some significant changes from what I can tell.

enderv:
Started working on bumping to their rc .0.31.0 as I have time. It is indeed a major change.

muvaf:
@enderv FWIW, we use AWS SDK v1 as well, which is GA as opposed to v2. So, if you don't want to deal with updating SDK, you can opt for using v1 in the controller you want to make a change.

negz:
We're continuing to use SDK v1 because ACK uses SDK v1 - we will revisit this later once there's ACK support for v2.

enderv:
Question on this: all of the current non ack resources use v2 already.  Just gonna stick on the old v2 version until they go to ack generated resources?

muvaf:
@enderv I think we're indifferent about what SDK they use but unless necessary, I wouldn't spend much effort updating them to newer v2 versions. In fact, we can think about downgrading them to v1 if the feature that's available in newer v2 versions is available there.

muvaf:
v2 SDK has gone GA and the current released version is [v1.3.0](https://github.com/aws/aws-sdk-go-v2/releases/tag/v1.3.0) which I think we should upgrade to as soon as we can.

enderv:
Have done a lot of work on this in my [aws-update branch](https://github.com/enderv/provider-aws/tree/aws-update) to get to 1.1. Just keep getting distracted with other things and new resources keep coming in. Did another rebase this morning and a few tests are failing but just wanted to call it out here in case someone else had more time and didn't want to start from scratch.

muvaf:
Fixed with https://github.com/crossplane/provider-aws/pull/602 and https://github.com/crossplane/provider-aws/pull/855

PR #384, opened 2020-10-14
s3: type corrections and fixes

muvaf:

I have tested most of the subresources manually and fixed the errors as I go. I couldn't get to success in some of them like replication configuration due to either me not knowing how to configure them properly or https://github.com/crossplane/provider-aws/issues/383

I believe we should do a patch release after this PR is merged.

I have:

Manually.


apis/s3/v1beta1/notificationConfiguration_types.go
```diff
@@ -111,9 +111,8 @@ type TopicConfiguration struct {
 
 	// The Amazon Resource Name (ARN) of the Amazon SNS topic to which Amazon S3
 	// publishes a message when it detects events of the specified type.
-	//
-	// TopicArn is a required field
-	TopicArn string `json:"topicArn"`
+	// At least one of topicArn, topicArnRef or topicSelector is required.
+	TopicArn string `json:"topicArn,omitempty"`
```
hasheddan, line 115: Is this field not required? Should it be `*string`?

apis/s3/v1beta1/lifecycleConfiguration_types.go
```diff
@@ -46,7 +46,7 @@ type LifecycleRule struct {
 	Filter *LifecycleRuleFilter `json:"filter,omitempty"`
 
 	// Unique identifier for the rule. The value cannot be longer than 255 characters.
-	ID *string `json:"ID,omitempty"`
+	ID *string `json:"id,omitempty"`
```
hasheddan, line 49: @muvaf this would be a breaking change and I think would necessitate a minor release, rather than a patch.

krishchow:
Working on some changes, I'll open a PR in your fork @muvaf :)

apis/s3/v1beta1/lifecycleConfiguration_types.go
muvaf: @hasheddan Do you think we should bump it to `v1beta2`? Also, I can separate that commit from the others and save jsontag changes for the major release.

hasheddan: Separating commits sounds good :) And I think the correct versioning would be to bump to `v1beta2`. I wonder if this would be a good opportunity to experiment with conversion webhooks...

muvaf: I think we can get away with providing manual instructions for now per [Kubernetes versioning](https://github.com/kubernetes/community/blob/3302b37/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions).

muvaf: So I went ahead and removed the actual breaking changes and the diff is as following:
image

However, I realized that the sub-resources that those fields belong didn't really work in the last release anyway. So, while technically it's more correct to bump to `v1beta2`, in practice I feel like we won't be breaking anyone using `Bucket` resource if we changed those fields and did a patch release. @hasheddan what do you think?

muvaf: @hasheddan pinging

hasheddan: Hmm I am not entirely sure the current behavior of these subresources, my only reservation would be if this could cause a case where a currently valid S3 bucket instance now failed (i.e. if we change field names that are required then a valid instance would become invalid). Since we don't have webhooks yet, maybe we should keep as `v1beta1` but just do a `v0.14` release?

apis/s3/v1beta1/notificationConfiguration_types.go
hasheddan: following up here

apis/s3/v1beta1/lifecycleConfiguration_types.go
muvaf: > if this could cause a case where a currently valid S3 bucket instance now failed

A valid working bucket wouldn't fail because in order to be working, it shouldn't include those fields anyway. If it did, then it's failing because of the errors fixed.

muvaf: @hasheddan Oh by the way, a full release sounds good.

muvaf:
@krishchow merged your commit into this PR. @hasheddan all your feedback should be addressed.

krishchow:
Did a second pass, LGTM :)

muvaf:
@hasheddan rebased. Can you take a final look? I think we're good to go for a release after this is merged.

hasheddan (approved):
LGTM, thanks @muvaf! Looks like DCO may be missing on a commit

PR #389, opened 2020-10-17
S3 Acceleration Parameter Optional

### LINKED ISSUE #388: S3 Acceleration not available in Region
### What happened?
Applying any s3 object against crossplane configured against a AWS Gov region results in an eternal retry loop.
https://github.com/crossplane/provider-aws/blob/master/pkg/controller/s3/bucket/accelerateConfig.go#L69

```
status code: 400, request id:  xx, host id: xx
  Warning  CannotObserveExternalResource  2m21s (x15 over 2m25s)  managed/bucket.s3.aws.crossplane.io  (combined from similar events): cannot get Bucket accelerate configuration: UnsupportedArgument: The request contained an unsupported argument.
```

### How can we reproduce it?
Get credentials for an AWS GovCloud region or see that terraform confirmed a similar issue here:
https://github.com/terraform-providers/terraform-provider-aws/issues/11556

Crossplane version: v0.13

Promaethius:

Rig the `AccelerateConfigurationClient` such that it will not panic if the method or parameter is not allows for Acceleration Configuration similar to the Terraform Provider. This is necessary for some regions which does not have S3 acceleration support.

Fixes #388

I have:

Manually tested with bucket creation in region `us-gov-west-1`.
s3.yaml
```
kind: Bucket
metadata:
  name: crossplane-bucket
spec:
  forProvider:
    locationConstraint: us-gov-west-1
    acl: private
    versioningConfiguration:
      status: Enabled
  providerConfigRef:
    name: default
```
Output:
```
  Normal   CreatedExternalResource  5s                 managed/bucket.s3.aws.crossplane.io  Successfully requested creation of external resource
```

### DIFF
--- a/pkg/clients/s3/bucket.go
+++ b/pkg/clients/s3/bucket.go
@@ -43,6 +43,10 @@ var (
 	TaggingErrCode = "NoSuchTagSet"
 	// WebsiteErrCode is the error code sent by AWS when the website config does not exist
 	WebsiteErrCode = "NoSuchWebsiteConfiguration"
+	// MethodNotAllowed is the error code sent by AWS when the request method for an object is not allowed
+	MethodNotAllowed = "MethodNotAllowed"
+	// UnsupportedArgument is the error code sent by AWS when the request fields contain an argument that is not supported
+	UnsupportedArgument = "UnsupportedArgument"
 )
 
 // BucketClient is the interface for Client for making S3 Bucket requests.
@@ -192,6 +196,22 @@ func TaggingNotFound(err error) bool {
 // WebsiteConfigurationNotFound is parses the aws Error and validates if the website configuration does not exist
 func WebsiteConfigurationNotFound(err error) bool {
 	if s3Err, ok := err.(awserr.Error); ok && s3Err.Code() == WebsiteErrCode {
+		return true
+	}
+	return false
+}
+
+// MethodNotSupported is parses the aws Error and validates if the method is allowed for a request
+func MethodNotSupported(err error) bool {
+	if s3Err, ok := err.(awserr.Error); ok && s3Err.Code() == MethodNotAllowed {
+		return true
+	}
+	return false
+}
+
+// ArgumentNotSupported is parses the aws Error and validates if parameters are now allowed for a request
+func ArgumentNotSupported(err error) bool {
+	if s3Err, ok := err.(awserr.Error); ok && s3Err.Code() == UnsupportedArgument {
 		return true
 	}
 	return false
--- a/pkg/controller/s3/bucket/accelerateConfig.go
+++ b/pkg/controller/s3/bucket/accelerateConfig.go
@@ -42,6 +42,10 @@ type AccelerateConfigurationClient struct {
 func (in *AccelerateConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
 	if err != nil {
+		// Short stop method for requests in a region without Acceleration Support
+		if s3.MethodNotSupported(err) || s3.ArgumentNotSupported(err) {
+			return nil
+		}
 		return errors.Wrap(err, accelGetFailed)
 	}
 
@@ -69,6 +73,10 @@ func NewAccelerateConfigurationClient(client s3.BucketClient) *AccelerateConfigu
 func (in *AccelerateConfigurationClient) Observe(ctx context.Context, bucket *v1beta1.Bucket) (ResourceStatus, error) {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
 	if err != nil {
+		// Short stop method for requests in a region without Acceleration Support
+		if s3.MethodNotSupported(err) || s3.ArgumentNotSupported(err) {
+			return Updated, nil
+		}
 		return NeedsUpdate, errors.Wrap(err, accelGetFailed)
 	}
 	if bucket.Spec.ForProvider.AccelerateConfiguration != nil &&


pkg/controller/s3/bucket/accelerateConfig.go
```diff
@@ -68,7 +69,7 @@ func NewAccelerateConfigurationClient(client s3.BucketClient) *AccelerateConfigu
 // Observe checks if the resource exists and if it matches the local configuration
 func (in *AccelerateConfigurationClient) Observe(ctx context.Context, bucket *v1beta1.Bucket) (ResourceStatus, error) {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
-	if err != nil {
+	if err != nil || strings.Contains(err.Error(), "MethodNotAllowed") || strings.Contains(err.Error(), "UnsupportedArgument") {
```
muvaf, line 72: For classifying errors, we use a specific type of function so that it can be used in our other helper functions like `resource.Ignore`. You can see some examples [here](https://github.com/crossplane/provider-aws/blob/b935c3e/pkg/clients/s3/bucket.go#L153).

pkg/controller/s3/bucket/accelerateConfig.go
```diff
@@ -68,7 +69,7 @@ func NewAccelerateConfigurationClient(client s3.BucketClient) *AccelerateConfigu
...
 		return NeedsUpdate, errors.Wrap(err, accelGetFailed)
```
muvaf, line 73: I believe if the error is `MethodNotAllowed`, the returned status should be `Updated` since there is no operation to be done here.

pkg/controller/s3/bucket/accelerateConfig.go
muvaf: Additionally, are we certain that `UnsupportedArgument` happens if and only if the accelerate config is not available in the region? It feels like we could get the same error for having unnecessary additional fields in the HTTP request, which isn't really possible since SDK designates it, but I wanted to know if there is a documentation somewhere that we can be sure about it. Also a comment here in the code would definitely help future readers.

muvaf (commented):
Thank you @Promaethius ! Your tests & fixes help a lot!

pkg/controller/s3/bucket/accelerateConfig.go
Promaethius: Yes there is! In the linked issue, there is a link to how the Terraform provider for aws handles it. I lifted this method from there cause I had the same concerns.

pkg/controller/s3/bucket/accelerateConfig.go
Promaethius: Good point.

Promaethius: Ah I messed up the logic, I did this late last night. Should be and does not contain.

pkg/clients/s3/bucket.go
```diff
@@ -197,6 +201,14 @@ func WebsiteConfigurationNotFound(err error) bool {
 	return false
 }
 
+// AllowedParameters is parses the aws Error and validates if the method and parameters are now allowed for a request
+func AllowedParameters(err error) bool {
```
muvaf, line 205: I'd prefer to have two error functions instead of one and have the `OR` statement in where they are necessary, i.e. `accelerateConfig.go`, but I guess that's also fine if you think these two errors happens interchangeably in all cases without difference.

Promaethius: That's fair, these seem like general errors that could happen anywhere.

Promaethius:
@muvaf g2g afaik

pkg/controller/s3/bucket/accelerateConfig.go
```diff
@@ -68,7 +68,7 @@ func NewAccelerateConfigurationClient(client s3.BucketClient) *AccelerateConfigu
 // Observe checks if the resource exists and if it matches the local configuration
 func (in *AccelerateConfigurationClient) Observe(ctx context.Context, bucket *v1beta1.Bucket) (ResourceStatus, error) {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
-	if err != nil && !s3.AllowedParameters(err) {
+	if err != nil && !s3.MethodNotSupported(err) && !s3.ArgumentNotSupported(err) {
 		return NeedsUpdate, errors.Wrap(err, accelGetFailed)
 	}
 	if bucket.Spec.ForProvider.AccelerateConfiguration != nil &&
```
muvaf, line 74: Would `external.Status` result in `nil` pointer exception if one of those errors are returned? I vaguely remember that even in error case, this API returns an object that's not `nil` but I'm not sure whether that behavior is consistent between different AWS endpoints.

muvaf: @Promaethius just pinging to check whether you had the chance to validate that. FWIW, you can also just:
```
external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
if s3.MethodNotSupported(err) || s3.ArgumentNotSupported(err){
  return Updated, nil
}
if err != nil {
...
```
to make sure it doesn't proceed if the error is of those kinds.

pkg/controller/s3/bucket/accelerateConfig.go
```diff
@@ -69,6 +73,10 @@ func NewAccelerateConfigurationClient(client s3.BucketClient) *AccelerateConfigu
 func (in *AccelerateConfigurationClient) Observe(ctx context.Context, bucket *v1beta1.Bucket) (ResourceStatus, error) {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
 	if err != nil {
+		// Short stop method for requests in a region without Acceleration Support
+		if !s3.MethodNotSupported(err) && !s3.ArgumentNotSupported(err) {
```
muvaf, line 77:
```suggestion
		if s3.MethodNotSupported(err) || s3.ArgumentNotSupported(err) {
```

I think that's what you want to achieve, right? Skip if error is of one of those kinds?

pkg/controller/s3/bucket/accelerateConfig.go
```diff
@@ -42,6 +42,10 @@ type AccelerateConfigurationClient struct {
 func (in *AccelerateConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
 	external, err := in.client.GetBucketAccelerateConfigurationRequest(&awss3.GetBucketAccelerateConfigurationInput{Bucket: aws.String(meta.GetExternalName(bucket))}).Send(ctx)
 	if err != nil {
+		// Short stop method for requests in a region without Acceleration Support
+		if !s3.MethodNotSupported(err) && !s3.ArgumentNotSupported(err) {
```
muvaf, line 46:
```suggestion
		if s3.MethodNotSupported(err) || s3.ArgumentNotSupported(err) {
```

Promaethius:
@muvaf you were right an or works just as well

muvaf (approved):
LGTM! Thanks @Promaethius ! Could you squash the commits before I merge?

Promaethius:
@muvaf done!

PR #390, opened 2020-10-20
s3 bucket: adds sorting for tagging lists, minor refactoring around tags

### LINKED ISSUE #385: S3 Bucket stuck in Ready state
Created new S3 Bucket and it stuck in the Creating phase, try to check event logs and everything look normal.

Although the managed resource is created successfully in AWS UI

```
status:
  atProvider:
    arn: arn:aws:s3:::mybucket-1
  conditions:
  - lastTransitionTime: "2020-10-15T08:12:56Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2020-10-15T08:12:58Z"
    reason: ReconcileSuccess
    status: "True"
    type: Synced
```

`Spec` as follow:
```
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: us-west-1
    objectLockEnabledForBucket: false
      rules:
          sseAlgorithm: AES256
    tagging:
      tagSet:
      - key: key1
        value: value1
      - key: key2
        value: value2
    versioningConfiguration:
      status: Enabled
  providerConfigRef:
    name: aws-provider
  writeConnectionSecretToRef:
    name: mybucket-1
    namespace: crossplane-system

```

Crossplane version: provider-aws:v0.12.0-10.g71a09e8

krishchow:

Fixes #385 

I have:

Added additional unit tests, manual testing with an S3 bucket (updated example.yaml)

### DIFF
--- a/pkg/clients/s3/bucket.go
+++ b/pkg/clients/s3/bucket.go
@@ -213,41 +213,21 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
 	return err
 }
 
-// CopyTag converts a local v1beta.Tag to an S3 Tag
-func CopyTag(tag *v1beta1.Tag) *s3.Tag {
-	if tag == nil {
-		return nil
-	}
-	return &s3.Tag{
-		Key:   aws.String(tag.Key),
-		Value: aws.String(tag.Value),
-	}
-}
-
 // CopyTags converts a list of local v1beta.Tags to S3 Tags
 func CopyTags(tags []v1beta1.Tag) []s3.Tag {
-	if tags == nil {
-		return nil
-	}
-	out := make([]s3.Tag, len(tags))
-	for i := range tags {
-		out[i] = *CopyTag(&tags[i])
+	out := make([]s3.Tag, 0)
+	for _, one := range tags {
+		out = append(out, s3.Tag{Key: aws.String(one.Key), Value: aws.String(one.Value)})
 	}
 	return out
 }
 
 // SortS3TagSet stable sorts an external s3 tag list by the key and value.
 func SortS3TagSet(tags []s3.Tag) []s3.Tag {
-	if len(tags) == 0 {
-		return tags
-	}
-	sort.SliceStable(tags, func(i, j int) bool {
-		if aws.StringValue(tags[i].Key) < aws.StringValue(tags[j].Key) {
-			return true
-		} else if aws.StringValue(tags[i].Key) == aws.StringValue(tags[j].Key) {
-			return aws.StringValue(tags[i].Value) < aws.StringValue(tags[j].Value)
-		}
-		return false
+	outTags := make([]s3.Tag, len(tags))
+	copy(outTags, tags)
+	sort.SliceStable(outTags, func(i, j int) bool {
+		return aws.StringValue(outTags[i].Key) < aws.StringValue(outTags[j].Key)
 	})
-	return tags
+	return outTags
 }
--- a/pkg/clients/s3/bucket.go
+++ b/pkg/clients/s3/bucket.go
@@ -19,6 +19,7 @@ package s3
 import (
 	"context"
 	"fmt"
+	"sort"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
@@ -210,4 +211,43 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
 	}
 	_, err := client.PutBucketAclRequest(config).Send(ctx)
 	return err
+}
+
+// CopyTag converts a local v1beta.Tag to an S3 Tag
+func CopyTag(tag *v1beta1.Tag) *s3.Tag {
+	if tag == nil {
+		return nil
+	}
+	return &s3.Tag{
+		Key:   aws.String(tag.Key),
+		Value: aws.String(tag.Value),
+	}
+}
+
+// CopyTags converts a list of local v1beta.Tags to S3 Tags
+func CopyTags(tags []v1beta1.Tag) []s3.Tag {
+	if tags == nil {
+		return nil
+	}
+	out := make([]s3.Tag, len(tags))
+	for i := range tags {
+		out[i] = *CopyTag(&tags[i])
+	}
+	return out
+}
+
+// SortS3TagSet stable sorts an external s3 tag list by the key and value.
+func SortS3TagSet(tags []s3.Tag) []s3.Tag {
+	if len(tags) == 0 {
+		return tags
+	}
+	sort.SliceStable(tags, func(i, j int) bool {
+		if aws.StringValue(tags[i].Key) < aws.StringValue(tags[j].Key) {
+			return true
+		} else if aws.StringValue(tags[i].Key) == aws.StringValue(tags[j].Key) {
+			return aws.StringValue(tags[i].Value) < aws.StringValue(tags[j].Value)
+		}
+		return false
+	})
+	return tags
 }
--- a/pkg/controller/s3/bucket/lifecycleConfig.go
+++ b/pkg/controller/s3/bucket/lifecycleConfig.go
@@ -28,7 +28,6 @@ import (
 	"github.com/crossplane/provider-aws/apis/s3/v1beta1"
 	aws "github.com/crossplane/provider-aws/pkg/clients"
 	"github.com/crossplane/provider-aws/pkg/clients/s3"
-	"github.com/crossplane/provider-aws/pkg/controller/s3/testing"
 )
 
 const (
@@ -63,6 +62,8 @@ func (in *LifecycleConfigurationClient) Observe(ctx context.Context, bucket *v1b
 		}
 		return NeedsUpdate, errors.Wrap(resource.Ignore(s3.LifecycleConfigurationNotFound, err), lifecycleGetFailed)
 	}
+
+	sortFilterTags(external.Rules)
 
 	switch {
 	case len(external.Rules) != 0 && config == nil:
@@ -129,13 +130,14 @@ func GenerateRules(in *v1beta1.BucketLifecycleConfiguration) []awss3.LifecycleRu
 		if local.Filter != nil {
 			rule.Filter = &awss3.LifecycleRuleFilter{
 				Prefix: local.Filter.Prefix,
-				Tag:    testing.CopyTag(local.Filter.Tag),
+				Tag:    s3.CopyTag(local.Filter.Tag),
 			}
 			if local.Filter.And != nil {
 				rule.Filter.And = &awss3.LifecycleRuleAndOperator{
 					Prefix: local.Filter.And.Prefix,
-					Tags:   testing.CopyTags(local.Filter.And.Tags),
+					Tags:   s3.CopyTags(local.Filter.And.Tags),
 				}
+				s3.SortS3TagSet(rule.Filter.And.Tags)
 			}
 		}
 		rules[i] = rule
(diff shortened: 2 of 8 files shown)


pkg/clients/s3/bucket.go
```diff
@@ -211,3 +212,42 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
 	_, err := client.PutBucketAclRequest(config).Send(ctx)
 	return err
 }
+
+// CopyTag converts a local v1beta.Tag to an S3 Tag
+func CopyTag(tag *v1beta1.Tag) *s3.Tag {
```
muvaf, line 217: I'm guessing that you implemented this function to not have to have that `if tag == nil` check each time you need it. In such cases where the logic that's being abstracted is really simple, I usually lean towards doing it in place so that it's easier to read and know what it does immediately without changing the context and checking what that function does, hence, would prefer duplication over reuse.

pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -163,3 +165,13 @@ func (in *LifecycleConfigurationClient) Delete(ctx context.Context, bucket *v1be
 	).Send(ctx)
 	return errors.Wrap(err, lifecycleDeleteFailed)
 }
+
+func sortFilterTags(rules []awss3.LifecycleRule) {
+	if len(rules) != 0 {
```
muvaf, line 170: If `len(rules) == 0`, `for` loop wouldn't run at all so the function would be no-op. So, I think we can remove that `if` check here.

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -222,3 +216,13 @@ func (in *ReplicationConfigurationClient) Delete(ctx context.Context, bucket *v1
 	).Send(ctx)
 	return errors.Wrap(err, replicationDeleteFailed)
 }
+
+func sortReplicationRules(rules []awss3.ReplicationRule) {
+	if len(rules) != 0 {
```
muvaf, line 221: If `len(rules) == 0`, `for` loop wouldn't run at all so the function would be no-op. So, I think we can remove that `if` check here.

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -81,13 +80,8 @@ func GenerateTagging(config *v1beta1.Tagging) *awss3.Tagging {
...
 		return &awss3.Tagging{TagSet: make([]awss3.Tag, 0)}
 	}
-	conf := &awss3.Tagging{TagSet: make([]awss3.Tag, len(config.TagSet))}
-	for i, v := range config.TagSet {
-		conf.TagSet[i] = awss3.Tag{
-			Key:   aws.String(v.Key),
-			Value: aws.String(v.Value),
-		}
-	}
+	conf := &awss3.Tagging{TagSet: nil}
```
muvaf, line 83: Is this line necessary?

muvaf (commented):
A few nitpicks, but mostly good to go.

pkg/controller/s3/bucket/lifecycleConfig.go
krishchow: This was mainly here since `len(nil)` would also resolve to 0, but I have changed this to something more clear.

pkg/clients/s3/bucket.go
krishchow: Sounds good, I have made this change for CopyTag and CopyTags

pkg/clients/s3/bucket.go
```diff
@@ -211,3 +212,42 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
...
+	out := make([]s3.Tag, len(tags))
+	for i := range tags {
+		out[i] = *CopyTag(&tags[i])
+	}
+	return out
+}
+
+// SortS3TagSet stable sorts an external s3 tag list by the key and value.
+func SortS3TagSet(tags []s3.Tag) []s3.Tag {
+	if len(tags) == 0 {
```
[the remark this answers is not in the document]
krishchow: This is also here to handle the case where `tags == nil`, but I changed this to be more clear as well.

pkg/clients/s3/bucket.go
```diff
@@ -211,3 +212,42 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
...
+		return tags
+	}
+	sort.SliceStable(tags, func(i, j int) bool {
+		if aws.StringValue(tags[i].Key) < aws.StringValue(tags[j].Key) {
+			return true
+		} else if aws.StringValue(tags[i].Key) == aws.StringValue(tags[j].Key) {
+			return aws.StringValue(tags[i].Value) < aws.StringValue(tags[j].Value)
```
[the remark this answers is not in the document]
krishchow: for `ExampleSliceStable`, the comment states "Sort by name, preserving original order", however, in this scenario we can't really trust the sorting from AWS at all, since I couldn't find any documentation that clearly indicated how it is ordered. So instead, I opted to sort by both since we can handle duplicate keys inside of a list of tags.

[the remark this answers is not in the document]
muvaf: Oh sorry, I thought the line compared keys but it seems it compares values. What I meant was this line should handle it:
```
return aws.StringValue(tags[i].Key) < aws.StringValue(tags[j].Key)
```

I think it's fair to expect that keys are unique in the array.

pkg/clients/s3/bucket.go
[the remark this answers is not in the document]
muvaf: Both `nil` and `len(tags) == 0` should be [handled](https://play.golang.org/p/B5WE7Hf1lzs) by `sort.SliceStable`

pkg/clients/s3/bucket.go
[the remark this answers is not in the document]
krishchow: https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html - nvm! did some more research, keys do **need** to be unique inside the array.

pkg/clients/s3/bucket.go
muvaf: Since we get that `if` out, I think we can get rid of this function and just write the following where it's called  like:
```
rule.Filter.Tag = &s3.Tag{
		Key:   aws.String(tag.Key),
		Value: aws.String(tag.Value),
	}
```

pkg/clients/s3/bucket.go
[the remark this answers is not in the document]
krishchow: Thanks! I was not aware of this behavior.

pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -163,3 +169,14 @@ func (in *LifecycleConfigurationClient) Delete(ctx context.Context, bucket *v1be
...
+func sortFilterTags(rules []awss3.LifecycleRule) {
+	if rules == nil {
+		return
+	}
+	for i := range rules {
+		if rules[i].Filter != nil && rules[i].Filter.And != nil {
+			s3.SortS3TagSet(rules[i].Filter.And.Tags)
+		}
+	}
+}
```
muvaf, line 182:
```suggestion
func sortFilterTags(rules []awss3.LifecycleRule) {
	for i := range rules {
		if rules[i].Filter != nil && rules[i].Filter.And != nil {
			s3.SortS3TagSet(rules[i].Filter.And.Tags)
		}
	}
}
```

That was what I meant. `for` loop won't run at all if `rules == nil` or `len(rules) == 0`. So, we don't need that check there.

pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -129,12 +130,17 @@ func GenerateRules(in *v1beta1.BucketLifecycleConfiguration) []awss3.LifecycleRu
...
+			if local.Filter.Tag != nil {
+				rule.Filter.Tag = s3.CopyTag(local.Filter.Tag)
 			}
 			if local.Filter.And != nil {
 				rule.Filter.And = &awss3.LifecycleRuleAndOperator{
 					Prefix: local.Filter.And.Prefix,
-					Tags:   testing.CopyTags(local.Filter.And.Tags),
+				}
+				if local.Filter.And.Tags != nil {
+					rule.Filter.And.Tags = s3.CopyTags(local.Filter.And.Tags)
```
muvaf, line 142:
```suggestion
					rule.Filter.And.Tags = &s3.Tag{Key:   aws.String(tag.Key),Value: aws.String(tag.Value)}
```

That was what I meant by removing `CopyTag` function.

pkg/clients/s3/bucket.go
```diff
@@ -211,3 +212,28 @@ func UpdateBucketACL(ctx context.Context, client BucketClient, bucket *v1beta1.B
...
+func CopyTags(tags []v1beta1.Tag) []s3.Tag {
+	out := make([]s3.Tag, 0)
+	for _, one := range tags {
+		out = append(out, s3.Tag{Key: aws.String(one.Key), Value: aws.String(one.Value)})
+	}
+	return out
+}
+
+// SortS3TagSet stable sorts an external s3 tag list by the key and value.
+func SortS3TagSet(tags []s3.Tag) []s3.Tag {
```
muvaf, line 234: This operation is in-place and I see that in some places we use the returned value, like ina `switch` statement, and in some cases we ignore the returned value since we know that the array we supplied is changed.

I'd propose making a choice here; if the function makes changes in-place then no value should be returned or we can copy the array, make the operation and return the sorted copy of the array. Because as a user of this function, I wouldn't expect the operation to be in-place if a value is returned. It's fine either way but I think one should be chosen.

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -222,3 +218,14 @@ func (in *ReplicationConfigurationClient) Delete(ctx context.Context, bucket *v1
...
+
+func sortReplicationRules(rules []awss3.ReplicationRule) {
+	if rules == nil {
+		return
+	}
+	for i := range rules {
+		if rules[i].Filter != nil && rules[i].Filter.And != nil {
+			s3.SortS3TagSet(rules[i].Filter.And.Tags)
+		}
+	}
```
muvaf, line 230:
```suggestion
	for i := range rules {
		if rules[i].Filter != nil && rules[i].Filter.And != nil {
			s3.SortS3TagSet(rules[i].Filter.And.Tags)
		}
	}
```

That was what I meant. `for` loop won't run at all if `rules == nil` or `len(rules) == 0`. So, we don't need that check there.

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -81,13 +80,7 @@ func GenerateTagging(config *v1beta1.Tagging) *awss3.Tagging {
 	if config == nil || config.TagSet == nil {
 		return &awss3.Tagging{TagSet: make([]awss3.Tag, 0)}
```
muvaf, line 81: Not changed in this PR, but `nil` is preferred if we know for sure that the array is empty, unless SDK has some weird behavior change between 0-length array vs `nil`.

pkg/controller/s3/bucket/lifecycleConfig.go
krishchow: Ah gotcha, fixed!

pkg/clients/s3/bucket.go
krishchow: Sounds good to me, I had written this way originally because I there were certain cases where I wanted the list returned, and others were it was sufficient if we sorted the list in place, but both cases can be simplified if we just always return a new list.

muvaf (approved):
LGTM! Thanks for working through the review passes! I assume this has been manually tested. Let me know when it's ready to merge.

krishchow:
Hey @muvaf, I did some more testing today with larger tagsets (e.g., 100 tags), and it seems like everything is working as expected :) Should be good to merge

PR #391, opened 2020-10-20
s3 bucket policy to v1alpha2

### LINKED ISSUE #374: S3 Bucket Policy to v1beta
We should bring the bucket policy for S3 to v1beta1 by adding full support for all policy fields.

krishchow:

Fixes #374

I have:

Additional unit tests, and manual testing


krishchow:
The main reason I opened this draft PR early on in the process is to get feedback from the community. There are a few points of discussion I wanted to try and get a dialog on:

- I haven't changed the API group yet, because I am not sure if this qualifies as ready for v1beta1, or if it should be v1alpha2.
- Currently, I am templating the _Resource_ field of the bucket policy with the ARN prefix, and then the rest of the bucket path. This isn't required, but it is more of a QOL feature, since it is generally less to write for users, and I have not been able to find a case where the resource is not an S3 bucket. However, I can see a clear reason for removing it, since it may be unclear to new users of the provider-aws since the input syntax is different from a regular AWS s3 bucket policy.

krishchow:
- Ended up moving the bucket to v1alpha2 to stay on the safe side.
- I am no longer templating the resource ARN, I don't think it was necessary, and it could make it harder to understand the bucket policy initially.

hasheddan (commented):
@krishchow apologies for the delay in review here! This looks pretty good to me :) It looks like the main change is that we reference the IAM user directly instead of extracting it from the `Bucket`, are there any other major differences here?

@muvaf could you take a peek at this as well?

krishchow:
@hasheddan that's one of the main changes yes :). Instead of referencing a single IAMUser for the entire policy, each statement can refer to an IAMUser. 

I have also added support for a few other identity types supported by the bucket policy, also the other big thing support for Condition in the policy. This was one key feature that wasn't previously supported.

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Crossplane Authors.
+Copyright 2020 The Crossplane Authors.
```
muvaf, line 2: We don't update license of existing files.

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
 	// BucketNameSelector selects a reference to an S3Bucket to retrieve its bucketName
```
muvaf, line 47: I think some of the parameters are immutable like `BucketName`

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
 }
 
 // BucketPolicyStatement defines an individual statement within the
 // BucketPolicyBody
 type BucketPolicyStatement struct {
 	// Optional identifier for this statement, must be unique within the
 	// policy if provided.
-	StatementID string `json:"sid,omitempty"`
+	// +optional
+	StatementID *string `json:"sid,omitempty"`
```
muvaf, line 58:
```suggestion
	StatementID *string `json:"statementId,omitempty"`
```

Is there a specific reason you want the jsontag be different than Go field name?

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
 type BucketPrincipal struct {
 	// This flag indicates if the policy should be made available
 	// to all anonymous users.
 	AllowAnon bool `json:"allowAnon,omitempty"`
 
 	// This list contains the all of the AWS IAM users which are affected
-	// by the policy statement
-	AWSPrincipal []string `json:"aws,omitempty"`
+	// by the policy statement.
+	AWSPrincipals []AWSPrincipal `json:"aws,omitempty"`
```
muvaf, line 111:
```suggestion
	AWSPrincipals []AWSPrincipal `json:"awsPrincipals,omitempty"`
```

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
+	// IAMUserARNSelector queries for an IAMUser to retrieve its userName
+	// +optional
+	IAMUserARNSelector *runtimev1alpha1.Selector `json:"IAMUserArnSelector,omitempty"`
+
+	// AWSAccountID identifies an AWS account as the principal
+	// +optional
+	AWSAccountID *string `json:"awsAccountId,omitempty"`
+
+	// IAMRoleARN contains the ARN of an IAM role
+	// +optional
```
muvaf, line 143:
```suggestion
	// +optional
	// +immutable
```

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
-func tryFirst(slc []string) interface{} {
-	if len(slc) == 1 {
-		return slc[0]
-	}
-	return slc
+// AWSPrincipal wraps the potential values a policy
+// principal can take. Only one of the values should be set.
+type AWSPrincipal struct {
+	// IAMUserARN contains the ARN of an IAM user
+	// +optional
```
muvaf, line 127:
```suggestion
	// +optional
	// +immutable
```

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
+	// IAMUserARNRef contains the reference to an IAMUser
+	// +optional
+	IAMUserARNRef *runtimev1alpha1.Reference `json:"IAMUserArnRef,omitempty"`
+
+	// IAMUserARNSelector queries for an IAMUser to retrieve its userName
+	// +optional
+	IAMUserARNSelector *runtimev1alpha1.Selector `json:"IAMUserArnSelector,omitempty"`
+
+	// AWSAccountID identifies an AWS account as the principal
+	// +optional
```
muvaf, line 139:
```suggestion
	// +optional
	// +immutable
```

apis/s3/v1alpha2/referencers.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// ResolvePrincipal resolves all the IAMUser and IAMRole references in a BucketPrincipal
+func ResolvePrincipal(ctx context.Context, r *reference.APIResolver, principal *BucketPrincipal, statementIndex int) error {
+	if principal == nil || principal.AWSPrincipals == nil {
```
muvaf, line 66:
```suggestion
	if principal == nil {
```

If the array is nil, for loop would just exit and `return nil` would be executed.

examples/s3/bucket.yaml
```diff
@@ -3,8 +3,8 @@ kind: Bucket
 metadata:
   name: test-bucket
 spec:
+  deletionPolicy: Delete
```
muvaf, line 6:
```suggestion
```

nitpick: other examples don't have it.

examples/s3/bucket.yaml
```diff
@@ -36,4 +38,7 @@ spec:
           expiration:
             days: 15
   providerConfigRef:
-    name: example
\ No newline at end of file
+    name: example
+  writeConnectionSecretToRef:
+    name: mybucket-1
```
muvaf, line 43: What does bucket write to the connection details?

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -24,7 +24,6 @@ import (
 
 // BucketPolicyParameters define the desired state of an AWS BucketPolicy.
 type BucketPolicyParameters struct {
-
 	// Region is where the Bucket referenced by this BucketPolicy resides.
 	Region string `json:"region"`
```
muvaf, line 29: I think `PolicyStatement []BucketPolicyStatement json:"statement"` should be `Statements []BucketPolicyStatement json:"statements"`

apis/s3/v1alpha2/bucketpolicy_types.go
```diff
@@ -48,147 +47,133 @@ type BucketPolicyParameters struct {
...
 
 	// Used with the S3 policy to specify the users which are not included
 	// in this policy
+	// +optional
 	NotPrincipal *BucketPrincipal `json:"notPrincipal,omitempty"`
 
 	// Each element of the PolicyAction array describes the specific
 	// action or actions that will be allowed or denied with this PolicyStatement.
+	// +optional
 	PolicyAction []string `json:"action,omitempty"`
```
muvaf, line 78: We should match the Go field names with corresponding jsontag unless there is a big reason for not doing so.

muvaf (commented):
Sorry for the delayed review. There are a few things to address and I think it's the right call to make it `v1alpha2` for now. Thanks @krishchow !

apis/s3/v1alpha2/bucketpolicy_types.go
krishchow: It was more for clarity, in the actual S3 bucket policy json, the field is "sid" and not "statementId", but I can switch it.

examples/s3/bucket.yaml
krishchow: Nothing anymore, I think this is a remanent of the previous bucket.yaml for the S3Bucket. Removed now.

apis/s3/v1alpha2/bucketpolicy_types.go
muvaf: Gotcha. Then it seems like changing Go field names makes more sense since we'd like to keep the UX of IAM policy as is. Do all the jsontags map to classic policy fields right now after the last push?

krishchow: Yeah, that is what I ended up doing, I think this makes it more clear from the developer/maintainer experience as well. These changes have been pushed :)

muvaf (approved):
thanks @krishchow !

muvaf:
@krishchow could you resolve the conflict so that we can merge it?

krishchow:
@muvaf Rebased and fixed all the conflicts

Issue #210, opened 2020-05-01
VPC with dependencies marked as deleted, but not actually deleted when dependencies are removed


prasek:
UPDATE 10/20/2020: See related https://github.com/upbound/platform-ref-aws/issues/3
- this may be due to a `RouteTable` update instead of delete: 
https://github.com/upbound/platform-ref-aws/issues/3#issuecomment-713025391

### Problem
Deleted a `VPC` managed resource that had a bunch of dependencies that were eventually automatically deleted by AWS, but it's been maybe an hour and the `VPC` managed resource detect that it can be deleted, and keeps reporting:

```
    - lastTransitionTime: '2020-04-30T23:58:55Z'
      message: "delete failed: failed to delete the VPC resource: DependencyViolation: The vpc 'vpc-07e0865ec163c4cda' has dependencies and cannot be deleted.\n\tstatus code: 400, request id: 219066b3-53e3-48f4-8fb6-629813bc4732"
      reason: Encountered an error during resource reconciliation
      status: 'False'
      type: Synced
```

If I go into the AWS console to delete the VPC:
![image](https://user-images.githubusercontent.com/284840/80770461-3ea01f80-8b05-11ea-88b8-eba1f153ac5d.png)

I can manually delete the VPC via the AWS console without issue:
![image](https://user-images.githubusercontent.com/284840/80770499-537cb300-8b05-11ea-91cb-4b0c11559e35.png)

After I manually delete the VPC in the AWS console, Crossplane detects it and the `VPC` CR is removed.

Would be nice to Crossplane to successfully delete the VPC once all the dependencies have been removed.

### Details
`VPC`
```
apiVersion: network.aws.crossplane.io/v1alpha3
kind: VPC
metadata:
  annotations:
    crossplane.io/external-name: vpc-07e0865ec163c4cda
  creationTimestamp: '2020-04-30T19:05:35Z'
  deletionGracePeriodSeconds: 0
  deletionTimestamp: '2020-04-30T23:22:47Z'
  finalizers:
    - finalizer.managedresource.crossplane.io
  generation: 818
  labels:
    core.crossplane.io/parent-group: aws.stacks.crossplane.io
    core.crossplane.io/parent-kind: AWSSample
    core.crossplane.io/parent-name: my-cool-stack
    core.crossplane.io/parent-namespace: ''
    core.crossplane.io/parent-uid: 5462d834-d846-4aaf-9188-14c2ed1e3e6b
[...]

PR #393, opened 2020-10-21
routetable: do not issue status update before create 

### LINKED ISSUE #392: Routetable controller can lose track of created external resource
### What happened?

In some cases, creation of `RouteTable` succeeds but its ID cannot be written to spec when [this call](https://github.com/crossplane/provider-aws/blob/c0bacda/pkg/controller/ec2/routetable/controller.go#L185) returns error. As an example:
```
Events:
  Type     Reason                           Age                    From                                      Message
  ----     ------                           ----                   ----                                      -------
  Warning  CannotResolveResourceReferences  5m52s (x3 over 5m54s)  managed/routetable.ec2.aws.crossplane.io  cannot resolve references: spec.forProvider.vpcId: referenced field was empty (referenced resource may not yet be ready)
  Warning  CannotResolveResourceReferences  5m22s (x2 over 5m22s)  managed/routetable.ec2.aws.crossplane.io  cannot resolve references: spec.forProvider.routes[0].gatewayId: referenced field was empty (referenced resource may not yet be ready)
  Warning  CannotCreateExternalResource     4m51s                  managed/routetable.ec2.aws.crossplane.io  cannot update spec of the RouteTable custom resource: Operation cannot be fulfilled on routetables.ec2.aws.crossplane.io "network-smzhg-j2vvc": the object has been modified; please apply your changes to the latest version and try again
  Normal   CreatedExternalResource          4m49s                  managed/routetable.ec2.aws.crossplane.io  Successfully requested creation of external resource
  Normal   UpdatedExternalResource          39s (x20 over 4m47s)   managed/routetable.ec2.aws.crossplane.io  Successfully requested update of external resource
```

As you can see, the first `CannotCreateExternalResource` event says update failed, which means the ID wasn't saved and in the reconcile a new one is created and its ID is successfully stored.

The tricky part is it doesn't always happen because the error says the object is not up-to-date, however, we update it right before the `CreateRouteTableRequest` call via `kube.Status().Update`. It seems like in some cases (api-server too busy, etcd slowness etc.), this status update call doesn't get the most up-to-date object from the server so we end up in a local variable `cr` that is old to be used for an update.

### How can we reproduce it?

Create `RouteTable` objects in a constrained.

Crossplane version: 0.13.0
AWS: 83bf642

muvaf:

Because the object returned by the api-server may not be latest, hence, causing spec update to fail which in turn creates additional routetable whose track is lost.

Keep in mind that this should not happen as status update already updates the local `cr` variable. However, we see this happening in some environments.

The actual solution to this problem is to have `Create` operations whose result has to be saved to be atomic, i.e. if spec update didn't go through then delete the created external resource.

Fixes #392
Fixes https://github.com/crossplane/provider-aws/issues/210
Fixes https://github.com/upbound/platform-ref-aws/issues/3

I have:

Manually.

### DIFF
--- a/pkg/controller/ec2/routetable/controller.go
+++ b/pkg/controller/ec2/routetable/controller.go
@@ -53,7 +53,6 @@ const (
 	errAssociateSubnet    = "failed to associate subnet %v to the RouteTable resource"
 	errDisassociateSubnet = "failed to disassociate subnet %v from the RouteTable resource"
 	errSpecUpdate         = "cannot update spec of the RouteTable custom resource"
-	errStatusUpdate       = "cannot update status of the RouteTable custom resource"
 	errCreateTags         = "failed to create tags for the RouteTable resource"
 )
 
@@ -162,26 +161,16 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
-
-	cr.Status.SetConditions(runtimev1alpha1.Creating())
-	if err := e.kube.Status().Update(ctx, cr); err != nil {
-		return managed.ExternalCreation{}, errors.Wrap(err, errStatusUpdate)
-	}
-
 	result, err := e.client.CreateRouteTableRequest(&awsec2.CreateRouteTableInput{
 		VpcId: cr.Spec.ForProvider.VPCID,
 	}).Send(ctx)
-
 	if err != nil {
 		return managed.ExternalCreation{}, errors.Wrap(err, errCreate)
 	}
-
 	if result.RouteTable == nil {
 		return managed.ExternalCreation{}, errors.New(errCreate)
 	}
-
 	meta.SetExternalName(cr, aws.StringValue(result.RouteTable.RouteTableId))
-
 	return managed.ExternalCreation{}, errors.Wrap(e.kube.Update(ctx, cr), errSpecUpdate)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/ec2/routetable/controller.go
```diff
@@ -162,26 +161,16 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
 	if !ok {
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
-
-	cr.Status.SetConditions(runtimev1alpha1.Creating())
-	if err := e.kube.Status().Update(ctx, cr); err != nil {
-		return managed.ExternalCreation{}, errors.Wrap(err, errStatusUpdate)
-	}
```
hasheddan, line 169: @muvaf just want to check and make sure I am understanding this correctly --> this status update is bumping resource version, which is causing our `meta.SetExternalName` update to fail, meaning we try to `Create` again on next reconcile and end up with two Route Tables?

muvaf: No, if status update works as expected, there is no problem because the local `cr` object would be updated to reflect the latest state and the next statements would work on the latest object, hence no error. However, api-server doesn't always return the updated object in that call for some reason that could be low limit, too busy or slow etcd.

hasheddan: So we get a successful update but the returned object is not latest?

muvaf: For example, this problem happens in a constrained UbC environment 50% of the time.

hasheddan (approved):
Noting here that we investigated further and think the likely issue is actually that the status update here is triggering the composition controller and so the resource version gets bumped between our status and spec updates, causing the failing spec update (and thus the double route table creation). This is just a partial fix to make this scenario less likely, but doesn't eliminate this problem, and this will need to be addressed further. Thanks for working on this @muvaf!

PR #401, opened 2020-10-26
AWS Go SDK v1 connection method for ACK-generated resources

muvaf:

We will be using ACK code generation pipeline to add new [resources](https://github.com/aws/aws-controllers-k8s/pull/444) to `provider-aws` but ACK and generated code works with AWS Go SDK v1 whereas we use v2. One place they differ is how they authorize. This PR adds a function specific to v1.

Fixes #

I have:

Not tested yet with an actual controller.

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -27,12 +27,12 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/crossplane/provider-aws/apis/v1beta1"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	awsv1 "github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


pkg/clients/aws.go
```diff
@@ -235,6 +236,109 @@ func UsePodServiceAccount(ctx context.Context, _ []byte, _, region string) (*aws
...
+	if err := c.Get(ctx, types.NamespacedName{Name: mg.GetProviderConfigReference().Name}, pc); err != nil {
+		return nil, errors.Wrap(err, "cannot get referenced Provider")
+	}
+
+	t := resource.NewProviderConfigUsageTracker(c, &v1beta1.ProviderConfigUsage{})
+	if err := t.Track(ctx, mg); err != nil {
+		return nil, errors.Wrap(err, "cannot track ProviderConfig usage")
+	}
+
+	// TODO(muvaf): Support IRSA.
```
hasheddan, line 258: Looks like you are supporting IRSA already?

hasheddan (approved):
LGTM! Thanks @muvaf! One comment on your `TODO` that looks like it has already been implemented.

pkg/clients/aws.go
muvaf: Ah, good catch!

PR #403, opened 2020-10-26
iam: iam user access key resource v1alpha1

### LINKED ISSUE #394: IAM User Access Key
I would like to be able to create Access Keys for IAM User I create.

Crossplane (and the AWS Provider) should automatically create an access key for a new IAM user.

krishchow:

Fixes #394 

I have:

1. Manual testing w/ an IAM User and an Access Key (both under example/identity)
2. Unit testing for the IAM Access Key resource


krishchow:
Something I wanted to add to this PR - you cannot delete an IAM User without removing all the Access Keys. So currently, the user would need to manually remove their access keys prior to deleting an IAMUser CR. I think this behavior is okay, and the correct approach for now. However, to make this more clear, it might make sense to add a finalizer.

krishchow:
cc: @muvaf

pkg/controller/aws.go
```diff
@@ -75,6 +76,8 @@ func Setup(mgr ctrl.Manager, l logging.Logger) error {
 		nodegroup.SetupNodeGroup,
 		s3.SetupBucket,
 		bucketpolicy.SetupBucketPolicy,
+		iamaccesskey.SetupIAMAccessKey,
+		iamuser.SetupIAMUser,
 		iamuser.SetupIAMUser,
```
hasheddan, line 81:
```suggestion
		iamuser.SetupIAMUser,
```

hasheddan (commented):
> you cannot delete an IAM User without removing all the Access Keys... However, to make this more clear, it might make sense to add a finalizer.

If both the IAM User and the Access Keys are represented in Crossplane, then a user should see an error when deleting an IAM User that indicates keys exist, then they could check for `IAMAccessKeys` present in the cluster right? When you mention adding a finalizer, how would you imagine that working? We don't want to block deletion of an access key on an IAM user, and the IAM user will automatically block because it fails to delete right?

pkg/controller/aws.go
krishchow: Fixed! :)

krishchow:
> If both the IAM User and the Access Keys are represented in Crossplane, then a user should see an error when deleting an IAM User that indicates keys exist, then they could check for `IAMAccessKeys` present in the cluster right? When you mention adding a finalizer, how would you imagine that working? We don't want to block deletion of an access key on an IAM user, and the IAM user will automatically block because it fails to delete right?

Yes you are correct, you are correct in that it isn't necessary for the IAMUser (and I would prefer not having it), but I just wanted to bring this up. I was thinking about it from a clarity perspective, having a finalizer would make it clear there are resources connected to an IAMUser.

apis/identity/v1alpha1/iamaccesskey_types.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+)
+
+// IAMAccessKeyParameters define the desired state of an AWS IAM Access Key.
+type IAMAccessKeyParameters struct {
+	// IAMUsername contains the name of the IAMUser.
+	// +optional
```
muvaf, line 28:
```suggestion
	// +optional
	// +immutable
```

apis/identity/v1alpha1/iamaccesskey_types.go
```diff
@@ -0,0 +1,87 @@
...
+	ForProvider                  IAMAccessKeyParameters `json:"forProvider"`
+}
+
+// IAMAccessKeyObservation keeps the state for the external resource
+type IAMAccessKeyObservation struct {
+	// The ID of the access key that unique identifies this resource
+	AccessKeyID string `json:"accessKeyId,omitempty"`
+
+	// The current status of this IAMAccessKey
+	// +kubebuilder:validation:Enum=Active;Inactive
```
muvaf, line 52:
```suggestion
```

Since this is not entered by the user, I'd not put a validation there so that if AWS changes anything, we don't care.

apis/identity/v1alpha1/referencers.go
```diff
@@ -1,9 +1,12 @@
 /*
-Copyright 2019 The Crossplane Authors.
+Copyright 2020 The Crossplane Authors.
+
```
muvaf, line 3:
```suggestion
Copyright 2019 The Crossplane Authors.
```

We don't make changes on licenses of existing files.

apis/identity/v1alpha1/register.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Crossplane Authors.
+Copyright 2020 The Crossplane Authors.
```
muvaf, line 2:
```suggestion
Copyright 2019 The Crossplane Authors.
```

examples/iam/user.yaml
```diff
@@ -3,6 +3,8 @@ apiVersion: identity.aws.crossplane.io/v1alpha1
 kind: IAMUser
 metadata:
   name: someuser
+  labels:
+    example: "true"
```
muvaf, line 7: nitpick: using a ref instead of selector could make examples look simpler.

pkg/controller/identity/iamaccesskey/controller.go
```diff
@@ -0,0 +1,167 @@
...
+func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.ExternalCreation, error) {
+	cr, ok := mgd.(*v1alpha1.IAMAccessKey)
+	if !ok {
+		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
+	}
+
+	cr.Status.SetConditions(runtimev1alpha1.Creating())
+
+	response, err := e.client.CreateAccessKeyRequest(&awsiam.CreateAccessKeyInput{UserName: aws.String(cr.Spec.ForProvider.IAMUsername)}).Send(ctx)
+
```
muvaf, line 119: We should handle the error right where it happens and then proceed onto next steps.

pkg/controller/identity/iamaccesskey/controller.go
```diff
@@ -0,0 +1,167 @@
...
+	var conn managed.ConnectionDetails = nil
+	if response != nil && response.AccessKey != nil {
+		conn = managed.ConnectionDetails{}
+		conn[runtimev1alpha1.ResourceCredentialsSecretUserKey] = []byte(aws.StringValue(response.AccessKey.AccessKeyId))
+		conn[runtimev1alpha1.ResourceCredentialsSecretPasswordKey] = []byte(aws.StringValue(response.AccessKey.SecretAccessKey))
+
+		cr.Status.AtProvider.Status = string(response.AccessKey.Status)
+		cr.Status.AtProvider.AccessKeyID = aws.StringValue(response.AccessKey.AccessKeyId)
```
muvaf, line 127: I think we need to use external-name annotation for access key id. See VPC controller for handling provider-assigned name.

apis/identity/v1alpha1/iamaccesskey_types.go
muvaf: Well, it seems like `Status` is a configuration that user assigns as far as I can tell from `Update`. So, it should go to spec and we can have the validation there.

muvaf (commented):
There are a few issues but they should be straight-forward to fix. I think this will make it to the release. Thanks @krishchow !

pkg/controller/identity/iamaccesskey/controller.go
krishchow: Ah, so it shouldn't be inside of the status?

krishchow: Ok all done!

pkg/controller/identity/iamaccesskey/controller.go
```diff
@@ -0,0 +1,175 @@
...
+
+		meta.SetExternalName(cr, aws.StringValue(response.AccessKey.AccessKeyId))
+
+		err := e.kube.Update(ctx, cr)
+		if err != nil {
+			return managed.ExternalCreation{}, errors.Wrap(err, errKubeUpdateFailed)
+		}
+	}
+
+	return managed.ExternalCreation{ConnectionDetails: conn}, errors.Wrap(err, errCreate)
```
muvaf, line 143:
```suggestion
	return managed.ExternalCreation{ConnectionDetails: conn}, nil
```

pkg/controller/identity/iamaccesskey/controller.go
```diff
@@ -0,0 +1,175 @@
...
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	_, err := e.client.UpdateAccessKeyRequest(&awsiam.UpdateAccessKeyInput{
+		AccessKeyId: aws.String(meta.GetExternalName(cr)),
+		Status:      awsiam.StatusType(cr.Spec.ForProvider.Status),
+		UserName:    aws.String(cr.Spec.ForProvider.IAMUsername),
+	}).Send(ctx)
+
+	return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(iam.IsErrorNotFound, err), errUpdate)
```
muvaf, line 158:
```suggestion
	return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
```

If the resource does not exist, we should return an error here.

muvaf (approved):
LGTM! There are two small things need to be fixed. Let me know when you push a change that fixes them and I'll merge the PR. Thanks @krishchow !!

krishchow:
Fixed @muvaf, should be good to merge :)

hasheddan:
@krishchow looks like you need a `make generate` here.

krishchow:
@hasheddan I ran make generate locally, and there weren't any changes in any of the CRDs, I am not sure why check-diff failed (it's succeeding local). Is there a way to run the CI action again?

muvaf:
go mod tidy could help

krishchow:
Should be all fixed now, @hasheddan @muvaf, but I think pushing while the action was running led to this push skipping the CI step

PR #408, opened 2020-10-30
Enable leader election for controller manager

khos2ow:

This PR enables leader election for AWS provider controller manager. This can be controlled by `--leader-election` CLI flag or `LEADER_ELECTION=true` environment variables. This can be achieved for deployment by applying the following:

```yaml
apiVersion: pkg.crossplane.io/v1alpha1
kind: ControllerConfig
metadata:
  name: aws-config
spec:
  replicas: 2
  env:
    - name: "LEADER_ELECTION"
      value: "true"
```

Partial fix for https://github.com/crossplane/crossplane/issues/5

I have:

TBA

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -33,9 +33,10 @@ import (
 
 func main() {
 	var (
-		app        = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
-		debug      = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncPeriod = app.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").Duration()
+		app            = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
+		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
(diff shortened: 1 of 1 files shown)


negz (approved):

cmd/provider/main.go
```diff
@@ -53,7 +54,11 @@ func main() {
 	cfg, err := ctrl.GetConfig()
 	kingpin.FatalIfError(err, "Cannot get API server rest config")
 
-	mgr, err := ctrl.NewManager(cfg, ctrl.Options{SyncPeriod: syncPeriod})
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
+		LeaderElection:   *leaderElection,
+		LeaderElectionID: "crossplane-leader-election-provider-aws",
```
khos2ow, line 59: @negz  @hasheddan I believe it is possible to install multiple versions of a same provider at the same time, right? Is it possible for them to end up in the same namespace? If yes, this line will cause a collision.

hasheddan: This could only be possible if someone tried to make it happen (i.e. installed the same provider twice under different names) and none of the CRDs overlapped (a revision would fail to start if there were any CRDs in conflict), so we are safe from conflict here 👍

hasheddan (approved):

cmd/provider/main.go
khos2ow: Ah good to know, thanks! I will update the description of this and other PRs for test results and they should be good to go.

Issue #411, opened 2020-11-02
EKS sync fails after being ready


muvaf:
### What happened?

I created a `Cluster.eks` resource using https://github.com/upbound/platform-ref-aws and got the following conditions after a while it's ready:
```yaml
  conditions:
  - lastTransitionTime: "2020-11-02T21:18:45Z"
    reason: Available
    status: "True"
    type: Ready
  - lastTransitionTime: "2020-11-02T21:28:08Z"
    message: "update failed: cannot update EKS cluster configuration: InvalidParameterException:
      subnetIds and securityGroupIds cannot be updated.\n\tstatus code: 400, request
      id: d01c974e-95bd-4172-9ebf-aa5309abe86d"
    reason: ReconcileError
    status: "False"
    type: Synced
```

I didn't change anything in spec after creation.

### How can we reproduce it?

Create a `Cluster.eks` using the `subnetIds` and `securityGroupIds` fields.

### What environment did it happen in?
Crossplane version: 0.14.0-rc.115.g6a7731d8
AWS: v0.13.0-2.gd840deb7

negz:
Noticing the same issue, also using platform-ref-aws.

PR #413, opened 2020-11-03
eks.cluster: do not include subnetIds and securityGroupIds in Update

muvaf:

Fixes https://github.com/crossplane/provider-aws/issues/411

These fields are rejected in all Update calls so we shouldn't include them.

I have:

Manually.


hasheddan (approved):
Thanks @muvaf! I wonder if this is in API change that is reflected in newer versions of SDK 🤔

muvaf:
> I wonder if this is in API change that is reflected in newer versions of SDK

Not sure, though, I'm not surprised that they don't allow you to change the subnetIds.

PR #416, opened 2020-11-05
12 new API Gateway managed resources generated by ACK

muvaf:

Using https://github.com/aws/aws-controllers-k8s/pull/449 , we are able to generate all APIs under `apigatewayv2` that is made available in ACK. Note that update is not available in the generated code and I haven't added it manually yet.

TODOs:

- [x] Add examples to `examples` directory.
- [x] Manually test every one of the new APIs.
- [ ] (Optional) Implement update functionality in provided `hooks.go` files.

What do you guys think about including these resources in this week's release? If we're good I can complete the TODO list above and make it ready to review. I don't think we need to wait for https://github.com/aws/aws-controllers-k8s/pull/449 to be merged before using the generated code. After all, it's pretty similar to what we'd have written by hand.

Depends on https://github.com/crossplane/provider-aws/pull/401

Fixes #

I have:

I have yet to test them with a real cluster.


go.mod
```diff
@@ -39,5 +38,5 @@ require (
 	k8s.io/apimachinery v0.18.8
 	k8s.io/client-go v0.18.8
 	sigs.k8s.io/controller-runtime v0.6.2
-	sigs.k8s.io/controller-tools v0.2.4
+	sigs.k8s.io/controller-tools v0.4.0
```
hasheddan, line 41: @muvaf this is bumping all CRDs to v1 because v0.4.0 uses v1 by default. Is it your intention to make that change in this PR?

muvaf: No, I had to update to `v0.4.0` because that's the minimum supported by ACK-generated CRDs.

muvaf:
I just pushed some updates that include example YAMLs of `API` and `Stage` CRDs. Also, you can see how we reference `API` object from `Stage` via inlined `CustomStageParameters` struct.

Also, we needed a field ignore feature in ACK, which exists in https://github.com/aws/aws-controllers-k8s/pull/449 right now, hence `generator-config.yaml` includes `ignore.field_paths` that is used to specify what fields to ignore during generation. We need it to exclude identifier fields, which we use external-name annotation for, and referenced fields which we add to custom parameters struct because in many cases they are marked as required but in Crossplane we don't mark them required at CRD level since objects can be created without them having a value.

Right now, both `API` and `Stage` CRDs work as expected without update. It took a bit more than I expected due to changes necessary in ACK PR to accommodate unexpected `ignore.shape_name` behavior that used to work for ECR and different nature of `ReadMany` used in API Gateway resources.

muvaf:
Right now, we got 6 of the resources working end-to-end with references. The process of making them work is pretty mechanical:
* The field that will be used as external-name should be ignored in `generator-config.yaml`.
* The fields that are required and can be referenced should be ignored in `generator-config.yaml`.
* Write referenced fields into the resource's `Custom*Parameters` struct.
* Implement the referencer resolver function, which is mostly copy-paste.
* Implement `filterList` so that the correct object is chosen from returned list.
* Make sure ignored fields end up in the AWS request input objects, like `CreateApiInput`, via hooks.
* If the object's external-name is assigned by AWS, then we need to set external name in `postCreate` hook.

Note that there is no update at all since ACK doesn't currently provide any mechanisms for it. I'll complete the rest of the resources tomorrow and try to get updates implemented for some resources but I believe we can still merge this without updates.

muvaf:
@negz @hasheddan @kasey This is ready to go. I believe the places to review are mostly `custom_types.go`, `references.go` and `generator-config.yaml` in API folder. Controller custom hooks live under `pkg/controller/apigatewayv2/*/hooks.go` files. You can test them manually if you'd like to with YAMLs in `examples/apigatewayv2` folder.

Generation is based on current latest form of https://github.com/aws/aws-controllers-k8s/pull/449 which is this [commit](https://github.com/aws/aws-controllers-k8s/pull/449/commits/2efc52930e5690b52a67e289a72e7429139fb757)

apis/apigatewayv2/v1alpha1/zz_vpc_link.go
```diff
@@ -0,0 +1,102 @@
...
+	runtimev1alpha1 "github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// VPCLinkParameters defines the desired state of VPCLink
+type VPCLinkParameters struct {
+	// Region is which region the VPCLink will be created.
+	// +kubebuilder:validation:Required
+	Region string `json:"region,omitempty"`
```
hasheddan, line 31: Should this be `omitempty` if it is required?

pkg/controller/apigatewayv2/deployment/zz_controller.go
```diff
@@ -0,0 +1,161 @@
...
+func (c *connector) Connect(ctx context.Context, mg cpresource.Managed) (managed.ExternalClient, error) {
+	cr, ok := mg.(*svcapitypes.Deployment)
+	if !ok {
+		return nil, errors.New(errUnexpectedObject)
+	}
+	cfg, err := awsclient.GetConfigV1(ctx, c.kube, mg, cr.Spec.ForProvider.Region)
+	if err != nil {
+		return nil, err
+	}
+	sess, err := session.NewSession(cfg)
```
hasheddan, line 63: looks like this line is reused across `Connect` methods, does it make sense to move to `GetConfigV1`?

pkg/controller/apigatewayv2/domainname/hooks.go
```diff
@@ -0,0 +1,115 @@
...
+func SetupDomainName(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(svcapitypes.DomainNameGroupKind)
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&svcapitypes.DomainName{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(svcapitypes.DomainNameGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient()}),
+			managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),
+			managed.WithConnectionPublishers(),
```
hasheddan, line 46: Do we need to override this? I am fairly certain it is just a no-op if no secret is defined

hasheddan (commented):
This looks awesome @muvaf! I have left a few comments with the context that I have (which is somewhat minimal).

pkg/controller/apigatewayv2/domainname/hooks.go
muvaf: With most of the resources not publishing anything, I thought this would be a safe default. In fact, we may want to make this default in runtime, too.

hasheddan: The `APISecretPublisher` effectively accomplishes the same thing though right? So it seems like a more powerful default: "If you provide a secret we will publish it, if not we will do nothing and not error."

muvaf:
@hasheddan just pushed a commit addressing your concerns.

package/crds/apigatewayv2.aws.crossplane.io_deployments.yaml
```diff
@@ -0,0 +1,204 @@
...
+  creationTimestamp: null
+  name: deployments.apigatewayv2.aws.crossplane.io
+spec:
+  group: apigatewayv2.aws.crossplane.io
+  names:
+    categories:
+    - crossplane
+    - managed
+    - aws
+    kind: Deployment
```
rozcietrzewiacz, line 15: Could this create a problem when listing deployments (from k8s `v1` api)? I can definitely imagine scripts naively wrapping around `kubectl get deployments` without filtering by api version.

pkg/controller/apigatewayv2/integration/hooks.go
```diff
@@ -0,0 +1,124 @@
...
+// SetupIntegration adds a controller that reconciles Integration.
+func SetupIntegration(mgr ctrl.Manager, l logging.Logger) error {
+	name := managed.ControllerName(svcapitypes.IntegrationGroupKind)
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(name).
+		For(&svcapitypes.Integration{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(svcapitypes.IntegrationGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient()}),
+			managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),
```
hasheddan, line 46: nitpick: I believe this reference resolver will be used by default right?

package/crds/apigatewayv2.aws.crossplane.io_deployments.yaml
muvaf: Kubernetes kinds take precedence. So, users would have to call it like `kubectl get deployments.apigatewayv2` or like others `kubectl get managed` or `kubectl get aws` etc.

hasheddan (approved):
Thanks for this awesome work @muvaf!

PR #421, opened 2020-11-09
securitygroup: add reference and selector fields to UserIDGroupPair

muvaf:

I just realized that we have `VPCID` but no reference or selector fields for it.

Fixes #

I have:

Not tested manually yet.


muvaf:
@hasheddan I think we can merge this. I didn't specifically test this case because it's a hard combination to get right but the code compiles and it's fairly similar to existing implementations.

hasheddan (approved):
@muvaf SGTM

PR #426, opened 2020-11-13
Do not panic on missing date in bucket lifecycleConfig

hasheddan:

This updates the GenerateLifecycleConfig method to not assume that a
non-nil Date is provided for every transition. This was causing a panic
if Days and/or StorageClass was provided but no Date was provided.

I have:

Unit tests.


pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -137,10 +137,12 @@ func GenerateLifecycleRules(in []v1beta1.LifecycleRule) []awss3.LifecycleRule {
 			rule.Transitions = make([]awss3.Transition, len(local.Transitions))
 			for tIndex, transition := range local.Transitions {
 				rule.Transitions[tIndex] = awss3.Transition{
-					Date:         &transition.Date.Time,
```
hasheddan, line 140: You can test that this does indeed panic by removing the date from the input of the test for `GenerateLifecycleConfig()` and then test that it works correctly now by adding it back and seeing `Date` populated

muvaf (approved):

Issue #443, opened 2020-11-19
SQS: redrivePolicy.deadLetterQueue typo 


druyang:
### What happened?

When trying to create a new queue with Crossplane, the following error appears: 

    cannot create Queue: InvalidParameterValue: Value {"**deadLetterQueueARN**:" ... maxReceiveCount":5} for parameter RedrivePolicy is invalid. Reason: Redrive policy does not contain mandatory attribute: **deadLetterTargetArn**. status code: 400

[The JSON Crossplane is sending to AWS seems to have an invalid format. **deadLetterQueueARN** should be **deadLetterTargetArn**](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html)

### How can we reproduce it?
Create a deadletter queue and have another queue select it via `redrivePolicy.deadLetterQueueArnSelector.matchLabels` 

### What environment did it happen in?
Crossplane version: 

- provider-aws:master 

[Relevant PR](https://github.com/crossplane/provider-aws/pull/440/
)

PR #444, opened 2020-11-20
queue: fixed the format of spec.forProvider.redrivePolicy

muvaf:

@druyang tested `Queue` resource with redrivePolicy defined and got the following error:
```
cannot create Queue: InvalidParameterValue: Value {"deadLetterQueueARN":"arn:aws:sqs:us-west-2:618659985493:poc-ay-sheet-workflow-dead-letter","deadLetterQueueArnRef":{"name":"poc-ay-sheet-workflow-dead-letter"},"deadLetterQueueArnSelector":{"matchLabels":{"sqs-label":"poc-ay-sheet-workflow-dead-letter"},"matchControllerRef":true},"maxReceiveCount":5} for parameter RedrivePolicy is invalid. Reason: Redrive policy does not contain mandatory attribute: deadLetterTargetArn. status code: 400
```

`deadLetterQueueARN` should be `deadLetterTargetArn`. The error doesn't tell much about it but I believe ref and selectors fields should be removed as well. This PR does the necessary fixes.

Also, there were a few small fixes related to the fact that AWS literally expects a JSON blob.

Fixes https://github.com/crossplane/provider-aws/issues/443

I have:

Manually tested with one `Queue` referring to the other one as its dead-letter queue.


apis/sqs/v1beta1/queue_types.go
```diff
@@ -51,20 +49,19 @@ type RedrivePolicy struct {
...
+	// DeadLetterTargetARNSelector selects reference to a Queue to retrieve its ARN
 	// +optional
-	DeadLetterQueueARNSelector *runtimev1alpha1.Selector `json:"deadLetterQueueArnSelector,omitempty"`
+	DeadLetterTargetARNSelector *runtimev1alpha1.Selector `json:"deadLetterTargetArnSelector,omitempty"`
 
 	// The number of times a message is delivered to the source queue before
 	// being moved to the dead-letter queue.
-	// +optional
-	MaxReceiveCount *int64 `json:"maxReceiveCount,omitempty"`
+	MaxReceiveCount int64 `json:"maxReceiveCount"`
```
hasheddan, line 64: I am guessing that this is only required if the dead letter target is also?

apis/sqs/v1beta1/queue_types.go
```diff
@@ -51,20 +49,19 @@ type RedrivePolicy struct {
 	// The Amazon Resource Name (ARN) of the dead-letter queue to which Amazon
 	// SQS moves messages after the value of maxReceiveCount is exceeded.
 	// +optional
-	DeadLetterQueueARN *string `json:"deadLetterQueueARN,omitempty"`
+	DeadLetterTargetARN *string `json:"deadLetterTargetArn,omitempty"`
```
hasheddan, line 52: This would be a breaking change to a `v1beta1` resource, but given that this field does not work currently I don't feel as though bumping the version would be required here. Does that sound right to you?

apis/sqs/v1beta1/queue_types.go
```diff
@@ -38,8 +38,6 @@ const (
 	AttributeDelaySeconds                          string = "DelaySeconds"
 	AttributeReceiveMessageWaitTimeSeconds         string = "ReceiveMessageWaitTimeSeconds"
 	AttributeRedrivePolicy                         string = "RedrivePolicy"
-	AttributeDeadLetterQueueARN                    string = "DeadLetterQueueARN"
-	AttributeMaxReceiveCount                       string = "MaxReceiveCount"
```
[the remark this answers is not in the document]
muvaf: These were used as attribute names but they are actually not. The string we get from AWS is a JSON blob. So, I removed those from the attribute list. The actual strings used in that JSON blob are used only during unmarshal/marshal so I thought having them there would be OK but I can change them to `const`s.

apis/sqs/v1beta1/queue_types.go
muvaf: Yes, but `deadLetterTargetArn` is referenced so we can't mark it as required.

apis/sqs/v1beta1/queue_types.go
muvaf: Yes, that's why I didn't bump it.

apis/sqs/v1beta1/queue_types.go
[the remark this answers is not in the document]
hasheddan: Gotcha, I am fine either way 👍

apis/sqs/v1beta1/queue_types.go
hasheddan: Okay. I think we typically would remove the `// +optional` off of `deadLetterTargetArn` in this case then (i.e. when it is required but it is a reference). My concern here with `MaxReceiveCount` becoming required is that now this PR technically could break an existing implementation. It would be very unlikely, but if someone had tried to use `deadLetterQueueArn` and did not use `MaxReceiveCount` and it was ignored by AWS (which, judging by the error if the value was supplied and no `MaxReceiveCount` was supplied it looks like it _might_ ignore it, but maybe not) so they go a queue just with the `RedrivePolicy` missing, then their YAML that "worked" would now be in violation of the schema here.

It could be that AWS would reject this already anyway, in which case this is not an issue. Even if it doesn't the Queue would not be configured in the way the user expected. But I did want to surface the fact that a previously valid YAML could now be invalid, and we are not bumping versions.

muvaf: > if someone had tried to use deadLetterQueueArn and did not use MaxReceiveCount and it was ignored by AWS (which, judging by the error if the value was supplied and no MaxReceiveCount was supplied it looks like it might ignore it, but maybe not)

It's not ignored. I realized that during the manual test that if you send any kind of JSON blob for `redrivePolicy`, it has to include both `deadLetterTargetArn` and `maxReceiveCount`. Also keep in mind that there is no way YAML can work with `deadLetterQueueArn` field name. I don't know where that came from but AWS outright rejects it and expects `deadLetterTargetArn`.

hasheddan: Gotcha, so the only break here is that a YAML that was being rejected by AWS is now rejected by the API server. That seems okay to me without version bump 👍 Thanks!

hasheddan (approved):

apis/sqs/v1beta1/queue_types.go
muvaf: Yeah. I was kind of on the fence about whether we should just expect a JSON blob as string like AWS but didn't go for that.

PR #445, opened 2020-11-20
API Gateway v2 resources updated with ReadOne template

muvaf:

I am implementing `ReadOne` class of APIs in ACK and realized that API Gateway v2 has singular read calls and they are preferred by ACK, so I updated them with that which made them simpler. Also, there were some flaws regarding external-name where external-name was used as name but it wasn't unique identifier so filters weren't working as expected. This PR fixes those too by making all external-names `ID`s wherever possible.

The first commit is generation changes and the rest is my manual changes.

I have:

Manually with the examples.


hasheddan (approved):
Sorry for the delay here @muvaf! LGTM :)

PR #452, opened 2020-11-20
Update AUTHENTICATION.md to use ControllerConfig

### LINKED ISSUE #437: Update AUTHENTICATION.md to use ControllerConfig

Using EKS IRSA has become much easier with the introduction of `ControllerConfig`. However, [AUTHENTICATION.md](https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md) still uses manual patches on the Service Account.

Update `AUTHENTICATION.md` with workflow using `ControllerConfig`.

hasheddan:

Update `AUTHENTICATION.md` with workflow using `ControllerConfig`.

Fixes #437 

I have:

Doing full e2e test so keeping in draft for a moment.

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -28,7 +28,18 @@ large node pool.
 aws eks --region <region> update-kubeconfig --name <cluster-name>
 ```
 
-2. Install Crossplane and provider-aws
+2. Get AWS account information
+
+Get AWS account information and pick an IAM role name. These will be used to
+setup an OIDC provider and inject credentials into the provider-aws controller
+pod.
+
+```
+export AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query "Account" --output text)
+export IAM_ROLE_NAME=provider-aws # name for IAM role, can be anything you want
+```
+
+3. Install Crossplane
 
 Install Crossplane from `alpha` channel:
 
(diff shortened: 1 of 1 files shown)


hasheddan:
This guide has been verified by a community member :heavy_check_mark: 

@muvaf PTAL :)

AUTHENTICATION.md
```diff
@@ -39,14 +50,16 @@ helm repo add crossplane-alpha https://charts.crossplane.io/alpha
...
-Install `provider-aws` (assumes you have [installed the Crossplane
-CLI](https://crossplane.io/docs/v0.13/getting-started/install-configure.html#install-crossplane-cli)):
+`provider-aws` can be installed with the [Crossplane
+CLI](https://crossplane.io/docs/v0.13/getting-started/install-configure.html#install-crossplane-cli),
+but we will do so manually so that we can also create and reference a
+`ControllerConfig`:
 
 ```
-kubectl crossplane install provider crossplane/provider-aws:master
+envsubst < irsa-controller-config.yaml | kubectl apply -f -
```
muvaf, line 59: Can we have the YAML snippet (or as an addition) for doing this quickly? Otherwise, they would need to go out of this doc to find the file.

AUTHENTICATION.md
```diff
@@ -65,7 +78,7 @@ SERVICE_ACCOUNT_NAMESPACE=crossplane-system
 SERVICE_ACCOUNT_NAME=provider-aws-<YOUR-SERVICE-ACCOUNT-EXTENSION>
```
muvaf, line 78: Can we do everything AWS-related first and then Crossplane so that there wouldn't be much context switch?

hasheddan: Unfortunately no adding the annotation for the ServiceAccount requires some AWS information, but then we need to create permissions based on the name of the ServiceAccount that is created.

AUTHENTICATION.md
hasheddan: Yeah I could add it inline. Could also update this to point to raw github URL, which might be easiest, but some folks may be averse to

muvaf: Inline YAML sounds like a good way to also let them see what's happening right here. Raw github URL might require some branch/release name in the URL and that may not be worth the hassle

AUTHENTICATION.md
rozcietrzewiacz: I ran into the same issue trying to wrap crossplane setup in a helm chart. A simple way to do it would be to hardcode the serviceAccount name somewhere, for instance through a new `serviceAccountName` parameter next to the  `useServiceAccount` boolean: https://github.com/crossplane/provider-aws/blob/67203492ddd76bcd481683b51f6c22b92e65e182/package/crds/aws.crossplane.io_providers.yaml#L68

hasheddan: @rozcietrzewiacz thanks for your thoughts here :) I think it would be more likely that we solve this by setting the `ServiceAccount` in the `ControllerConfig`

AUTHENTICATION.md
```diff
@@ -74,14 +87,12 @@ Console](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-s
...
 ```
 
-5. Create IAM Role that provider-aws will use
+6. Create IAM Role that provider-aws will use
 
 Set environment variables that will be used in subsequent commands:
 
 ```
-AWS_ACCOUNT_ID=$(aws2 sts get-caller-identity --query "Account" --output text)
 OIDC_PROVIDER=$(aws2 eks describe-cluster --name <cluster-name> --region <region> --query "cluster.identity.oidc.issuer" --output text | sed -e "s/^https:\/\///")
```
rozcietrzewiacz, line 95: Why is `aws2` command used in some places (not all)?  I don't get the `aws2` binary (/link? alias?) installed with `awscli` and all of the commands work for me with `aws` in its place.

muvaf (approved):
Sorry, I meant to approve in my first review. Comments were just some nitpicks.

AUTHENTICATION.md
rozcietrzewiacz: @hasheddan Ah, if that can be done - perfect!

PR #460, opened 2020-12-02
bucketpolicy: multiple condition values

### LINKED ISSUE #455: S3 BucketPolicy not able to put multiple StringEquals conditions
### What happened?

While creating s3.aws.crossplane.io/v1alpha2 BucketPolicy conditions for StringEquals. Seem I can only put 1 key-value pair. As for certain scenario, multi key-value pairs are needed

### How can we reproduce it?

Sample manifest:

```
apiVersion: s3.aws.crossplane.io/v1alpha2
kind: BucketPolicy
metadata:
  name: mybucketname.bucketpolicy
spec:
  forProvider:
    version: "2012-10-17"
    bucketName: mybucketname
    region: us-west-1
    statements:
      - sid: DownloadandUpload
        action:
          - "s3:GetObject"
          - "s3:GetObjectAcl"
          - "s3:GetObjectVersion"
          - "s3:PutObject"
          - "s3:PutObjectAcl"
          - "s3:DeleteObject"
          - "s3:DeleteObjectVersion"
        effect: Allow
        principal:
          allowAnon: true
        resource: 
          - "mybucketname"
          - "mybucketname/*"  
        condition:
          StringEquals:
            - key: "aws:Key1"
              stringValue: "value1"
            - key: "aws:Key2"
              stringValue: "value2"
            - key: "aws:Key3"
              stringValue: "value3"
            - key: "aws:Key4"
              stringValue: "value4"
  providerConfigRef:
    name: aws-provider
  deletionPolicy: Delete
```

Crossplane version: v0.14.0
Provider-aws version: v0.15.0

krishchow:

Fixes #455 

I have:

This PR has been tested through updated unit tests and also manual testing with an S3 bucket (updated examples).


krishchow:
@muvaf This PR aims to fix the issue that @kelvinwijaya brought up specifically. We should be able to support multiple values in the condition now. I will open up a second PR/Issue to make the other modifications we discussed in #455.

kelvinwijaya:
Hi @krishchow @muvaf ,

Just to check if this PR will be merged soon?

Thanks,
Kelvin

kelvinwijaya:
Hi @krishchow @muvaf ,

Apologise to ask again on this PR, will this item be included anytime soon?

krishchow:
Hi @kelvinwijaya, apologies! I'll rebase this, review the PR today and try to get a review on it ASAP :)

kelvinwijaya:
> Hi @kelvinwijaya, apologies! I'll rebase this, review the PR today and try to get a review on it ASAP :)

No worries and thanks in advance :)

hasheddan (commented):
This looks pretty reasonable to me, though I wonder if we should promote to `v1alpha3` due to the fact that this breaks the schema for bucket policy. Would also like @muvaf to weigh in here before proceeding.

muvaf:
>This looks pretty reasonable to me, though I wonder if we should promote to v1alpha3 due to the fact that this breaks the schema for bucket policy.

It seems like a breaking change to me as well.

@krishchow I had thought we captured the full schema before. Does this PR make sure that the fields we have fully comply with the current policy schema? I would like to make sure we implement the schema completely before bumping to `v1alpha3`

krishchow:
This would be a breaking change, however, I am opening up another PR later today with support for stringified JSON input for the S3 bucket policy. Would you suggest updating to v1alpha3 in this PR, or in that one?

Also @muvaf, I think this PR should be the last of the changes to the actual shape of the bucket policy API, based on [this site](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html) I added support for list values as well.

examples/s3/bucketpolicy.yaml
```diff
@@ -21,12 +21,17 @@ spec:
         resource:
           - "arn:aws:s3:::test-bucket"
         condition:
+          StringEquals:
```
muvaf, line 24: When I look at the [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_multi-value-conditions.html) you referenced, I see the following example:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "dynamodb:GetItem",
            "Resource": "arn:aws:dynamodb:*:*:table/Thread",
            "Condition": {
                "ForAllValues:StringEquals": {
                    "dynamodb:Attributes": [
                        "ID",
                        "Message",
                        "Tags"
                    ]
                }
            }
        }
    ]
}
```

If I'm not mistaken, `Condition` here is a map of maps, i.e. the value of `ForAllValues:StringEquals` also seems to be a map object with single key `dynamodb:Attributes`. But in this PR, we have only one level map and then we use an array with `key` field. Was this intentional?

Kubernetes API patterns actually [discourage](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps) the use of maps whose value type is `struct`. On the other hand, we're trying to stick to how provider wants to represent their data structures. I feel closer to avoiding the maps with struct value and go with arrays but I can be convinced either way. Though I believe both levels should be either map or array instead of array inside of a map.

muvaf:
> Would you suggest updating to v1alpha3 in this PR, or in that one?

I'm OK either way with a slight preference to do it here since this is the PR that's breaking.

examples/s3/bucketpolicy.yaml
krishchow: Each object in the array represents a key/value pair under the "condition operator" (the map string key). I wasn't sure if there was a better way to represent this design. I will look into switching the design to use arrays on both sides.

krishchow:
> I'm OK either way with a slight preference to do it here since this is the PR that's breaking.

That makes sense, I'll move this to v1alpha3, and then if we can get the other PR through before the next release they can both be v1alpha3

muvaf (approved):
LGTM! Thanks @krishchow !

PR #465, opened 2020-12-07
Optional ACL, Endpoint secret

krishchow:

This PR fixes an issue outlined on the Crossplane slack channel regarding the ACL field in the bucket resource. Additionally, one piece of functionality that we lost when bringing the s3 bucket to v1beta1 was the connection secret containing the endpoint of the bucket. This is something that I am adding back in this PR.

Slack Conversation: https://crossplane.slack.com/archives/CEG3T90A1/p1607034420017100?thread_ts=1607033622.011100&cid=CEG3T90A1

I have:

Updated unit tests, manual testing via the examples manifest.

### DIFF
--- a/pkg/controller/s3/bucket.go
+++ b/pkg/controller/s3/bucket.go
@@ -49,6 +49,8 @@ const (
 	errCreateOrUpdate   = "cannot create or update"
 	errDelete           = "cannot delete"
 	errKubeUpdateFailed = "cannot update S3 custom resource"
+	// ResourceCredentialsSecretRegionKey is the key for region that the S3 bucket is located
+	ResourceCredentialsSecretRegionKey = "region"
 )
 
 // SetupBucket adds a controller that reconciles Buckets.
@@ -94,6 +96,7 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
+
 	if _, err := e.s3client.HeadBucketRequest(&awss3.HeadBucketInput{Bucket: aws.String(meta.GetExternalName(cr))}).Send(ctx); err != nil {
 		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(s3.IsNotFound, err), errHead)
 	}
@@ -132,6 +135,10 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	return managed.ExternalObservation{
 		ResourceExists:   true,
 		ResourceUpToDate: true,
+		ConnectionDetails: map[string][]byte{
+			runtimev1alpha1.ResourceCredentialsSecretEndpointKey: []byte(meta.GetExternalName(cr)),
+			ResourceCredentialsSecretRegionKey:                   []byte(cr.Spec.ForProvider.LocationConstraint),
+		},
 	}, nil
 }
 
--- a/pkg/controller/s3/bucket_test.go
+++ b/pkg/controller/s3/bucket_test.go
@@ -117,6 +117,10 @@ func TestObserve(t *testing.T) {
 				result: managed.ExternalObservation{
 					ResourceExists:   true,
 					ResourceUpToDate: true,
+					ConnectionDetails: map[string][]byte{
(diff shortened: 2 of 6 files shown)


pkg/controller/s3/bucket.go
```diff
@@ -132,6 +135,9 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	return managed.ExternalObservation{
 		ResourceExists:   true,
 		ResourceUpToDate: true,
+		ConnectionDetails: map[string][]byte{
+			runtimev1alpha1.ResourceCredentialsSecretEndpointKey: []byte(meta.GetExternalName(cr)),
```
muvaf, line 139: Is bucket name enough here for the apps to connect?

pkg/controller/s3/bucket.go
```diff
@@ -183,3 +189,9 @@ func (e *external) Delete(ctx context.Context, mg resource.Managed) error {
 	_, err := e.s3client.DeleteBucketRequest(&awss3.DeleteBucketInput{Bucket: aws.String(meta.GetExternalName(cr))}).Send(ctx)
 	return resource.Ignore(s3.IsNotFound, err)
 }
+
+func lateInit(mg *v1beta1.Bucket) {
+	if mg.Spec.ForProvider.ACL == nil {
+		mg.Spec.ForProvider.ACL = aws.String("private")
```
muvaf, line 195: Can we use `GetBucketAcl` request after `HeadBucketRequest` to get the current state from AWS and do late-init against that info?

hasheddan: We can also default at the CRD level, which I think may be preferable.

pkg/controller/s3/bucket.go
krishchow: I can add the region as well

pkg/controller/s3/bucket.go
krishchow: The reason I did it this way, as opposed to changing the CRD was just prevented any breaking changes to the spec. But I guess if someone had the ALC missing previously, it wouldn't have worked regardless.

krishchow: Hm, actually, upon further inspection of the docs, this seems like the intended behavior.

> You can use either a canned ACL or specify access permissions explicitly. You cannot do both.

So maybe we don't need to change anything for this

hasheddan: Just to make sure I understand correctly, if we submit a bucket with no ACL then it is going to fail right?

pkg/controller/s3/bucket.go
hasheddan: Maybe the ARN would make sense?

pkg/controller/s3/bucket.go
krishchow: You can either use a canned ACL or individual permissions, in the conversation linked in this PR, the manifest didn't use either, which is why it failed

hasheddan: @krishchow gotcha, sounds like we should just add a note in the field descriptions then

pkg/controller/s3/bucket.go
krishchow: For connecting to an s3 bucket, only the region and bucket name is required. Ex., https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro

pkg/controller/s3/bucket.go
krishchow: All done now :) @hasheddan

muvaf (approved):
Thanks @krishchow !

pkg/controller/s3/bucket_test.go
```diff
@@ -117,6 +117,9 @@ func TestObserve(t *testing.T) {
 				result: managed.ExternalObservation{
 					ResourceExists:   true,
 					ResourceUpToDate: true,
+					ConnectionDetails: map[string][]byte{
+						corev1alpha1.ResourceCredentialsSecretEndpointKey: []byte(s3Testing.BucketName),
+					},
```
hasheddan, line 122: Looks like you need to add the region key here also :)

krishchow: Thanks, my bad! :smile:

PR #469, opened 2020-12-11
Implement aws secrets manager

### LINKED ISSUE #405: Support for AWS Secrets Manager

I want to manage AWS Secrets Manager via crossplane provider-aws, e.g. create, update and delete secrets

Implement a Secrets Manager resource: https://docs.aws.amazon.com/sdk-for-go/v2/api/service/secretsmanager

bjpw:

Fixes #405 

I have:

- [x] Implement review TODO comments

Unit tests and manual testing


muvaf:
@bjpw would you be interested in updating this with [ACK code generation](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md)? I just tested and it generates the `Secret` controller when `secretsmanager` is given as argument.

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+
+// LateInitialize fills the empty fields in *v1alpha1.SecretParameters with the
+// values seen in secretsmanager.DescribeSecretOutput.
+func LateInitialize(in *v1alpha1.SecretParameters, so *secretsmanager.DescribeSecretOutput) { // nolint:gocyclo
+	if so == nil {
+		return
+	}
+	if in.Description == nil && so.Description != nil {
+		in.Description = so.Description
+	}
```
muvaf, line 175:
```suggestion
	in.Description = awsclients.LateInitializeStringPtr(in.Description, so.Description)
```

You can use late-init helper functions if needed.

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+	if in.KmsKeyID == nil && so.KmsKeyId != nil {
+		in.KmsKeyID = so.KmsKeyId
+	}
+
+	r := int64(GetRecoveryWindowInDays(in.RecoveryWindowInDays))
+	in.RecoveryWindowInDays = &r
```
muvaf, line 181: It seems we're doing a defaulting here but I'd advise against it since they can change the default without us knowing. So, if the describe result doesn't contain this information, we can leave it empty without trying to late-init with a known value.

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+
+	if len(in.Tags) == 0 && len(so.Tags) > 0 {
+		in.Tags = make([]v1alpha1.Tag, len(so.Tags))
+		for i, t := range so.Tags {
+			in.Tags[i] = v1alpha1.Tag{
+				Key:   *t.Key,
+				Value: *t.Value,
```
muvaf, line 188:
```suggestion
				Key:   awsclients.StringValue(t.Key),
				Value: awsclients.StringValue(t.Value),
```

nitpick: we have these small helper functions for these cases. It's not really a big deal here but for example if you were to do operations on pointers in the for loop, then it could add wrong element to the array repeatedly since these `i` and `t` pointers are static and only their values are changing. But helper functions operate in separate stack, they don't have this problem.

apis/secretsmanager/v1alpha1/common.go
```diff
@@ -0,0 +1,78 @@
...
+			delete(add, k)
+		default:
+			delete(add, k)
+		}
+	}
+	addTags = []secretsmanager.Tag{}
+	for key, value := range add {
+		value := value
+		key := key
+		addTags = append(addTags, secretsmanager.Tag{Key: &key, Value: &value})
```
muvaf, line 75:
```suggestion
		addTags = append(addTags, secretsmanager.Tag{Key: awsclients.String(key), Value: awsclients.String(value)})
```

You can to use this helper function here.

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	secret, err := secretsmanager.GetSecret(ctx, e.kube, cr)
+	if err != nil {
+		return managed.ExternalUpdate{}, err
+	}
+
+	_, err = e.client.UpdateSecretRequest(secretsmanager.GenerateUpdateSecretInput(meta.GetExternalName(cr), cr.Spec.ForProvider, secret)).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(secretsmanager.IsErrorNotFound, err), errUpdateFailed)
```
muvaf, line 195: Why ignore this error?

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	if cmp.Equal(v1alpha1.GenerateSecretTags(cr.Spec.ForProvider.Tags), req.Tags) {
```
muvaf, line 198: Aren't tags handled by `e.updateTags` in the beginning of `Update`?

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+
+	cr, ok := mg.(*v1alpha1.Secret)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errNotSecret)
+	}
+
+	req, err := e.client.DescribeSecretRequest(&awssecretsmanager.DescribeSecretInput{SecretId: awscommon.String(meta.GetExternalName(cr))}).Send(ctx)
+	if err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(resource.Ignore(secretsmanager.IsErrorNotFound, err), errDescribeSecretFailed)
```
muvaf, line 180: Why ignore this error?

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+
+	cr, ok := mg.(*v1alpha1.Secret)
+	if !ok {
+		return errors.New(errNotSecret)
+	}
+
+	cr.SetConditions(runtimev1.Deleting())
+
+	if cr.Status.AtProvider.DeletionDate != nil {
+		// only request a new deletion if the user has changed the recovery window
```
muvaf, line 221: I couldn't get this part. Why do we need to do this calculation?

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+		if oldDeletionDate.Equal(newDeletionDate) {
+			return nil
+		}
+	}
+
+	rsp, err := e.client.DeleteSecretRequest(secretsmanager.GenerateDeleteSecretInput(meta.GetExternalName(cr), cr.Spec.ForProvider)).Send(ctx)
+	if err != nil {
+		return errors.Wrap(resource.Ignore(secretsmanager.IsErrorNotFound, err), errDeleteFailed)
+	}
+	cr.Status.AtProvider.DeletionDate = &metav1.Time{Time: *rsp.DeletionDate}
```
muvaf, line 234: nitpick: this will be assigned in `Observe` anyway.

pkg/controller/secretsmanager/controller.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// TODO(knappek): split this out as it is used in several controllers
+func (t *tagger) Initialize(ctx context.Context, mg resource.Managed) error {
+	cr, ok := mg.(*v1alpha1.Secret)
+	if !ok {
+		return errors.New(errNotSecret)
+	}
+	tagMap := map[string]string{}
+	for _, t := range cr.Spec.ForProvider.Tags {
```
muvaf, line 249: `t` is already used in line 243 `(t *tagger)`

muvaf (commented):
Looking pretty good. I've got a few comments but it's close to merge. Thanks @bjpw !

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+	errGetSecretFailed     = "failed to get Kubernetes secret"
+	errSecretMarshalFailed = "cannot marshal secret"
+)
+
+// Client defines Secretsmanager Client operations
+type Client secretsmanageriface.ClientAPI
+
+// NewSecretsmanagerClient creates new Secretsmanager Client with provided AWS Configurations/Credentials.
+func NewSecretsmanagerClient(cfg aws.Config) Client {
+	return secretsmanager.New(cfg)
```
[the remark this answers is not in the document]
bjpw: We reused the structure of existing client code for other controllers. Plus, using the New() function directly gave me the following error:

```
cannot use awssecretsmanager.New (value of type func(config "github.com/aws/aws-sdk-go-v2/aws".Config) *"github.com/aws/aws-sdk-go-v2/service/secretsmanager".Client) as func("github.com/aws/aws-sdk-go-v2/aws".Config) "github.com/crossplane/provider-aws/pkg/clients/secretsmanager".Client value in struct literal
```

apis/secretsmanager/v1alpha1/referencers.go
```diff
@@ -0,0 +1,49 @@
...
+func (mg *Secret) ResolveReferences(ctx context.Context, c client.Reader) error {
+	r := reference.NewAPIResolver(c, mg)
+
+	// Resolve spec.forProvider.kmsKeyId
+	if mg.Spec.ForProvider.KmsKeyIDRef != nil || mg.Spec.ForProvider.KmsKeyIDSelector != nil {
+		rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+			CurrentValue: *mg.Spec.ForProvider.KmsKeyID,
+			Reference:    mg.Spec.ForProvider.KmsKeyIDRef,
+			Selector:     mg.Spec.ForProvider.KmsKeyIDSelector,
+			To:           reference.To{Managed: &Secret{}, List: &SecretList{}},
```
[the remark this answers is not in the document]
bjpw: I'm going to wait with this until #505 is merged.

pkg/controller/secretsmanager/controller.go
bjpw: Indeed, however, only after trying to retrieve the secret value from AWS. We experienced an error when doing so with a deleted secret. For that reason, we make sure that `DeletionDate`is added here, so during the next observation, the secret is skipped and the error does not occur.

pkg/controller/secretsmanager/controller.go
bjpw: This calculation was necessary, because we combined `ForceDeleteWithoutRecovery` and `RecoveryWindowInDays`. I split that up again, so this code isn't needed anymore.

pkg/controller/secretsmanager/controller.go
bjpw: `DeletedDate` is set in `Observe`. `DeletionDate` is only returned by the `DeleteSecretRequest` call.

pkg/controller/secretsmanager/controller.go
bjpw: As mentioned below `DeletionDate` is set based on the response of the `DeleteSecretRequest`. After it is set once, we check whether the `RecoveryWindowInDays` has changed as the resource in Kubernetes is only deleted after the grace period is over. If it has not changed, there is no need to call `DeleteSecretRequest` again and update the `DeletionDate`.

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+		Namespace: s.Spec.ForProvider.SecretBinarySecretRef.Namespace,
+	}
+	sc := &corev1.Secret{}
+	if err := kube.Get(ctx, nn, sc); err != nil {
+		return nil, errors.Wrap(err, errGetSecretFailed)
+	}
+	if len(sc.Data) == 0 {
+		return nil, nil
+	}
+	scb, err := json.Marshal(sc.Data)
```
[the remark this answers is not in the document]
bjpw: The Secrets Manager itself does not do any distinction. On the contrary, you are not even able to retrieve your secret value in the AWS Secrets Manager via the console, but only using the SDK or CLI.

pkg/controller/secretsmanager/controller.go
bjpw: Another comment from my end: We had a look what would happen if you don't add a `RecoveryWindowInDays` and do not add a `ForceDeleteWithoutRecovery`. In that case, the secret is still deleted and the `DeletionDate` is set to 30 days. `RecoveryWindowInDays` is therefore implicit.

pkg/clients/secretsmanager/secret.go
```diff
@@ -0,0 +1,267 @@
...
+func UpdateObservation(o *v1alpha1.SecretObservation, svr *secretsmanager.GetSecretValueResponse, do *secretsmanager.DescribeSecretOutput) {
+	if svr != nil {
+		svo := svr.GetSecretValueOutput
+		if svo != nil {
+			arn := svo.ARN
+			if svo.ARN == nil {
+				arn = do.ARN
+			}
+			o.ARN = awsclients.StringValue(arn)
+			o.VersionID = awsclients.StringValue(svo.VersionId)
```
muvaf, line 153: This seems to be set at every reconcile, hence triggers a new reconcile, so it reconciles repeatedly. Can we remove this field if not used in a functional logic in the code?

pkg/clients/secretsmanager/secret.go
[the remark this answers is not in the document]
bjpw: We've added `*runtimev1.SecretKeySelector`, so you have the option to select a specific value in your secret data that is uploaded "as it is" to the Secrets Manager. If you don't add a key to your YAML, the secret data is uploaded as a map in its entirety.

apis/secretsmanager/v1alpha1/types.go
```diff
@@ -0,0 +1,224 @@
...
+	//
+	// We do not use SecretString from the AWS API (see https://docs.aws.amazon.com/sdk-for-go/api/service/secretsmanager/#CreateSecretInput)
+	// at all as we always reference a Kubernetes secret.
+	//
+	// SecretBinary is automatically base64 encoded/decoded by the SDK.
+	// +optional
+	SecretBinarySecretRef *runtimev1.SecretReference `json:"secretBinarySecretRef,omitempty"`
+
+	// (Optional) Specifies the number of days that Secrets Manager waits before
+	// it can delete the secret. This value can be 0 to force deletion without recovery or
```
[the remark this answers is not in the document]
muvaf: I had another look and I think what we need to do here is to include the field that deletion call has for this operation in the spec, i.e. `ForceDeleteWithoutRecovery`, and keep the functionality of `RecoveryWindowInDays` same as SDK.

muvaf (approved):
Thank you @bjpw ! I'll work on the missing parts when I migrate the boilerplate parts to code generated output.

PR #471, opened 2020-12-12
Bump crossplane-runtime to v0.12.0

negz:

I have:

It has not - it's a very mechanical change.

### DIFF
--- a/pkg/clients/ec2/routetable.go
+++ b/pkg/clients/ec2/routetable.go
@@ -6,9 +6,10 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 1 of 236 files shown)


negz:
@muvaf Note that I may have changed the import style of a bunch of the generated packages - we should probably update the AWS codegen to use the new `xpv1` import convention.

pkg/clients/ec2/routetable.go
```diff
@@ -6,10 +6,11 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
-	"github.com/crossplane/crossplane-runtime/apis/core/v1alpha1"
```
hasheddan, line 9: Note to self: looks like this package is not being linted correctly

hasheddan (approved):

PR #475, opened 2020-12-17
Attempt to remove requestId from aws errs, align on awsclient name

### LINKED ISSUE #461: AWS Request ID seems to cause fast reconciliation
### What happened?

I'ved noticed on Crossplane resources which fail to create/delete via AWS API end up causing excess API calls. Rather than the  short wait reconciliation of 30 seconds, they can reconcile multiple times per second due to each request getting a unique request ID back from AWS. This might be related to https://github.com/crossplane/provider-aws/issues/412

```
  - lastTransitionTime: "2020-12-03T20:45:58Z"
    message: "create failed: failed to create the VPC resource: InvalidParameterValue:
      Value (172.16.0.0/34) for parameter cidrBlock is invalid. This is not a valid
      CIDR block.\n\tstatus code: 400, request id: 7273cf82-7bd3-4846-9bc0-6d55898786a1"
---
  - lastTransitionTime: "2020-12-03T20:45:59Z"
    message: "create failed: failed to create the VPC resource: InvalidParameterValue:
      Value (172.16.0.0/34) for parameter cidrBlock is invalid. This is not a valid
      CIDR block.\n\tstatus code: 400, request id: cd6de9ef-969d-4a68-84d8-b6d0e49c6136"

```

### How can we reproduce it?

Simplest I've found is an invalid VPC, although any accepted Kubernetes resource which is invalid in AWS should repro.

```yaml
kind: VPC
metadata:
  name: test-vpc
spec:
  forProvider:
    cidrBlock: 172.16.0.0/34
    enableDnsHostNames: true
    enableDnsSupport: true
    instanceTenancy: default
    region: us-east-2
    tags:
      - key: Name
        value: test-vpc
  providerConfigRef:
    name: <provider>
```

Crossplane version: 0.14.0
Kubernetes version: 1.17.12
Kubernetes distribution: EKS

enderv:

Attempt to strip AWS request ID's from errors to prevent fast reconcilation. While going through aligned all the various controllers to awsclient naming (since looks like that is in the ack template going forward)

**DID NOT TOUCH ACK GENERATED CONTROLLERS** - will need another MR for this did not know if should touch or update the ack template

Fixes #461 

I have:

- [x ] Read and followed Crossplane's [contribution process].
- [x ] Run `make reviewable test` to ensure this PR is ready for review.

Tested using the VPC example in #461. Ran unit tests. 

### DIFF
--- a/cluster/local/integration_tests.sh
+++ b/cluster/local/integration_tests.sh
@@ -302,7 +302,7 @@ echo_step "uninstalling ${PROJECT_NAME}"
 echo "${INSTALL_YAML}" | "${KUBECTL}" delete -f -
 
 # check pods deleted
-sleep 30
+sleep 60
 echo_step "check only crossplane pods remain"
 echo "--- pods ---"
 "${KUBECTL}" -n "${CROSSPLANE_NAMESPACE}" get pods
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -28,9 +28,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws/session"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 	"github.com/aws/aws-sdk-go-v2/aws/endpoints"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
@@ -38,15 +37,15 @@ import (
 	awsv1 "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	endpointsv1 "github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/go-ini/ini"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
-	"github.com/crossplane/crossplane-runtime/pkg/resource"
 
 	"github.com/crossplane/provider-aws/apis/v1alpha3"
 	"github.com/crossplane/provider-aws/apis/v1beta1"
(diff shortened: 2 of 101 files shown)


pkg/clients/aws.go
```diff
@@ -632,3 +631,27 @@ func DiffLabels(local, remote map[string]string) (addOrModify map[string]string,
...
+	message string
+}
+
+func (e cleanedError) Error() string {
+	return fmt.Sprintf("%s - %s", e.code, e.message)
+}
+
+func cleanError(err error) error {
+	if awsErr, ok := err.(awserr.Error); ok {
+		return cleanedError{awsErr.Code(), awsErr.Message()}
```
muvaf, line 646: What I feel kind of uneasy about this solution is that we can't use any `Is*Error` check functions we have for each resource since it doesn't satisfy `awserr.Error` interface anymore. The usage of this `awsclient.Wrap` seems to be mostly when returned from the controller anyway, but I feel that we'll see more use of it across the codebase.

Is there a way to keep the returned object from `Wrap` satisfy `awserr.Error` and close to the original error so that we don't need to think about this additional layer of abstraction? Maybe `awserr.New`?

enderv: Did some testing and was able to get `cleanError` to return a `awserr.Error` but the underlying `errors.Wrap` makes it so it won't satisfy `awserr.Error` anymore.

Thoughts on if I just allow cleanError to be exported and it will return a `awserr.Error` so that is available if you just want to strip the requestId? But still keep the wrap function for ease of use?

muvaf: Yeah I think that'd be nice.
>But still keep the wrap function for ease of use?

I'm not sure. Assuming I got what you're saying right, the usage would be something like this `errors.Wrap(awsclient.CleanError(err), "some error")`, right?

Could you share a code snippet of how it'd look like just to be sure?

enderv: Oh sorry, was just trying to say wasn't sure If I should keep the `awsclient.Wrap` I had added or remove it and go through and remove  them all and just change to the now exported `CleanError`? 


So itd still be
```
	if err != nil {
		return managed.ExternalObservation{}, awsclient.Wrap(resource.Ignore(iam.IsErrorNotFound, err), errGet)
	}
```
vs
```
if err != nil {
		return managed.ExternalObservation{}, errrors.Wrap(awsclient.CleanError(resource.Ignore(iam.IsErrorNotFound, err)), errGet)
	}
```

Just wasn't sure which way to go and wanted to double check before I go through all of them again

muvaf: No worries. I feel like the second one is easier to grasp for a newcomer to see what we're doing but I don't feel strongly so I'm fine with what you choose here.

muvaf:
@enderv is this ready to go? I don't have further comments.

muvaf (approved):

enderv:
@muvaf Good from my side. Just fixed conflicts and the e2e is failing though, Currently trying to run e2e locally to see if I can repo it. Any chance that is a flaky test though?

cluster/local/integration_tests.sh
```diff
@@ -85,7 +85,7 @@ check_pods(){
         sleep 10
         "${KUBECTL}" get events -A
         sleep 20
-        echo_error "unexpected number of pods ${count} ${1}"
```
muvaf: Why do we need this change?

muvaf (commented):
Could you squash the commits?

pkg/clients/aws.go
```diff
@@ -632,3 +631,20 @@ func DiffLabels(local, remote map[string]string) (addOrModify map[string]string,
...
+	}
+	return err
+}
+
+// Wrap Attempts to remove requestID from awserr before calling Wrap
+func Wrap(err error, msg string) error {
+	if err != nil {
+		return errors.Wrap(CleanError(err), msg)
+	}
+	return errors.Wrap(err, msg)
```
muvaf, line 649:
```suggestion
	return errors.Wrap(CleanError(err), msg)
```

`errors.Wrap` is no-op if error is `nil`

cluster/local/integration_tests.sh
enderv: Yeah wasn't needed, removed already was just trying to figure out the e2e tests and accidentally added

pkg/clients/aws.go
enderv: went ahead and added

muvaf:
@enderv thank you!

PR #478, opened 2020-12-24
Adding destinations to routetable, fixing update issues

### LINKED ISSUE #308: Support for NAT Gateways (and others) and Reference them with RouteTable
I am unable to create a NAT Gateway and add it as a route to my RouteTables. The RouteTable Should be able to reference a NAT Gateway as a route just like how a RouteTable can route to an InternetGateway. AWS provides quite a bit of targets to route to, Crossplane only provides one, Internet Gateway.

The route tables routes have these properties: `gatewayId`, `gatewayIdRef`, and a `gatewayIdSelector`. I believe the naming should change from `gateway` to `target`. 
```
gatewayId = targetId
gatewayIdRef = targetIdRef
gatewayIdSelector = targetIdSelector
```

There should be one more field added to define what kind of Target you are selecting. 

```yaml
kind: RouteTable
metadata:
  name: private-us-west-2a
spec:
  forProvider:
    routes:
    - destinationCidrBlock: 10.101.0.0/16
      targetKind: NATGateway # or InternetGateway
      targetIdRef:
         name: my-cool-nat
```

Ultimately the following targetKinds are missing in Crossplane for a full featured experience:
 - Local
 - Egress Only Gateway
 - Instance
 - Network Interface
 - Outpost Local Gateway
 - Peering Connection
 - Transit Gateway
 - Virtual Private Gateway

Even if Crossplane were to omit one of the above targets as a Crossplane resource kind, it should still be available as a target option so one could at least manually set the `targetId`. 

Lastly, if the `targetKind` is not set then the selectors should not do anything and the value for the target should simply default to `local`. 
e.g. If I only set the `destinationCidrBlock` then the target in the AWS console should say `local`.

enderv:

Adding more destinations to routetable. Fixing some update issues with tags/routes and increasing test coverage. Bumping routetable to v1alpha5 as other components will need to go to beta1 at the same time to prevent circular dependencies.

Fixes #308 

I have:

Updated the unit tests. Ran locally creating/updating/deleting new routetable with natgateway selector and internet gateway selector.


pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,114 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
...
+
+	if err != nil {
+		return errors.Wrap(err, errDeleteRoute)
+	}
+	err = e.createRoutes(ctx, tableID, desired, observed)
+
+	if err != nil {
+		return errors.Wrap(err, errCreateRoute)
+	}
+	return nil
```
negz, line 334:
```suggestion
	if err := e.deleteRoutes(ctx, tableID, desired, observed); err != nil {
		return errors.Wrap(err, errDeleteRoute)
	}
	return errors.Wrap(e.createRoutes(ctx, tableID, desired, observed), errCreateRoute)
```

Nit: This and the equivalent associations method below could be simplified to the above. `errors.Wrap` returns `nil` if the supplied error is `nil`.

pkg/controller/ec2/routetable/controller.go
```diff
@@ -192,24 +194,35 @@ func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.Ex
...
+				return managed.ExternalUpdate{}, errors.Wrap(err, errCreateTags)
+			}
+		}
+		if len(removeTags) > 0 {
+			if _, err := e.client.DeleteTagsRequest(&awsec2.DeleteTagsInput{
+				Resources: []string{meta.GetExternalName(cr)},
+				Tags:      removeTags,
+			}).Send(ctx); err != nil {
+				return managed.ExternalUpdate{}, errors.Wrap(err, errDeleteTags)
+			}
```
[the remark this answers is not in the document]
enderv: Yeah discovered issues with tag updates as I was testing. Went ahead and updated the description

pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,114 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
 	return errors.Wrap(resource.Ignore(ec2.IsRouteTableNotFoundErr, err), errDelete)
 }
 
-func (e *external) createRoutes(ctx context.Context, tableID string, desired []v1alpha4.Route, observed []v1alpha4.RouteState) error {
+func (e *external) deleteRoutes(ctx context.Context, tableID string, desired []v1alpha5.Route, observed []v1alpha5.RouteState) error { // nolint:gocyclo
+	for _, rt := range observed {
+		found := false
+		for _, ds := range desired {
```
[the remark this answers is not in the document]
enderv: There's a cap of 50 routes per table.

pkg/clients/ec2/routetable.go
```diff
@@ -77,29 +79,36 @@ func IsAssociationIDNotFoundErr(err error) bool {
...
+				LocalGatewayID:           aws.StringValue(rt.LocalGatewayId),
+				NatGatewayID:             aws.StringValue(rt.NatGatewayId),
+				NetworkInterfaceID:       aws.StringValue(rt.NetworkInterfaceId),
+				TransitGatewayID:         aws.StringValue(rt.TransitGatewayId),
+				VpcPeeringConnectionID:   aws.StringValue(rt.VpcPeeringConnectionId),
 			}
 		}
 	}
 
 	if len(rt.Routes) > 0 {
```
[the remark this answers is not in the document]
enderv: Good catch yeah think its supposed to be associations, I'll try it out

pkg/clients/ec2/routetable.go
```diff
@@ -143,46 +158,67 @@ func LateInitializeRT(in *v1alpha4.RouteTableParameters, rt *ec2.RouteTable) { /
...
+	SortRoutes(target.Routes, in.Routes)
 
 	LateInitializeRT(currentParams, &in)
 
+	for i := 0; i < len(target.Routes); i++ {
+		target.Routes[i].ClearRefSelectors()
+	}
+	for i := 0; i < len(target.Associations); i++ {
+		target.Associations[i].ClearRefSelectors()
+	}
```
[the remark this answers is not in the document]
enderv: Yeah intention is to clear out the refs/selectors which is also why I don't use the range loops above as want to mutate in place for the patch compare. I can go ahead and swap in a copy just in case

pkg/clients/ec2/routetable.go
```diff
@@ -143,46 +158,67 @@ func LateInitializeRT(in *v1alpha4.RouteTableParameters, rt *ec2.RouteTable) { /
...
-	return cmp.Equal(&v1alpha4.RouteTableParameters{}, patch,
+	return cmp.Equal(&v1alpha5.RouteTableParameters{}, patch,
 		cmpopts.EquateEmpty(),
 		cmpopts.IgnoreTypes(&xpv1.Reference{}, &xpv1.Selector{}),
-		cmpopts.IgnoreFields(v1alpha4.RouteTableParameters{}, "Region"),
+		cmpopts.IgnoreFields(v1alpha5.RouteTableParameters{}, "Region"),
 	), nil
 }
+
+// SortRoutes sorts array of Routes on DestinationCIDR
```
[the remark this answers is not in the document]
enderv: oops it sure can will add check

pkg/controller/ec2/routetable/controller.go
enderv: Think I got em all

enderv:
@negz kinda forgot about this for a minute but addressed all the comments I believe.

muvaf:
It seems that all fields are additional so it doesn't seem like we need to bump the `apiVersion`. @enderv is there something I'm missing?

pkg/clients/ec2/routetable.go
```diff
@@ -143,46 +158,68 @@ func LateInitializeRT(in *v1alpha4.RouteTableParameters, rt *ec2.RouteTable) { /
...
-		if *val.GatewayId == LocalGatewayID {
-			target.Routes = append([]v1alpha4.Route{{
+		if val.GatewayId != nil && *val.GatewayId == LocalGatewayID {
+			target.Routes = append([]v1alpha5.Route{{
 				GatewayID:            val.GatewayId,
 				DestinationCIDRBlock: val.DestinationCidrBlock,
 			}}, target.Routes...)
 		}
 	}
+	SortRoutes(targetCopy.Routes, in.Routes)
```
muvaf, line 179: nitpick: you can sort the routes in `cmp.Equal` by using `cmpopts.SortSlices()` option instead of sorting them separately.

pkg/controller/ec2/routetable/controller.go
```diff
@@ -94,7 +96,7 @@ type external struct {
 }
 
 func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.ExternalObservation, error) { // nolint:gocyclo
-	cr, ok := mgd.(*v1alpha4.RouteTable)
+	cr, ok := mgd.(*v1alpha5.RouteTable)
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
```
muvaf, line 102: Unrelated to this PR but the err string used in the result of `ec2.IsRtUpToDate(cr.Spec.ForProvider, observed)` is `errDescribe`, which is misleading.

pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,110 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
 	return awsclient.Wrap(resource.Ignore(ec2.IsRouteTableNotFoundErr, err), errDelete)
 }
 
-func (e *external) createRoutes(ctx context.Context, tableID string, desired []v1alpha4.Route, observed []v1alpha4.RouteState) error {
+func (e *external) deleteRoutes(ctx context.Context, tableID string, desired []v1alpha5.Route, observed []v1alpha5.RouteState) error { // nolint:gocyclo
```
muvaf, line 253:
```suggestion
func (e *external) deleteRoutes(ctx context.Context, cr *v1alpha5.Route) error { // nolint:gocyclo
```

Seems like this would be enough for `deleteRoutes` to work.

pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,110 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
...
+	}
+	return nil
+}
+
+func (e *external) createRoutes(ctx context.Context, tableID string, desired []v1alpha5.Route, observed []v1alpha5.RouteState) error { // nolint:gocyclo
 	for _, rt := range desired {
 		isObserved := false
 		for _, ob := range observed {
-			if ob.GatewayID == aws.StringValue(rt.GatewayID) && ob.DestinationCIDRBlock == aws.StringValue(rt.DestinationCIDRBlock) {
+			if ob.DestinationCIDRBlock == aws.StringValue(rt.DestinationCIDRBlock) && (ob.GatewayID == aws.StringValue(rt.GatewayID) ||
```
muvaf, line 288: I'm not sure about this condition. What happens when user changes only one or a few fields but not all? In such cases, we should delete and recreate the route since there is no update call but from what I can tell, having one field match is OK to skip doing any operation. Should we use `&&` instead of `||`?

pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,110 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
 	return awsclient.Wrap(resource.Ignore(ec2.IsRouteTableNotFoundErr, err), errDelete)
 }
 
-func (e *external) createRoutes(ctx context.Context, tableID string, desired []v1alpha4.Route, observed []v1alpha4.RouteState) error {
+func (e *external) deleteRoutes(ctx context.Context, tableID string, desired []v1alpha5.Route, observed []v1alpha5.RouteState) error { // nolint:gocyclo
+	for _, rt := range observed {
+		found := false
+		for _, ds := range desired {
+			if aws.StringValue(ds.DestinationCIDRBlock) == rt.DestinationCIDRBlock && (aws.StringValue(ds.GatewayID) == rt.GatewayID ||
```
muvaf, line 257: See my comment on `createRoutes`.

pkg/controller/ec2/routetable/controller.go
```diff
@@ -237,33 +250,110 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
...
+		found := false
+		for _, ds := range desired {
+			if asc.SubnetID == aws.StringValue(ds.SubnetID) {
+				found = true
+				break
+			}
+		}
+		if !found {
+			// No longer needed so delete
+			_, err := e.client.DisassociateRouteTableRequest(&awsec2.DisassociateRouteTableInput{
```
muvaf, line 345: Can we reuse `deleteAssociations` function here?

pkg/controller/ec2/routetable/controller.go
```diff
@@ -192,24 +194,35 @@ func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.Ex
 
 	if len(patch.Tags) != 0 {
 		// tagging the RouteTable
-		if _, err := e.client.CreateTagsRequest(&awsec2.CreateTagsInput{
-			Resources: []string{meta.GetExternalName(cr)},
-			Tags:      v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags),
-		}).Send(ctx); err != nil {
-			return managed.ExternalUpdate{}, awsclient.Wrap(err, errCreateTags)
+		addTags, removeTags := awsclient.DiffEC2Tags(v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags), table.Tags)
```
muvaf, line 197: Thanks for adding this!

apis/ec2/v1alpha5/referencers.go
```diff
@@ -62,6 +63,22 @@ func (mg *RouteTable) ResolveReferences(ctx context.Context, c client.Reader) er
...
+	for i := range mg.Spec.ForProvider.Routes {
+		rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+			CurrentValue: aws.StringValue(mg.Spec.ForProvider.Routes[i].NatGatewayID),
+			Reference:    mg.Spec.ForProvider.Routes[i].NatGatewayIDRef,
+			Selector:     mg.Spec.ForProvider.Routes[i].NatGatewayIDSelector,
+			To:           reference.To{Managed: &ec2v1alpha1.NATGateway{}, List: &ec2v1alpha1.NATGatewayList{}},
+			Extract:      reference.ExternalName(),
+		})
+		if err != nil {
+			return errors.Wrapf(err, "spec.forProvider.routes[%d].gatewayId", i)
```
muvaf, line 76:
```suggestion
			return errors.Wrapf(err, "spec.forProvider.routes[%d].natGatewayId", i)
```

pkg/clients/ec2/routetable.go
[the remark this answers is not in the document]
enderv: As near as I could tell since they are nested under Associations they weren't being ignored and couldn't find a way to set the options to ignore them so just cleared them out.

pkg/controller/ec2/routetable/controller.go
enderv: This is a fun one. Will try my best to explain but it should be checking that the destinationCIDR AND a destination is still the same. Since there should only be one destination per route ( which is why i was doing the `||`) I think `&&` should work though and better enforce the check will try it out.

pkg/controller/ec2/routetable/controller.go
enderv: went ahead and added better message while i'm at it

pkg/clients/ec2/routetable.go
enderv: Was using this to make sure they were sorted before creating a patch instead of at the compare bit (if that makes sense) was seeing issues on the patch if they weren't in the same order.

enderv:
@muvaf Think I fixed and addressed everything. Took back to v1alpha4 or left a comment for why i didn't do a change 🤞

pkg/controller/ec2/routetable/controller.go
muvaf: This seems to be marked as resolved. Is there a specific reason why we can't accumulate all associations into a `[]v1alpha4.AssociationState` and pass that to `deleteAssociations` after the loop finishes? Having one implementation is better because for example here `NotFound` case is not ignored while `deleteAssociations` takes care of that.

muvaf (approved):
Thank you for working on this challenging API @enderv !!

LGTM. I have some non-blocking comments. Let me know when this is ready to merge.

pkg/controller/ec2/routetable/controller.go
enderv: Oops sorry swear I had done this already. Went ahead and got it added.

enderv:
@muvaf  Thanks! Added the last thing and did another rebase to change the signing email but should be good to go.

PR #486, opened 2021-01-08
Documentation to add new resource support using ACK

muvaf:

I was planning to write this after I get referencer field generation as well but several community members are asking how they can contribute to provider-aws using ACK so I decided to put it up there and make changes as we go.

Fixes https://github.com/crossplane/provider-aws/issues/482

I have:

Not needed.

### DIFF
--- a/CODE_GENERATION.md
+++ b/CODE_GENERATION.md
@@ -1,41 +1,57 @@
 # Contributing New Resource Using ACK
 
-Provider AWS uses ACK code generation pipeline to generate the boilerplate for
-CRDs and their controllers. Most of the generated code is ready to go but there
-are places where we need Crossplane-specific functionality or some hacks to work
-around non-generic behavior of the AWS API.
+[AWS Controllers for Kubernetes (ACK)](https://aws.github.io/aws-controllers-k8s/)
+is an open source project built by AWS that has Kubernetes controllers for AWS
+services and a code generation pipeline built to generate those controllers from
+AWS Go SDK. We are collaborating with ACK community to get more coverage of AWS
+services in provider-aws by using some parts of the ACK code generation pipeline.
+
+Even though we are able to use the same code generation pipeline, Crossplane and
+ACK differs in terms of how their CRDs look like, what common reconciler they use
+and how the controllers are deployed etc. What we generate apart from generic
+Crossplane boilerplate can be summarized into conversion functions between AWS SDK
+structs and local structs and the CRD fields under `spec.forProvider`/`status.atProvider`
+structs.
+
+Most of the generated code is ready to go but there are places where we need
+Crossplane-specific functionality, or some hacks to work around non-generic
+behavior of the AWS API.
 
 This guide shows how to get a new resource support up and running step by step.
 
 ## Code generation
 
-This is the very first step. You need to clone ACK repository and run the following
+This is the very first step. You need to clone [ACK repository](https://github.com/aws/aws-controllers-k8s/) and run the following
 command in the root directory:
 
 ```console
-go -tags generate run cmd/ack-generate/main.go crossplane <ServiceID> --provider-dir <provider-aws directory>
+go run -tags codegen cmd/ack-generate/main.go crossplane <ServiceID> --provider-dir <provider-aws directory>
 ```
 
-`ServiceID` list is found here: https://github.com/aws/aws-sdk-go/tree/master/models/apis
+`ServiceID` list is found [here](https://github.com/aws/aws-sdk-go/tree/v1.34.32/models/apis).
 For example, in order to generate Lambda resources, you need to use `lambda` as
 Service ID.
 
-Once this is completed, you'll see new folders in `apis` and `pkg/controller` directories.
-By default, it generates every resource it can recognize but that's not what you
-want if you'd like to add support for only a few resource. In order to ignore some
-of the generated resources, you need to create a `apis/<serviceid>/v1alpha1/generator-config.yaml`
+The first run will take a while since it clones AWS SDK. Once it is completed,
+you'll see new folders in `apis` and `pkg/controller` directories. By default,
+it generates every resource it can recognize but that's not what you want if you'd
+like to add support for only a few resources. In order to ignore some of the
+generated resources, you need to create a `apis/<serviceid>/v1alpha1/generator-config.yaml`
 file with the following content:
 ```yaml
 ignore:
   resource_names:
-    - <CRD Name you want to ignore>
+    - DBInstance
+    - DBCluster
+    - <other CRD kinds you want to ignore>
 ```
 
 When you re-run the generation with this configuration, existing files won't be
 deleted. So you may want to delete everything in `apis/<serviceid>/v1alpha1` except
 `generator-config.yaml` and then re-run the command.
 
-If the generation fails for any reason, please open an issue.
+If this step fails for some reason, please raise an issue in [ACK](https://github.com/aws/aws-controllers-k8s/)
+and mention that you're using Crossplane pipeline.
 
 ### Crossplane Code Generation
 
@@ -44,7 +60,8 @@ which includes kubebuilder and crossplane-tools processes. However, currently
 there are some missing structs in `apis/<serviceid>/v1alpha1` folder called
 `Custom<CRDName>Parameters` which we'll use in the next steps. For now, create a
 file called `apis/<serviceid>/v1alpha1/custom_types.go` with empty structs to get
-the code to compile like the following:
+the code to compile. For example, let's say you have a CRD whose kind is `Stage`.
+The empty struct would look like following:
 ```golang
 // CustomStageParameters includes the custom fields of Stage.
 type CustomStageParameters struct{}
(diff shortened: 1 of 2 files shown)


CODE_GENERATION.md
```diff
@@ -0,0 +1,284 @@
...
+	func isUpToDate(*svcapitypes.Stage, *svcsdk.DescribeStageOutput) bool {
+return true
+}
+```
+
+#### Update
+
+ACK has partial support for update calls. You need to inspect what's generated
+and inject custom code wherever necessary. In a lot of cases, Update calls are tricky
+and may have to be done by several calls. You can see an injected [example here](https://github.com/crossplane/provider-aws/pull/485/files#diff-8a38ae8bcdbb080fbc35feb71701c1ada99de0c36c474d42899d1ce628cf3927R277)
```
muvaf, line 277: I will make it ready to merge after this PR is merged.

CODE_GENERATION.md
```diff
@@ -0,0 +1,288 @@
...
+Provider AWS uses ACK code generation pipeline to generate the boilerplate for
+CRDs and their controllers. Most of the generated code is ready to go but there
+are places where we need Crossplane-specific functionality or some hacks to work
+around non-generic behavior of the AWS API.
+
+This guide shows how to get a new resource support up and running step by step.
+
+## Code generation
+
+This is the very first step. You need to clone ACK repository and run the following
```
hasheddan, line 12: might be worth linking the URL of ACK repository here

CODE_GENERATION.md
```diff
@@ -0,0 +1,288 @@
...
+
+In Crossplane, fields whose value can be taken from another CRD have two additional
+fields called `*Ref` and `*Selector`. At the moment, ACK doesn't know about them
+so we add them manually.
+
+You'll see that `<CRDName>Parameters` struct has an inline field called `Custom<CRDName>Parameters`
+whose struct we left empty earlier. We'll add the additional fields there.
+
+Note that some fields are required and reference-able but we cannot mark them required
+since when a reference is given, their value is resolver after the creation. In
```
hasheddan, line 99:
```suggestion
since when a reference is given, their value is resolved after the creation. In
```

CODE_GENERATION.md
```diff
@@ -0,0 +1,288 @@
...
+such cases, we need to omit that field and put it under `Custom<CRDName>Parameters`
+alongisde with its referencer and selector fields. To ignore, you need to add the
```
hasheddan, line 101:
```suggestion
alongside with its referencer and selector fields. To ignore, you need to add the
```

CODE_GENERATION.md
```diff
@@ -0,0 +1,288 @@
...
+### Reference Resolvers
+
+We need to implement the resolver functions for every reference-able field. They
+are mostly identical and you can see examples in the existing implementations like
+[`Stage`](https://github.com/crossplane/provider-aws/blob/c269977/apis/apigatewayv2/v1alpha1/referencers.go#L30) resource.
+
+### External Name
+
+Crossplane has the notion of external name that we put under annotations. It corresponds
+to the identifier field of the resource. We try to stick with name if available
```
hasheddan, line 127: nipick:
```suggestion
to the identifier field of the resource in AWS. We try to stick with name if available
```

CODE_GENERATION.md
```diff
@@ -0,0 +1,288 @@
...
+## Custom Parts for Beta Quality
+
+The earlier section is sufficient to get a controller into a working state. But
+there are some functionalities missing like late-initialization and update. These
+are not supported by ACK and even if there is support for update, it usually has
+some caveats that cannot be made generic. So, we need to implement them manually.
+
+### Late Initialization
+
+Late initialization means that if user doesn't have a desired value for a field
```
hasheddan, line 229:
```suggestion
Late initialization means that if a user doesn't have a desired value for a field
```

hasheddan (approved):
Thanks for this @muvaf! Added a few non-blocking comments, feel free to merge when you are ready :)

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
+# Contributing New Resource Using ACK
```
negz, line 1: Nit: I bet 90% of folks reading this will know already, but it may be worth using 1-2 sentences to explain what ACK is and how it relates to Crossplane (e.g. that they share a codegen pipeline but the actual Amazon controllers are a separate thing).

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
...
+The first run will take a while since it clones AWS SDK. Once it is completed,
+you'll see new folders in `apis` and `pkg/controller` directories.
+By default, it generates every resource it can recognize but that's not what you
+want if you'd like to add support for only a few resource. In order to ignore some
+of the generated resources, you need to create a `apis/<serviceid>/v1alpha1/generator-config.yaml`
+file with the following content:
+```yaml
+ignore:
+  resource_names:
+    - <CRD Name you want to ignore>
```
negz, line 32: Is this a `kind` name (e.g. `DBInstance`) or the fully qualified name (e.g. `dbinstances.rds.aws.crossplane.io`)?

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+When you re-run the generation with this configuration, existing files won't be
+deleted. So you may want to delete everything in `apis/<serviceid>/v1alpha1` except
+`generator-config.yaml` and then re-run the command.
+
+If the generation fails for any reason, please open an issue.
```
negz, line 39: Should we open issues against ACK, or Crossplane?

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
...
+
+After ACK generation process completes, we need to run Crossplane generation procedure
+which includes kubebuilder and crossplane-tools processes. However, currently
+there are some missing structs in `apis/<serviceid>/v1alpha1` folder called
+`Custom<CRDName>Parameters` which we'll use in the next steps. For now, create a
+file called `apis/<serviceid>/v1alpha1/custom_types.go` with empty structs to get
+the code to compile like the following:
+```golang
+// CustomStageParameters includes the custom fields of Stage.
+type CustomStageParameters struct{}
```
negz, line 51: Does `Stage` have special meaning to ACK, or is this just an example kind? I thought the former at first, but now I suspect it's the latter. I think it's worth clarifying in the document.

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
...
+ACK has partial support for update calls. You need to inspect what's generated
+and inject custom code wherever necessary. In a lot of cases, Update calls are tricky
+and may have to be done by several calls. You can see an injected [example here](https://github.com/crossplane/provider-aws/pull/485/files#diff-8a38ae8bcdbb080fbc35feb71701c1ada99de0c36c474d42899d1ce628cf3927R277)
+with custom logic to work around an API quirk.
+
+## Testing
+
+You need to thoroughly test the controller by using example YAMLs. Make sure
+every functionality you implement works by checking with AWS console. Once all looks
+good, add the example YAML you used to `examples/<serviceid>/<crd-name>.yaml`
```
negz, line 290: What are our unit testing practices for ACK-generated controllers? It makes sense to skip unit tests for generated code, but should we be writing tests for the custom parts that are written by humans?

CODE_GENERATION.md
hiteshghia-splunk: I believe  it is the `kind` name - DBInstance

CODE_GENERATION.md
```diff
@@ -0,0 +1,290 @@
...
+file called `apis/<serviceid>/v1alpha1/custom_types.go` with empty structs to get
+the code to compile like the following:
+```golang
+// CustomStageParameters includes the custom fields of Stage.
+type CustomStageParameters struct{}
+```
+
+Now the code compiles, run the following:
+```
+make generate
```
hiteshghia-splunk, line 56: Some explanation of what  this command is suppose to do will be helpful. for eg creates the crd.yaml.

hiteshghia-splunk: Also do we manually manually need to run https://github.com/crossplane/provider-aws/blob/master/apis/generate.go?

CODE_GENERATION.md
muvaf: It should be ACK since the code is there. Added some clarification.

CODE_GENERATION.md
muvaf: It's just an example kind, added a sentence to make it clear.

CODE_GENERATION.md
muvaf: Just added a section there. I think we need to write tests for every custom code we put in but other than that, tests for generated code should be OK to skip.

CODE_GENERATION.md
muvaf: Just added some details.

muvaf:
This is ready to go. @negz @hiteshghia-splunk let me know if you need another look, otherwise I'm planning to merge this in a few hours as @hasheddan already gave his thumbs up.

PR #499, opened 2021-01-20
Support custom Docker registry when building the Docker images

bandesz:

I would like to run a custom build of the AWS provider and for this I would like to upload the built Docker images to my own Docker registry.

The current setup has the following limitations:
 - the value of DOCKER_REGISTRY can not be overridden when running make targets - this is just an inconvenience as it's easy to tag the built images
 - the Docker image name for the controller is baked in as `crossplane/provider-aws-controller:VERSION` - this is again a minor inconvenience because you can override the image using the ControllerConfig CRD.

Still it's more convenient if I can do a full build, upload both images and I only have to update the spec.package on the Provider object.

For reference I was running the following commands:

```
PLATFORMS=linux_amd64 DOCKER_REGISTRY=quay.io/acme VERSION=v0.16.0-p1 make build publish

docker tag quay.io/acme/provider-aws-amd64:v0.16.0-p1 quay.io/acme/provider-aws:v0.16.0-p1
docker push quay.io/acme/provider-aws:v0.16.0-p1
docker tag quay.io/acme/provider-aws-controller-amd64:v0.16.0-p1 quay.io/acme/provider-aws-controller:v0.16.0-p1
docker push quay.io/acme/provider-aws-controller:v0.16.0-p1
```

I have:

I deployed my own images and verified that the provider works.

Also I've ran the make targets without setting DOCKER_REGISTRY and verified that it uses "crossplane" correctly everywere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ GO111MODULE = on
 # ====================================================================================
 # Setup Images
 
-DOCKER_REGISTRY = crossplane
+DOCKER_REGISTRY ?= crossplane
 IMAGES = provider-aws provider-aws-controller
 -include build/makelib/image.mk
 
--- a/cluster/images/provider-aws/Makefile
+++ b/cluster/images/provider-aws/Makefile
@@ -6,6 +6,7 @@ include ../../../build/makelib/common.mk
 
 # ====================================================================================
 #  Options
+DOCKER_REGISTRY ?= crossplane
 IMAGE = $(BUILD_REGISTRY)/provider-aws-$(ARCH)
 OSBASEIMAGE = scratch
 include ../../../build/makelib/image.mk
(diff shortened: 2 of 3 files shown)


muvaf (approved):
LGTM but I'd want @hasheddan to weigh in here since he dealt with our build pipelines recently. Maybe we can enable this in some form in `build` submodule so that it applies to all repos.

negz:
I'd certainly like to keep our makefiles as consistent as possible, even if it's as simple as manually replicating these changes in our various other repos.

hasheddan (approved):
This seems reasonable to me 👍 I plan on rounding through our repos and potentially moving to buildx tooling across the board instead of the submodule image.mk commands. Until then @bandesz would you mind making this same change in other provider repos as @negz suggested?

bandesz:
@hasheddan @negz  I won't be able to do this for the other providers in the near future, sorry.

Issue #477, opened 2020-12-18
EKS NodeGroup scalingConfig.desiredSize can conflict with autoscaler


enderv:
### What happened?
I created a node group for an already existing EKS cluster running cluster autoscaler. I never deployed anything to the nodes and cluster autoscaler adjusted the desired size down to the group minimum but the nodegroup reconciler bumped it back up to what was defined in the spec.

### How can we reproduce it?
Need an already existing EKS cluster with [cluster autoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md) installed.

```
apiVersion: eks.aws.crossplane.io/v1alpha1
kind: NodeGroup
metadata:
  name: testgroupb
spec:
  deletionPolicy: Delete
  forProvider:
    amiType: AL2_x86_64
    clusterName: <your_cluster_name>
    diskSize: 20
    instanceTypes:
    - t2.medium
    nodeRole: <node-arn>
    region: us-east-1
    scalingConfig:
      maxSize: 5
      minSize: 1
      desiredSize: 3
    subnets:
    - <your_subnets>
    version: "1.17"
  providerConfigRef:
    name: aws-provider
```

### What environment did it happen in?
Crossplane version: 1.0.0
Provider-aws: v0.16.0
Kubernetes Version: EKS 1.17
Cluster Autoscaler Version:1.17.3

hasheddan:
@enderv would it make sense to omit your `scalingConfig` in this case?

muvaf:
FWIW, Crossplane deems to be the source of truth so it will correct any drift. In this case, I believe autoscaler changes the desired size, but not min/max. From what I can tell, `desiredSize` is optional. So, if you'd like autoscaler to manage that property, leaving `desiredSize` empty might help.

enderv:
@hasheddan - Just tried with no scaling config and ended up with max: 2 min: 1 desired: 2. I'm assuming it wouldn't try to correct desired anymore but can't really do much with a max 2 unfortunately :( 

@muvaf - unfortunately looks like if one value is supplied have to give all of them. Tried with no desiredSize and would get 400 back from AWS saying its a required value. Only docs I could actually find saying that though is the [cloudformation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-nodegroup-scalingconfig.html) docs.

enderv:
[...]

PR #502, opened 2021-01-22
ignore nodegroup desiredSize

mcavoyk:
Signed-off-by: Kellin McAvoy <kellinmcavoy@gmail.com>

Fixes https://github.com/crossplane/provider-aws/issues/477

I have:

Added new unit tests. Checked e2e tested with a nodeGroup without `desiredSize` set and modified min/max to ensure create and update worked with nil `desiredSize`.


mcavoyk:
@muvaf This PR was my thinking for a fix based on the related issue. AWS requires that this field is set, the CRD already allows the field to be nil. If the desiredSize of the NodeGroup CRD is nil, then desiredSize is ignored in `IsUpToDate` and during update requests the previously seen desiredSize is used as the value.

This should allow another controller like Cluster Autoscaler the ability to adjust `desiredSize` if the value is nil in the NodeGroup CRD.

pkg/clients/eks/nodegroup.go
```diff
@@ -54,6 +54,11 @@ func GenerateCreateNodeGroupInput(name string, p *v1alpha1.NodeGroupParameters)
 			MinSize:     p.ScalingConfig.MinSize,
 			MaxSize:     p.ScalingConfig.MaxSize,
 		}
+
+		// If desiredSize is not set, default to minSize
+		if p.ScalingConfig.DesiredSize == nil {
+			c.ScalingConfig.DesiredSize = p.ScalingConfig.MinSize
```
muvaf, line 60: Hmm I don't really feel comfortable having this controller make that decision but it seems we're required to find a value for API request while allowing user to leave it as `nil` to handle the autoscaler problem and in creation case, it becomes a chicken-egg problem. Could you explain that as comment? Maybe refer to the issue as well?

pkg/clients/eks/nodegroup.go
```diff
@@ -78,6 +83,10 @@ func GenerateUpdateNodeGroupConfigInput(name string, p *v1alpha1.NodeGroupParame
 			MinSize:     p.ScalingConfig.MinSize,
 			MaxSize:     p.ScalingConfig.MaxSize,
 		}
+
```
muvaf, line 86: A comment here explaining why we do this would help the future readers.

pkg/clients/eks/nodegroup.go
```diff
@@ -177,7 +194,7 @@ func IsNodeGroupUpToDate(p *v1alpha1.NodeGroupParameters, ng *eks.Nodegroup) boo
 		return true
 	}
 	if p.ScalingConfig != nil && ng.ScalingConfig != nil {
-		if !cmp.Equal(p.ScalingConfig.DesiredSize, ng.ScalingConfig.DesiredSize) {
+		if p.ScalingConfig.DesiredSize != nil && !cmp.Equal(p.ScalingConfig.DesiredSize, ng.ScalingConfig.DesiredSize) {
```
muvaf, line 197:
```suggestion
		if p.ScalingConfig.DesiredSize != nil && awsclients.Int64Value(p.ScalingConfig.DesiredSize) != awsclients.Int64Value(ng.ScalingConfig.DesiredSize) {
```

apis/eks/v1alpha1/nodegroup_types.go
```diff
@@ -224,6 +224,11 @@ type NodeGroupObservation struct {
 	// and security groups for remote access.
 	Resources NodeGroupResources `json:"resources,omitempty"`
 
+	// The scaling configuration details for the Auto Scaling group that is created
+	// for your node group.
+	// +optional
```
muvaf, line 229:
```suggestion
```

muvaf (approved):
LGTM! I have a few small comments, let me know when it's ready to merge. @mcavoyk thanks for tackling this problem!

pkg/clients/eks/nodegroup.go
mcavoyk: Good point, should be resolved

mcavoyk:
@muvaf Should be good now, just finished up e2e testing :+1:

apis/eks/v1alpha1/nodegroup_types.go
```diff
@@ -204,6 +204,12 @@ type NodeGroupScalingConfig struct {
 	MinSize *int64 `json:"minSize,omitempty"`
 }
 
+// NodeGroupScalingConfigStatus is the observed scaling configuration for a node group.
+type NodeGroupScalingConfigStatus struct {
```
mcavoyk, line 208: I broke out desiredSize into a status struct based on https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#embedded-structs-with-mixed-fields

pkg/clients/eks/nodegroup.go
```diff
@@ -78,6 +86,20 @@ func GenerateUpdateNodeGroupConfigInput(name string, p *v1alpha1.NodeGroupParame
 			MinSize:     p.ScalingConfig.MinSize,
 			MaxSize:     p.ScalingConfig.MaxSize,
 		}
+
+		// If desiredSize is not set, derive the value from either the
+		// current observed desiredSize, or the min/max if observed is out of bounds.
+		if p.ScalingConfig.DesiredSize == nil {
+			// The min/max size set the floor/ceiling for the desiredSize
+			switch desiredSizeVal := awsclients.Int64Value(ng.ScalingConfig.DesiredSize); {
```
muvaf, line 94: I wouldn't really do those checks because we should aim for making these logic functions as dumb as possible. So, if desired size is to be ignored, we should just use the observed size.

apis/eks/v1alpha1/nodegroup_types.go
```diff
@@ -204,6 +204,12 @@ type NodeGroupScalingConfig struct {
 	MinSize *int64 `json:"minSize,omitempty"`
 }
```
muvaf, line 205: Could you add a comment to `DesiredSize` field about what happens if you leave it empty and/or what you should do if you want to use autoscaler so that it appears in documentation?

pkg/clients/eks/nodegroup.go
mcavoyk: That was my initial implementation. When testing I ran into some issues where the crd had say `minSize` of 1 and `maxSize` of 10, and the nodeGroup had been scaled by another entity to a desired size of 4. Because of the value of the desiredSize, increasing the `minSize` of the nodeGroup  > 4 will return errors from AWS as this would cause desired size to be less than `minSize`.

My thinking is that because the user has provided a min/max size without a desired size, when creating we can derive a suitable desired size using the min size as an initial value.. When updating we can derive one as well, either the current observed value for desired size, or the bounds of the min/max size.

muvaf (approved):
LGTM. Thank you @mcavoyk !

PR #528, opened 2021-02-05
iamrole: add update capability for tags

muvaf:

Fixes https://github.com/crossplane/crossplane/issues/2118

I have:

Unit test and manually.


pkg/clients/iam/iamrole.go
```diff
@@ -151,3 +153,27 @@ func IsRoleUpToDate(in v1beta1.IAMRoleParameters, observed iam.Role) (bool, erro
 
 	return cmp.Equal(desired, &observed, cmpopts.IgnoreInterfaces(struct{ resource.AttributeReferencer }{})), nil
 }
+
+// DiffIAMTags returns the lists of tags that need to be removed and added according
+// to current and desired states.
+func DiffIAMTags(local []v1beta1.Tag, remote []iam.Tag) (add []iam.Tag, remove []string) {
```
hasheddan, line 159: I would add same comments as the ECR PR here. I was thinking if it would be useful to have a tag interface since we are doing this in a number of places, but I also realize that some of this has to be repeated due to the AWS structs :/

muvaf: Yes, each service has its own `Tag` implementation. Generics could help here in the future.

hasheddan (approved):

PR #529, opened 2021-02-05
ecr: simplify tagging

muvaf:

Noticed when I was working on https://github.com/crossplane/provider-aws/pull/528

Fixes #

I have:

Unit test.


pkg/clients/ecr/repository.go
```diff
@@ -125,3 +126,54 @@ func GenerateCreateRepositoryInput(name string, params *v1alpha1.RepositoryParam
...
+			continue
+		}
+		removeMap[aws.StringValue(t.Key)] = aws.StringValue(t.Value)
+	}
+	for k, v := range addMap {
+		addTags = append(addTags, ecr.Tag{Key: aws.String(k), Value: aws.String(v)})
+	}
+	for k := range removeMap {
+		remove = append(remove, k)
+	}
```
hasheddan, line 177: nitpick: can we pre-allocate the size of `addTags` and `remove` since they are known at this point?

pkg/clients/ecr/repository.go
```diff
@@ -125,3 +126,54 @@ func GenerateCreateRepositoryInput(name string, params *v1alpha1.RepositoryParam
...
+	})
+
+	sort.Slice(ecrTags, func(i, j int) bool {
+		return *ecrTags[i].Key < *ecrTags[j].Key
+	})
+}
+
+// DiffTags returns tags that should be added or removed.
+func DiffTags(spec []v1alpha1.Tag, current []ecr.Tag) (addTags []ecr.Tag, remove []string) {
+	addMap := map[string]string{}
```
hasheddan, line 160: nitpick: we know `addMap` max size
```suggestion
	addMap := make(map[string]string, len(spec))
```

pkg/clients/ecr/repository.go
```diff
@@ -125,3 +126,54 @@ func GenerateCreateRepositoryInput(name string, params *v1alpha1.RepositoryParam
...
+	for _, t := range spec {
+		addMap[t.Key] = t.Value
+	}
+	removeMap := map[string]string{}
```
hasheddan, line 164: Since we are just using `removeMap` to guarantee uniqueness of keys, it would be more efficient to just use a `map[string]struct{}{}` as an empty struct has zero width.

pkg/clients/ecr/repository.go
muvaf: If we pre-allocate, we can't use `append` because it adds to the end so we'd need indexes in each of the for loops and that's not readily available because we iterate on map. So, I just preferred `append` instead of having to do bookkeeping separately.

muvaf: Just tested https://play.golang.org/p/jmLT-dbhnNA

hasheddan: Right, I was suggesting:
```go
i := 0
for k := range removeMap {
  remove[i] = k
  i++
}
```

But I am fine with moving forward as-is if it is your preference.

pkg/clients/ecr/repository.go
muvaf: I guess `bool` would take even less space, right?

hasheddan: @muvaf a `bool` is 1 byte, `struct{}{}` is 0, so empty struct would technically be optimal

hasheddan (approved):

PR #530, opened 2021-02-05
Fix flaky tag comparison

mcavoyk:
Signed-off-by: Kellin McAvoy <kellinmcavoy@gmail.com>

Fixes flaky unit tests

I have:

Increase test count in Makefile to bump the default/CI test count.

### DIFF
--- a/pkg/clients/aws_test.go
+++ b/pkg/clients/aws_test.go
@@ -353,11 +353,14 @@ func TestDiffEC2Tags(t *testing.T) {
 
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -28,6 +28,9 @@ NPROCS ?= 1
 # to half the number of CPU cores.
 GO_TEST_PARALLEL := $(shell echo $$(( $(NPROCS) / 2 )))
 
+# increase test count to reduce chance of missing flaky unit tests
+GO_TEST_FLAGS    ?= -count=5
```
hasheddan, line 32: We haven't had too much of a problem with flakes on unit tests so far, so I think I would probably leave this off for now. If we do include it we should use `-failfast` as well

hasheddan (approved):
Thanks @mcavoyk!

PR #536, opened 2021-02-10
S3 Bucket Late Initialization

### LINKED ISSUE #507: Optional fields are missing after import existing AWS S3 bucket into crossplane
### What happened?

Tried to import existing aws s3 bucket by following steps in https://crossplane.io/docs/v0.14/introduction/managed-resources.html#importing-existing-resources. However, all optional fields (ie, tagging, cors, encryption, policy etc) are missing, and the optional fileds in the actual aws s3 bucket has been reset to be default (in this case, all empty). which is not desired.  

### How can we reproduce it?

1. create an aws s3 bucket
    - add some tagging.
    - add a policy and cors rule.
    - and other settings like encryption.
2. apply following yaml to import:
    make sure the external-name and all mandatory fields match the existing aws s3 bucket settings.
```
kind: Bucket
metadata:
  name: bucket_name
  namespace: default
  annotations:
    crossplane.io/external-name: existed_bucket_in_aws
spec:
  forProvider:
    locationConstraint: us-west-2
    acl: "private"
```

3.  What I expected is the crossplane will import the existing bucket as a managed resources, also keep all field values from aws bucket. however, what happened is that all the optional fields (like tagging, encryption, cors, policy etc) become empty, and the actual aws bucket also got updated.

Crossplane version: 
1.0

Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"darwin/amd64"}

krishchow:

Fixes #507 
Also relevant: https://crossplane.slack.com/archives/CEG3T90A1/p1612889482408500

I have:

Manual testing, updated unit tests


pkg/controller/s3/bucket.go
```diff
@@ -98,43 +98,45 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	}
 
 	if _, err := e.s3client.HeadBucketRequest(&awss3.HeadBucketInput{Bucket: aws.String(meta.GetExternalName(cr))}).Send(ctx); err != nil {
-		return managed.ExternalObservation{}, awsclient.Wrap(resource.Ignore(s3.IsNotFound, err), errHead)
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(s3.IsNotFound, err), errHead)
```
negz, line 101: Should this definitely change? iirc `awsclient.Wrap` extends `errors.Wrap` to trim UIDs from AWS errors so that we don't consider two instances of the same error to be two different errors.

pkg/controller/s3/bucket.go
```diff
@@ -168,11 +185,11 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
-				return managed.ExternalUpdate{}, awsclient.Wrap(err, errDelete)
+				return managed.ExternalUpdate{}, errors.Wrap(err, errDelete)
 			}
 		case bucket.NeedsUpdate:
 			if err := awsClient.CreateOrUpdate(ctx, cr); err != nil {
-				return managed.ExternalUpdate{}, awsclient.Wrap(err, errCreateOrUpdate)
+				return managed.ExternalUpdate{}, errors.Wrap(err, errCreateOrUpdate)
 			}
 		}
 	}
```
muvaf, line 195: Unrelated: managed reconciler already takes care of reconcile success condition, so you can remove line `196`.

examples/s3/late-init-bucket.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: s3.aws.crossplane.io/v1beta1
```
muvaf, line 1: Do we need two examples? Which one do you think shows off the most functionality?

pkg/clients/s3/bucket.go
```diff
@@ -242,6 +243,15 @@ func CopyTags(tags []v1beta1.Tag) []s3.Tag {
 	return out
 }
 
+// CopyAWSTags converts a list of external s3.Tags to local Tags
+func CopyAWSTags(tags []s3.Tag) []v1beta1.Tag {
+	out := make([]v1beta1.Tag, 0)
+	for _, one := range tags {
+		out = append(out, v1beta1.Tag{Key: aws.StringValue(one.Key), Value: aws.StringValue(one.Value)})
+	}
```
muvaf, line 251:
```suggestion
	out := make([]v1beta1.Tag, len(tags))
	for i, one := range tags {
		out[i] = v1beta1.Tag{Key: aws.StringValue(one.Key), Value: aws.StringValue(one.Value)}
	}
```

nitpick

pkg/controller/s3/bucket.go
```diff
@@ -98,43 +98,45 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 	for _, awsClient := range e.subresourceClients {
-		err := awsClient.LateInitialize(ctx, cr)
-		if err != nil {
-			return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false}, err
+		if awsClient.SubresourceExists(cr) {
+			err := awsClient.LateInitialize(ctx, cr)
+			if err != nil {
+				return managed.ExternalObservation{ResourceExists: true, ResourceUpToDate: false}, err
+			}
+			lateInit = true
```
muvaf, line 114: The `ResourceLateInitialized` field in return variable triggers a spec update which has to delete the status subresource, so we treat carefully and not call it unless there is a change in spec we need to save. In this instance, it seems like subresource existence causes `ResourceLateInitialized` to be returned as `true`, which might render status subresource useless. I'd suggest returning `ResourceLateInitialized: true` only if spec has changed.

pkg/controller/s3/bucket/CORSConfig.go
```diff
@@ -131,3 +143,36 @@ func (in *CORSConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.B
...
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()), "external output", external.GetBucketCorsOutput)
+
+	if bucket.Spec.ForProvider.CORSConfiguration == nil {
+		bucket.Spec.ForProvider.CORSConfiguration = &v1beta1.CORSConfiguration{}
+	}
+
+	bucket.Spec.ForProvider.CORSConfiguration.CORSRules = GenerateCORSRule(external.CORSRules)
```
muvaf, line 170: We can assign the observed state only if `len(bucket.Spec.ForProvider.CORSConfiguration.CORSRules) == 0` so that we don't override user's desired state.

pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -193,3 +190,147 @@ func sortFilterTags(rules []awss3.LifecycleRule) {
...
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	if bucket.Spec.ForProvider.LifecycleConfiguration == nil {
+		bucket.Spec.ForProvider.LifecycleConfiguration = &v1beta1.BucketLifecycleConfiguration{}
+	}
+
+	createLifecycleRulesFromExternal(external.Rules, bucket.Spec.ForProvider.LifecycleConfiguration)
```
muvaf, line 217: Unless there is a unique identifier in the type, `id` seems to exists but it's optional, we can late-init only if the array in spec is empty because otherwise, it's not possible to know which element corresponds to which in cloud and whether we should delete or late-init. So, I'd suggest having a simple `if len(bucket.Spec.ForProvider.LifecycleConfiguration.Rules) == 0` condition here and calling a raw creation function similar to `GenerateCORSRule`.

pkg/controller/s3/bucket/loggingConfig.go
```diff
@@ -167,3 +130,50 @@ func (in *LoggingConfigurationClient) CreateOrUpdate(ctx context.Context, bucket
...
+		// There is no value send by AWS to initialize
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	if config == nil {
+		// We need the configuration to exist so we can initialize
+		bucket.Spec.ForProvider.LoggingConfiguration = &v1beta1.LoggingConfiguration{}
+		config = bucket.Spec.ForProvider.LoggingConfiguration
```
muvaf, line 151: This is already set in line `140`

pkg/controller/s3/bucket/loggingConfig.go
```diff
@@ -167,3 +130,50 @@ func (in *LoggingConfigurationClient) CreateOrUpdate(ctx context.Context, bucket
...
+	}
+	// Late initialize the target Bucket and target prefix
+	config.TargetBucket = awsclient.LateInitializeStringPtr(config.TargetBucket, external.LoggingEnabled.TargetBucket)
+	config.TargetPrefix = awsclient.LateInitializeString(config.TargetPrefix, external.LoggingEnabled.TargetPrefix)
+	// If the there is an external target grant list, and the local one does not exist
+	// we create the target grant list
+	if external.LoggingEnabled.TargetGrants != nil && len(config.TargetGrants) == 0 {
```
muvaf, line 158:
```suggestion
	if len(external.LoggingEnabled.TargetGrants) != 0 && len(config.TargetGrants) == 0 {
```

Length check is always safer than nil check for arrays since array with 0 elements is not nil but empty for us.

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -331,3 +292,54 @@ func (in *NotificationConfigurationClient) CreateOrUpdate(ctx context.Context, b
...
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	config := bucket.Spec.ForProvider.NotificationConfiguration
+	if config == nil {
+		// We need the configuration to exist so we can initialize
+		bucket.Spec.ForProvider.NotificationConfiguration = &v1beta1.NotificationConfiguration{}
+		config = bucket.Spec.ForProvider.NotificationConfiguration
```
muvaf, line 313: Already set in line `309`

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -331,3 +292,54 @@ func (in *NotificationConfigurationClient) CreateOrUpdate(ctx context.Context, b
...
+	}
+
+	// A list is provided by AWS
+	if external.LambdaFunctionConfigurations != nil {
+		if config.LambdaFunctionConfigurations == nil {
+			config.LambdaFunctionConfigurations = make([]v1beta1.LambdaFunctionConfiguration, len(external.LambdaFunctionConfigurations))
+		}
+		LateInitializeLambda(external.LambdaFunctionConfigurations, config.LambdaFunctionConfigurations)
```
muvaf, line 321: Instead of doing this and having to check `if i >= len(local)` in `LateInitializeLambda` function, you can assign the output of `LateInitializeLambda` to `config.LambdaFunctionConfigurations` directly here since you know that it's empty.

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -331,3 +292,54 @@ func (in *NotificationConfigurationClient) CreateOrUpdate(ctx context.Context, b
...
+	config := bucket.Spec.ForProvider.NotificationConfiguration
+	if config == nil {
+		// We need the configuration to exist so we can initialize
+		bucket.Spec.ForProvider.NotificationConfiguration = &v1beta1.NotificationConfiguration{}
+		config = bucket.Spec.ForProvider.NotificationConfiguration
+	}
+
+	// A list is provided by AWS
+	if external.LambdaFunctionConfigurations != nil {
+		if config.LambdaFunctionConfigurations == nil {
```
muvaf, line 318:
```suggestion
	if len(external.LambdaFunctionConfigurations) != 0 && len(config.LambdaFunctionConfigurations) == 0 {
```

We can late-init only in this case.

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -331,3 +292,54 @@ func (in *NotificationConfigurationClient) CreateOrUpdate(ctx context.Context, b
...
+			config.LambdaFunctionConfigurations = make([]v1beta1.LambdaFunctionConfiguration, len(external.LambdaFunctionConfigurations))
+		}
+		LateInitializeLambda(external.LambdaFunctionConfigurations, config.LambdaFunctionConfigurations)
+	}
+
+	// A list is provided by AWS
+	if external.QueueConfigurations != nil {
```
muvaf, line 325: See my comments for `LateInitializeLambda`

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -331,3 +292,54 @@ func (in *NotificationConfigurationClient) CreateOrUpdate(ctx context.Context, b
...
+		if config.QueueConfigurations == nil {
+			config.QueueConfigurations = make([]v1beta1.QueueConfiguration, len(external.QueueConfigurations))
+		}
+		LateInitializeQueue(external.QueueConfigurations, config.QueueConfigurations)
+	}
+
+	// A list is provided by AWS
+	if external.TopicConfigurations != nil {
```
muvaf, line 333: See my comments for `LateInitializeLambda`

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -214,3 +213,154 @@ func sortReplicationRules(rules []awss3.ReplicationRule) {
...
 }
+
+// LateInitialize does nothing because the resource might have been deleted by
+// the user.
+func (in *ReplicationConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
+	external, err := in.client.GetBucketReplicationRequest(&awss3.GetBucketReplicationInput{Bucket: awsclient.String(meta.GetExternalName(bucket))}).Send(ctx)
+	if err != nil {
+		if s3.ReplicationConfigurationNotFound(err) {
+			return nil
+		}
```
muvaf, line 224:
```suggestion
```

Already ignored in `errors.Wrap`

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -214,3 +213,154 @@ func sortReplicationRules(rules []awss3.ReplicationRule) {
...
+	return nil
+}
+
+func createReplicationRulesFromExternal(external *awss3.ReplicationConfiguration, config *v1beta1.ReplicationConfiguration) { // nolint:gocyclo
+	config.Role = awsclient.LateInitializeStringPtr(config.Role, external.Role)
+	if config.Rules == nil {
+		config.Rules = make([]v1beta1.ReplicationRule, 0)
+	}
+
+	for i, rule := range external.Rules {
```
muvaf, line 249: Same comment as other late-inited arrays. Unless we can sort them reliably with a unique identifier (could be a field or combination of fields), we can late-init only if the desired array is empty.

pkg/controller/s3/bucket/requestPaymentConfig.go
```diff
@@ -104,3 +89,27 @@ func (in *RequestPaymentConfigurationClient) CreateOrUpdate(ctx context.Context,
...
+		return awsclient.Wrap(err, paymentGetFailed)
+	}
+	if len(external.Payer) == 0 {
+		return nil
+	}
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+	config := bucket.Spec.ForProvider.PayerConfiguration
+	if config == nil {
+		bucket.Spec.ForProvider.PayerConfiguration = &v1beta1.PaymentConfiguration{}
+		config = bucket.Spec.ForProvider.PayerConfiguration
```
muvaf, line 106: Already assigned in line `103` and honestly I don't really get the point of this redirection, why not just keep using `bucket.Spec.ForProvider.PayerConfiguration`?

pkg/controller/s3/bucket/sseConfig.go
```diff
@@ -121,3 +135,36 @@ func (in *SSEConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.Bu
...
 	return awsclient.Wrap(err, sseDeleteFailed)
 }
+
+// LateInitialize does nothing because the resource might have been deleted by
+// the user.
+func (in *SSEConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
+	external, err := in.client.GetBucketEncryptionRequest(&awss3.GetBucketEncryptionInput{Bucket: awsclient.String(meta.GetExternalName(bucket))}).Send(ctx)
+	if err != nil {
+		// Short stop method for requests without a server side encryption
+		if s3.SSEConfigurationNotFound(err) {
```
muvaf, line 145: We can use `resource.Ignore` in return statement.

pkg/controller/s3/bucket/sseConfig.go
```diff
@@ -121,3 +135,36 @@ func (in *SSEConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.Bu
...
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	if bucket.Spec.ForProvider.ServerSideEncryptionConfiguration == nil {
+		bucket.Spec.ForProvider.ServerSideEncryptionConfiguration = &v1beta1.ServerSideEncryptionConfiguration{}
+	}
+
+	bucket.Spec.ForProvider.ServerSideEncryptionConfiguration.Rules = GenerateLocalBucketEncryption(external.ServerSideEncryptionConfiguration)
```
muvaf, line 162: only if `len(bucket.Spec.ForProvider.ServerSideEncryptionConfiguration.Rules) == 0 && len(external.ServerSideEncryptionConfiguration) != 0`

pkg/controller/s3/bucket/subresources.go
```diff
@@ -29,22 +31,23 @@ type SubresourceClient interface {
 	CreateOrUpdate(ctx context.Context, bucket *v1beta1.Bucket) error
 	Delete(ctx context.Context, bucket *v1beta1.Bucket) error
 	LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error
+	SubresourceExists(bucket *v1beta1.Bucket) bool
 }
 
 // NewSubresourceClients creates the array of all clients for a given BucketProvider
-func NewSubresourceClients(client s3.BucketClient) []SubresourceClient {
+func NewSubresourceClients(client s3.BucketClient, l logging.Logger) []SubresourceClient {
```
muvaf, line 38: I'm not sure we want to log every late init call, even if it's debug but I'm fine either way.

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -110,3 +118,36 @@ func (in *TaggingConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
 	return awsclient.Wrap(err, taggingDeleteFailed)
 }
+
+// LateInitialize does nothing because the resource might have been deleted by
+// the user.
+func (in *TaggingConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
+	external, err := in.client.GetBucketTaggingRequest(&awss3.GetBucketTaggingInput{Bucket: awsclient.String(meta.GetExternalName(bucket))}).Send(ctx)
+	if err != nil {
+		// Short stop method for requests in a region without Tags
+		if s3.TaggingNotFound(err) {
```
muvaf, line 128: We can use `resource.Ignore` in return statement.

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -110,3 +118,36 @@ func (in *TaggingConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
+	// We need the second check here because by default the tags are not set
+	if external.GetBucketTaggingOutput == nil || len(external.TagSet) == 0 {
+		return nil
+	}
+
+	if bucket.Spec.ForProvider.BucketTagging == nil {
+		bucket.Spec.ForProvider.BucketTagging = &v1beta1.Tagging{}
+	}
+
+	bucket.Spec.ForProvider.BucketTagging = GenerateLocalTagging(external.TagSet)
```
muvaf, line 145: only if `len(bucket.Spec.ForProvider.BucketTagging.TagSet) == 0 && len(external.TagSet) != 0`

pkg/controller/s3/bucket/versioningConfig.go
```diff
@@ -103,3 +87,31 @@ func (in *VersioningConfigurationClient) CreateOrUpdate(ctx context.Context, buc
...
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	config := bucket.Spec.ForProvider.VersioningConfiguration
+	if config == nil {
+		bucket.Spec.ForProvider.VersioningConfiguration = &v1beta1.VersioningConfiguration{}
+		config = bucket.Spec.ForProvider.VersioningConfiguration
+	}
+	config.Status = awsclient.LateInitializeStringPtr(config.Status, awsclient.String(string(external.Status)))
+	config.MFADelete = awsclient.LateInitializeStringPtr(config.MFADelete, awsclient.String(string(external.MFADelete)))
```
muvaf, line 110:
```suggestion
	if bucket.Spec.ForProvider.VersioningConfiguration == nil {
		bucket.Spec.ForProvider.VersioningConfiguration = &v1beta1.VersioningConfiguration{}
	}
	bucket.Spec.ForProvider.VersioningConfiguration.Status = awsclient.LateInitializeStringPtr(bucket.Spec.ForProvider.VersioningConfiguration.Status, awsclient.String(string(external.Status)))
	bucket.Spec.ForProvider.VersioningConfiguration.MFADelete = awsclient.LateInitializeStringPtr(bucket.Spec.ForProvider.VersioningConfiguration.MFADelete, awsclient.String(string(external.MFADelete)))
```

pkg/controller/s3/bucket/websiteConfig.go
```diff
@@ -149,3 +148,106 @@ func (in *WebsiteConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
 }
+
+// LateInitialize does nothing because the resource might have been deleted by
+// the user.
+func (in *WebsiteConfigurationClient) LateInitialize(ctx context.Context, bucket *v1beta1.Bucket) error {
+	external, err := in.client.GetBucketWebsiteRequest(&awss3.GetBucketWebsiteInput{Bucket: awsclient.String(meta.GetExternalName(bucket))}).Send(ctx)
+	if err != nil {
+		if s3.ReplicationConfigurationNotFound(err) {
+			return nil
+		}
```
muvaf, line 159:
```suggestion
```

pkg/controller/s3/bucket/websiteConfig.go
```diff
@@ -149,3 +148,106 @@ func (in *WebsiteConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
+		config.RedirectAllRequestsTo.Protocol = awsclient.LateInitializeString(
+			config.RedirectAllRequestsTo.Protocol,
+			awsclient.String(string(external.RedirectAllRequestsTo.Protocol)),
+		)
+		config.RedirectAllRequestsTo.HostName = awsclient.LateInitializeString(
+			config.RedirectAllRequestsTo.HostName,
+			external.RedirectAllRequestsTo.HostName,
+		)
+	}
+	if external.RoutingRules != nil {
```
muvaf, line 203: Same as my comments regarding array late-init.

muvaf (changes requested):
Thanks for tackling this @krishchow !

Because of [this issue](https://github.com/crossplane/provider-aws/issues/461), we need to use `awsclients.Wrap` for every error that's returned from AWS API. I have left some other blocking comments as well.

pkg/controller/s3/bucket.go
```diff
@@ -98,43 +98,45 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
...
 
 	return managed.ExternalObservation{
-		ResourceExists:   true,
-		ResourceUpToDate: true,
+		ResourceExists:          true,
+		ResourceUpToDate:        true,
+		ResourceLateInitialized: lateInit,
 		ConnectionDetails: map[string][]byte{
 			xpv1.ResourceCredentialsSecretEndpointKey: []byte(meta.GetExternalName(cr)),
 			ResourceCredentialsSecretRegionKey:        []byte(cr.Spec.ForProvider.LocationConstraint),
```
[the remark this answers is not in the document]
krishchow: Although not highlighted, I think you are referring to ResourceCredentialsSecretRegionKey, which has now been updated

examples/s3/late-init-bucket.yaml
krishchow: The main difference between the two is the inclusion of the external-name annotation. The old bucket probably shows more functionality, and I can remove the late-init file if you think that's a good idea

pkg/controller/s3/bucket.go
krishchow: Hmm, so could one potential change to this could include returning a boolean value from the `LateInitialize` which reflects if some changes have been made? If that condition is met, then we return `ResourceLateInitialized: true`?

pkg/controller/s3/bucket/notificationConfig.go
krishchow: I have changed the function to reflect the other behavior we talked about. Essentially if the local list ever exists (and is non zero length), we return early, otherwise, we actually go through with the late init

pkg/controller/s3/bucket/requestPaymentConfig.go
krishchow: Yup, this is definitely an ugly pattern, I've gone through and fixed it where I've noticed it now. I am not sure exactly what motivated me to write it this way before

pkg/controller/s3/bucket/subresources.go
krishchow: I think it was just useful to add this in so it is available to add more detailed logging in the future as well. This way when debugging these resources going forward, logging statements can be added in easily.

pkg/controller/s3/bucket/taggingConfig.go
krishchow: Is not sufficient to check `fp.BucketTagging == nil || len(fp.BucketTagging.TagSet) != 0` since we check `len(external.TagSet) == 0` above?

muvaf:
@krishchow is this ready for another review pass?

krishchow:
Yes @muvaf, I think it's ready for another review, the main thing left to add is updated testing for the lateinit

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -214,3 +212,153 @@ func sortReplicationRules(rules []awss3.ReplicationRule) {
...
+	}
+
+	createReplicationRulesFromExternal(external.ReplicationConfiguration, fp.ReplicationConfiguration)
+	return nil
+}
+
+func createReplicationRulesFromExternal(external *awss3.ReplicationConfiguration, config *v1beta1.ReplicationConfiguration) { // nolint:gocyclo
+	config.Role = awsclient.LateInitializeStringPtr(config.Role, external.Role)
+	if config.Rules == nil {
+		config.Rules = make([]v1beta1.ReplicationRule, 0)
```
muvaf, line 244:
```suggestion
		config.Rules = make([]v1beta1.ReplicationRule, len(external.Rules))
```

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -214,3 +212,153 @@ func sortReplicationRules(rules []awss3.ReplicationRule) {
...
+	}
+
+	for i, rule := range external.Rules {
+		if i == len(config.Rules) {
+			config.Rules = append(config.Rules, v1beta1.ReplicationRule{})
+		}
```
muvaf, line 250:
```suggestion
```

pkg/controller/s3/bucket/replicationConfig.go
```diff
@@ -214,3 +212,153 @@ func sortReplicationRules(rules []awss3.ReplicationRule) {
...
+		config.Rules[i] = v1beta1.ReplicationRule{
+			ID:       awsclient.LateInitializeStringPtr(config.Rules[i].ID, rule.ID),
+			Priority: awsclient.LateInitializeInt64Ptr(config.Rules[i].Priority, rule.Priority),
+			Status:   awsclient.LateInitializeString(config.Rules[i].Status, awsclient.String(string(rule.Status))),
+		}
+		if rule.Filter != nil {
+			if config.Rules[i].Filter == nil {
```
muvaf, line 257:
```suggestion
```

We don't need nil guards or use of LateInitialize* function usage for `config.Rules[i]` objects since we initialize them in the beginning of the function, so it's safe to assign directly.

pkg/controller/s3/bucket/sseConfig.go
```diff
@@ -121,3 +135,34 @@ func (in *SSEConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.Bu
...
+	}
+
+	// We need the second check here because by default the SSE is not set
+	if external.GetBucketEncryptionOutput == nil || external.ServerSideEncryptionConfiguration == nil || len(external.ServerSideEncryptionConfiguration.Rules) == 0 {
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+	fp := &bucket.Spec.ForProvider
+	if fp.ServerSideEncryptionConfiguration == nil || len(fp.ServerSideEncryptionConfiguration.Rules) == 0 {
```
muvaf, line 154:
```suggestion
	if fp.ServerSideEncryptionConfiguration == nil {
```

pkg/controller/s3/bucket/sseConfig.go
```diff
@@ -121,3 +135,34 @@ func (in *SSEConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.Bu
...
+		fp.ServerSideEncryptionConfiguration = &v1beta1.ServerSideEncryptionConfiguration{}
+	} else {
+		return nil
+	}
+
+	fp.ServerSideEncryptionConfiguration.Rules = GenerateLocalBucketEncryption(external.ServerSideEncryptionConfiguration)
```
muvaf, line 160: We can do this only if `len(fp.ServerSideEncryptionConfiguration.Rules) == 0`

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -110,3 +118,35 @@ func (in *TaggingConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()))
+
+	// We need the second check here because by default the tags are not set
+	if external.GetBucketTaggingOutput == nil || len(external.TagSet) == 0 {
+		return nil
+	}
+
+	fp := &bucket.Spec.ForProvider
+	if fp.BucketTagging == nil || len(fp.BucketTagging.TagSet) != 0 {
```
muvaf, line 138:
```suggestion
	if fp.BucketTagging == nil {
```

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -74,13 +72,23 @@ func (in *TaggingConfigurationClient) Observe(ctx context.Context, bucket *v1bet
...
+}
+
+// GenerateLocalTagging creates the v1beta1.Tagging from the AWS SDK tagging
+func GenerateLocalTagging(config []awss3.Tag) *v1beta1.Tagging {
+	if len(config) == 0 {
+		if config == nil {
+			return nil
+		}
+		return &v1beta1.Tagging{TagSet: make([]v1beta1.Tag, 0)}
+	}
```
muvaf, line 90:
```suggestion
	if config == nil {
			return nil
		}
	if len(config) == 0 {
		return &v1beta1.Tagging{TagSet: make([]v1beta1.Tag, 0)}
	}
```

pkg/controller/s3/bucket/taggingConfig.go
```diff
@@ -110,3 +118,35 @@ func (in *TaggingConfigurationClient) Delete(ctx context.Context, bucket *v1beta
...
+
+	fp := &bucket.Spec.ForProvider
+	if fp.BucketTagging == nil || len(fp.BucketTagging.TagSet) != 0 {
+		fp.BucketTagging = &v1beta1.Tagging{}
+	} else {
+		return nil
+	}
+
+	fp.BucketTagging = GenerateLocalTagging(external.TagSet)
```
muvaf, line 144: We can do this only if `len(fp.BucketTagging.TagSet) == 0`

pkg/controller/s3/bucket/CORSConfig.go
```diff
@@ -131,3 +143,34 @@ func (in *CORSConfigurationClient) Delete(ctx context.Context, bucket *v1beta1.B
...
+	if err != nil {
+		return awsclient.Wrap(resource.Ignore(s3.CORSConfigurationNotFound, err), corsGetFailed)
+	}
+
+	// We need the second check here because by default the CORS is not set
+	if external.GetBucketCorsOutput == nil || len(external.CORSRules) == 0 {
+		return nil
+	}
+
+	in.logger.Debug(fmt.Sprintf("called LateInitialize for %s", reflect.TypeOf(in).Elem().Name()), "external output", external.GetBucketCorsOutput)
```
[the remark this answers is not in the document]
krishchow: I added the log statement here because I thought late-inits happen infrequently enough that it would not produce an excessive number of log statements. Also, since it's a debug log there shouldn't be any effect for (the average) end-user, right?

pkg/controller/s3/bucket/lifecycleConfig.go
```diff
@@ -193,3 +190,131 @@ func sortFilterTags(rules []awss3.LifecycleRule) {
...
+		return nil
+	}
+
+	createLifecycleRulesFromExternal(external.Rules, fp.LifecycleConfiguration)
+
+	return nil
+}
+
+func createLifecycleRulesFromExternal(external []awss3.LifecycleRule, config *v1beta1.BucketLifecycleConfiguration) { // nolint:gocyclo
+	config.Rules = make([]v1beta1.LifecycleRule, len(external))
```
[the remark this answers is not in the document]
krishchow: Makes sense!

pkg/controller/s3/bucket/taggingConfig.go
krishchow: In general, it would be better to check if a list is `nil` as opposed to `len() == 0`, since in some cases the user might intend the value to be an empty list, right?

muvaf (approved):
Thanks @krishchow for getting this to the finish line! Could you remove loggers from subresource client structs since they are not used? After that, it should be ready to be merged.

PR #538, opened 2021-02-12
Update API Gateway controller to new codegen structure

muvaf:

In order to get the latest features, we need to update all ACK-generated controllers to new hook functions.

The first commit is only generation, the second one is manually done and that's the part that needs to be reviewed.

Fixes #

I have:

Tested `API` resource just to be sure but didn't go through all of them since there isn't much functionality change.


pkg/controller/apigatewayv2/api/setup.go
```diff
@@ -36,30 +36,36 @@ import (
...
+func preObserve(_ context.Context, cr *svcapitypes.API, obj *svcsdk.GetApiInput) error {
+	obj.ApiId = aws.String(meta.GetExternalName(cr))
 	return nil
 }
-func (*external) postObserve(_ context.Context, cr *svcapitypes.API, _ *svcsdk.GetApiOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+
+func postObserve(_ context.Context, cr *svcapitypes.API, _ *svcsdk.GetApiOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
 	cr.SetConditions(xpv1.Available())
 	return obs, err
 }
```
hasheddan, line 66: Do we need to check error before setting `Available` here?

hasheddan: It does seem like in practice we are always passing `nil` error here, but I see it is checked in other places, so not sure what pattern we want to follow

hasheddan (commented):
Looks pretty good to me @muvaf! Awesome work, per usual. I did leave one comment. Overall, I like the model of generating the no-op's then overriding with options rather than having to hand write all of the methods that just return the `obj` without any operations. Have you had the opportunity to do manual testing here?

pkg/controller/apigatewayv2/api/setup.go
muvaf: Yeah, good catch!

muvaf:
@hasheddan I just updated the testing section, I tested the `API` resource for sanity check but not all of them.

hasheddan (approved):

PR #544, opened 2021-02-15
Use rate limiting in all controllers

hasheddan:

Updates all controllers to use a shared provider rate limiter alongside
a per item exponential backoff limiter to improve requeue behavior.

I have:

Tested this out with a misconfigured `Bucket` and observed proper backoff:

s3prom

### DIFF
--- a/pkg/controller/acmpca/certificateauthority/controller.go
+++ b/pkg/controller/acmpca/certificateauthority/controller.go
@@ -23,13 +23,16 @@ import (
 	awsacmpca "github.com/aws/aws-sdk-go-v2/service/acmpca"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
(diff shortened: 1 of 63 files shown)


pkg/controller/acmpca/certificateauthority/controller.go
```diff
@@ -54,11 +57,14 @@ const (
 )
 
 // SetupCertificateAuthority adds a controller that reconciles ACMPCA.
-func SetupCertificateAuthority(mgr ctrl.Manager, l logging.Logger) error {
+func SetupCertificateAuthority(mgr ctrl.Manager, l logging.Logger, limiter workqueue.RateLimiter) error {
```
negz, line 60:
```suggestion
func SetupCertificateAuthority(mgr ctrl.Manager, l logging.Logger, global workqueue.RateLimiter) error {
```

I wonder if something like this that indicates what the rate limiter we're being passed would be worth it? Perhaps not - we don't do that for the `logging.Logger`. If we're going to stick with something generic I'd prefer something shorter - e.g. `rl workqueue.RateLimiter`. `limiter` seems redundant given it's always followed by the type, and given that the variable doesn't live very long lines-from-definition wise.

negz: I also wonder whether we're getting close to the point that we might want to change the `Setup` signatures (again) to accomodate more options like this - e.g. `SetupFoo(ctrl.Manager, setup.Option...)`. No need to bring that in scope here though.

negz (approved):

pkg/controller/acmpca/certificateauthority/controller.go
hasheddan: `rl` sounds good to me 👍

PR #545, opened 2021-02-15
secretsmanager.secret: migrated to generated code

muvaf:

Migrated to code generation approach. However, this API has quite a lot of quirks and some changes were necessary in code-generator. Hence this is a custom build I have [here](https://github.com/muvaf/code-generator/tree/latest) built with my open PRs.

Fixes https://github.com/crossplane/provider-aws/issues/570

I have:

Manually.


examples/secretsmanager/secret.yaml
```diff
@@ -1,12 +1,13 @@
...
-  name: example-secret
+  name: example-secret-3
 spec:
   forProvider:
     region: us-east-1
-    description: "secrets of test cluster created with writeConnectionSecretToRef"
-    kmsKeyIdRef: example-key-id
+    description: "test"
+#    kmsKeyIDRef:
+#      name: example-key-id
```
hasheddan, line 10: looks like we haven't included this in a release yet so no need for API version bump :+1:

hasheddan (approved):
LGTM

MisterMX:
@muvaf 
The previous implementation of the controller was able to store a whole secret with multiple key-value-pairs in AWS. The `key` field in `secretRef` was optional, now it is mandatory.

Additionally, the decoded base64 value was stored in AWS using `SecretString` instead of `SecretBinary` as it is done now.

Are both of these intended changes?

muvaf:
> The previous implementation of the controller was able to store a whole secret with multiple key-value-pairs in AWS.

I'm planning to implement this in a separate field. So users will choose to use whichever they want.

> Additionally, the decoded base64 value was stored in AWS using SecretString instead of SecretBinary as it is done now.

Does using one or the other have an impact on the functionality?

muvaf:
@MisterMX I added a new field called `PayloadType` to allow users to specify which part of the Kubernetes Secret should be used. The options are `SingleValue` for value of a single key and `Map` to include the whole map as JSON string map. Does that cover your use cases?

MisterMX:
> @MisterMX I added a new field called `PayloadType` to allow users to specify which part of the Kubernetes Secret should be used. The options are `SingleValue` for value of a single key and `Map` to include the whole map as JSON string map. Does that cover your use cases?

What speaks against a `key` field in `secretSelector` as it was done before? To me it would make more sense to have `name`, `namespace` and `key` in one group since they are used together to identify the value of the secret that should be stored in AWS.

Also, I don't think `PayloadType` is the right name for the variable considering what it does. It implies that it specifies the *type* of the value that should be shipped (so only either binary or string) and not where to look for the value.

> > Additionally, the decoded base64 value was stored in AWS using SecretString instead of SecretBinary as it is done now.

> Does using one or the other have an impact on the functionality?

We also use [kubernetes-external-secrets](https://github.com/external-secrets/kubernetes-external-secrets) in our project where in some cases you have to explicitly specify that a secret is a binary value. So this might cause some conflicts.

Do you think it is possible to add a config option, say `secretType`, where the user can tell the secretsmanager to either use `secretString` or `secretBinary` with a default fallback to the latter?

muvaf:
> What speaks against a key field in secretSelector as it was done before? To me it would make more sense to have name, namespace and key in one group since they are used together to identify the value of the secret that should be stored in AWS.

This is still the case, see the type of [`SingleValueSecretRef`](https://github.com/crossplane/provider-aws/pull/545/files#diff-89f02592bca9e2fe62869c6ae35f2c8b79d13ca7f5465b40b677c16c3ac06e92R33). The only difference is that `key` is required in that struct.

> Also, I don't think PayloadType is the right name for the variable considering what it does. It implies that it specifies the type of the value that should be shipped (so only either binary or string) and not where to look for the value.

> Do you think it is possible to add a config option, say secretType, where the user can tell the secretsmanager to either use secretString or secretBinary with a default fallback to the latter?

Yeah we can use `PayloadType` with `String|Binary(default)` options and then have a separate `InputSource` with `SingleValue|Map` options. I'm wary of naming it `SecretType` since it can be confused with secret types in Kubernetes.

muvaf:
@hasheddan could you take another look for the last commit?

apis/secretsmanager/v1alpha1/custom_types.go
```diff
@@ -20,10 +20,21 @@ import xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
...
-	// SecretRef is used to reference the secret and the key whose value you'd
-	// like to be stored in AWS.
-	// The value is stored as is with no change.
-	SecretRef xpv1.SecretKeySelector `json:"secretRef"`
+	// PayloadType lets you choose which part of the input secret should be used.
+	// SingleValue: Only value of the given key.
+	// Map: The whole content of the input secret will be used after converting
+	// it to JSON map.
+	// +kubebuilder:validation:Enum=SingleValue;Map
+	PayloadType string `json:"payloadType"`
```
hasheddan, line 28: @muvaf instead of using the `crossplane-runtime` `SecretKeySelector`, can we just use a custom implementation where the `key` field is optional and if it is not provided then we use the whole secret map?

MisterMX:
> Yeah we can use `PayloadType` with `String|Binary(default)` options and then have a separate `InputSource` with `SingleValue|Map` options. I'm wary of naming it `SecretType` since it can be confused with secret types in Kubernetes.

What about `secretDataType`, `secretStoreType` or `secretStoreFormat`? I think `payloadType` is also misleading as the term `payload` is not used anywhere else in the yaml.

> This is still the case, see the type of [`SingleValueSecretRef`](https://github.com/crossplane/provider-aws/pull/545/files#diff-89f02592bca9e2fe62869c6ae35f2c8b79d13ca7f5465b40b677c16c3ac06e92R33). The only difference is that `key` is required in that struct.

This might be my personal taste but a single type with an optional `key` field looks cleaner to me, having a yaml like this:

```yaml
secretRef:
      name: my-cluster
      namespace: my-namespace
      key: example-key               # optional
```

I also would go with a custom type for `SecretKeySelector` as suggested by @hasheddan.

apis/secretsmanager/v1alpha1/custom_types.go
muvaf: That would be closer to the initial implementation where `key` was optional but in case `key` wasn't given and there was only one key, the value of that one was used. If there were more, only then a new JSON map would be constructed.

Anyway, the reason I ended up with a discriminator is that the shape of the content changes. For example, for the following secret:
```yaml
k8skey2:  value1
k8skey:   value2
```
`SingleKey` pointing to `k8skey` would send only `value2` (just like other secret inputs) but if you choose to send the whole secret a JSON map that includes the keys are sent (I'm describing this because I was confused initially, thinking in all cases we send a map with keys). The latter is also a valid use case because AWS Secret Manager support both key-value pairs as well as simple raw strings with the same mechanism.

So I can go with your suggestion, just wanted to make sure you know this is the case.

MisterMX:
@muvaf 
Found a possible bug in your implementation:

If the `kmsKeyIDRef` of an existing secret is changed, the controller log `External resource is up to date` and the respective entry in AWS secretmanager is not updated (`KmsKeyId` is not changed).

Used yaml:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
type: Opaque
data:
  test1: SGVsbG8gV29ybGQ=
  test2: SGVsbG8gYWdhaW4h
---
apiVersion: kms.aws.crossplane.io/v1alpha1
kind: Key
metadata:
  name: "test-key-1"
  annotations:
    crossplane.io/external-name: <kms-key-id>
  labels:
    example: "true"
spec:
  deletionPolicy: Orphan
  forProvider:
    region: eu-central-1
  providerConfigRef:
    name: <provider-config>
---
apiVersion: kms.aws.crossplane.io/v1alpha1
kind: Key
metadata:
  name: "test-key-2"
  annotations:
    crossplane.io/external-name: <kms-key-id>
  labels:
    example: "true"
spec:
  deletionPolicy: Orphan
  forProvider:
    region: eu-central-1
  providerConfigRef:
    name: <provider-config>
---
apiVersion: secretsmanager.aws.crossplane.io/v1alpha1
kind: Secret
metadata:
  name: test-secret
spec:
  forProvider:
    region: eu-central-1
    description: "Test secret"
    kmsKeyIDRef:
      name: test-key-1
      # name: test-key-2
    forceDeleteWithoutRecovery: true
    payloadType: Map
    mapSecretRef:
      name: test-secret
      namespace: default
    tags:
      - key: cluster
        value: "my-cluster"
  providerConfigRef:
    name: <provider-config>
```

MisterMX:
Had another look into this. The reason is the resolver in `apis/secretsmanager/v1alpha1/referencers.go` which does not look for an updated `key` reference if `CurrentValue` is set to an non-empty string (see `IsNoOp` function in `crossplane-runtime`).

```golang
        rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
		CurrentValue: reference.FromPtrValue(mg.Spec.ForProvider.KMSKeyID),
		Reference:    mg.Spec.ForProvider.KMSKeyIDRef,
		Selector:     mg.Spec.ForProvider.KMSKeyIDSelector,
		To:           reference.To{Managed: &kms.Key{}, List: &kms.KeyList{}},
		Extract:      reference.ExternalName(),
	})
```

I tried setting `CurrentValue: ""` and the KMS ID is updated in AWS successfully.

So this line should be replaced.

muvaf:
Just pushed a commit that implements the suggestion from @MisterMX and @hasheddan . `Key` is now optional and it decides whether the full map is used or not. The difference from the earlier implementation is that the author of the CR decides whether to use the full map or not instead of relying on how many keys exist in the source Kubernetes Secret.

@MisterMX This is an [expected behavior](https://github.com/crossplane/crossplane-runtime/blob/master/pkg/reference/reference.go#L116) of reference resolver. Here is an issue to track the discussion about it https://github.com/crossplane/crossplane-runtime/issues

apis/secretsmanager/v1alpha1/custom_types.go
hasheddan: Thanks for explaining this @muvaf :) If someone did want to send a single kv pair could they do something like the following with a `key` of `k8skey`?

```
k8skey: "smkey: smvalue"
```

apis/secretsmanager/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,68 @@
...
+
+// A SecretReference is a reference to a secret in an arbitrary namespace.
+type SecretReference struct {
+	// Name of the secret.
+	Name string `json:"name"`
+
+	// Namespace of the secret.
+	Namespace string `json:"namespace"`
+
+	// Key whose value will be used. If not give, the whole map in the Secret
```
hasheddan, line 65:
```suggestion
	// Key whose value will be used. If not given, the whole map in the Secret
```

apis/secretsmanager/v1alpha1/custom_types.go
muvaf: Yes, they are free to put anything they want there, including JSON map.

muvaf:
Manually tested the latest changes.

PR #557, opened 2021-02-22
bucket: versioning + replication configuration sub resource ordering fix

### LINKED ISSUE #553: S3 bucket replication rules can't be applied when the bucket is being created
### What happened?

When we apply the following yaml to create a S3 bucket with replication rules, we found that the replication rules were not applied.  We got the following error when describing the "Bucket"

```
Warning  CannotUpdateExternalResource  4m1s (x32175 over 98m)  managed/bucket.s3.aws.crossplane.io  (combined from similar events): cannot create or update: cannot put Bucket replication: InvalidRequest: Versioning must be 'Enabled' on the bucket to apply a replication configuration
       status code: 400, request id: BDA631A781251CFF, host id: 7U2kQhRheWko/QpJjpabk9hV0UuAsJ/G+W9QdWmWMGU+R5SOlBEiu6x92zDCVnSoMH6SnzWYT2I=
```

However, we do have the "versioning" set to "Enabled" in the yaml file (See below). 

In addition, if we do this in two steps: 
first, create a bucket without replication rules.  
Second, we apply the yaml with replication rules and it did succeed. 

### How can we reproduce it?

Apply the following yaml

```
kind: Bucket
metadata:
  name: xxxx-claimed-crossplane-test-yl
  namespace: warrior
  annotations:
    crossplane.io/external-name: xxxx-claimed-crossplane-test-yl
spec:
  forProvider:
    acl: private
    locationConstraint: us-west-2
    versioningConfiguration:
      status: Enabled
    replicationConfiguration:
      role: "arn:aws:iam::00000000:role/crossplane-s3-test-xxxx"
      rules:
        - destination:
            bucket: arn:aws:s3:::test-bucket-xxxxx-crossplane-yl
            storageClass: "STANDARD_IA" 
          status: Enabled
          priority: 10
          deleteMarkerReplication:
            Status: Disabled
          id: "test"
  providerConfigRef:
    name: aws
```

Crossplane version: 
crossplane/provider-aws-controller:v0.16.0

krishchow:

Fixes #553 

I have:

- [] Run `make reviewable test` to ensure this PR is ready for review.

Previously the reconciliation would be blocked since the versioning could never complete because of a replication error based on the fact that there is no versioning. This is resolved now by fixing the ordering in the `subresource.go` folder.

I did manual testing with two S3 buckets, I can provide the gists if that would be helpful. But essentially, the first was a regular S3 bucket, and the second was an S3 bucket with versioning enabled, as well as a replication configuration pointing to the first bucket. There was a role that had access to both buckets, set up on the replication configuration as well. 

Previously, the 2nd bucket would never finish reconciling, but that is resolved now.

### DIFF
--- a/pkg/controller/s3/bucket/subresources.go
+++ b/pkg/controller/s3/bucket/subresources.go
@@ -34,6 +34,9 @@ type SubresourceClient interface {
 // NewSubresourceClients creates the array of all clients for a given BucketProvider
 func NewSubresourceClients(client s3.BucketClient) []SubresourceClient {
 	return []SubresourceClient{
+		// Note: Moved VersioningClient up, since ReplicationConfiguration may be blocked
+		// by an invalid VersioningConfig, see https://github.com/crossplane/provider-aws/issues/553
+		NewVersioningConfigurationClient(client),
 		NewAccelerateConfigurationClient(client),
 		NewCORSConfigurationClient(client),
 		NewLifecycleConfigurationClient(client),
@@ -43,7 +46,6 @@ func NewSubresourceClients(client s3.BucketClient) []SubresourceClient {
 		NewRequestPaymentConfigurationClient(client),
 		NewSSEConfigurationClient(client),
 		NewTaggingConfigurationClient(client),
-		NewVersioningConfigurationClient(client),
 		NewWebsiteConfigurationClient(client),
 	}
 }


krishchow:
cc @yshu2017

muvaf (approved):
Thanks @krishchow !

PR #487, opened 2021-01-08
Lambda Function generated by ACK

muvaf:
### Description of your changes

Opening this so that @duncaan and @smcavallo can complete the missing custom parts for Lambda Function.

Here is the draft guide to follow: https://github.com/crossplane/provider-aws/pull/486/

Fixes #

I have:

- [ ] Read and followed Crossplane's [contribution process].
- [ ] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

Issue #173, opened 2020-03-26
Add Lambda support


infinitecompute:
### What problem are you facing?
Would like add a AWS Lambda support.  

### How could Crossplane help solve your problem?
Add Lambda as a managed resource.  It is unclear how this would be supported within Crossplane without design input.

### References
*

jbw976:
This was fixed by #559

PR #559, opened 2021-02-23
lambda.function: code generated and custom parts are implemented 

smcavallo:

Replaces PR #
https://github.com/crossplane/provider-aws/pull/487
Fixes #
https://github.com/crossplane/provider-aws/issues/173

I have:

* Unit tests
* Run in local crossplane using many different configurations
  - ensured resources were created/updated accordingly
  - ensured resources were updated accordingly
  - ensured resources did not continually update


pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,500 @@
+package function
+
+import (
```
hasheddan, line 3: @smcavallo order here goes:
```
<stdlib pkgs>

<external pkgs (anything not in github.com/crossplane)>

<crossplane org pkgs>

<local to this repo pkgs>
```

apis/lambda/v1alpha1/zz_types.go
```diff
@@ -0,0 +1,202 @@
...
+type FileSystemConfig struct {
+	ARN *string `json:"arn,omitempty"`
+
+	LocalMountPath *string `json:"localMountPath,omitempty"`
+}
+
+type FunctionCode struct {
+	ImageURI *string `json:"imageURI,omitempty"`
+
+	S3Bucket *string `json:"s3Bucket,omitempty"`
```
muvaf, line 89: Seems like this field could have referencer to `Bucket` object.

apis/lambda/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,109 @@
...
+	KMSKeyARNRef *xpv1.Reference `json:"kmsKeyARNRef,omitempty"`
+
+	// KMSKeyARNSelector selects references to kms key arn used
+	// to set the KMSKeyARN.
+	// +optional
+	KMSKeyARNSelector *xpv1.Selector `json:"kmsKeyARNSelector,omitempty"`
+
+	// Role is the arn of the iam role.
+	// +immutable
+	Role *string `json:"role,omitempty"`
```
muvaf, line 69:
```suggestion
```

We can use the generated one in the reference resolver.

apis/lambda/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,109 @@
...
+	SubnetIDRefs []xpv1.Reference `json:"vpcConfig.subnetIdRefs,omitempty"`
+
+	// SubnetIDsSelector selects references to Subnets used
+	// to set the SubnetIDs.
+	// +optional
+	SubnetIDSelector *xpv1.Selector `json:"vpcConfig.subnetIdSelector,omitempty"`
+
+	// KMSKeyARN is the arn of the KMS Key.
+	// +immutable
+	KMSKeyARN *string `json:"kmsKeyARN,omitempty"`
```
muvaf, line 55:
```suggestion
	// KMSKeyARN is the arn of the KMS Key.
	// +immutable
	KMSKeyARN *string `json:"kmsKeyARN,omitempty"`
```

We can use the generated one in the reference resolver.

examples/lambda/function.yaml
```diff
@@ -0,0 +1,26 @@
...
+  forProvider:
+    code:
+      s3Bucket: "my-s3-bucket"
+      s3Key: "prefix/path_to_key.zip"
+    environment:
+      variables:
+        envKey: envValue
+    functionName: "test-function"
+    handler: "index.handler"
+    role: arn:aws:iam::123456789012:role/test-function-role
```
muvaf, line 15:
```suggestion
    roleRef:
      name: somerole
```

We try to use references to other example YAMLs.

examples/lambda/function.yaml
```diff
@@ -0,0 +1,26 @@
...
+    runtime: nodejs12.x
+    region: us-east-1
+    vpcConfig:
+      securityGroupIDs:
+        - sg-id
+      subnetIDs:
+        - subnet-id
```
muvaf, line 22:
```suggestion
      securityGroupIDRefs:
        - name: sample-cluster-sg
      subnetIDRefs:
        - name: sample-subnet1
```

We try to use references to other example YAMLs.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+}
+
+func postCreate(_ context.Context, cr *svcapitypes.Function, resp *svcsdk.FunctionConfiguration, cre managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	meta.SetExternalName(cr, aws.StringValue(resp.FunctionName))
+	cre.ExternalNameAssigned = true
+	return cre, nil
+}
```
muvaf, line 85:
```suggestion
func preCreate(_ context.Context, cr *svcapitypes.Function, obj *svcsdk.CreateFunctionInput) error {
	obj.FunctionName = aws.String(meta.GetExternalName(cr))
	return nil
}
```

`FunctionName` is used as identifier in all calls and not generated by AWS. So, we should use external name annotation as the source for it. In order for this to work, you need to ignore `FunctionName` in all calls in `generator-config.yaml`.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	return false, nil
+}
+
+// nolint:gocyclo
+func isUpToDate(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) (bool, error) {
+
+	// Compare CODE
+	if !isUpToDateCode(cr, obj) {
+		return false, nil
+	}
```
muvaf, line 98:
```suggestion
```

We can ignore no-op checks and leave a note here about why they are ignored.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// Compare CONFIGURATION
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Description), aws.StringValue(obj.Configuration.Description)) != 0 {
```
muvaf, line 101:
```suggestion
	if aws.StringValue(cr.Spec.ForProvider.Description) != aws.StringValue(obj.Configuration.Description) {
```

Does `strings.Compare` have an advantage over equality check in this context? Seems like it does an equality check and then another check for lexicographically order, which is not necessary in this context.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	return len(addTags) == 0 && len(removeTags) == 0, nil
+
+}
+
+// isUpToDateEnvironment checks if FunctionConfiguration EnvironmentResponse Variables are up to date
+func isUpToDateEnvironment(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// Handle nil pointer refs
+	envVars := map[string]*string{}
+	awsVars := map[string]*string{}
+	if !reflect.ValueOf(cr.Spec.ForProvider.Environment).IsZero() &&
```
muvaf, line 164: Can we check for length equality and then iterate over one of the maps and return in case there is any mismatch with the other? `reflect` incurs some performance cost and unpredictability in the code in general.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	if len(envVars) > 0 &&
+		!reflect.DeepEqual(envVars, awsVars) {
+		return false
+	}
+	return true
+}
+
+func isUpToDateFileSystemConfigs(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// Handle nil pointer refs
+	var fileSystemConfigs []*svcsdk.FileSystemConfig
```
muvaf, line 186: What do you think about doing a length check first, and then sort both arrays by ARN and then compare if the fields are equal for each pair?

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Role), aws.StringValue(obj.Configuration.Role)) != 0 {
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Runtime), aws.StringValue(obj.Configuration.Runtime)) != 0 {
+		return false, nil
+	}
```
muvaf, line 139: Same comment as line 101.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Handler), aws.StringValue(obj.Configuration.Handler)) != 0 {
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.KMSKeyARN), aws.StringValue(obj.Configuration.KMSKeyArn)) != 0 {
+		return false, nil
+	}
+
+	// The function's layers (https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html).
+	if !isUpToDateLayers(cr, obj) {
```
muvaf, line 123: Same comment as line 95.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Handler), aws.StringValue(obj.Configuration.Handler)) != 0 {
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.KMSKeyARN), aws.StringValue(obj.Configuration.KMSKeyArn)) != 0 {
+		return false, nil
+	}
```
muvaf, line 120: Same comment as line 101.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// The function's layers (https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html).
+	if !isUpToDateLayers(cr, obj) {
+		return false, nil
+	}
+
+	// set default
+	if aws.Int64Value(cr.Spec.ForProvider.MemorySize) != 0 &&
```
muvaf, line 128: Why do we need to check for `0` specifically?

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Role), aws.StringValue(obj.Configuration.Role)) != 0 {
+		return false, nil
+	}
+
+	if strings.Compare(aws.StringValue(cr.Spec.ForProvider.Runtime), aws.StringValue(obj.Configuration.Runtime)) != 0 {
+		return false, nil
+	}
+	// Set default
+	if aws.Int64Value(cr.Spec.ForProvider.Timeout) != 0 &&
```
muvaf, line 141: Same comment as line 128.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+}
+
+func isUpToDateTracingConfig(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// Handle nil pointer refs
+	mode := svcsdk.TracingModePassThrough
+	if !reflect.ValueOf(cr.Spec.ForProvider.TracingConfig).IsZero() &&
+		!reflect.ValueOf(cr.Spec.ForProvider.TracingConfig.Mode).IsZero() {
+		mode = *cr.Spec.ForProvider.TracingConfig.Mode
+	}
+	if !reflect.DeepEqual(mode, *obj.Configuration.TracingConfig.Mode) {
```
muvaf, line 236: Let's go with a simple string check using `aws.StringValue`.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+func isUpToDateLayers(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// The generator is creating layers with type of
+	// 	Layers []*string `json:"layers,omitempty"`
+	// Instead of []*svcsdk.Layer
+	// We can't properly implement a comparison until that is fixed
+	return true
+}
+
+func isUpToDateTracingConfig(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// Handle nil pointer refs
```
muvaf, line 230: You can make sure both pointers are non-nil before doing the string comparison. If any one or both of them are nil, then it can return `false`.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+	if !reflect.DeepEqual(mode, *obj.Configuration.TracingConfig.Mode) {
+		return false
+	}
+	return true
+}
+
+func isUpToDateSecurityGroupIDs(cr *svcapitypes.Function, obj *svcsdk.GetFunctionOutput) bool {
+	// Handle nil pointer refs
+	var securityGroupIDs []*string
```
muvaf, line 244: Same comment as line 186.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+	return true
+}
+
+// returns which AWS Tags exist in the resource tags and which are outdated and should be removed
+func diffTags(spec map[string]*string, current map[string]*string) (map[string]*string, []*string) {
+	addMap := make(map[string]*string, len(spec))
+	removeTags := make([]*string, 0)
+	for k, v := range current {
+		if reflect.DeepEqual(spec[k], v) {
```
muvaf, line 271:
```suggestion
		if aws.StringValue(spec[k]) == aws.StringValue(v) {
```

We don't need to use `reflect` for simple string comparisons.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+			continue
+		}
+		removeTags = append(removeTags, aws.String(k))
+	}
+	for k, v := range spec {
+		if reflect.DeepEqual(current[k], v) {
```
muvaf, line 277: Same comment as line 271.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,499 @@
...
+func (u *updater) update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+	cr, ok := mg.(*svcapitypes.Function)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	// https://docs.aws.amazon.com/sdk-for-go/api/service/lambda/#Lambda.UpdateFunctionCode
+	updateFunctionCodeInput := GenerateUpdateFunctionCodeInput(cr)
+	if _, err := u.client.UpdateFunctionCodeWithContext(ctx, updateFunctionCodeInput); err != nil {
+		return managed.ExternalUpdate{}, errors.Wrap(err, errUpdate)
```
muvaf, line 298:
```suggestion
		return managed.ExternalUpdate{}, aws.Wrap(err, errUpdate)
```

All errors coming from AWS should be wrapped by our custom wrapper.

muvaf (commented):
Thank you @smcavallo ! This is a pretty good start!

pkg/controller/lambda/function/setup.go
smcavallo: Because of  `omitempty` - security groups will initialize as nil instead of an empty list if they aren't set.
The current behavior treats nil as empty list and will remove security groups in aws if they they aren't defined in crossplane.
Otherwise it would always end up returning false and be in a constant update loop.  Is there a better way to initialize nil to empty value?  Is there a better way to handle this?

apis/lambda/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,109 @@
...
+	// to set the lambda Role.
+	// +optional
+	RoleSelector *xpv1.Selector `json:"roleSelector,omitempty"`
+}
+
+// CustomFunctionVPCConfigParameters includes custom fields for FunctionVPCConfigParameters.
+type CustomFunctionVPCConfigParameters struct {
+
+	// SecurityGroupIDs is the list of IDs for the SecurityGroups.
+	SecurityGroupIDs []string `json:"vpcConfig.securityGroupIDs,omitempty"`
```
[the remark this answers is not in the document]
smcavallo: @muvaf - mind taking a look to make sure I did the right thing here?

apis/lambda/v1alpha1/zz_types.go
smcavallo: Unfortunately the `FunctionCodeLocation` property returned as part of `GetFunctionOutput` does not map in any way to this `FunctionCode` object.   The properties on this object are used when creating the lambda function and can't be retrieved, although you can make an update request to change them.
https://docs.aws.amazon.com/sdk-for-go/api/service/lambda/#FunctionCodeLocation

apis/lambda/v1alpha1/referencers.go
```diff
@@ -0,0 +1,111 @@
...
+		To:           reference.To{Managed: &iamv1beta1.IAMRole{}, List: &iamv1beta1.IAMRoleList{}},
+		Extract:      iamv1beta1.IAMRoleARN(),
+	})
+	if err != nil {
+		return errors.Wrap(err, "spec.forProvider.role")
+	}
+	mg.Spec.ForProvider.Role = reference.ToPtrValue(rsp.ResolvedValue)
+	mg.Spec.ForProvider.RoleRef = rsp.ResolvedReference
+
+	// Resolve spec.forProvider.VPCConfig.subnetIds
```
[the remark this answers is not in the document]
smcavallo: In the referencer I initialized a new config instead - if this is a better approach vs nil handling?
```
	// Initialize a new VPCConfig
	if mg.Spec.ForProvider.VPCConfig == nil {
		mg.Spec.ForProvider.VPCConfig = &VPCConfig{}
	}
```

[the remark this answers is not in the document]
muvaf: `nil` handling would be better because we should make changes in spec only if necessary.

pkg/controller/lambda/function/setup.go
muvaf: `len` returns `0` if the array is `nil`, so it's usually a safer approach to check for length of arrays/maps instead of a nil check since it covers both empty and nil cases.

smcavallo: makes sense.  Since I had to do this several places I use the method you mention in `compareSliceStringStringPointer`

pkg/controller/lambda/function/setup.go
smcavallo: yes - it has been updated for better comparison

pkg/controller/lambda/function/setup.go
smcavallo: What you suggested is the best option.  I have made the change, thanks again.

smcavallo:
@muvaf - ready for re-review

apis/lambda/v1alpha1/referencers.go
```diff
@@ -0,0 +1,95 @@
...
+		// Resolve spec.forProvider.VPCConfig.securityGroupIds
+		mrsp, err = r.ResolveMultiple(ctx, reference.MultiResolutionRequest{
+			CurrentValues: reference.FromPtrValues(mg.Spec.ForProvider.VPCConfig.SecurityGroupIDs),
+			References:    mg.Spec.ForProvider.CustomFunctionParameters.CustomFunctionVPCConfigParameters.SecurityGroupIDRefs,
+			Selector:      mg.Spec.ForProvider.CustomFunctionParameters.CustomFunctionVPCConfigParameters.SecurityGroupIDSelector,
+			To:            reference.To{Managed: &ec2.SecurityGroup{}, List: &ec2.SecurityGroupList{}},
+			Extract:       reference.ExternalName(),
+		})
+		if err != nil {
+			return errors.Wrap(err, "spec.forProvider.VPCConfig.securityGroupIds")
```
muvaf, line 74:
```suggestion
			return errors.Wrap(err, "spec.forProvider.vpcConfig.securityGroupIds")
```

apis/lambda/v1alpha1/referencers.go
```diff
@@ -0,0 +1,95 @@
...
+		// Resolve spec.forProvider.VPCConfig.subnetIds
+		mrsp, err := r.ResolveMultiple(ctx, reference.MultiResolutionRequest{
+			CurrentValues: reference.FromPtrValues(mg.Spec.ForProvider.VPCConfig.SubnetIDs),
+			References:    mg.Spec.ForProvider.CustomFunctionParameters.CustomFunctionVPCConfigParameters.SubnetIDRefs,
+			Selector:      mg.Spec.ForProvider.CustomFunctionParameters.CustomFunctionVPCConfigParameters.SubnetIDSelector,
+			To:            reference.To{Managed: &ec2.Subnet{}, List: &ec2.SubnetList{}},
+			Extract:       reference.ExternalName(),
+		})
+		if err != nil {
+			return errors.Wrap(err, "spec.forProvider.VPCConfig.subnetIds")
```
muvaf, line 60:
```suggestion
			return errors.Wrap(err, "spec.forProvider.vpcConfig.subnetIds")
```

apis/lambda/v1alpha1/generator-config.yaml
```diff
@@ -0,0 +1,9 @@
+ignore:
+  field_paths:
+    - CreateFunctionInput.VPCConfig.SecurityGroupIDs
+    - CreateFunctionInput.VPCConfig.SubnetIDs
+    - UpdateFunctionInput.VPCConfig.SecurityGroupIDs
+    - UpdateFunctionInput.VPCConfig.SubnetIDs
+  resource_names:
+    - Alias
+    - EventSourceMapping
```
muvaf, line 9:
```suggestion
    - EventSourceMapping
    - CodeSigningConfig
```

Seems we need `CodeSigningConfig` to be ignored as well. Note that ignore doesn't delete already generated files, so you'd need to manually remove `pkg/controller/lambda/codesigningconfig` and related files in `apis/lambda`

examples/lambda/function.yaml
```diff
@@ -0,0 +1,27 @@
...
+        envKey: envValue
+    functionName: "test-function"
+    handler: "index.handler"
+    roleRef:
+      name: somerole
+    runtime: nodejs12.x
+    region: us-east-1
+    vpcConfig:
+      securityGroupIDRefs:
+        - name: sample-sg
```
muvaf, line 21:
```suggestion
        - name: sample-cluster-sg
```

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,516 @@
...
+
+	if aws.StringValue(cr.Spec.ForProvider.Runtime) != aws.StringValue(obj.Configuration.Runtime) {
+		return false, nil
+	}
+
+	if aws.Int64Value(cr.Spec.ForProvider.Timeout) != aws.Int64Value(obj.Configuration.Timeout) {
+		return false, nil
+	}
+
+	if aws.StringValue(cr.Spec.ForProvider.TracingConfig.Mode) != aws.StringValue(obj.Configuration.TracingConfig.Mode) {
```
muvaf, line 167: `cr.Spec.ForProvider.TracingConfig` or `obj.Configuration.TracingConfig.Mode` might be `nil` and this could result in panic. We need to account for that case, too.

pkg/controller/lambda/function/setup.go
```diff
@@ -0,0 +1,516 @@
...
+		return false
+	}
+	for k, v := range obj {
+		if aws.StringValue(cr[k]) == aws.StringValue(v) {
+			continue
+		}
+		return false
+	}
+	return true
+}
```
muvaf, line 225:
```suggestion
```

nitpick, you don't have to do it: What do you think about having loops in `isUpToDateEnvironment` and `isUpToDateSecurityGroupIDs` that fills up `map[string]string`s and then compare two maps using `cmp.Equal`? That's take care of pointers as well as ordering issue with slices where the elements are same but order is different, hence false negative.

muvaf: Or an approach similar to `isUpToDateFileSystemConfigs`. Dealing with pointers in more than one place sort of makes me feel unsafe because of the panic possibility.

muvaf (commented):
Looking pretty good in general. Though from what I can tell, we're able to use external name as is with this resource and not store ARN in its place. If that's possible, it would be the preferred approach instead of using provider-defined string as external name (ARN in this case).

pkg/controller/lambda/function/setup.go
smcavallo: added a comment in the code - it can never be nil as it will be set via LateInitialize with the default value from aws, if not set

pkg/controller/lambda/function/setup.go
smcavallo: good point - i've updated based on the comment to use `cmp.Equal` which works great.  IMO now that I see  this, for all comparisons should always use `cmp.Equal` unless there is a compelling reason not to.

muvaf (approved):
LGTM! Thank you @smcavallo !!

PR #563, opened 2021-02-25
common: add float64 helpers

muvaf:

These were missing and they are necessary for some types in AWS Go SDK.

Fixes #

I have:

Not necessary.

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -413,37 +413,6 @@ func CreateJSONPatch(source, destination interface{}) ([]byte, error) {
 	return patchJSON, nil
 }
 
-// String converts the supplied string for use with the AWS Go SDK.
-func String(v string, o ...FieldOption) *string {
-	for _, fo := range o {
-		if fo == FieldRequired && v == "" {
-			return &v
-		}
-	}
-	if v == "" {
-		return nil
-	}
-	return &v
-}
-
-// StringValue converts the supplied string pointer to a string, returning the
-// empty string if the pointer is nil.
-func StringValue(v *string) string {
-	if v == nil {
-		return ""
-	}
-	return *v
-}
-
-// Int64Value converts the supplied int64 pointer to a int64, returning
-// 0 if the pointer is nil.
-func Int64Value(v *int64) int64 {
-	if v != nil {
-		return *v
-	}
-	return 0
-}
-
 // LateInitializeStringPtr returns in if it's non-nil, otherwise returns from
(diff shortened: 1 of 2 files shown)


pkg/clients/aws.go
```diff
@@ -517,6 +517,21 @@ func LateInitializeInt64Ptr(in *int64, from *int64) *int64 {
 	return from
 }
 
+// Float64Value returns the value of the float64 pointer passed in or
+// 0 if the pointer is nil.
+func Float64Value(v *float64) float64 {
+	return aws.Float64Value(v)
+}
```
hasheddan, line 524: It looks like we already do this for some other types, but what is the rationale for wrapping the AWS SDK function here? Also, will we actually need `resource.Quantity` to `float64` given that CRDs do not support `float64`?

muvaf: > what is the rationale for wrapping the AWS SDK function here?

I am not sure TBH, just copied the content of `Int64Value` to get EFS generated code to work for TBS today.

> Also, will we actually need resource.Quantity to float64 given that CRDs do not support float64?

It seems to give `type: number` in OpenAPIv3Schema but I haven't tried giving that field a value. Let's keep this unmerged for now and I'll see if `float64` is OK or a change is needed in ACK to use `resource.Quantity` strictly

muvaf: Seems it was accepting integers in practice anyway, so I ignored that field and used `*int64` but we still need some helpers to do the casting from and to `*float64`. Updated the PR to remove the `LateInitializeFloat64Ptr` function, which is not needed now.

hasheddan: @muvaf I think I am still confused on why we would use `Float64Value` here instead of just `aws.Float64Value` directly?

muvaf: Personally, I think we should have the actual implementations here in this file without depending on AWS but I was in a rush when I opened this PR. Let me change all these small functions to have the actual implementations.

muvaf: Haha just saw my own note `// TODO(muvaf): is this really meaningful? why not implement it?`

muvaf: >why we would use Float64Value here instead of just aws.Float64Value directly?

Sorry I realized I didn't answer this. There are cases where we need to give `0` if the value is `nil` since they are required, hence we needed the `FieldOption` struct in for example `String` function in this file. But that's not implemented in some of the resources, so I'm updating that now.

PR #576, opened 2021-03-04
database.rdsinstance: add DBSubnetParameterGroupName to late init

muvaf:

Seems like it wasn't used in late-init because observation object returns an array. It's safe to assign the first value since there is more than value only during a change and that doesn't happen in creation where late-init is used.

Fixes https://github.com/crossplane/crossplane/issues/2187

I have:

Manually tested.

### DIFF
--- a/examples/database/rdsinstance.yaml
+++ b/examples/database/rdsinstance.yaml
@@ -4,6 +4,7 @@ metadata:
   name: example-rds
 spec:
   forProvider:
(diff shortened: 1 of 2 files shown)


hasheddan (approved):
Thanks for fixing this @muvaf! I don't love how you can only supply one parameter group at creation time but the describe DB call gives you a slice back :disappointed:

Issue #567, opened 2021-03-01
Resource for RDS PostgreSQL DB Parameter Groups


Dkaykay:
Hi, 
besides the RDSInstance CRD, it should be possible to create / update / delete RDS Parametergroups with crossplane:
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithParamGroups.html

### What problem are you facing?
Currently it is only possible to provide the name of an existing parameter group to RDSInstance. I need to create a new parameter group with crossplane. 

### How could Crossplane help solve your problem?
Implement a CRD for db parameter groups

jbhennin:
No only should you be able to create parameter groups, the ability to update parameter group parameters is also needed.
This should exist for both RDS Clusters and Instances and

muvaf:
Hi @Dkaykay ! Thanks for opening the issue! It seems like our [code generation pipeline](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md) supports generating this resource. Would you be interested in contributing?

Dkaykay:
Implemented basic functionalities for creating, updating and deleting RDS DB Parametergroups and created above PR

chlunde:
I think this is solved? Closing.

PR #577, opened 2021-03-05
Add feature for RDS DB Parametergroups

Dkaykay:

Add rds/dbparametergroup resources as described in 
https://github.com/crossplane/provider-aws/issues/567
by following the guidelines in https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md

ToDo: 
Add the possibility to pass in json / parameters that will be added/changed in the created DB parametergroup. It seems like these parameters cant be passed in at creation of the parametergroup. They need to be changed after it is created:
https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBParameterGroup.html


pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,79 @@
...
+	return obs, err
+}
+
+func postCreate(_ context.Context, cr *svcapitypes.DBParameterGroup, obj *svcsdk.CreateDBParameterGroupOutput, cre managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	meta.SetExternalName(cr, awsclients.StringValue(obj.DBParameterGroup.DBParameterGroupName))
+	return managed.ExternalCreation{ExternalNameAssigned: true}, nil
+}
```
muvaf, line 69: We need this logic only if the ID/name is decided by AWS, but in this case we give it to AWS. Similar to `preUpdate`, `preDelete` and `preObserve`, you need to implement `preCreate` as well that gives the external name to the payload object.

pkg/controller/rds/dbparametergroup/zz_controller.go
```diff
@@ -0,0 +1,236 @@
...
+func nopFilterList(_ *svcapitypes.DBParameterGroup, list *svcsdk.DescribeDBParameterGroupsOutput) *svcsdk.DescribeDBParameterGroupsOutput {
+	return list
+}
+
+func nopLateInitialize(*svcapitypes.DBParameterGroupParameters, *svcsdk.DescribeDBParameterGroupsOutput) error {
+	return nil
+}
+func alwaysUpToDate(*svcapitypes.DBParameterGroup, *svcsdk.DescribeDBParameterGroupsOutput) (bool, error) {
+	return true, nil
+}
```
muvaf, line 217: In order to enable update, you need to create a `isUpToDate` function in `setup.go` whose signature is same as this one. Then by comparing the values in `DescribeDBParameterGroupsOutput` and `DBParameterGroup` objects, the function will return `true` or `false`. If it decides that an update is due, only then an update will be done, i.e. `ModifyDBParametersGroupWithContext` call. Though I see that `DBParameterGroup` CRD does not contain parameters field, so you need to add it to `CustomDBParameterGroup`.

So, the next steps would be:
* Implement `isUpToDate`.
* Add the `Parameters` field to `CustomDBParameterGroup`. You can copy the struct from `ModifyDBParametersGroupInput`.
* In `preUpdate`, set the values of `obj` from the field you added in `CustomDBParameterGroup`.
* Manually test all the functionality with an example YAML.

apis/rds/v1alpha1/zz_db_parameter_group.go
```diff
@@ -0,0 +1,118 @@
...
+	//
+	//    * Must be 1 to 255 letters, numbers, or hyphens.
+	//
+	//    * First character must be a letter
+	//
+	//    * Can't end with a hyphen or contain two consecutive hyphens
+	//
+	// This value is stored as a lowercase string.
+	// +kubebuilder:validation:Required
+	DBParameterGroupName *string `json:"dbParameterGroupName"`
```
muvaf, line 57: We are using external name annotation for this field, so we should ignore it in `generator-config.yaml` like [this](https://github.com/crossplane/provider-aws/blob/7273d40/apis/apigatewayv2/v1alpha1/generator-config.yaml#L4). It needs to be ignored in all AWS calls in `zz_controller.go`, i.e. `CreateDBParameterGroupInput.DBParameterGroupName` etc.

muvaf (commented):
Thanks @Dkaykay ! That's a pretty good start!

pkg/controller/rds/dbparametergroup/zz_controller.go
Dkaykay: Hi @muvaf,
in preUpdate, when trying to set the values of obj, i get the following error:

muvaf: Yes the structs are different. You need to iterate over the array and add to the target array one by one by constructing an `*rds.Parameter` object per entry.

Dkaykay:
@muvaf can you review, please?

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+	input := &rds.DescribeDBParametersInput{
+		DBParameterGroupName: awsclients.String(meta.GetExternalName(cr)),
+		MaxRecords:           awsclients.Int64(20),
+	}
+	pageNum := 0
+	var results []*svcsdk.Parameter
+	err := e.client.DescribeDBParametersPagesWithContext(ctx, input, func(page *rds.DescribeDBParametersOutput, lastPage bool) bool {
+		pageNum++
+		results = append(results, page.Parameters...)
+		return pageNum <= 20
```
muvaf, line 144:
```suggestion
		results = append(results, page.Parameters...)
		return !lastPage
```

From my reading of the API, we need to return `false` if we'd like to stop and it seems `!lastPage` would be suitable for this. Is there a specific reason for stopping at page 20?

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (e *custom) isUpToDate(cr *svcapitypes.DBParameterGroup, obj *svcsdk.DescribeDBParameterGroupsOutput) (bool, error) {
+	// TODO(Dkaykay): We need isUpToDate to have context.
+	ctx := context.TODO()
+	results, err := e.getCurrentDBParameters(ctx, cr)
+
+	// compare CR with currently set Parameters
+	for _, v := range cr.Spec.ForProvider.Parameters {
+		for _, w := range results {
+			if *v.ParameterName == *w.ParameterName {
```
muvaf, line 119: You can use `awsclients.StringValue(v.ParameterName)` function and worry about only the value, checking against `""` instead of `nil`.

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+	obj.DBParameterGroupName = awsclients.String(meta.GetExternalName(cr))
+	return false, nil
+}
+
+func (e *custom) isUpToDate(cr *svcapitypes.DBParameterGroup, obj *svcsdk.DescribeDBParameterGroupsOutput) (bool, error) {
+	// TODO(Dkaykay): We need isUpToDate to have context.
+	ctx := context.TODO()
+	results, err := e.getCurrentDBParameters(ctx, cr)
+
+	// compare CR with currently set Parameters
```
muvaf, line 116: Since it's possible for the parameters to be many, what do you think about using a `map[string]svcapitypes.Parameter`? First you would check the length of both arrays and then one loop would fill the map and the second loop would do the checks against fetched parameters by using `ParameterName` as the key.

muvaf (commented):
There are a few small things but I think it's pretty close to merge. Thanks a lot @Dkaykay !

apis/rds/v1alpha1/generator-config.yaml
```diff
@@ -9,14 +9,16 @@ ignore:
     - CreateDBClusterInput.VpcSecurityGroupIds
     - ModifyDBClusterInput.VpcSecurityGroupIds
     - DBCluster.PendingModifiedValues
+    - CreateDBParameterGroupInput.DBParameterGroupName
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

examples/rds/dbparametergroup.yaml
```diff
@@ -0,0 +1,18 @@
...
+    dbParameterGroupName: example-dbparametergroup
+    dbParameterGroupFamily: postgres9.6
+    description: test
+    parameters:
+      - parameterName: application_name
+        parameterValue: "example"
+        applyMethod: immediate
+
+  providerConfigRef:
+    name: aws-provider-config
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// func postCreate(_ context.Context, cr *svcapitypes.DBParameterGroup, obj *svcsdk.CreateDBParameterGroupOutput, cre managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+// 	if err != nil {
+// 		return managed.ExternalCreation{}, err
+// 	}
+// 	meta.SetExternalName(cr, awsclients.StringValue(obj.DBParameterGroup.DBParameterGroupName))
+// 	return managed.ExternalCreation{ExternalNameAssigned: true}, nil
+// }
+
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

apis/rds/v1alpha1/custom_types.go
```diff
@@ -18,6 +18,13 @@ package v1alpha1
 
 import xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 
+// CustomDBParameterGroupParameters are custom parameters for DBParameterGroup
+type CustomDBParameterGroupParameters struct {
+	// A list of parameters to associate with this DB parameter group
+	// +optional
+	Parameters []*Parameter `json:"parameters,omitempty"`
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"context"
+
+	"k8s.io/client-go/util/workqueue"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	"github.com/aws/aws-sdk-go/service/rds"
+	svcsdk "github.com/aws/aws-sdk-go/service/rds"
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+
+	for i, v := range cr.Spec.ForProvider.Parameters {
+		// check if mandatory parameters are set (ApplyMethod, ParameterName, ParameterValue)
+		if (v.ApplyMethod == nil) || (v.ParameterName == nil) || (v.ParameterValue == nil) {
+			return errors.New("ApplyMethod, ParameterName and ParameterValue are mandatory fields and can not be nil")
+		}
+		obj.Parameters[i] = &rds.Parameter{
+			ApplyMethod:    awsclients.String(*v.ApplyMethod),
+			ParameterName:  awsclients.String(*v.ParameterName),
+			ParameterValue: awsclients.String(*v.ParameterValue),
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion, now filling all fields

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+// 	if err != nil {
+// 		return managed.ExternalCreation{}, err
+// 	}
+// 	meta.SetExternalName(cr, awsclients.StringValue(obj.DBParameterGroup.DBParameterGroupName))
+// 	return managed.ExternalCreation{ExternalNameAssigned: true}, nil
+// }
+
+func preCreate(_ context.Context, cr *svcapitypes.DBParameterGroup, obj *svcsdk.CreateDBParameterGroupInput) error {
+	obj.DBParameterGroupName = awsclients.String(meta.GetExternalName(cr))
+
```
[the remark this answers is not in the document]
Dkaykay: changed following suggestion

pkg/controller/rds/dbparametergroup/setup.go
Dkaykay: I think i dont get this one. The getCurrentDBParameters function returns an array of *svcsdk.Parameter, so all parameters currently set on AWS (about 300). In isUpToDate all parameters are compared with the ones set in the dbparametergroup claim

muvaf: It'd look like the following:
```golang
func (e *custom) isUpToDate(cr *svcapitypes.DBParameterGroup, obj *svcsdk.DescribeDBParameterGroupsOutput) (bool, error) {
	// TODO(Dkaykay): We need isUpToDate to have context.
	ctx := context.TODO()
	results, err := e.getCurrentDBParameters(ctx, cr)
	if err != nil {
		return false, err
	}
	observed := make(map[string]svcsdk.Parameter, len(results))
	for _, p := range results {
		observed[awsclients.StringValue(p.ParameterName)] = *p
	}

	// compare CR with currently set Parameters
	for _, v := range cr.Spec.ForProvider.Parameters {
		existing, ok := observed[awsclients.StringValue(v.ParameterName)]
		if !ok {
			return false, nil
		}
		switch {
		case awsclients.StringValue(existing.ParameterValue) != awsclients.StringValue(v.ParameterName):
			return false, nil
		case awsclients.StringValue(existing.ApplyMethod) != awsclients.StringValue(v.ApplyMethod):
			return false, nil
		}
	}
	return true, nil
}
```

Dkaykay: Gotcha! will do

apis/rds/v1alpha1/zz_db_parameter_group.go
muvaf: You need to re-run the code generator so that this field is removed.

examples/rds/dbparametergroup.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: rds.aws.crossplane.io/v1alpha1
+kind: DBParameterGroup
+metadata:
+  name: example-dbparametergroup
+spec:
+  forProvider:
+    region: eu-central-1
+    dbParameterGroupName: example-dbparametergroup
```
muvaf, line 8:
```suggestion
```

examples/rds/dbparametergroup.yaml
```diff
@@ -0,0 +1,17 @@
...
+    dbParameterGroupFamily: postgres9.6
+    description: test
+    parameters:
+      - parameterName: application_name
+        parameterValue: "example"
+        applyMethod: immediate
+
```
muvaf, line 15:
```suggestion
```

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/pkg/errors"
+
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+
+	svcapitypes "github.com/crossplane/provider-aws/apis/rds/v1alpha1"
+
+	awsclients "github.com/crossplane/provider-aws/pkg/clients"
```
muvaf, line 25:
```suggestion
	"sigs.k8s.io/controller-runtime/pkg/controller"
	"github.com/pkg/errors"
	svcsdk "github.com/aws/aws-sdk-go/service/rds"
	svcsdkapi "github.com/aws/aws-sdk-go/service/rds/rdsiface"
	
	"github.com/crossplane/crossplane-runtime/pkg/event"
	"github.com/crossplane/crossplane-runtime/pkg/logging"
	"github.com/crossplane/crossplane-runtime/pkg/meta"
	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
	"github.com/crossplane/crossplane-runtime/pkg/resource"
	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"

	svcapitypes "github.com/crossplane/provider-aws/apis/rds/v1alpha1"
	awsclients "github.com/crossplane/provider-aws/pkg/clients"
```

Imports are grouped from the top in the order of `std -> other -> same-org -> same repo`

muvaf (approved):
LGTM! A few things:
* ACK needs to run again to remove the `DBParameterGroupName` from CRD.
* `make reviewable` should work.
* Could you squash the commits?
* Lastly, can you confirm given example YAML works? create, update, delete scenarios.

examples/rds/dbparametergroup.yaml
Dkaykay: Why do we need to remove the dbParamterGroupName and how to exactly do this? if i run make generate, it is not removed

Dkaykay: Also,  it seems that i am not authorized to squash & merge due to the branch being protected. 

make reviewable test was green as well as create, update and delete scenarios.

muvaf: You can squash your commits locally and then `git push --force`  because you’re pushing to your own fork

Dkaykay:
closing PR and re-creating from non-broken branch

PR #582, opened 2021-03-08
Update authentication documentation

smcavallo:
Update authentication documentation with serviceAccount recommendation for provider upgrade

We use pod-iam which requires assigning the service account name to the aws role.  The problem is that the service account name is the provider-aws revision and it changes during the release, breaking crossplane permissions
Ex.  v.0.16.0 - provider-aws-6b36ca882d8e and v.0.17.0 - provider-aws-4ebd5d8c7cb7
Using a static `StringEquals` condition with a hard coded provider will break when the `SERVICE_ACCOUNT_NAME` changes when the provider is upgraded
```
"StringEquals": {
          "${OIDC_PROVIDER}:sub": "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
        }
```
This condition can be avoided by using an alternative IAM policy which uses `StringLike` and a wildcard `SERVICE_ACCOUNT_NAME`:
```
"StringLike": {
          "${OIDC_PROVIDER}:sub": "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:provider-aws-*"
        }
```
This trust policy would cover both service account names - `provider-aws-6b36ca882d8e` and `provider-aws-4ebd5d8c7cb7`
As long as provider-aws- will not change as part of a release the above should cover future cases of provider names changing during a release without needing to update the role trust policy.

Fixes #

I have:

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -130,8 +130,8 @@ read -r -d '' TRUST_RELATIONSHIP <<EOF
       },
       "Action": "sts:AssumeRoleWithWebIdentity",
       "Condition": {
-        "StringEquals": {
-          "${OIDC_PROVIDER}:sub": "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:${SERVICE_ACCOUNT_NAME}"
+        "StringLike": {
+          "${OIDC_PROVIDER}:sub": "system:serviceaccount:${SERVICE_ACCOUNT_NAMESPACE}:provider-aws-*"
         }
       }
     }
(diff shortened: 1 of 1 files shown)


AUTHENTICATION.md
```diff
@@ -141,6 +141,9 @@ EOF
 echo "${TRUST_RELATIONSHIP}" > trust.json
 ```
 
+*The default service account name is the provider-aws revision and changes with every provider release. The conditional above wildcard matches the default service account name in order to keep the role consistent across provider releases.
+The above policy assumes a service account name of `provider-aws-*`
```
hasheddan, line 145: nitpick: Mind using > to make this format as a note?

hasheddan (approved):
Thanks so much @smcavallo! This looks great, one nitpick below and I think you need DCO as well.

PR #585, opened 2021-03-09
Implemented User Creation for ElasticCache

### LINKED ISSUE #548: Implement User Creation for ElasticCache
We would like to create an initial user for an ElasticCache Instance, similar to the steps given below via CrossPlane 

https://docs.aws.amazon.com/cli/latest/reference/elasticache/create-user.html

Implementing within the provider-aws  could help us providing a user per customer when an elasticcache  instance is provisioned under AWS with CrossPlane

namku:
Resolves #548

- Add CRD and Controller for Elasticache user
- Add example file in elasticache folder

I have:

Following this steps:
https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md#local


pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	svcapitypes "github.com/crossplane/provider-aws/apis/elasticache/v1alpha1"
+)
+
+// SetupUser adds a controller that reconciles User.
+func SetupUser(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
```
hasheddan, line 18: @namku I am guessing you are going to need to define additional functions for this type ([example](https://github.com/crossplane/provider-aws/blob/a8e175bc572c441188ee370a5023f6d726f68157/pkg/controller/rds/dbcluster/setup.go#L54)). Have you tested out using this resource?

apis/aws.go
```diff
@@ -33,6 +33,7 @@ import (
 	efsv1alpha1 "github.com/crossplane/provider-aws/apis/efs/v1alpha1"
 	eksv1alpha1 "github.com/crossplane/provider-aws/apis/eks/v1alpha1"
 	eksv1beta1 "github.com/crossplane/provider-aws/apis/eks/v1beta1"
+	elasticache1alpha1 "github.com/crossplane/provider-aws/apis/elasticache/v1alpha1"
```
hasheddan, line 36: @namku due to legacy naming patterns, we currently have elasticache types in the `cache` group. @muvaf do you think we should continue with the old naming pattern?

pkg/controller/elasticache/user/setup.go
namku: Hi @hasheddan and thanks for the review, yes I tried to create this resource with this steps (https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md#local) and I could create the user, and delete too, but I going to review the example you share with me.

apis/aws.go
namku: I think it make sense to use the same serviceid name as in the list (https://github.com/aws/aws-sdk-go/tree/v1.34.32/models/apis), I think is better solution to change the others resources from cache to elasticache. What do you think? @hasheddan @muvaf 
I can change to cache serviceid name if you think is better solution keep this serviceid name.

pkg/controller/elasticache/user/setup.go
namku: I tried to add PreObserver function but when I add it and I try to create the User, the User is create but, all the time is spamming that message:
```
"message": "cannot create User in AWS: UserAlreadyExists: User exampleid already exists."
```
When I try to delete the User don't delete it from AWS and the message is:
```
"error": "User.elasticache.aws.crossplane.io\"userelasticache\" not found"
```
And if I try the same without that function, working properly, creating and deleting the user. can you give me some light about this?

the process I follow to try is:

1- Create cluster with kind.
2- Create ns crossplane-system
3- Create secret with aws credentials
4- Deploy providerconfig.yaml (AWS)
5- Execute Make generate
6- Deploy packages/crds
7- Execute "go run cmd/provider/main.go --debug"
8- Deploy example/elasticache/user.yaml
9- Delete example/elasticache/user.yaml

apis/aws.go
muvaf: The code generator decides on the [folder name here](https://github.com/aws-controllers-k8s/code-generator/blob/907b085/cmd/ack-generate/command/crossplane.go#L56), which would be whatever AWS decided to use as service name.

muvaf:
@hasheddan would you mind taking another look at this?

pkg/controller/elasticache/user/setup.go
hasheddan: @namku I think what you want to do here is first attempt to `Get` the user. If the user does not exist then you can `Create`. Likewise, with the delete operation, if you call `Delete` and the user no longer exists then there is no need to return an error because we have reached the desired state :)

apis/aws.go
```diff
@@ -81,6 +82,7 @@ func init() {
 		kmsv1alpha1.SchemeBuilder.AddToScheme,
 		efsv1alpha1.SchemeBuilder.AddToScheme,
 		rdsv1alpha1.SchemeBuilder.AddToScheme,
+		elasticache1alpha1.SchemeBuilder.AddToScheme,
```
turkenh, line 85: nit: `elasticache1alpha1` -> `elasticachev1alpha1`

apis/aws.go
turkenh: I would agree we need to be consistent under `elasticache` which would mean moving existing ones under `cache`. But, this would be an API change assuming we would also want to change the group name at CRD level (e.g. `apiVersion`). 
Maybe we can just create a follow-up issue for that and handle with a separate PR.

pkg/controller/elasticache/user/setup.go
turkenh: @hasheddan do we always need to override hooks for [external names](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md#external-name)? Do we have some defaults for external names which could be just working here without overriding?

turkenh: @namku I believe we at least have to set a postObserve to indicate our [resource as ready](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md#readiness-check), with something like:

```
func postObserve(_ context.Context, cr *v1alpha1.User, resp *elasticache.DescribeUsersOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
	if err != nil {
		return managed.ExternalObservation{}, err
	}
	if len(resp.Users) != 1 || resp.Users[0] == nil {
		return managed.ExternalObservation{}, errors.New("no user info found in response")
	}
	switch aws.StringValue(resp.Users[0].Status) {
	case "active":
	case "modifying","deleting":
		cr.SetConditions(xpv1.Unavailable())
	}

	return obs, nil
}
```

linuxbsdfreak:
Any update regarding this?

pkg/controller/elasticache/user/setup.go
haarchri: @turkenh @hasheddan fixed

apis/aws.go
haarchri: fixed

apis/aws.go
haarchri: we will open new PR to change cache to elasticache but it is breaking change as mentioned

haarchri:
@turkenh  @hasheddan  i added a new commit to fix the open issues 

tested 
- create
- update
- delete 
```

kubectl describe user.elasticache.aws.crossplane.io/example
Name:         example
Namespace:    
Labels:       <none>
Annotations:  crossplane.io/external-create-pending: 2021-11-25T17:27:56+01:00
              crossplane.io/external-create-succeeded: 2021-11-25T17:27:56+01:00
              crossplane.io/external-name: example
API Version:  elasticache.aws.crossplane.io/v1alpha1
Kind:         User
  Resource Version:  748000
  UID:               a8625015-0f6b-45bf-9604-ccc8be7023ec
Spec:
  Deletion Policy:  Delete
  For Provider:
    Access String:  on
    Engine:         redis
    Password Secret Ref:
      Key:        password
      Name:       example-elasticache
      Namespace:  crossplane-system
    Region:       us-east-1
  Provider Config Ref:
    Name:  example
  Write Connection Secret To Ref:
    Name:       example-elasticache
    Namespace:  default
Status:
  At Provider:
    Arn:  arn:aws:elasticache:us-east-1:255932642927:user:example
    Authentication:
      Password Count:  1
      type_:           password
    Status:            active
    User ID:           example
    User Name:         example
  Conditions:
    Last Transition Time:  2021-11-25T16:27:57Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2021-11-25T16:27:56Z
    Status:                True
    Type:                  Synced
Events:
  Type    Reason                   Age                From                                        Message
  ----    ------                   ----               ----                                        -------
  Normal  CreatedExternalResource  21m                managed/user.elasticache.aws.crossplane.io  Successfully requested creation of external resource
```

image

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+	kube     client.Client
+	client   svcsdkapi.ElastiCacheAPI
+	external *external
+}
+
+func filterList(cr *svcapitypes.User, obj *svcsdk.DescribeUsersOutput) *svcsdk.DescribeUsersOutput {
+	userIdentifier := awsclients.String(meta.GetExternalName(cr))
+	resp := &svcsdk.DescribeUsersOutput{}
+	for _, user := range obj.Users {
+		if awsclients.StringValue(user.UserId) == awsclients.StringValue(userIdentifier) {
```
muvaf, line 72:
```suggestion
		if awsclients.StringValue(user.UserId) == meta.GetExternalName(cr) {
```

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+			resp.Users = append(resp.Users, user)
+			break
+		}
+	}
+	return resp
+}
+
+func (e *custom) preCreate(ctx context.Context, cr *svcapitypes.User, obj *svcsdk.CreateUserInput) error {
+	obj.UserId = awsclients.String(cr.Name)
+	obj.UserName = awsclients.String(cr.Name)
```
muvaf, line 82:
```suggestion
	obj.UserId = awsclients.String(meta.GetExternalName(cr))
	obj.UserName = awsclients.String(meta.GetExternalName(cr))
```

`cr.Name` should never be used since it may not be mutable by the user in composition.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+	pw, _, err := elasticache.GetPassword(ctx, e.kube, &cr.Spec.ForProvider.PasswordSecretRef, cr.Spec.WriteConnectionSecretToReference)
+	if resource.IgnoreNotFound(err) != nil {
+		return errors.Wrap(err, "cannot get password from the given secret")
+	}
+
+	if pw != "" {
+		obj.NoPasswordRequired = awsclients.Bool(false)
+		obj.Passwords = append(obj.Passwords, awsclients.String(pw))
+	} else {
+		obj.NoPasswordRequired = awsclients.Bool(true)
```
muvaf, line 93: Why don't we make this part of spec?

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return nil
+}
+
+func postCreate(_ context.Context, cr *svcapitypes.User, obj *svcsdk.CreateUserOutput, cre managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	meta.SetExternalName(cr, awsclients.StringValue(&cr.Name))
```
muvaf, line 103: Please use the value from the `obj` here.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+	return cre, nil
+}
+
+func preObserve(_ context.Context, cr *svcapitypes.User, obj *svcsdk.DescribeUsersInput) error {
+	obj.UserId = awsclients.String(meta.GetExternalName(cr))
```
muvaf, line 108: I'm a bit confused. We should use a single field for external name. If you think there is no way to use `UserName` since the queries are done via `UserId` and that's calculated, then external name should be `UserId` and `UserName` should exist in the spec.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func postObserve(_ context.Context, cr *svcapitypes.User, obj *svcsdk.DescribeUsersOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+	if err != nil {
+		return managed.ExternalObservation{}, err
+	}
+
+	if obj.Users[0] == nil {
+		return managed.ExternalObservation{}, errors.New("no user info found in response")
+	}
```
muvaf, line 119:
```suggestion
```
If the code comes to that point, we're sure that the user exists because we already did the filtering.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+	switch awsclients.StringValue(obj.Users[0].Status) {
+	case "active":
+		cr.SetConditions(xpv1.Available())
+	case "modifying":
+		cr.SetConditions(xpv1.Unavailable())
```
muvaf, line 124: Is it unavailable during `modifying` state?

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+
+	pw, _, err := elasticache.GetPassword(ctx, e.kube, &cr.Spec.ForProvider.PasswordSecretRef, cr.Spec.WriteConnectionSecretToReference)
+	if err != nil {
+		return upd, err
+	}
+
+	var conn = managed.ConnectionDetails{
+		xpv1.ResourceCredentialsSecretPasswordKey: []byte(pw),
+	}
+	return managed.ExternalUpdate{ConnectionDetails: conn}, nil
```
muvaf, line 184:
```suggestion
	return managed.ExternalUpdate{ConnectionDetails: managed.ConnectionDetails{
		xpv1.ResourceCredentialsSecretPasswordKey: []byte(pw),
	}}, nil
```

apis/elasticache/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+
+// CustomUserParameters contains the addtionals fields for UserParameters
+type CustomUserParameters struct {
+	PasswordSecretRef xpv1.SecretKeySelector `json:"passwordSecretRef,omitempty"`
```
muvaf, line 23: It seems like the API takes an array of passwords, so I think this should be an array of secret key selectors.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,185 @@
...
+			return nil
+		}
+		return err
+	}
+	return err
+}
+
+func (e *custom) isUpToDate(cr *svcapitypes.User, obj *svcsdk.DescribeUsersOutput) (bool, error) {
+	ctx := context.Background()
+	_, pwChanged, err := elasticache.GetPassword(ctx, e.kube, &cr.Spec.ForProvider.PasswordSecretRef, cr.Spec.WriteConnectionSecretToReference)
```
muvaf, line 150: Isn't there any other thing that might be updated? Keep in mind that once you return `false, nil` here, ACK-generated code will populate all parameters in `ModifyUserInput`.

pkg/controller/elasticache/user/setup.go
haarchri: done

pkg/controller/elasticache/user/setup.go
haarchri: done

pkg/controller/elasticache/user/setup.go
haarchri: done

pkg/controller/elasticache/user/setup.go
haarchri: i can change it - done

pkg/controller/elasticache/user/setup.go
haarchri: yes fair enough

pkg/controller/elasticache/user/setup.go
haarchri: done

pkg/controller/elasticache/user/setup.go
haarchri: done

pkg/controller/elasticache/user/setup.go
haarchri: added other fields here

apis/elasticache/v1alpha1/custom_types.go
haarchri: i will think about more then one password - in general we have PasswordSecretRef for array on no other ressource if i am correct

hint:
_Warning! Both passwords will be updated. If you are only updating one password, please enter the second one as well in order to retain it._

haarchri: @muvaf can we go with followup PR for Passwords and merge this PR ? i will focus on followup to have array of passwords

haarchri: @muvaf can we go with followup PR for Passwords and merge this PR ? i will focus on followup to have array of passwords

haarchri: fixed!

haarchri:
rebased & implemented secretKeyRef as array

```
NAME                                         READY   SYNCED   EXTERNAL-NAME
user.elasticache.aws.crossplane.io/example   True    True     example
```

with two password:
```
Spec:
  Deletion Policy:  Delete
  For Provider:
    Access String:         on
    Engine:                redis
    No Password Required:  false
    Password Secret Ref:
      Key:        password1
      Name:       example-elasticache
      Namespace:  crossplane-system
      Key:        password2
      Name:       example-elasticache
      Namespace:  crossplane-system
    Region:       us-east-1
    User Name:    example-user
  Provider Config Ref:
    Name:  default
  Write Connection Secret To Ref:
    Name:       example-elasticache
    Namespace:  default
Status:
  At Provider:
    Arn:  arn:aws:elasticache:us-east-1:12345678910:user:example
    Authentication:
      Password Count:  2
      type_:           password
    Status:            active
    User ID:           example
  Conditions:
    Last Transition Time:  2022-09-23T10:28:35Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2022-09-23T10:10:36Z
    Status:                True
    Type:                  Synced
```

with one password:

```
Spec:
  Deletion Policy:  Delete
  For Provider:
    Access String:         on
    Engine:                redis
    No Password Required:  false
    Password Secret Ref:
      Key:        password1
      Name:       example-elasticache
      Namespace:  crossplane-system
    Region:       us-east-1
    User Name:    example-user
  Provider Config Ref:
    Name:  default
  Write Connection Secret To Ref:
    Name:       example-elasticache
    Namespace:  default
Status:
  At Provider:
    Arn:  arn:aws:elasticache:us-east-1:12345678910:user:example
    Authentication:
      Password Count:  1
      type_:           password
    Status:            active
    User ID:           example
  Conditions:
    Last Transition Time:  2022-09-23T09:10:35Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2022-09-23T09:09:36Z
    Status:                True
    Type:                  Synced
```

tested all comibinations 1 to 2 passwords, 2 to 1 password

haarchri:
@muvaf @chlunde   any chance to get a review ? i fixed all openpoints

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,200 @@
...
+
+func preDelete(_ context.Context, cr *svcapitypes.User, obj *svcsdk.DeleteUserInput) (bool, error) {
+	obj.UserId = awsclients.String(meta.GetExternalName(cr))
+	return false, nil
+}
+
+func postDelete(_ context.Context, cr *svcapitypes.User, obj *svcsdk.DeleteUserOutput, err error) error {
+	if err != nil {
+		if strings.Contains(err.Error(), svcsdk.ErrCodeInvalidUserStateFault) {
+			// skip: failed to delete User: InvalidUserState: User has status deleting.
```
MisterMX, line 137: Is there a reason why we silently drop this error?

If the user is already in a deleting state, it is better to check that during `preDelete` and skip the delete call in this case.

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,200 @@
...
+			return nil
+		}
+		return err
```
MisterMX, line 140: Redundant return statement

pkg/controller/elasticache/user/setup.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	connectionDetails := managed.ConnectionDetails{}
+
+	if len(pwds) > 0 {
+		connectionDetails[elasticache.PasswordKey1] = []byte(awsclients.StringValue(pwds[0]))
+		if len(pwds) > 1 {
+			connectionDetails[elasticache.PasswordKey2] = []byte(awsclients.StringValue(pwds[1]))
+		} else {
+			connectionDetails[elasticache.PasswordKey2] = nil
```
MisterMX, line 194: I don't think this is a good approach. If we want to support more than one password we should write all passwords into the secret using a loop and `key := fmt.Sprintf("%s%d", prefix, i)`.

And in addition to that, we can set the first password to the default key `password`.

MisterMX (changes requested):

pkg/controller/elasticache/user/setup.go
haarchri: only possible is 1 or 2 password 
if you remove the second password we remove it

pkg/controller/elasticache/user/setup.go
haarchri: if we drop - the resource is in deletion - we drop the resource in postDelete - we don't need to wait .. its fine ?

pkg/controller/elasticache/user/setup.go
haarchri: we return err if we get some other erros like our ErrCode

pkg/controller/elasticache/user/setup.go
MisterMX: I would prefer to check if the resource can be deleted in `preDelete` - if that is possible.

PR #601, opened 2021-03-19
Secondary vpc cidr

### LINKED ISSUE #581: Secondary VPC CIDR support
To take advantage of [CNI custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) on EKS need the ability to add secondary CIDR's to the vpc

Add support for adding secondary CIDR's to a VPC

enderv:

Adds support for adding more cidr blocks to a vpc.

Fixes #581

I have:

Unit tests. Made sure could create and delete a cidrblock


apis/ec2/v1alpha1/vpc_types.go
```diff
@@ -0,0 +1,150 @@
+/*
```
muvaf, line 1: We should probably name this file `vpccidrblock_types.go` instead of `vpc_types.go`

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+	// in a successful response, there should be one and only one object
+	if len(response.Vpcs) != 1 {
+		return managed.ExternalObservation{}, errors.New(errMultipleItems)
+	}
+
+	observed := response.Vpcs[0]
+
+	// update the CRD spec for any new values from provider
+	current := cr.Spec.ForProvider.DeepCopy()
+	ec2.LateInitializeVPCCIDRBlock(&cr.Spec.ForProvider, &observed)
```
muvaf, line 122:
```suggestion
```

`LateInitializeVPCCIDRBlock` seems to be no-op.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	currentStatusCode, err := ec2.FindVPCCIDRBlockStatus(meta.GetExternalName(cr), observed)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
+	}
+
+	switch *currentStatusCode { //nolint:exhaustive
```
muvaf, line 129:
```suggestion
	switch awsclient.StringValue(currentStatusCode) { //nolint:exhaustive
```

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	observed := response.Vpcs[0]
+
+	// update the CRD spec for any new values from provider
+	current := cr.Spec.ForProvider.DeepCopy()
+	ec2.LateInitializeVPCCIDRBlock(&cr.Spec.ForProvider, &observed)
+
+	currentStatusCode, err := ec2.FindVPCCIDRBlockStatus(meta.GetExternalName(cr), observed)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
```
muvaf, line 126:
```suggestion
		return managed.ExternalObservation{}, errors.Wrap(err, errDescribe)
```

This is actually a valid error case for this controller. We can ignore `NotFound` error only if we are the one creating the resource, in this case `IsCIDRNotFound` does that.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+	cr.Status.AtProvider = ec2.GenerateVpcCIDRBlockObservation(meta.GetExternalName(cr), observed)
+
+	isUpToDate, err := ec2.IsVpcCidrBlockUpToDate(meta.GetExternalName(cr), cr.Spec.ForProvider, observed)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
+	}
+	return managed.ExternalObservation{
+		ResourceExists:          true,
+		ResourceUpToDate:        isUpToDate,
+		ResourceLateInitialized: !cmp.Equal(current, &cr.Spec.ForProvider),
```
muvaf, line 151:
```suggestion
```
There is no late-init implementation so we don't really need to make a comparison.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	if result != nil {
+		if result.CidrBlockAssociation != nil {
+			meta.SetExternalName(cr, aws.StringValue(result.CidrBlockAssociation.AssociationId))
+		}
+		if result.Ipv6CidrBlockAssociation != nil {
+			meta.SetExternalName(cr, aws.StringValue(result.Ipv6CidrBlockAssociation.AssociationId))
+		}
+	}
+	cr.Status.SetConditions(xpv1.Creating())
```
muvaf, line 181: We don't need to do that anymore, runtime will set this condition on its own.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+	return managed.ExternalCreation{ExternalNameAssigned: true}, nil
+}
+
+func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.ExternalUpdate, error) {
+	_, ok := mgd.(*v1alpha1.VPCCIDRBlock)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+	return managed.ExternalUpdate{}, errors.New(errCannotUpdate)
+}
```
muvaf, line 192:
```suggestion
func (e *external) Update(_ context.Context, _ resource.Managed) (managed.ExternalUpdate, error) {
	return managed.ExternalUpdate{}, nil
}
```

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+
+	cr.Status.AtProvider = ec2.GenerateVpcCIDRBlockObservation(meta.GetExternalName(cr), observed)
+
+	isUpToDate, err := ec2.IsVpcCidrBlockUpToDate(meta.GetExternalName(cr), cr.Spec.ForProvider, observed)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
+	}
+	return managed.ExternalObservation{
+		ResourceExists:          true,
+		ResourceUpToDate:        isUpToDate,
```
muvaf, line 150:
```suggestion
		ResourceUpToDate:        true,
```

Seems like there is no update API for this resource so we don't need to check whether an update is due.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,210 @@
...
+func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
+	cr, ok := mgd.(*v1alpha1.VPCCIDRBlock)
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+	if ec2.IsVpcCidrDeleting(cr.Status.AtProvider) {
+		return nil
+	}
+
+	cr.Status.SetConditions(xpv1.Deleting())
```
muvaf, line 203:
```suggestion
```

pkg/clients/ec2/vpccidrblock.go
```diff
@@ -0,0 +1,142 @@
...
+		return aws.StringValue(spec.IPv6CIDRBlock) == aws.StringValue(IPv6.Ipv6CidrBlock) &&
+			aws.StringValue(spec.IPv6Pool) == aws.StringValue(IPv6.Ipv6Pool) &&
+			aws.StringValue(spec.IPv6CIDRBlockNetworkBorderGroup) == aws.StringValue(IPv6.NetworkBorderGroup), nil
+	}
+	return false, &CIDRNotFoundError{}
+}
+
+// IsVpcCidrDeleting returns true if the CIDR Block is already disassociated or disassociating
+func IsVpcCidrDeleting(observation v1alpha1.VPCCIDRBlockObservation) bool {
+	if observation.CIDRBlockState == nil && observation.IPv6CIDRBlockState == nil {
```
muvaf, line 60: Can we use a single `switch` for all three conditions in this function for better readability?

pkg/clients/ec2/vpccidrblock.go
```diff
@@ -0,0 +1,142 @@
...
+	return nil, &CIDRNotFoundError{}
+}
+
+// LateInitializeVPCCIDRBlock fills the empty fields in *v1alpha1.VPCCIDRBlockParameters with
+// the values seen in ec2.Vpc.
+func LateInitializeVPCCIDRBlock(in *v1alpha1.VPCCIDRBlockParameters, v *ec2.Vpc) { // nolint:gocyclo
+	if v == nil {
+		return
+	}
+}
```
muvaf, line 123:
```suggestion
```

muvaf (commented):
Thanks @enderv ! Pretty close to merging!

pkg/controller/ec2/vpccidrblock/controller.go
enderv: this is actually the `VpcCidrBlockStateCode` type but I'll update to avoid using pointer which I am guessing is what you were wanting here?

pkg/controller/ec2/vpccidrblock/controller.go
enderv: Cool yeah wasn't sure about this one will update.

apis/ec2/v1alpha1/referencers.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/crossplane/provider-aws/apis/ec2/v1beta1"
+)
+
+// ResolveReferences of this VPCCIDRBlock
+func (mg *VPCCIDRBlock) ResolveReferences(ctx context.Context, c client.Reader) error {
+	r := reference.NewAPIResolver(c, mg)
+
+	// Resolve spec.forProvider.vpcId
+	rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+		CurrentValue: aws.StringValue(mg.Spec.ForProvider.VPCID),
```
muvaf, line 36:
```suggestion
		CurrentValue: reference.FromPtrValue(mg.Spec.ForProvider.VPCID),
```

apis/ec2/v1alpha1/referencers.go
```diff
@@ -0,0 +1,48 @@
...
+		Reference:    mg.Spec.ForProvider.VPCIDRef,
+		Selector:     mg.Spec.ForProvider.VPCIDSelector,
+		To:           reference.To{Managed: &v1beta1.VPC{}, List: &v1beta1.VPCList{}},
+		Extract:      reference.ExternalName(),
+	})
+	if err != nil {
+		return errors.Wrap(err, "spec.forProvider.vpcId")
+	}
+	mg.Spec.ForProvider.VPCID = aws.String(rsp.ResolvedValue)
```
muvaf, line 45:
```suggestion
	mg.Spec.ForProvider.VPCID = reference.ToPtrValue(rsp.ResolvedValue)
```

We don't want to import AWS SDK library in `apis` packages so that consumers of API types don't need that dependency.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,195 @@
...
+		cr.Status.SetConditions(xpv1.Deleting())
+	default:
+		cr.SetConditions(xpv1.Unavailable())
+	}
+
+	cr.Status.AtProvider = ec2.GenerateVpcCIDRBlockObservation(meta.GetExternalName(cr), observed)
+
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
+	}
```
muvaf, line 140:
```suggestion
```

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	if ec2.IsVpcCidrDeleting(cr.Status.AtProvider) {
+		return nil
+	}
+
+	_, err := e.client.DisassociateVpcCidrBlockRequest(&awsec2.DisassociateVpcCidrBlockInput{
+		AssociationId: aws.String(meta.GetExternalName(cr)),
+	}).Send(ctx)
+
+	return awsclient.Wrap(resource.Ignore(ec2.IsVPCNotFoundErr, err), errDisassociate)
```
muvaf, line 194: This is actually a valid error for this resource. If it exists, we should ignore `NotAssociated`

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,195 @@
...
+		return managed.ExternalObservation{
+			ResourceExists: false,
+		}, nil
+	}
+
+	response, err := e.client.DescribeVpcsRequest(&awsec2.DescribeVpcsInput{
+		VpcIds: []string{aws.StringValue(cr.Spec.ForProvider.VPCID)},
+	}).Send(ctx)
+	if err != nil {
+		return managed.ExternalObservation{}, awsclient.Wrap(resource.Ignore(ec2.IsVPCNotFoundErr, err), errDescribe)
```
muvaf, line 108:
```suggestion
		return managed.ExternalObservation{}, awsclient.Wrap(err, errDescribe)
```

The error we should ignore would be `NotAssociated`, if it exists.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	// in a successful response, there should be one and only one object
+	if len(response.Vpcs) != 1 {
+		return managed.ExternalObservation{}, errors.New(errMultipleItems)
+	}
+
+	observed := response.Vpcs[0]
+
+	currentStatusCode, err := ec2.FindVPCCIDRBlockStatus(meta.GetExternalName(cr), observed)
+	if err != nil {
+		return managed.ExternalObservation{}, errors.Wrap(err, errDescribe)
```
muvaf, line 120: Here we should ignore `&CIDRNotFoundError{}` so that in case the association isn't there, we return `ExternalObservation{ResourceExists: false}` and trigger creation.

muvaf (approved):
I have some small comments but the rest should be good to go after a squash. Let me know when it's ready to merge. Thanks!!

enderv:
@muvaf should be good to go. couldn't find a notassociated error so just changed it to check the aws association not found error.

pkg/controller/ec2/vpccidrblock/controller.go
muvaf: I think we have to do this otherwise it won't report back `ResourceExists: false` without an error, which means it won't call `Create` after finding the VPC.

enderv: doh this is my bad, missed this one when i was going through on it

enderv: alright now should be fixed and good to go. sorry about that.

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -0,0 +1,192 @@
...
+	// in a successful response, there should be one and only one object
+	if len(response.Vpcs) != 1 {
+		return managed.ExternalObservation{}, errors.New(errMultipleItems)
+	}
+
+	observed := response.Vpcs[0]
+
+	currentStatusCode, err := ec2.FindVPCCIDRBlockStatus(meta.GetExternalName(cr), observed)
+	if err != nil && !ec2.IsCIDRNotFound(err) {
+		return managed.ExternalObservation{}, errors.Wrap(err, errDescribe)
```
muvaf, line 120: No worries!

```suggestion
	if err != nil {
		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
```

We need to return if we can't find it so that `Create` is called.

enderv: ah sorry misunderstood ok _now_ it should be ready. maybe...

muvaf:
Thanks a lot @enderv !

Issue #596, opened 2021-03-12
IAM and Route53 resources fail to Reconcile in non aws partitions


smcavallo:
### What happened?

In non-aws partitions (such as gov cloud and cn) iam and route53 resources fail to reconcile.
While using aws credentials the exception is `InvalidClientTokenId: The security token included in the request is invalid.`
While using pod-iam it is not able to find the OIDC provider.

### How can we reproduce it?

Attempt to create any iam or route53 resource in gov cloud or cn or non-aws based partition.

### What environment did it happen in?
Crossplane version:  0.17.0 (all)

The issue is that these clients have a global region hardcoded as `aws-global` which is only valid for the aws partition.
https://github.com/crossplane/provider-aws/blob/35d9474782ee0f85a3820ea8175e5d59c2c3e4b8/pkg/clients/aws.go#L59

I was able to successfully create iam resources in gov cloud by changing the GlobalRegion to `aws-us-gov-global` per the below.

aws partition
https://github.com/aws/aws-sdk-go-v2/blob/59774b18837c670a2c5327ae244c664459462697/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json#L3048-L3062

gov partition
https://github.com/aws/aws-sdk-go-v2/blob/59774b18837c670a2c5327ae244c664459462697/codegen/smithy-aws-go-codegen/src/main/resources/software/amazon/smithy/aws/go/codegen/endpoints.json#L8103-L8120

Supported partitions according to the aws-sdk
```
const GlobalRegion = "aws-global"
const GlobalRegionGov = "aws-us-gov-global"
const GlobalRegionCn = "aws-cn-global"
const GlobalRegionIso = "aws-iso-global"
const GlobalRegionIsoB = "aws-iso-b-global"
```

hasheddan:
@smcavallo Thanks for identifying this! Are credentials for non-`aws-global` global regions scoped only to their single global region? If so, this feels like a `ProviderConfig` setting.

smcavallo:
> 
> 
> @smcavallo Thanks for identifying this! Are credentials for non-`aws-global` global regions scoped only to their single global region? If so, this feels like a `ProviderConfig` setting.

[...]

PR #605, opened 2021-03-24
allow region configuration using ProviderConfig

smcavallo:

Allow configuration of region in ProviderConfig.  Region set in ProviderConfig overrides all other region settings (with the exception of region being set via annotations.)  
This mirrors some of the functionality of the aws sdk where a default region can be set for all clients/calls.
This fixes cases where it's not feasible/possible to override the region at the resource level.  Ex. Calls to STS and other non-resource based requests where the region is not directly exposed or cannot specifically be configured.

Calls to resources where the region is not exposed (iam, route53) can be worked around today using pod annotations to set the endpoint and region, but having to set annotations on each and every resource is not ideal.

Fixes https://github.com/crossplane/provider-aws/issues/596

I have:

* Unit tests
* Creating IAM roles in gov cloud using access keys from local kubernetes environment

### DIFF
--- a/apis/v1beta1/providerconfig_types.go
+++ b/apis/v1beta1/providerconfig_types.go
@@ -36,7 +36,7 @@ type ProviderCredentials struct {
 
 	// Region the provider credentials should be used in.
 	// This applies to all clients and resources - iam, sts, anywhere a region is used
-	Region string `json:"region",omitempty`
+	Region string `json:"region,omitempty"`
 
 	xpv1.CommonCredentialSelectors `json:",inline"`
 }
--- a/apis/v1beta1/providerconfig_types.go
+++ b/apis/v1beta1/providerconfig_types.go
@@ -34,6 +34,10 @@ type ProviderCredentials struct {
 	// +kubebuilder:validation:Enum=None;Secret;InjectedIdentity;Environment;Filesystem
 	Source xpv1.CredentialsSource `json:"source"`
 
+	// Region the provider credentials should be used in.
+	// This applies to all clients and resources - iam, sts, anywhere a region is used
+	Region string `json:"region",omitempty`
+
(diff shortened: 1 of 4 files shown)


apis/v1beta1/providerconfig_types.go
```diff
@@ -34,6 +34,10 @@ type ProviderCredentials struct {
 	// +kubebuilder:validation:Enum=None;Secret;InjectedIdentity;Environment;Filesystem
 	Source xpv1.CredentialsSource `json:"source"`
 
+	// Region the provider credentials should be used in.
+	// This applies to all clients and resources - iam, sts, anywhere a region is used
+	Region string `json:"region,omitempty"`
```
negz, line 39: Should this be `+optional`? Perhaps we should document the fact that it overrides the region modelled on each resource in their spec?

negz (commented):
There is some history here in that we used to have a (required) `spec.region` in the AWS ProviderConfig, but then removed it. See https://github.com/crossplane/provider-aws/pull/339 and https://github.com/crossplane/provider-aws/issues/38.

I've always personally been hesitant to expose the region here because it feels like we're mirroring the behaviour of CLI/SDK rather than mirroring the behaviour of the AWS API, but I think I'm very much the minority here.

muvaf:
> I've always personally been hesitant to expose the region here because it feels like we're mirroring the behaviour of CLI/SDK rather than mirroring the behaviour of the AWS API, but I think I'm very much the minority here.

@negz I agree that treating region as an authorization information in `ProviderConfig` is an implementation detail. That's why we came to [this conclusion](https://github.com/crossplane/provider-aws/issues/596#issuecomment-807148216) and I think @smcavallo will open another PR superseding this one.

smcavallo:
Replaced by https://github.com/crossplane/provider-aws/pull/608

PR #608, opened 2021-03-27
Optional region for iam and route53 resources to support non-aws partitions

smcavallo:

Expose optional region on iam and route53 resources.  Can be used in non aws partitions (gov|cn) to override the global region default which only applies to aws partition.

Fixes #
https://github.com/crossplane/provider-aws/issues/596

I have:

* Unit tests
* Testing IAMRole against gov partition
* Testing IAMRole in aws partition without region set
* Testing IAMRole in aws partition with us-east-[1-2] as the region

### DIFF
--- a/apis/identity/v1alpha1/iamaccesskey_types.go
+++ b/apis/identity/v1alpha1/iamaccesskey_types.go
@@ -37,6 +37,12 @@ type IAMAccessKeyParameters struct {
 	// +optional
 	IAMUsernameSelector *xpv1.Selector `json:"userNameSelector,omitempty"`
 
+	// Region can be used to supply an optional region.
+	// If no region is defined, aws-global region will be used
+	// which is the default region for IAM resources in the aws partition
+	// +optional
+	Region string `json:"region,omitempty"`
+
 	// The current status of this IAMAccessKey on the AWS
 	// Must be either Active or Inactive.
 	// +kubebuilder:validation:Enum=Active;Inactive
--- a/pkg/controller/identity/iamaccesskey/controller.go
+++ b/pkg/controller/identity/iamaccesskey/controller.go
@@ -72,7 +72,15 @@ type connector struct {
 }
 
 func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
-	cfg, err := awsclient.GetConfig(ctx, c.kube, mg, awsclient.GlobalRegion)
+	cr, ok := mg.(*v1alpha1.IAMAccessKey)
+	region := awsclient.GlobalRegion
+	if cr.Spec.ForProvider.Region != "" {
+		region = cr.Spec.ForProvider.Region
+	}
+	if !ok {
+		return nil, errors.New(errUnexpectedObject)
+	}
+	cfg, err := awsclient.GetConfig(ctx, c.kube, mg, region)
 	if err != nil {
 		return nil, err
 	}
--- a/apis/identity/v1alpha1/iamgroup_types.go
+++ b/apis/identity/v1alpha1/iamgroup_types.go
@@ -27,6 +27,12 @@ type IAMGroupParameters struct {
 	// The path for the group name.
 	// +optional
 	Path *string `json:"path,omitempty"`
+
+	// Region can be used to supply an optional region.
+	// If no region is defined, aws-global region will be used
+	// which is the default region for IAM resources in the aws partition
+	// +optional
+	Region string `json:"region,omitempty"`
 }
 
 // An IAMGroupSpec defines the desired state of an IAM Group.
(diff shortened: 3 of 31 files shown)


apis/identity/v1alpha1/iamaccesskey_types.go
```diff
@@ -37,6 +37,12 @@ type IAMAccessKeyParameters struct {
 	// +optional
 	IAMUsernameSelector *xpv1.Selector `json:"userNameSelector,omitempty"`
 
+	// Region can be used to supply an optional region.
+	// If no region is defined, aws-global region will be used
+	// which is the default region for IAM resources in the aws partition
+	// +optional
+	Region string `json:"region,omitempty"`
```
muvaf, line 44:
```suggestion
	Region *string `json:"region,omitempty"`
```

Using pointer type is [recommended](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#pointer-types-and-markers) for optional fields.

apis/identity/v1alpha1/iamaccesskey_types.go
```diff
@@ -37,6 +37,12 @@ type IAMAccessKeyParameters struct {
 	// +optional
 	IAMUsernameSelector *xpv1.Selector `json:"userNameSelector,omitempty"`
 
+	// Region can be used to supply an optional region.
```
muvaf, line 40: Can we be more specific in this comment? I believe people might try to supply regions like `us-east-1` for this field, which will return an error from AWS. Something along the lines of `Region should be supplied only if you are targeting non-AWS partitions such as CN or Gov`. We could even consider naming this as `Partition` maybe if that's the term used to differentiate.

pkg/controller/identity/iamaccesskey/controller.go
```diff
@@ -72,7 +72,15 @@ type connector struct {
 }
 
 func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
-	cfg, err := awsclient.GetConfig(ctx, c.kube, mg, awsclient.GlobalRegion)
+	cr, ok := mg.(*v1alpha1.IAMAccessKey)
+	region := awsclient.GlobalRegion
+	if cr.Spec.ForProvider.Region != "" {
+		region = cr.Spec.ForProvider.Region
+	}
+	if !ok {
```
muvaf, line 80: This check needs to happen right after `cr, ok := mg.(*v1alpha1.IAMAccessKey)` otherwise cr is `nil`

apis/identity/v1alpha1/iamaccesskey_types.go
smcavallo: `us-east-1` will work as will other actual regions.  The endpoint resolver in the aws sdk will find the correct endpoint to create the region.  But that brings up the question of - model representation of the sdk vs model representation of the api.
The IAM resource will not actually live in the region defined in that property.  The sdk will use that region to find the correct endpoint where that resource should live (implementation detail?)

As an alternative to this PR, what are your thoughts on simply adding a `Partition` property on the `ProviderConfig` ?
A single provider cannot make requests across aws partitions.  aws-provider can look up the global region per-partition.
The thought being that your provider is only valid for a single partition, and consumers can define where that is, if not aws.
Ex. 
```
func GetGlobalRegion(partition string) string {
	// Global regions are convention based by partition
	return partition + "-global"
}
````
From a consumer perspective this is easy to implement.  Add the partition on the providerConfig and everything just works.   It may just not be apparent to non-aws partition users that they have to do this.

Another alternative is since endpoint annotations can be used as a workaround in non-aws partitions, propose a solution where the non-exposed configuration sts calls (for pod iam) can be managed.  Example annotations which if set on the `ProviderConfig` could be honored by `provider-aws` -
```
    eks.amazonaws.com/audience: "sts.us-gov-west-1.amazonaws.com"
    eks.amazonaws.com/sts-regional-endpoints: "true"
```
per https://github.com/aws/amazon-eks-pod-identity-webhook#aws_sts_regional_endpoints-injection
the nice thing here is that it gives users more control over the sts call.

muvaf: I see `region` and `partition` in the same bucket of configuration parameters and we had decided to keep the `region` in the managed resources instead of `ProviderConfig`, so I feel like that decision applies to here as well.

I'm open to either `region` or `partition` on managed resources with a preference of `partition` since normally we expect the input in the format of `aws-global` or `gov-global`, not an actual region like `us-gov-west-1` or `us-west-1` in IAM and Route53. Even if the normal regions work, it's used only to infer the partition from my reading of your comments.

muvaf: We have synced with @smcavallo and it seems that as opposed to `region`, you cannot use the same credential set with more than one partition, i.e. there is no cross-partition IAM like cross-account. The partition and the credentials are tightly coupled. Hence, the flexibility of choosing the partition on managed resource level is not much meaningful because the chosen `ProviderConfig` credential has to belong to that partition anyway.

Except IAM and Route53, `partition` field is optional because SDK gets a region string like `us-gov-east-1` and figures out the partition. For IAM and Route53, one has to specify the partition (we do this by hard-coding `aws-global`). Implementation-wise, I think having optional `spec.partition` in `ProviderConfig` and using it in all calls if given would be the best UX here. If it's empty, it'll fall back to `aws-global`.

smcavallo:
closing in favor of https://github.com/crossplane/provider-aws/pull/623

PR #614, opened 2021-04-07
ec2.securitygroup: add security group resolver

### LINKED ISSUE #592: Reference SecurityGroups from SecurityGroups without hardcoding IDs
When creating `SecurityGroups` with Crossplane, we are not able to reference other `SecurityGroups` without hardcoding the `SecurityGroupIDs`.

For every application, we want a `SecurityGroup` for only that app. This should be referenced when creating a new AWS service, and service `SecurityGroup`, for example a new CacheCluster.

Reason Zero Trust Networking.

We would like to be able to reference other SecurityGroups in `userIdGroupPairs` in both `Ingress` and `Egress`.

An example where we create a app SG and a CacheCluster SG and want to link them together. This is _not_ necessary in the same composition.

```yaml
kind: SecurityGroup
metadata:
  name: applicationname
  labels:
    app: appname
spec:
  forProvider:
    description: application security group attached to pod
    groupName: applicationname
    egress:
    - ipProtocol: tcp
      securityGroupObjectRef:
      - groupIDSelector:
          matchLabels:
            appservices: applicationname
---
kind: SecurityGroup
metadata:
  name: cacheClusterSecurityGroup
  labels:
    appservices: applicationname
spec:
  forProvider:
    description: CacheCluster security group for inbound access from application
    groupName: appCacheCluster
    ingress:
      - ipProtocol: tcp
        userIdGroupPairs:
          - groupIDSelector:
              matchLabels:
                app: appname
               #matchControllerRef: [true|false]
            # or a new object ref would be nice (since we only want one without labels in this specific use-case)
            securityGroupObjectRef:
              namespace: app
              name: applicationname
              kind: SecurityGroup
```

chlunde:
Add possibility to refer to security group kubernetes objects in ingress/egress rules.

Fixes #592 

I have:

Tested in AWS by creating two security groups and using a named reference. Resolvers do not seem to have

### DIFF
--- a/apis/ec2/v1beta1/referencers.go
+++ b/apis/ec2/v1beta1/referencers.go
@@ -89,10 +89,54 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
 				Extract:      reference.ExternalName(),
 			})
 			if err != nil {
-				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d]", i, j))
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].vpcId", i, j))
 			}
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCID = reference.ToPtrValue(rsp.ResolvedValue)
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCIDRef = rsp.ResolvedReference
+
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].groupId", i, j))
+			}
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupIDRef = rsp.ResolvedReference
+		}
+	}
+
+	for i, egr := range mg.Spec.ForProvider.Egress {
+		for j, pair := range egr.UserIDGroupPairs {
+			// Resolve spec.forProvider.egress[*].userIdGroupPairs[*]
+			rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.VPCID),
+				Reference:    pair.VPCIDRef,
+				Selector:     pair.VPCIDSelector,
+				To:           reference.To{Managed: &VPC{}, List: &VPCList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d].vpcId", i, j))
+			}
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].VPCID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].VPCIDRef = rsp.ResolvedReference
+
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d].groupId", i, j))
+			}
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].GroupID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].GroupIDRef = rsp.ResolvedReference
 		}
 	}
 
--- a/apis/ec2/v1beta1/referencers.go
+++ b/apis/ec2/v1beta1/referencers.go
@@ -80,7 +80,7 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
 
 	for i, ing := range mg.Spec.ForProvider.Ingress {
 		for j, pair := range ing.UserIDGroupPairs {
-			// Resolve spec.forProvider.ingress[*].userIdGroupPairs[*]
+			// Resolve spec.forProvider.ingress[*].userIdGroupPairs[*].vpcId
 			rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
 				CurrentValue: reference.FromPtrValue(pair.VPCID),
 				Reference:    pair.VPCIDRef,
@@ -94,6 +94,7 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCID = reference.ToPtrValue(rsp.ResolvedValue)
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCIDRef = rsp.ResolvedReference
 
+			// Resolve spec.forProvider.ingress[*].userIdGroupPairs[*].groupId
 			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
 				CurrentValue: reference.FromPtrValue(pair.GroupID),
 				Reference:    pair.GroupIDRef,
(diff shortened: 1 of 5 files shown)


apis/ec2/v1beta1/referencers.go
```diff
@@ -89,10 +89,56 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
...
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupIDRef = rsp.ResolvedReference
+		}
+	}
+
+	for i, ing := range mg.Spec.ForProvider.Egress {
+		for j, pair := range ing.UserIDGroupPairs {
+			// Resolve spec.forProvider.ingress[*].userIdGroupPairs[*]
+			/*
+				   why was this not implemented already?
```
chlunde, line 116: @muvaf do you know why this was was only implemented in ingress?

apis/ec2/v1beta1/referencers.go
```diff
@@ -89,10 +89,56 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
 				Extract:      reference.ExternalName(),
 			})
 			if err != nil {
-				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d]", i, j))
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d] VPC", i, j))
```
muvaf, line 92:
```suggestion
				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].vpcId", i, j))
```

apis/ec2/v1beta1/referencers.go
```diff
@@ -89,10 +89,56 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
...
+
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d] Group", i, j))
```
muvaf, line 105:
```suggestion
				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].groupId", i, j))
```

apis/ec2/v1beta1/referencers.go
```diff
@@ -89,10 +89,56 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
...
+			*/
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d] Group", i, j))
```
muvaf, line 138:
```suggestion
				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d].groupId", i, j))
```

apis/ec2/v1beta1/referencers.go
```diff
@@ -89,10 +89,56 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
...
+				   why was this not implemented already?
+				rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+					CurrentValue: reference.FromPtrValue(pair.VPCID),
+					Reference:    pair.VPCIDRef,
+					Selector:     pair.VPCIDSelector,
+					To:           reference.To{Managed: &VPC{}, List: &VPCList{}},
+					Extract:      reference.ExternalName(),
+				})
+				if err != nil {
+					return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d] VPC", i, j))
```
muvaf, line 125:
```suggestion
					return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].vpcId", i, j))
```

apis/ec2/v1beta1/referencers.go
muvaf: I don't know, probably egress is added after ingress.

chlunde:
> @chlunde let me know when this is ready for another review after manual test.

I will have to find out what happens here, it always produces a diff:
```
  Conditions:
    Last Transition Time:  2021-04-08T12:20:03Z
    Reason:                Creating
    Status:                False
    Type:                  Ready
    Last Transition Time:  2021-04-08T12:20:04Z
    Status:                True
    Type:                  Synced
```

It would be nice if --debug printed cmp.Diff in the log

muvaf:
You might need to explicitly ignore new ref and selector fields [here](https://github.com/crossplane/provider-aws/blob/25c6581/pkg/clients/ec2/securitygroup.go#L333) since they are nested.

chlunde:
@muvaf the problem seems to be that the real "diff" code is here:

jsonPatch, err := awsclients.CreateJSONPatch(*currentParams, target)

and the cmp.Diff just tries to compare that with an empty object. It is not compared with the real ingress rules.

Is this pattern specific for security groups, and has it been superseeded by something else?

If it must be done this way I wonder if we could nil out the refs in target before calling CreateJSONPatch

chlunde:
```go
	target = *target.DeepCopy()
	for i, spec := range target.Egress {
		for j := range spec.UserIDGroupPairs {
			target.Egress[i].UserIDGroupPairs[j].GroupIDRef = nil
			target.Egress[i].UserIDGroupPairs[j].GroupIDSelector = nil
		}
	}

	for i, spec := range target.Ingress {
		for j := range spec.UserIDGroupPairs {
			target.Ingress[i].UserIDGroupPairs[j].GroupIDRef = nil
			target.Ingress[i].UserIDGroupPairs[j].GroupIDSelector = nil
		}
	}
```
like this, in `CreateSGPatch`. It works... I know the current approach is to copy stuff from the k8s object to the "observed k8s-ified AWS object" but that does not work for arrays (like ingress/egress rules) and causes bugs if rules are added/removed/changed. This seems cleaner for in this case, we will create another issue for that other problem.

chlunde:
@muvaf we could also do a `cmp.Equal` - one time for each slice, and remove equal items from the `CreateSGPatch` return value.

muvaf:
@chlunde Assigning `nil` to ref/selectors of the copied object sounds good to me. I think I've seen that pattern in another resource, too. The problem happens when the ref/selectors are deep in the struct.

chlunde:
@muvaf cool, I will follow this example: https://github.com/crossplane/provider-aws/blob/master/apis/ec2/v1beta1/routetable_types.go#L140-L144

chlunde:
Added `func (u *UserIDGroupPair) ClearRefSelectors()`, also clearing VPC ref/selector

I also did `targetCopy := *target.DeepCopy()` to match the routetable code (but I wonder if `target = *target.DeepCopy()` might be equally good, because now it might be a bit to easy to ready/write to `target` instead of `targetCopy`).

Tested and it now is in sync "instantly".

chlunde:
@muvaf one problem with this implementation is that we can't match more than one security group. Do you think this should be added later or should it be redesigned to handle this?

chlunde:
@muvaf I think the only difference between this and #731 was a few missing comments, added them now. Do you think we can merge? I've been running this code since april :smile: 

We need `ClearRefSelectors` to avoid bugs in isUpToDate.

AaronME:
> @muvaf one problem with this implementation is that we can't match more than one security group. Do you think this should be added later or should it be redesigned to handle this?

@chlunde I ran a test with the example below. I saw some odd behavior:
1. I could not predict which of the three "extra" security groups would become the successful egress rule. (On four runs, I got group *-one, *-three, *-one, *-one)
1. Changing the ingress groupIdRef adds a rule instead of changing the rule. (I believe this is addressed by https://github.com/crossplane/provider-aws/pull/631). BUT, it also triggers the Selector and causes a second selector rule to be added.
1. Changing the groupIdRef on a group a second time can cause a failure to update if the Egress selector tries to re-add the rule that was not deleted in the first change.

I think we should definitely test again after https://github.com/crossplane/provider-aws/pull/631 is merged, and I think we should consider fixing the selector behavior or removing the selector and moving forward with only the Reference for now.

chlunde:
> 1. I could not predict which of the three "extra" security groups would become the successful egress rule. (On four runs, I got group *-one, *-three, *-one, *-one)

I think this is a general issue with selectors if you have multiple possible matches and not using the multiresolver.

> 2. Changing the ingress groupIdRef adds a rule instead of changing the rule. (I believe this is addressed by [ec2.securitygroup: fix add, implement revoke/update ingress and egress rules #631](https://github.com/crossplane/provider-aws/pull/631)). BUT, it also triggers the Selector and causes a second selector rule to be added.

Yeah, I think this just show cases one of the multiple issues i with complex security group rule without #631 (and [with late init](https://github.com/crossplane/provider-aws/issues/618)). It's not directly caused by this PR, it's just easier to provoke it.

> 3. Changing the groupIdRef on a group a second time can cause a failure to update if the Egress selector tries to re-add the rule that was not deleted in the first change.
> 
> I think we should definitely test again after #631 is merged, and I think we should consider fixing the selector behavior or removing the selector and moving forward with only the Reference for now.

Not sure if it is worth breaking this PR into two pieces, it might be just as good to wait for #631.

negz:
@chlunde I believe this is redundant now that https://github.com/crossplane/provider-aws/pull/631 is merged, but I could be wrong. Let me know if we need to reopen this.

Issue #618, opened 2021-04-09
ec2.securitygroup: Late init lists mixes lists 


chlunde:
### What happened?
* Create an object with a "late init" field set in an array, such as description in security groups in the yaml
* Remove that field from the keyless array, and add another element at the same time, without that "late init" field set 

The "late init" field might now end up in an unrelated element in the array.

Example:

`kubectl create` this object with description for one IP range:
```yaml
apiVersion: ec2.aws.crossplane.io/v1beta1
kind: SecurityGroup
metadata:
  name: sg-1
spec:
  forProvider:
    description: desc
    groupName: sg1
    region: eu-north-1
    ingress:
    - ipProtocol: "tcp"
      fromPort: 2
      toPort: 2
      ipRanges:
      - cidrIp: 10.0.0.0/8
    - ipProtocol: "tcp"
      fromPort: 1
      toPort: 1
      ipRanges:
      - cidrIp: 10.0.0.0/8
        description: "foo"
    vpcId: vpc-...
```

next, `kubectl apply` with a completely different rule, and without description:

```yaml
apiVersion: ec2.aws.crossplane.io/v1beta1
kind: SecurityGroup
metadata:
  name: sg-1
spec:
  forProvider:
    description: desc
    groupName: sg1
    region: eu-north-1
    ingress:
    - ipProtocol: "tcp"
      fromPort: 2
      toPort: 2
      ipRanges:
      - cidrIp: 10.0.0.0/8
    - ipProtocol: "tcp"
      fromPort: 3
      toPort: 3
      ipRanges:
      - cidrIp: 172.0.0.0/8
    vpcId: vpc-...
```

using `kubectl get -o yaml` we see now that the `description: "foo"` from 10.0.0.0/8:1 is now set in `172.0.0.0/8:3`:

```yaml
    - fromPort: 3
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 172.0.0.0/8
        description: foo
      toPort: 3
```

I believe this is because late init is applied instantly after a change, and the only check is that the array sizes are the same:

https://github.com/crossplane/provider-aws/blob/07b4af13224b32ac298d957a7e949cb33abacbdd/pkg/clients/ec2/securitygroup.go#L178-L185

Some ideas:
* Would it be possible to late init as few fields as possible. For examply, why late init description, is it useful?
* Avoid late init if the CR is changed since the last apply - is it possible?
* Detect mismatches for fields like port number, ip ranges etc and abort late init on mismatch

[...]

PR #620, opened 2021-04-12
ec2.securitygroup: Avoid late init if it seems unsafe

chlunde:
If the structure of the ingress/egress rules are different, or any port
numbers/protocols do not match, avoid late init because we will not be
able to do it safely. The array does not have a unique key.

Also avoid late init of description because I have not seen that AWS
sets description.

Updates #618

I have:

`kubectl create` this object with description for one IP range:
```yaml
kind: SecurityGroup
metadata:
  name: sg-1
spec:
  forProvider:
    description: desc
    groupName: sg1
    region: eu-north-1
    ingress:
    - ipProtocol: "tcp"
      fromPort: 2
      toPort: 2
      ipRanges:
      - cidrIp: 10.0.0.0/8
    - ipProtocol: "tcp"
      fromPort: 1
      toPort: 1
      ipRanges:
      - cidrIp: 10.0.0.0/8
        description: "foo"
    vpcId: vpc-...
```

next, `kubectl apply` with a completely different rule, and without description:

```yaml
kind: SecurityGroup
metadata:
  name: sg-1
spec:
  forProvider:
    description: desc
    groupName: sg1
    region: eu-north-1
    ingress:
    - ipProtocol: "tcp"
      fromPort: 2
      toPort: 2
      ipRanges:
      - cidrIp: 10.0.0.0/8
    - ipProtocol: "tcp"
      fromPort: 3
      toPort: 3
      ipRanges:
      - cidrIp: 172.0.0.0/8
    vpcId: vpc-...
```

With the old code, using `kubectl get -o yaml` we see now that the `description: "foo"` from 10.0.0.0/8:1 is now set in `172.0.0.0/8:3`. This is no longer the case.

```yaml
    - fromPort: 3
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 172.0.0.0/8
        description: foo
      toPort: 3
```

### DIFF
--- a/pkg/clients/ec2/securitygroup.go
+++ b/pkg/clients/ec2/securitygroup.go
@@ -2,6 +2,7 @@ package ec2
 
 import (
 	"encoding/json"
+	"log"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/clients/ec2/securitygroup.go
```diff
@@ -198,42 +199,55 @@ func LateInitializeIPPermissions(spec []v1beta1.IPPermission, o []ec2.IpPermissi
...
+	for i := range o {
+		spec[i].FromPort = awsclients.LateInitializeInt64Ptr(spec[i].FromPort, o[i].FromPort)
+		spec[i].ToPort = awsclients.LateInitializeInt64Ptr(spec[i].ToPort, o[i].ToPort)
+		spec[i].IPProtocol = awsclients.LateInitializeString(spec[i].IPProtocol, o[i].IpProtocol)
+
 		for j := range o[i].UserIdGroupPairs {
-			if len(spec[i].UserIDGroupPairs) == j {
-				spec[i].UserIDGroupPairs = append(spec[i].UserIDGroupPairs, v1beta1.UserIDGroupPair{})
-			}
 			spec[i].UserIDGroupPairs[j].Description = awsclients.LateInitializeStringPtr(
```
chlunde, line 251: Is there any more checks we should do here for UserIDGroupPairs

pkg/clients/ec2/securitygroup.go
```diff
@@ -2,6 +2,7 @@ package ec2
 
 import (
 	"encoding/json"
+	"log"
```
chlunde, line 5: if you like this we can remove the logging and move it to a new helper method

chlunde:
what do you think @muvaf

chlunde:
I can write a unit test if you agree that this change is needed.

muvaf:
From what I can tell, we rely on `protocol`, `fromPort` and `toPort` as key in this [other PR](https://github.com/crossplane/provider-aws/pull/631). Would we be able to rely to those as keys here, too?

Issue #562, opened 2021-02-25
Implement Amazon Cloud Map service


janwillies:
### What problem are you facing?
Amazon Cloud Map resources are missing.

https://docs.aws.amazon.com/sdk-for-go/api/service/servicediscovery/

### How could Crossplane help solve your problem?
Use ACK to generate servicediscovery and then fill the gaps.

haarchri:
implemented with https://github.com/crossplane/provider-aws/pull/627

PR #627, opened 2021-04-13
feat(servicediscovery): added privatednsnamespace API and controller …

blut:
…for AWS servicediscovery

Signed-off-by: Hannes Blut <hannes.blut-extern@deutschebahn.com>
(external expert on behalf of DB Netz AG)

* Added generated API code and controllers for AWS servicediscovery
* Added resource type: PrivateDNSNamespace

Fixes # https://github.com/crossplane/provider-aws/issues/562

I have:

Testing has thus far been done manually to establish a first servicediscovery controller and functional requirements.
Unit tests will follow after the resource handling has been discussed


apis/servicediscovery/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,16 @@
+package v1alpha1
+
+// CustomServiceParameters are custom parameters for Services.
+type CustomServiceParameters struct{}
+
+// CustomPrivateDNSNamespaceParameters are custom parameters for PrivateDNSNamespaces.
+type CustomPrivateDNSNamespaceParameters struct {
+	// HostedZoneID *string
+	// HTTPName     *string
```
MisterMX, line 9: These commented lines should be removed

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
+	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	cpresource "github.com/crossplane/crossplane-runtime/pkg/resource"
+
+	svcapitypes "github.com/crossplane/provider-aws/apis/servicediscovery/v1alpha1"
+	awsclients "github.com/crossplane/provider-aws/pkg/clients"
+)
+
+func (e *external) observeImpl(ctx context.Context, cpm cpresource.Managed) (managed.ExternalObservation, error) { // nolint:gocyclo
```
MisterMX, line 19: You should use a seperate `hook` type where you define your custom functions and pass them to `external` which is auto-generated and should not be modified.

Then you can rename this to `observe` as well.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	ret := managed.ExternalObservation{}
```
MisterMX, line 20: I think it would increase readability if a `managed.ExternalObservation` is created for every return statement. It is a bit hard to follow the state of the variable.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	cr, ok := cpm.(*svcapitypes.PrivateDNSNamespace)
```
MisterMX, line 21: I think `mg` is the convention for `cpresource.Managed` variables (at least that's how it is named in other controllers. So I would suggest to stick to it.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	if cr.Status.AtProvider.OperationID == nil || *cr.Status.AtProvider.OperationID == "" {
+		return ret, nil
+	}
+	opInput := &svcsdk.GetOperationInput{
+		OperationId: cr.Status.AtProvider.OperationID,
+	}
+
+	opReqResp, err := e.client.GetOperationWithContext(ctx, opInput)
```
MisterMX, line 37: I think it should be `opReq` and `opResp`, shouldn't it?

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return ret, errors.Wrap(err, "get-operation failed")
+	}
+
+	if *opReqResp.Operation.Status == "PENDING" || *opReqResp.Operation.Status == "SUBMITTED" {
```
MisterMX, line 42: `opReqResp.Operation.Status` is a pointer and might be nil in some edge cases. This will cause a panic if you try to deref it.
Better to use `awsclient.StringValue`

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+
+	nsInput := &svcsdk.GetNamespaceInput{
+		Id: namespaceID,
+	}
+	nsReqResp, err := e.client.GetNamespaceWithContext(ctx, nsInput)
+	if err != nil {
+		if ActualIsNotFound(err) || IsDuplicateRequest(err) {
+			cr.Status.SetConditions(xpv1.Unavailable())
+		}
+		return ret, awsclients.Wrap(cpresource.IgnoreAny(err, ActualIsNotFound, IsDuplicateRequest), "get-namespace failed")
```
MisterMX, line 69: Error messages should be constants, not magic strings

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+		ret.ResourceLateInitialized = true
+	}
+
+	currentSpec := cr.Spec.ForProvider.DeepCopy()
+	lateInitialize(&cr.Spec.ForProvider, nsReqResp)
+	ret.ResourceLateInitialized = ret.ResourceLateInitialized || !cmp.Equal(&cr.Spec.ForProvider, currentSpec)
+
+	upToDate, err := e.isUpToDate(cr, nsReqResp)
+	if err != nil {
+		return ret, errors.Wrap(err, "isUpToDate check failed")
```
MisterMX, line 85: Error messages should be constants, not magic strings

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	return ret, nil
+}
+
+func (e *external) updateImpl(ctx context.Context, mg cpresource.Managed) (managed.ExternalUpdate, error) {
+	// Update always returns an error, since the servicediscovery API does not allow updating namespaces
+	return managed.ExternalUpdate{}, nil
+}
+
+func (e *external) isUpToDate(cr *svcapitypes.PrivateDNSNamespace, resp *svcsdk.GetNamespaceOutput) (bool, error) { // nolint:unparam
+	if meta.GetExternalName(cr) != *resp.Namespace.Id {
```
MisterMX, line 98: Use `awsclient.StringValue` instead of raw deref

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	// one could use the HostedZone information to verify the VPC configuration
+	// through Route53 HostedZone records
+
+	// tags := map[string]string
+	// if cmp.Equal(cr.Spec.ForProvider.Tags, tags) {
+	// Where does the servicediscovery API provide Tag information?
+
+	// if cr.Spec.ForProvider.Region
+	// Region information is not available through servicediscovery API
+	return true, nil
```
MisterMX, line 120: Error in `isUpToDate` is always nil, so it can be removed.

pkg/controller/servicediscovery/privatednsnamespace/setup.go
```diff
@@ -0,0 +1,55 @@
...
+
+	svcapitypes "github.com/crossplane/provider-aws/apis/servicediscovery/v1alpha1"
+)
+
+// SetupPrivateDNSNamespace adds a controller that reconciles PrivateDNSNamespaces.
+func SetupPrivateDNSNamespace(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
+	name := managed.ControllerName(svcapitypes.PrivateDNSNamespaceGroupKind)
+	opts := []option{
+		func(e *external) {
+			e.observe = e.observeImpl
```
MisterMX, line 24: It would be best to create a `hook` type that has `observe`, `create`, `update` and `delete`.

See https://github.com/crossplane/provider-aws/blob/master/pkg/controller/secretsmanager/secret/setup.go as example

pkg/controller/servicediscovery/privatednsnamespace/setup.go
```diff
@@ -0,0 +1,55 @@
...
+		For(&svcapitypes.PrivateDNSNamespace{}).
+		Complete(managed.NewReconciler(mgr,
+			resource.ManagedKind(svcapitypes.PrivateDNSNamespaceGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+// ActualIsNotFound reimplements IsNotFound which doesn't do it's job
+func ActualIsNotFound(err error) bool {
```
MisterMX, line 45: Those two functions don't belong to `setup.go` and should rather be in `hooks.go` where they are used.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+		return ret, errors.Wrap(err, "isUpToDate check failed")
+	}
+
+	ret.ResourceUpToDate = upToDate
+	return ret, nil
+}
+
+func (e *external) updateImpl(ctx context.Context, mg cpresource.Managed) (managed.ExternalUpdate, error) {
+	// Update always returns an error, since the servicediscovery API does not allow updating namespaces
+	return managed.ExternalUpdate{}, nil
```
MisterMX, line 94: In contrast to the comment `updateImpl` does not return an error

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+func lateInitialize(forProvider *svcapitypes.PrivateDNSNamespaceParameters, nsReqResp *svcsdk.GetNamespaceOutput) {
+	if nsReqResp == nil {
+		return
+	}
+
+	if forProvider.Description == nil && nsReqResp.Namespace.Description != nil {
+		forProvider.Description = nsReqResp.Namespace.Description
+	}
+
+	if forProvider.CreatorRequestID == nil && nsReqResp.Namespace.CreatorRequestId != nil {
```
MisterMX, line 149: I think you can remove the second nil check from both if's. `Description` and `CreatorRequestID` will be set to nil anyway.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	cr, ok := cpm.(*svcapitypes.PrivateDNSNamespace)
+	if !ok {
+		return ret, errors.New(errUnexpectedObject)
+	}
+	externalName := meta.GetExternalName(cr)
+	if externalName == "" {
+		return ret, nil
+	}
+
+	if cr.Status.AtProvider.OperationID == nil || *cr.Status.AtProvider.OperationID == "" {
```
MisterMX, line 30: Might wanna use `awsclient.StringValue(cr.Status.AtProvider.OperationID) == ""`

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return ret, errors.Wrap(err, "get-operation failed")
+	}
+
+	if *opReqResp.Operation.Status == "PENDING" || *opReqResp.Operation.Status == "SUBMITTED" {
+		ret.ResourceExists = true
+		return ret, nil
+	}
+
+	if *opReqResp.Operation.Status == "FAIL" {
```
MisterMX, line 47: Better use `awsclient.StringValue`

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+		if cr.GetCondition(xpv1.TypeReady).Reason != xpv1.ReasonDeleting {
+			ret.ResourceExists = true
+		}
+		cr.Status.SetConditions(xpv1.Unavailable().WithMessage(*opReqResp.Operation.ErrorMessage))
```
MisterMX, line 51: Check for nil before deref

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		if ActualIsNotFound(err) || IsDuplicateRequest(err) {
+			cr.Status.SetConditions(xpv1.Unavailable())
+		}
+		return ret, awsclients.Wrap(cpresource.IgnoreAny(err, ActualIsNotFound, IsDuplicateRequest), "get-namespace failed")
+	}
+	ret.ResourceExists = true
+
+	cr.Status.SetConditions(xpv1.Available())
+	if externalName != *namespaceID {
```
MisterMX, line 74: Better use `awsclient.StringValue`

apis/servicediscovery/v1alpha1/generator-config.yaml
```diff
@@ -0,0 +1,6 @@
+ignore:
+  resource_names:
+    - PublicDNSNamespace
+    - Service
+    - HTTPNamespace
+# How to add +immutable to required fields Region, Name, and VPC
```
MisterMX, line 6: Comment should be removed

MisterMX (changes requested):

apis/servicediscovery/v1alpha1/custom_types.go
blut: Removed commented parameters

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: updated code to have separate hook struct

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: I'm not convinced that it helps readability, but I've changed the code to create the external observation on each return

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: moved all string pointer dereferences into awsclient.StringValue calls

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: This is also practice in some of the other controllers I used for reference. I've changed it to a constant.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: Error does not exist anymore

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/setup.go
blut: Thanks for the reference! I've updated the code to use a hooks struct.

pkg/controller/servicediscovery/privatednsnamespace/setup.go
blut: As far as I'm concerned they should be part of the generated code, which is why I've placed them in the setup.go.
Moved to hooks.go.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: This function was initially intended to always return an error but has since been reduced to this state. The function has been removed, as it double the nopUpdate function.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: Again, referenced from another controller, but I agree. Updated without the second test.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: opStatus is now determined apriori with StringValue.

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

pkg/controller/servicediscovery/privatednsnamespace/hooks.go
blut: done

apis/servicediscovery/v1alpha1/generator-config.yaml
blut: How would someone updating the code know that I'd previously added the immutable keyword to required fields in the generated code?

blut:
Squashed all commits

muvaf (approved):
Apologies for late review. There are a couple of things that would be good to fix but I'll merge this PR and open another one to fix them since this PR has been up for a long time now.

PR #629, opened 2021-04-14
cache.replicationgroup: Support pattern matching for engine version

### LINKED ISSUE #622: cache.replicationgroup: Engine version mismatch 6.x vs. 6.0.5, never ready
### What happened?
```yaml
apiVersion: cache.aws.crossplane.io/v1beta1
kind: ReplicationGroup
...
spec:
  forProvider:
    cacheNodeType: cache.t3.micro
    cacheParameterGroupName: default.redis6.x.cluster.on
    engine: redis
    engineVersion: 6.x
...
```

It looks like it never becomes ready:
```
k get replicationgroup -A
NAME                    READY   SYNCED   STATE       VERSION   AGE
my-redis2-c86xx-m6tbp   False   True     modifying   6.x       71m
```

If we patch the code to log why it needs an update, we can find what the difference is:
```go
func cacheClusterNeedsUpdate(kube v1beta1.ReplicationGroupParameters, cc elasticache.CacheCluster) bool { // nolint:gocyclo
	// AWS will set and return a default version if we don't specify one.
	if !reflect.DeepEqual(kube.EngineVersion, cc.EngineVersion) {
		log.Println("- engine version", aws.StringValue(kube.EngineVersion), aws.StringValue(cc.EngineVersion))
		return true
	}
```

```
crossplane-provider-aws-c2ea9b021877-745c78dbcc-kz6gs provider-aws 2021/04/12 14:13:00 - engine version 6.x 6.0.5
```

### How can we reproduce it?
Create a cluster with `6.x` as version.

See also: https://github.com/hashicorp/terraform-provider-aws/issues/15625

provider-aws: 0.17

chlunde:
At the moment, it is only possible to send in version 6.x for redis
engine version on redis 6, but AWS will return back a specific version
like 6.0.5.

In this case, redis will never become ready.

Change the controller to consider the resource up to date when we
specify 6.x and have 6.0.5.

Fixes #622

I have:

* added unit test
* created redis 6.x cluster and check that it is ready

```
k get replicationgroup 
NAME                               READY   SYNCED   STATE       VERSION   AGE
redis-s9c9r-vw7mx                  True    True     available   6.x       3h3m
```

### DIFF
--- a/pkg/clients/elasticache/elasticache.go
+++ b/pkg/clients/elasticache/elasticache.go
@@ -19,6 +19,7 @@ package elasticache
 import (
 	"reflect"
 	"strconv"
+	"strings"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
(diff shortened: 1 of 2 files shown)


pkg/clients/elasticache/elasticache.go
```diff
@@ -219,9 +220,21 @@ func automaticFailoverEnabled(af elasticache.AutomaticFailoverStatus) *bool {
 	return &r
 }
 
+func versionMatches(kubeVersion *string, awsVersion *string) bool {
+	if reflect.DeepEqual(kubeVersion, awsVersion) {
```
muvaf, line 224:
```suggestion
	if clients.StringValue(kubeVersion) == clients.StringValue(awsVersion) {
```

pkg/clients/elasticache/elasticache.go
```diff
@@ -219,9 +220,21 @@ func automaticFailoverEnabled(af elasticache.AutomaticFailoverStatus) *bool {
 	return &r
 }
 
+func versionMatches(kubeVersion *string, awsVersion *string) bool {
```
muvaf, line 223: What do you think about using `switch` instead of three separate conditions in this function? I think `switch` makes it easier to see what's going on when ordered if's are needed.

muvaf (approved):
LGTM with two small comments. Thanks @chlunde !

pkg/clients/elasticache/elasticache.go
chlunde: fixed

pkg/clients/elasticache/elasticache.go
chlunde: fixed

muvaf (approved):
LGTM! Thanks!

muvaf:
@chlunde could you squash the commits?

PR #631, opened 2021-04-14
ec2.securitygroup: fix add, implement revoke/update ingress and egress rules

### LINKED ISSUE #300: Not possible to remove rules from securitygroup
### What happened?

The controller supports only addition but not removal of egress/ingress rules.

AWS: 0.11.0

### LINKED ISSUE #503: Security Group doesn't add new ingress rule
### What happened?

I added a new ingress rule to a SecurityGroup but it wasn't applied to the resource.
After the update Ready = True and Sync = True, no errors in events.

### How can we reproduce it?

Create the following security group:

```
cat <<EOF | kubectl apply -f -
kind: SecurityGroup
metadata:
  name: crossplane-aws-rds
spec:
  deletionPolicy: Delete
  forProvider:
    description: Security group to allow traffic to databases created by Crossplane
    groupName: example-crossplane-aws-rds
    ingress:
    - fromPort: 5432
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 10.0.0.0/16
      toPort: 5432
    region: eu-west-2
    vpcId: vpc-xxx
  providerConfigRef:
    name: default
EOF
```

Add a new ingress rule:
```
cat <<EOF | kubectl apply -f -
kind: SecurityGroup
metadata:
  name: crossplane-aws-rds
spec:
  deletionPolicy: Delete
  forProvider:
    description: Security group to allow traffic to databases created by Crossplane
    groupName: example-crossplane-aws-rds
    ingress:
    - fromPort: 3306
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 10.0.0.0/16
      toPort: 3306
    - fromPort: 5432
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 10.0.0.0/16
      toPort: 5432
    region: eu-west-2
    vpcId: vpc-xxx
  providerConfigRef:
    name: default
EOF
```

Crossplane version: 1.0.0
AWS provider: 0.16.0

chlunde:

Implement a "diff rules" function.

When adding a rule, we should not send the complete rule set, only the
new rule.

When updating a rule, we must first delete it. And with that done, we
also support deleting rules.

Fixes #503
Fixes #300

I have:

- manually added/modified/removed simple ingress rules with cidr blocks.
- unit tests (for IP-ranges)

TODO
- [x] add unit tests
- [x] handle TODO in code
- [x] consider performance
- [x] handle tcp vs TCP in diff
- [x] ignore order in arrays in diff function ([1.1.1.1, 2.2.2.2] == [2.2.2.2, 1.1.1.1])
- [x] check how diff function handles all types, not just cidr blocks
- [x] ignore order in arrays in unit tests ([1.1.1.1, 2.2.2.2] == [2.2.2.2, 1.1.1.1])

### DIFF
--- a/apis/ec2/v1beta1/referencers.go
+++ b/apis/ec2/v1beta1/referencers.go
@@ -88,10 +88,54 @@ func (mg *SecurityGroup) ResolveReferences(ctx context.Context, c client.Reader)
 				Extract:      reference.ExternalName(),
 			})
 			if err != nil {
-				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d]", i, j))
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].vpcId", i, j))
 			}
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCID = reference.ToPtrValue(rsp.ResolvedValue)
 			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].VPCIDRef = rsp.ResolvedReference
+
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.ingress[%d].userIdGroupPairs[%d].groupId", i, j))
+			}
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Ingress[i].UserIDGroupPairs[j].GroupIDRef = rsp.ResolvedReference
+		}
+	}
+
+	for i, egr := range mg.Spec.ForProvider.Egress {
+		for j, pair := range egr.UserIDGroupPairs {
+			// Resolve spec.forProvider.egress[*].userIdGroupPairs[*]
+			rsp, err := r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.VPCID),
+				Reference:    pair.VPCIDRef,
+				Selector:     pair.VPCIDSelector,
+				To:           reference.To{Managed: &VPC{}, List: &VPCList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d].vpcId", i, j))
+			}
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].VPCID = reference.ToPtrValue(rsp.ResolvedValue)
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].VPCIDRef = rsp.ResolvedReference
+
+			rsp, err = r.Resolve(ctx, reference.ResolutionRequest{
+				CurrentValue: reference.FromPtrValue(pair.GroupID),
+				Reference:    pair.GroupIDRef,
+				Selector:     pair.GroupIDSelector,
+				To:           reference.To{Managed: &SecurityGroup{}, List: &SecurityGroupList{}},
+				Extract:      reference.ExternalName(),
+			})
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("spec.forProvider.egress[%d].userIdGroupPairs[%d].groupId", i, j))
+			}
+			mg.Spec.ForProvider.Egress[i].UserIDGroupPairs[j].GroupID = reference.ToPtrValue(rsp.ResolvedValue)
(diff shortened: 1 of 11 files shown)


chlunde:
@muvaf hi, this is ready for review now. I will have to do some more manual testing for prefix lists and ipv6, and get cmp.Diff to ignore the rule order (or make it stable) for the unit tests, but feedback on the approach would be cool!

chlunde:
@muvaf what do you think is the next step here?

pkg/controller/ec2/securitygroup/diff.go
```diff
@@ -0,0 +1,295 @@
...
+import (
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsec2 "github.com/aws/aws-sdk-go-v2/service/ec2"
+)
+
+// ruleKey represents the unique tuple (protocol, from-to port) in a
+// format supported as a map key
+type ruleKey struct {
```
muvaf, line 28: Have you had the chance to give some thoughts to my suggestion about using this as key in the map to find the rule and compare the matching rules manually by using `StringValue` and friends?

chlunde: yeah, I've tried this. It's about the same amount of code in total, but if we use lists to compare the rules manually, we have to sort them to prevent a performance degradation on large lists I. Because we don't have generics I think this makes some tricky code duplicated for each type:

```go
// for merge:

for _, r := range m.PrefixListIds {
        idx := sort.Search(len(i.prefixListIDs), func(idx int) bool {
                return prefixListIDCmp(i.prefixListIDs[idx], r) <= 0
        })

        if idx == len(i.prefixListIDs) { // nil or after last element
                i.prefixListIDs = append(i.prefixListIDs, r)
        } else if prefixListIDCmp(i.prefixListIDs[idx], r) != 0 {
                // not present, insert at idx
                i.prefixListIDs = append(i.prefixListIDs[:idx+1], i.prefixListIDs[idx:]...) // index < len(a)
                i.prefixListIDs[idx] = r
        }
}

// for diff
func (i ipPermissionMap) diffPrefixListIDs(other ipPermissionMap) []awsec2.PrefixListId {
        var ret []awsec2.PrefixListId
       for _, r := range i.prefixListIDs {
               idx := sort.Search(len(other.prefixListIDs), func(idx int) bool {
                       return prefixListIDCmp(other.prefixListIDs[idx], r) <= 0
               })
               if idx == len(other.prefixListIDs) || prefixListIDCmp(other.prefixListIDs[idx], r) != 0 {
                       ret = append(ret, r) // not found
                }
        }
        return ret

```

Later I realized there is a separate API to update descriptions only, and since CIDR etc. cannot be nil, it's  even better to rewrite as `map[cidrIPv4] -> description`, I thin, so I will do another rewrite - without any string intern magic...

avdhoot:
@chlunde are you still working on this?

chlunde:
@avdhoot I have a new version here: https://github.com/chlunde/provider-aws/commit/c20dbe278baaca765bea47e8112f679385b07d55

but one question is if we can use `map[string]UserIdGroupPair` for rules to/from security groups, where the string key will be the group id (if defined), falling back to the group name. To answer that I would have run with default VPCs.

Then, I *think* we could simplify a lot:

```go
func (i ipPermissionMap) diffUserIDGroupPair(other ipPermissionMap) []awsec2.UserIdGroupPair {
	var ret []awsec2.UserIdGroupPair
	for key, r := range i.groups {
		r2, ok := other.groups[key]
		if !ok || aws.StringValue(r.Description) != aws.StringValue(r2.Description) {
			// there are other fields here:
			// * peering connection id
			// * peering status
			// * VPC ID
			// * User ID
			// but all are immutable per `key` (group name/id), or one-way status fields
			ret = append(ret, r)
		}
	}
	return ret
```
	
Create a map of the UserIdGroupPairs:
```go
	for _, r := range m.UserIdGroupPairs {
		// a UserIdGroupPair must have a group id or group name, and
		// they are unique so we can use them for keys
		key := aws.StringValue(r.GroupId)
		if key == "" {
			key = aws.StringValue(r.GroupName)
		}
		i.groups[key] = r
	}
```

If this is correct, I think we could add back late init again for UserIdGroupPairs (and only that), for the fields mentioned above.

avdhoot:
@chlunde  I am not the best person to answer this. @muvaf can you please guide us here. This is an important feature.

AaronME:
Confirmed fixes #503.
Confirmed fixes #300.
 
 @chlunde It looks like this needs a rebase and some conflicts addressed. Are you planning another pass using the GroupID as a key?

chlunde:
@AaronME this version gets rid of string interning completely: https://github.com/chlunde/provider-aws/commit/c20dbe278baaca765bea47e8112f679385b07d55 - but has a complex `sort.Search` for inserting.

I have another branch somewhere with the [code above](https://github.com/crossplane/provider-aws/pull/631#issuecomment-903816157) which uses maps for `UserIdGroupPairs` but I'm not 100% sure if there are any cases where `GroupName` is insufficient as a key, but I think that is only possible for default VPCs, and I *assume* you can't refer to another account then.

Which would you prefer, `map[groupid]pair` or `sort.Search`?

Also, this does not introduce late init, but personally I think there are so many issues with the current security group implementation that I would rather introduce that in a later PR (but possibly in the same release), if we find out what we should late init. Alternatively I think we could do late init of these fields in `UserIdGroupPairs`, when we have an exact match:

```
			// there are other fields here:
			// * peering connection id
			// * peering status
			// * VPC ID
			// * User ID
```

My concerns with late init are that if you remove a security group rule, or all rules in a group object, we must avoid to "late init" it back to the the CR. So late init of this object is much more complex than the other objects.

Also I would want to avoid doing a late init PR (which I suspect could be fairly big in itself) before we decide on the data structures.

negz (approved):
Two nits, but nothing that need block this PR. Thank you for tackling this somewhat mind bending fix. :)

negz:
/backport

😬

PR #633, opened 2021-04-17
ec2.securitygroup: Add tests for tags

chlunde:

Add tests I missed to sg tags.

I have:

Test code only. Tested.


hasheddan (approved):
@chlunde does this need any updates before it is brought out of draft?

chlunde:
@hasheddan I was just a bit surprised that we did not have to delete a tag key before setting a new value. I wanted to double check that this was correct, because this is not how the security group rule API works.. (but I like atomic replace APIs better than delete + insert :smile:)

chlunde:
yeah, checked (again), CreateTags can overwrite tags

PR #646, opened 2021-05-03
Feature for creating VPC Peering Connections

### LINKED ISSUE #325: Support for AWS VPC Peering

There is no managed resource for creating VPC Peering

Implement a managed resource for VPC Peering.

Dkaykay:

Create VPC Peering Connections and approve the connection on AWS

Fixes #325

I have:

- create VPC Peering connection
- (partly update VPC Peering connection)
- delete VPC Peering connection


pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+			},
+		},
+	}
+	obj.TagSpecifications = append(obj.TagSpecifications, &spec)
+
+	return nil
+}
+
+func (e *custom) postCreate(ctx context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.CreateVpcPeeringConnectionOutput, cre managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+	// set peering connection id as label on the k8s object after creation
```
muvaf, line 156: Why do we need it as a label? Doesn't it exist as external-name annotation already?

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+	key := "Name"
+	value := obj.VpcPeeringConnection.VpcPeeringConnectionId
+	tags := []*svcsdk.Tag{
+		{
+			Key:   &key,
+			Value: value,
+		},
+	}
+	obj.VpcPeeringConnection.SetTags(tags)
```
muvaf, line 165: This seems to set tags on the create output object, which we don't use after this function completes.

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+type custom struct {
+	kube   client.Client
+	client svcsdkapi.EC2API
+}
+
+func preObserve(ctx context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPeeringConnectionsInput) error {
```
muvaf, line 58: I think overriding `filterList` would be more appropriate for this task.

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+	obj.SetFilters(filters)
+	return nil
+}
+
+func (e *custom) postObserve(_ context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPeeringConnectionsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+	if err != nil {
+		return managed.ExternalObservation{}, err
+	}
+
+	connectionCounter := -1
```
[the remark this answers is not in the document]
muvaf: A small function to find the target connection object could be helpful. You can reuse it in `postObserve`, `isUpToDate` and `filterList`.

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+		}
+	}
+	return false, nil
+}
+
+func preCreate(ctx context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.CreateVpcPeeringConnectionInput) error {
+	// set external name as tag on the vpc peering connection
+	resType := "vpc-peering-connection"
+	key := "Name"
```
muvaf, line 138:
```suggestion
	key := "crossplane-external-name"
```
so that it's more clear who set this. Also, we can have this as a `const` and use it in all functions.

pkg/controller/rds/dbparametergroup/setup.go
```diff
@@ -0,0 +1,148 @@
+package dbparametergroup
```
muvaf, line 1: `dbparametergroup` is already implemented, seems like files are mixed.

apis/ec2/v1alpha1/zz_vpc_peering_connection.go
```diff
@@ -0,0 +1,125 @@
...
+type VPCPeeringConnectionParameters struct {
+	// Region is which region the VPCPeeringConnection will be created.
+	// +kubebuilder:validation:Required
+	// +immutable
+	Region string `json:"region"`
+	// Checks whether you have the required permissions for the action, without
+	// actually making the request, and provides an error response. If you have
+	// the required permissions, the error response is DryRunOperation. Otherwise,
+	// it is UnauthorizedOperation.
+	DryRun *bool `json:"dryRun,omitempty"`
```
muvaf, line 37:
```suggestion
```

This isn't really applicable in a CRD, so we should make code-generator ignore this field via `generator-config.yaml`

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+func (e *custom) postObserve(_ context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPeeringConnectionsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+	if err != nil {
+		return managed.ExternalObservation{}, err
+	}
+
+	connectionCounter := -1
+	for _, v := range obj.VpcPeeringConnections {
+		connectionCounter++
+		for _, tag := range v.Tags {
+			if *tag.Key == "Name" {
```
muvaf, line 80:
```suggestion
			if awsclients.StringValue(tag.Key) == "Name" {
```

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+				if *tag.Value == cr.ObjectMeta.Name {
+					switch *v.Status.Code {
+					case "active":
+						return true, nil
+					case "deleted":
+						return false, nil
+					}
+
+					if *v.Status.Code == "pending-acceptance" && cr.Spec.ForProvider.AcceptRequest {
```
muvaf, line 115: Wouldn't `postObserve` be more suited for this task?

pkg/controller/ec2/vpcpeeringconnection/setup.go
```diff
@@ -0,0 +1,189 @@
...
+	spec := svcsdk.TagSpecification{
+		ResourceType: &resType,
+		Tags: []*svcsdk.Tag{
+			{
+				Key:   &key,
+				Value: &value,
+			},
+		},
+	}
+	obj.TagSpecifications = append(obj.TagSpecifications, &spec)
```
muvaf, line 150: I'd suggest checking whether it exists and if so, return error saying that they need to change the external name annotation to affect the name and that specific tag key is reserved.

pkg/controller/ec2/vpcpeeringconnection/setup.go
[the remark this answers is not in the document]
Dkaykay: @muvaf for my understanding, there is only one object that holds all the vpc peering connections, so we need to iterate through it

pkg/controller/ec2/vpcpeeringconnection/setup.go
Dkaykay: I was using the Name tag because this will also be the display name in aws console

pkg/controller/ec2/vpcpeeringconnection/setup.go
Dkaykay: No, external-name is the metadata.name from the claim but peering connection id is generated dynamically by aws. But we might also want to have the peering connection id available in the cluster, so we set it as a label

pkg/controller/ec2/vpcpeeringconnection/setup.go
Dkaykay: @muvaf Can you provide an example or further description?

AaronME:
@muvaf taking a look at this

pkg/controller/ec2/vpcpeeringconnection/setup.go
muvaf: Why not use the connection id as external name, similar to [other resources](https://github.com/crossplane/provider-aws/blob/7f29cf3/pkg/controller/apigatewayv2/apimapping/setup.go#L90) whose id is generated after creation?

apis/rds/v1alpha1/custom_types.go
```diff
@@ -18,6 +18,13 @@ package v1alpha1
 
 import xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 
+// CustomDBParameterGroupParameters are custom parameters for DBParameterGroup
```
AaronME, line 21: @Dkaykay Are these RDS changes required to implement the vpcpeeringconnection?

pkg/controller/ec2/vpcpeeringconnection/setup.go
AaronME: Take a look at https://github.com/crossplane/provider-aws/blob/959cc5b741fd7c2dafa7e20cac75ab55ae2c79bb/pkg/controller/rds/dbcluster/setup.go#L123

pkg/controller/ec2/vpcpeeringconnection/setup.go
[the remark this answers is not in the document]
AaronME: @Dkaykay Correct, but once you find the connection you want, you can break iteration and return.

As @muvaf pointed out, we can remove a lot of these loops if we set the external name to the generated id of the connection.

AaronME (changes requested):

Dkaykay:
i will recreate this PR, seems like my branch was not clean. I already implemented the proposed changes there:
https://github.com/crossplane/provider-aws/pull/688

PR #649, opened 2021-05-04
Bugfix/s3 website configuration

jack1902:

Slack Thread: https://crossplane.slack.com/archives/CEF5N8X08/p1620143654284900

Fixes https://github.com/crossplane/provider-aws/issues/650

I have:

- [x] Read and followed Crossplane's [contribution process]

To Note, i have removed `protocol` as a required parameter because even the description states it defaults to what the inbound request is yet the Spec states its required which goes against the given description

- I have been trying to use the website configuration on the bucket resource within a composition. I noticed that it worked if i passed routing rules but didn't when i only passed the index and error documents (things i have already done before in terraform but i am moving the bucket creation to crossplane 🚀 )

- `make test` - ✅ 
- `make e2e` - ✅ 
Screenshot 2021-05-04 at 21 09 59

- `make reviewable` - ✅ 
- `make` - ✅

### DIFF
--- a/package/crds/s3.aws.crossplane.io_buckets.yaml
+++ b/package/crds/s3.aws.crossplane.io_buckets.yaml
@@ -885,7 +885,6 @@ spec:
                             type: string
                         required:
                         - hostName
-                        - protocol
                         type: object
                       routingRules:
                         description: Rules that define when a redirect is applied and the redirect behavior.
@@ -920,8 +919,6 @@ spec:
                                 replaceKeyWith:
                                   description: The specific object key to use in the redirect request. For example, redirect request to error.html. Not required if one of the siblings is present. Can be present only if ReplaceKeyPrefixWith is not provided.
                                   type: string
-                              required:
-                              - protocol
                               type: object
                           required:
                           - redirect
--- a/pkg/controller/s3/bucket/websiteConfig.go
+++ b/pkg/controller/s3/bucket/websiteConfig.go
@@ -139,28 +139,35 @@ func GenerateWebsiteConfiguration(config *v1beta1.WebsiteConfiguration) *awss3.W
 	if config.RedirectAllRequestsTo != nil {
 		wi.RedirectAllRequestsTo = &awss3.RedirectAllRequestsTo{
 			HostName: awsclient.String(config.RedirectAllRequestsTo.HostName),
-			Protocol: awss3.Protocol(config.RedirectAllRequestsTo.Protocol),
+		}
+		if config.RedirectAllRequestsTo.Protocol != "" {
+			wi.RedirectAllRequestsTo.Protocol = awss3.Protocol(config.RedirectAllRequestsTo.Protocol)
 		}
 	}
-	wi.RoutingRules = make([]awss3.RoutingRule, len(config.RoutingRules))
-	for i, rule := range config.RoutingRules {
-		rr := awss3.RoutingRule{
-			Redirect: &awss3.Redirect{
-				HostName:             rule.Redirect.HostName,
-				HttpRedirectCode:     rule.Redirect.HTTPRedirectCode,
-				Protocol:             awss3.Protocol(rule.Redirect.Protocol),
-				ReplaceKeyPrefixWith: rule.Redirect.ReplaceKeyPrefixWith,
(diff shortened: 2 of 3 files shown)


package/crds/s3.aws.crossplane.io_buckets.yaml
```diff
@@ -885,7 +885,6 @@ spec:
                             type: string
                         required:
                         - hostName
-                        - protocol
```
muvaf, line 888: CRDs are automatically generated. You'll need to make [this field](https://github.com/crossplane/provider-aws/blob/master/apis/s3/v1beta1/websiteConfiguration_types.go#L64) 
```
Protocol *string `json:"protocol,omitempty"`
```
 to make it non-required then fix where it's used.

muvaf (commented):
Hi @jack1902 ! Congrats on your first contribution! One small caveat and I think we're good to go!

package/crds/s3.aws.crossplane.io_buckets.yaml
jack1902: How do i regen these files? Since i have made a change to the overall structs

muvaf: you can run `make reviewable`

jack1902: Running now, will push once its done (i forgot about that command as i ran `test` and `e2e` only 🤦 )

jack1902: Ran it locally, no difference in output (assuming this means my hand-crafted changes are correct)

pkg/controller/s3/bucket/websiteConfig.go
```diff
@@ -139,28 +139,37 @@ func GenerateWebsiteConfiguration(config *v1beta1.WebsiteConfiguration) *awss3.W
...
 		}
 	}
-	wi.RoutingRules = make([]awss3.RoutingRule, len(config.RoutingRules))
-	for i, rule := range config.RoutingRules {
-		rr := awss3.RoutingRule{
-			Redirect: &awss3.Redirect{
+	if config.RoutingRules != nil {
+		wi.RoutingRules = make([]awss3.RoutingRule, len(config.RoutingRules))
+		for i, rule := range config.RoutingRules {
+			redirect := &awss3.Redirect{
```
muvaf, line 150: Why did you need to switch from initializing with `RoutingRule` to `Redirect`? The former seems more straightforward to read to me.

pkg/controller/s3/bucket/websiteConfig.go
```diff
@@ -139,28 +139,37 @@ func GenerateWebsiteConfiguration(config *v1beta1.WebsiteConfiguration) *awss3.W
...
+		}
+		if config.RedirectAllRequestsTo.Protocol != "" {
+			wi.RedirectAllRequestsTo.Protocol = awss3.Protocol(config.RedirectAllRequestsTo.Protocol)
 		}
 	}
-	wi.RoutingRules = make([]awss3.RoutingRule, len(config.RoutingRules))
-	for i, rule := range config.RoutingRules {
-		rr := awss3.RoutingRule{
-			Redirect: &awss3.Redirect{
+	if config.RoutingRules != nil {
```
muvaf, line 147:
```suggestion
	if len(config.RoutingRules) != 0 {
```

This is usually a safer check for slices.

muvaf (commented):
Just small comments. Make sure to squash the commits when you're done so that I can merge right away :)

pkg/controller/s3/bucket/websiteConfig.go
jack1902: i did this becaue i wasn't 100% sure if i could do `rr.Redirect.Protocol` but i realise now that i can do that (i'm a bit rusty in Go as i haven't written it in ages)

pkg/controller/s3/bucket/websiteConfig.go
jack1902: done

pkg/controller/s3/bucket/websiteConfig.go
jack1902: updated

muvaf (approved):
Thanks @jack1902 !

PR #653, opened 2021-05-05
s3.bucketpolicy: Ignore string order

chlunde:
When checking if two policies are identical, we must ignore string
order. Idea for the fix from the ECR policy check.

Do we want to move this method to a shared library to avoid duplicating it for all controllers with resource based policies?

I have:

- Added unit tests

### DIFF
--- a/pkg/controller/s3/bucketpolicy/bucketpolicy_test.go
+++ b/pkg/controller/s3/bucketpolicy/bucketpolicy_test.go
@@ -188,6 +188,58 @@ func TestObserve(t *testing.T) {
 	}
 }
 
+func TestIsBucketPolicyUpToDate(t *testing.T) {
+
+	type want struct {
+		local    string
+		remote   string
+		uptodate bool
+	}
+
+	cases := map[string]struct {
+		want
+	}{
+		"Equal": {
+			want: want{
+				local:    policy,
+				remote:   policy,
+				uptodate: true,
+			},
+		},
+		"EqualWhiteSpace": {
+			want: want{
+				local:    policy,
+				remote:   policy + "  ",
+				uptodate: true,
+			},
+		},
+		"Differs": {
+			want: want{
+				local:    `{"Statement":[{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com"}],"Version":"2012-10-17"}`,
+				remote:   `{"Statement":[{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test2.s3.crossplane.com"}],"Version":"2012-10-17"}`,
+				uptodate: false,
+			},
+		},
+		"EqualUnordered": {
+			want: want{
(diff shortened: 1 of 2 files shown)


pkg/controller/s3/bucketpolicy/bucketpolicy_test.go
```diff
@@ -188,6 +188,58 @@ func TestObserve(t *testing.T) {
...
+				local:    `{"Statement":[{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com"}],"Version":"2012-10-17"}`,
+				remote:   `{"Statement":[{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test2.s3.crossplane.com"}],"Version":"2012-10-17"}`,
+				uptodate: false,
+			},
+		},
+		"EqualUnordered": {
+			want: want{
+				local:    `{"Statement":[{"Action":"s3:GetObject","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com/*"},{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com"}],"Version":"2012-10-17"}`,
+				remote:   `{"Version":"2012-10-17","Statement":[{"Action":"s3:ListBucket","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com"},{"Action":"s3:GetObject","Effect":"Allow","Principal":"*","Resource":"arn:aws:s3:::test.s3.crossplane.com/*"}]}`,
+				uptodate: false,
```
muvaf, line 227:
```suggestion
				uptodate: true,
```

Don't you want these two statements to be considered equal?

chlunde: Good catch, I did *intended* to write that when I wrote the test, as the name suggests :+1:

I think there are *four* different things that *can* happen, in theory:

- Indentation change (seen in real life, at least the AWS IAM Policy console)
- Map keys change order (seen in real life w/ bucketpolicy code). Top level map keys, like `Version`/`Statement` and `Action`/`Principal`/`Effect` within `Statement`s.
- The order of the array items I *think* we don't need to fix (i.e. Action: [] and Statement: []) - but I'm not sure. If I understand correctly, it does not matter when evaluating the policy, so in theory aws could re-order them. On the other hand, from my testing, they are preserved and I think it would be confusing for the user (and terraform) if AWS would re-shuffle them.
- Single item as array vs string. For example, `Action: ["s3:ListBucket"]` vs `Action: "s3:ListBucket"`.

This change handles the two first, which are the ones I have observed so far. I'm not sure if we want to proactively handle case 3/4. What do you think?

chlunde: @muvaf should we start with the known, simple case, indentation change and map key order (i.e. keep the code as it is but fix the testcases)?

muvaf: https://github.com/crossplane/provider-aws/pull/701 also has a similar discussion related to up to date check. Is there anything we can do to converge on one place to handle this tricky stuff for everyone and handle it well there?

chlunde: I think that is a good idea, I've written some notes in #704. Closing this issue to follow up there.

PR #654, opened 2021-05-06
Add CloudFront Distribution managed resource

### LINKED ISSUE #236: Add CloudFront as a managed resource

Cloudfront is a popular CDN offering from AWS and a lot of users do use it for their web services which are public.
It would be nice for provider-aws to support this

This provider can have a managed resource Cloudfront to provision and manage AWS Cloudfront

ulucinar:
This PR aims to introduce a managed resource for CloudFront's Distribution API. 

Fixes #236 

I have:

- [x] Implement tests
- [x] Implement update logic

Manually deployed a distribution via `v1alpha1.Distribution` resource. Updated its `comment` field. Disabled and reenabled it and disabled it via  its `enabled` field. Finally, deleted it.


pkg/controller/cloudfront/distribution/setup.go
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		cr.SetConditions(xpv1.Unavailable())
+	}
+
+	return eo, nil
+}
+
+func preUpdate(_ context.Context, cr *svcapitypes.Distribution, udi *svcsdk.UpdateDistributionInput) error {
+	udi.Id = awsclients.String(meta.GetExternalName(cr))
+	udi.SetIfMatch(awsclients.StringValue(cr.Status.AtProvider.ETag))
```
muvaf, line 144: Is using `ETag` required and how frequently does it change? If it's as frequent as each request, we might get re-queued constantly.

pkg/clients/lateinit.go
```diff
@@ -0,0 +1,373 @@
...
+		return s
+	}
+}
+
+// LateInitializeFromResponse Copy unset (nil) values from responseObject to crObject
+//   Both crObject and responseObject must be pointers to structs.
+//	 Otherwise, an error will be returned. Returns `true` if at least one field has been stored
+//   from source `responseObject` into a corresponding field of target `crObject`.
+// nolint:gocyclo
+func LateInitializeFromResponse(parentName string, crObject interface{}, responseObject interface{},
```
muvaf, line 165: It seems that this library is generic enough to be used by other resources as well but I'm not sure if folks would be comfortable with this level of complexity. From my experience, generalization through loosely typed methods, like reflect or json, leads to bugs that are very hard to debug for inexperienced contributors (and time-consuming for others), which happens once in a while with `CreateJSONPatch` implementation we have for updates. We put exceptions in place per-resource to fix them but this leads to either requiring all cases to be tested for that specific resource or letting customers run into them and report back, which is not an experience we'd like users to have. Additionally, we have not tested for it, but it's in the hot execution path (called in every reconcile), so the performance implication is probably higher than simple `if` conditions.

Because of these reasons, we're relying more on the code generation as much as possible to eliminate these disadvantages of implicit methods as much as possible. https://github.com/aws-controllers-k8s/code-generator/pull/9 is the PR that adds support to ACK for generating the `lateInitialize` functions and https://github.com/aws-controllers-k8s/code-generator/pull/17 adds support for `isUpToDate` check in a strongly typed manner. I'll get back to those PRs in the coming days and we'll hopefully be able to use them for all generated resources.

pkg/clients/aws.go
```diff
@@ -445,6 +445,13 @@ func Int64Value(v *int64) int64 {
 	return 0
 }
 
+// PInt64Value converts an int to a *int64
+func PInt64Value(v int) *int64 {
```
muvaf, line 449: This seems like duplicate of `awsclients.Int64`

pkg/clients/lateinit.go
muvaf: We have synced with @ulucinar offline. The code generated by https://github.com/aws-controllers-k8s/code-generator/pull/9 will supersede this functionality. Until that PR is merged, we'll keep this function private to cloudfront controller and once it's merged, we'll replace it with the generated code.

muvaf (commented):
@ulucinar Looks pretty close to merging, could you respond to the remaining comments?

pkg/controller/cloudfront/distribution/setup.go
```diff
@@ -0,0 +1,182 @@
...
+				},
+			}),
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+func preCreate(_ context.Context, cr *svcapitypes.Distribution, cdi *svcsdk.CreateDistributionInput) error {
+	cdi.DistributionConfig.CallerReference = awsclients.String(string(cr.UID))
+	cdi.DistributionConfig.Origins.Quantity =
+		awsclients.PInt64Value(len(cr.Spec.ForProvider.DistributionConfig.Origins.Items))
```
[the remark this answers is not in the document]
ulucinar: Hi @muvaf,
Thanks for the review. 
I have removed the `aws.PInt64Value` function in favor of `aws.Int64` as suggested. 

Regarding the nil check, do you mean `cr.DistributionConfig.Origins` instead of `cdi`? The SDK type for the fields `DistributionConfig` and `Origins` both have the yaml tags `required`. Does that guarantee them to be non-nil?

pkg/controller/cloudfront/distribution/setup.go
```diff
@@ -0,0 +1,182 @@
...
+
+	meta.SetExternalName(cr, awsclients.StringValue(cdo.Distribution.Id))
+	ec.ExternalNameAssigned = true
+
+	return ec, nil
+}
+
+func preObserve(_ context.Context, cr *svcapitypes.Distribution, gdi *svcsdk.GetDistributionInput) error {
+	gdi.Id = awsclients.String(meta.GetExternalName(cr))
+
```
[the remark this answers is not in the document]
ulucinar: Very old habit from Deitel & Deitel programming books. The idea is to have all declarations, control blocks, etc. stand out for "ease of reading". But this's not aligned with go conventions. The more space style is easier to my eyes and because `gofmt -s` does not remove them (except for repeated empty lines), I have liberally used them but agreed, I will be more careful the repo conventions.

pkg/controller/cloudfront/distribution/setup.go
ulucinar: Yes, it's required. If not set, the request is invalid and results in a HTTP-400 error if I don't remember wrong. My observation is that (again if I recall correctly), if the SDK object is updated, it's assigned a new ETag and you need to supply that ETag in your requests.

pkg/controller/cloudfront/distribution/setup.go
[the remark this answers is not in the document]
muvaf: I meant `cdi.DistributionConfig.Origins` which is defined in AWS SDK and a pointer, i.e. the code in our repo sets its value, hence can be `nil`. But I also checked `cr.Spec.ForProvider.DistributionConfig.Origins` and it seems that's [not marked as `required`](https://github.com/crossplane/provider-aws/pull/654/files#diff-92efa1c153bb84bc723bdcd87cc53bb9b584ab93e4a2bf9fe6b5dd244356ea95R378) in CRD YAML that's been generated (probably because we yet to produce `required` markers for types in `types.go` file)

One way to be sure could be counting `cdi.DistributionConfig.Origins.Quantity.Items` if `cdi.DistributionConfig.Origins` is not `nil` because `Quantity` field is bound to `Items` on that struct so you'd need to have two `if`s.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/crossplane/provider-aws
 go 1.13
 
 require (
+	github.com/aws-controllers-k8s/runtime v0.1.1
```
muvaf, line 6: Oh and you need to regenerate using the latest ACK code-generator that has my fixes that removes this dependency.

pkg/controller/cloudfront/distribution/setup.go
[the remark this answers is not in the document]
ulucinar: Good observation @muvaf. The assumption I made was  `cr.Spec.ForProvider.DistributionConfig.Origins` and in turn `cdi.DistributionConfig.Origins` can never be `nil` because it's marked as required in AWS API. Why don't we also mark it as required in generated code? We now check if the CR has non-nil `Origins`.

Issue #647, opened 2021-05-03
Key (kms.aws.crossplane.io) is not creating new resources


EladDolev:
### What happened?

Unable to create a new KMS key, not even with the committed example. Getting `failed to describe Key: NotFoundException: Invalid keyId dev-key`

If I set `metadata.name` to an already existing KeyId, it seems to work fine, and the same if I use an arbitrary name and set the `crossplane.io/external-name` annotation to an already existing KeyId

### How can we reproduce it?
Apply the example from `examples/kms/key.yaml`

### What environment did it happen in?
Crossplane version: v1.2.0
AWS provider: 0.18.0
Kubernetes version: v1.18.16
Kubernetes distribution: kops
OS: 20.04.2 LTS (Focal Fossa)
Kernel: 5.4.0-1038-aws

bandesz:
Same here, tried to use the resource in a composition:

```
  Warning  CannotObserveExternalResource  47s (x8 over 2m44s)  managed/key.kms.aws.crossplane.io  failed to describe Key: NotFoundException: Invalid keyId kms-test-cjs55-gzftn
```

Crossplane version: v1.2.1
AWS Provider: 0.18.0

PR #659, opened 2021-05-10
In the KMS generator config define a custom error for NotFoundException + regenerate code

bandesz:

The KMS Key resource is broken and you can't create a KMS key with it. Fixes: #647

After looking at the code it seems the IsNotFound() function was generated incorrectly:

https://github.com/crossplane/provider-aws/blob/1ad7872a1b8f19cbd56f020e780ee698198163dc/pkg/controller/kms/key/zz_conversions.go#L147 

I added a custom exception to the generator config. Although the API returns a 400 http code if the kms key does not exist, I still had to use 404 in the config, which seemed to do the trick.

I followed the instructions from CODE_GENERATION.md and ran these two commands:

In code-generator:
```
go run -tags codegen cmd/ack-generate/main.go crossplane kms --provider-dir /.../src/github.com/crossplane/provider-aws
```

In provider-aws:
```
make generate
```

I'm not sure I generated the code correctly as there are more changes that I would expect.

I have:

I've built and deployed the provider to a real environment and tested KMS key creation, with and without an external-name.


muvaf (approved):
LGTM! Thanks @bandesz !

PR #669, opened 2021-05-17
Secretsmanager: Implement LateInitialize of a K8s secret when AWS secret already exists

### LINKED ISSUE #668: Secretsmanager: Implement LateInitialize of a K8s secret when AWS secret already exists

As described in #641, Secretsmanager does currently only sync a secret from K8s. It would be great if the controller creates a K8s secret during `LateInitialize` with the value of the AWS secret (if it exists).

`provider-aws` should implement `lateInitialize` in the secrets controller that checks whether the referenced K8s secret does exist.

1. If it does not exist, the controller should create a new K8s secret from the existing AWS secret (which does exist since lateInitialize is called)
2. If it does not exist, the controller should behave as it does now, for example raise an error if the K8s secret reference is not found.

MisterMX:
Fixes #668

The controller now creates a K8s with the value from AWS if it does not already exist.

I have:

Manually

### DIFF
--- a/pkg/controller/secretsmanager/secret/setup.go
+++ b/pkg/controller/secretsmanager/secret/setup.go
@@ -25,6 +25,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -56,6 +57,12 @@ const (
 	errDeleteResourcePolicy = "cannot delete resource policy"
 	errInvalidSpecPolicy    = "spec policy is invalid"
 	errInvalidCurrentPolicy = "current policy is invalid"
+	errParseSecretValue     = "cannot parse AWS secret value"
+	errGetAWSSecretValue    = "cannot get AWS secret value"
+	errCreateK8sSecret      = "canoot create secret in K8s"
+	errNoAWSValue           = "neither SecretString nor SecretBinary field is filled in the returned object"
+	errNoSecretRef          = "neither binarySecretRef nor stringSecretRef is given"
+	errOnlyOneSecretRef     = "only one of binarySecretRef or stringSecretRef must be set"
 )
 
 // SetupSecret adds a controller that reconciles a Secret.
@@ -66,6 +73,7 @@ func SetupSecret(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, p
 			e.preObserve = preObserve
 			e.postObserve = postObserve
 			h := &hooks{client: e.client, kube: e.kube}
+			e.lateInitialize = h.lateInitialize
 			e.isUpToDate = h.isUpToDate
 			e.preUpdate = h.preUpdate
 			e.preCreate = h.preCreate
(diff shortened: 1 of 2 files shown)


haarchri:
@MisterMX  can you rebase the MR ? thanks

MisterMX:
@haarchri done

haarchri:
@MisterMX sorry we merged #907 can you rebase again

MisterMX:
@haarchri  Done ;-)

pkg/controller/secretsmanager/secret/setup.go
```diff
@@ -113,6 +119,86 @@ type hooks struct {
...
+
+		sc := &corev1.Secret{}
+		sc.ObjectMeta.Name = ref.Name
+		sc.ObjectMeta.Namespace = ref.Namespace
+
+		data, err := getAWSSecretData(ref, valueResp)
+		if err != nil {
+			return errors.Wrap(err, errGetAWSSecretValue)
+		}
+		sc.Data = data
```
muvaf, line 150:
```suggestion
	data, err := getAWSSecretData(ref, valueResp)
	if err != nil {
		return errors.Wrap(err, errGetAWSSecretValue)
	}
	sc := &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      ref.Name,
			Namespace: ref.Namespace,
		},
		Data: data,
	}
```
nitpick: easier to follow.

pkg/controller/secretsmanager/secret/setup.go
```diff
@@ -113,6 +119,86 @@ type hooks struct {
...
+	}
+
+	parsed := map[string]string{}
+	err := json.Unmarshal(raw, &parsed)
+	if err != nil {
+		return nil, errors.Wrap(err, errParseSecretValue)
+	}
+
+	payload := map[string][]byte{}
+	if v, exists := parsed[awsclients.StringValue(ref.Key)]; exists {
```
muvaf, line 191: We send the data as a `string` if the key is given, I'd expect the reverse of this to happen in late-init, i.e. look for a string only if the key is given. If not, then we should always treat that the whole map is requested. So, I think we should only iterate here and not account for the case where there is a key, blob in AWS is a map and user actually meant a specific key in that whole blob in AWS.

pkg/controller/secretsmanager/secret/setup.go
```diff
@@ -172,18 +258,13 @@ func (e *hooks) isUpToDate(cr *svcapitypes.Secret, resp *svcsdk.DescribeSecretOu
...
-	var ref *svcapitypes.SecretReference
-	switch {
-	case cr.Spec.ForProvider.StringSecretRef != nil:
-		ref = cr.Spec.ForProvider.StringSecretRef
-	case cr.Spec.ForProvider.BinarySecretRef != nil:
-		ref = cr.Spec.ForProvider.BinarySecretRef
-	default:
-		return nil, errors.New("neither binarySecretRef nor stringSecretRef is given")
+func (e *hooks) getPayload(ctx context.Context, cr *svcapitypes.SecretParameters) ([]byte, error) {
+	ref, err := getSecretRef(cr)
```
muvaf, line 265: I think it's easier to follow what's happening if we have duplication for that switch case instead of calling another function.

pkg/controller/secretsmanager/secret/setup.go
muvaf: Oh in fact, it will never be non-nil anyway since we checked it in the beginning.

pkg/controller/secretsmanager/secret/setup.go
```diff
@@ -172,18 +258,13 @@ func (e *hooks) isUpToDate(cr *svcapitypes.Secret, resp *svcsdk.DescribeSecretOu
...
-func (e *hooks) getPayload(ctx context.Context, cr *svcapitypes.Secret) ([]byte, error) {
-	var ref *svcapitypes.SecretReference
-	switch {
-	case cr.Spec.ForProvider.StringSecretRef != nil:
-		ref = cr.Spec.ForProvider.StringSecretRef
-	case cr.Spec.ForProvider.BinarySecretRef != nil:
-		ref = cr.Spec.ForProvider.BinarySecretRef
-	default:
-		return nil, errors.New("neither binarySecretRef nor stringSecretRef is given")
+func (e *hooks) getPayload(ctx context.Context, cr *svcapitypes.SecretParameters) ([]byte, error) {
```
muvaf, line 264: `cr` seems to be used to point to the whole `Secret` object in this file, i.e. as in `(c)ustom (r)esource`. So, I'd suggest either keeping this line as is or use `params` as variable name when it addresses `SecretParameters`.

muvaf (commented):
We haven't done late-init for an input `Secret` before but I think it makes sense given that it's very similar to other params except that it's stored differently. Thanks @MisterMX !

pkg/controller/secretsmanager/secret/setup.go
MisterMX: I would prefer to keep code encapsulated that is used more than once.

I modified the function so it will return an error in case `StringSecretRef` and `BinarySecretRef` are both set. Before that, `StringSecretRef` took preference over `BinarySecretRef` and I don't think that was intended. I also added a note to the API docs.

pkg/controller/secretsmanager/secret/setup.go
MisterMX: Renamed to `params`.

MisterMX (commented):
Thanks @muvaf for the review. I added (most of) your requested changes and rebased on current master.

One last proposal before bumping this to `v1beta1`:

Currently we have a `binarySecretRef` and `stringSecretRef` next to each other. Wouldn't it be better if we'd have a single `secretRef` and a `format` property instead? This way we would be able to support different file formats, such as `yaml`, `json` etc.

pkg/controller/secretsmanager/secret/setup.go
```diff
@@ -113,6 +121,79 @@ type hooks struct {
 	kube   client.Client
 }
 
+func (e *hooks) lateInitialize(spec *svcapitypes.SecretParameters, resp *svcsdk.DescribeSecretOutput) error {
+	payload, err := e.getPayload(context.TODO(), spec)
+	if err := client.IgnoreNotFound(err); err != nil || payload == nil {
```
muvaf, line 126:
```suggestion
	if err := client.IgnoreNotFound(err); err != nil || payload != nil {
```
I think my initial code suggestion was the opposite of what you're trying to do in that if statement, sorry about that. We should `return nil` in case payload is full since we don't need to take any action.

muvaf (approved):
I'm merging this to unblock https://github.com/crossplane/provider-aws/pull/1107 before the release. I'll open a follow-up PR if my concern about payload being nil is indeed a problem. Thanks @MisterMX !!

PR #680, opened 2021-05-27
Add DBInstance

PocketMobsters:
Signed-off-by: Trenton Lawrence <pitime2@outlook.com>

Adds DBInstance. This resource is basically just `RDSInstance` with API fields consistent with resources such as `DBCluster` and `GlobalCluster`. 

Fixes #

I have:

I've created a number of DBInstances but there's a lot of things in this API. That said, quite a lot of it is the same code as in RDSInstance.


apis/rds/v1alpha1/generator-config.yaml
```diff
@@ -9,6 +9,16 @@ ignore:
...
+    - DescribeDBInstancesInput.DBInstanceIndentifier
+    - CreateDBInstanceInput.DBInstanceIdentifier
+    - ModifyDBInstanceInput.DBInstanceIdentifier
+    - DeleteDBInstanceInput.DBInstanceIdentifier
+    - CreateDBInstanceInput.MasterUserPassword
+    - ModifyDBInstanceInput.MasterUserPassword
+    - CreateDBInstanceInput.VpcSecurityGroupIds
+    - ModifyDBInstanceInput.VpcSecurityGroupIds
+    - CreateDBInstanceInput.DBSecurityGroups
+    - ModifyDBInstanceInput.DBSecurityGroups
```
muvaf, line 21:
```suggestion
```

I think we should be able to resolve to the generated fields, similar to `DomainIAMRoleName`

apis/rds/v1alpha1/referencers.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	database "github.com/crossplane/provider-aws/apis/database/v1beta1"
 	network "github.com/crossplane/provider-aws/apis/ec2/v1beta1"
+	"github.com/crossplane/provider-aws/apis/identity/v1beta1"
 	iamv1beta1 "github.com/crossplane/provider-aws/apis/identity/v1beta1"
```
muvaf, line 25:
```suggestion
	iamv1beta1 "github.com/crossplane/provider-aws/apis/identity/v1beta1"
```

Looks like duplicate import entry.

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (e *custom) postCreate(ctx context.Context, cr *svcapitypes.DBInstance, _ *svcsdk.CreateDBInstanceOutput, _ managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
+	if err != nil {
+		return managed.ExternalCreation{}, err
+	}
+	conn := managed.ConnectionDetails{
+		xpv1.ResourceCredentialsSecretUserKey: []byte(aws.StringValue(cr.Spec.ForProvider.MasterUsername)),
+	}
+	pw, _, err := rds.GetPassword(ctx, e.kube, cr.Spec.ForProvider.MasterUserPasswordSecretRef, cr.Spec.WriteConnectionSecretToReference)
+	if resource.IgnoreNotFound(err) != nil {
```
muvaf, line 109:
```suggestion
	if err != nil {
```

We shouldn't actually ignore this error since that's the only way user can provide a password.

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,190 @@
...
+			resp.DBInstances = append(resp.DBInstances, dbInstance)
+			break
+		}
+	}
+	return resp
+}
+
+func (e *custom) savePasswordSecret(ctx context.Context, cr *svcapitypes.DBInstance, pw string) error {
+	if cr.Spec.ForProvider.MasterUserPasswordSecretRef == nil {
+		return errors.New("no MasterUserPasswordSecretRef given, unable to save password")
```
muvaf, line 155:
```suggestion
		return errors.New("no MasterUserPasswordSecretRef given, unable to store generated password")
```

nitpick

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,190 @@
...
+	nn := types.NamespacedName{
+		Name:      ref.Name,
+		Namespace: ref.Namespace,
+	}
+	sc := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nn.Name,
+			Namespace: nn.Namespace,
+		},
+	}
```
muvaf, line 167:
```suggestion
	sc := &corev1.Secret{}
```

Metadata here will be filled out by the `Get` call.

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,190 @@
...
+	err := e.kube.Get(ctx, nn, sc)
+	var create bool
```
muvaf, line 169: I'd suggest using [APIPatchingApplicator](https://github.com/crossplane/crossplane-runtime/blob/84dc59d/pkg/resource/api.go#L124) which handles all these cases around making sure the object makes it to api-server in the form that we want.

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,190 @@
...
+	name := managed.ControllerName(svcapitypes.DBInstanceGroupKind)
+	opts := []option{
+		func(e *external) {
+			c := &custom{client: e.client, kube: e.kube}
+			e.preObserve = preObserve
+			e.postObserve = postObserve
+			e.preCreate = c.preCreate
+			e.postCreate = c.postCreate
+			e.preDelete = preDelete
+			e.filterList = filterList
```
muvaf, line 48: Could you use existing [LateInitialize](https://github.com/crossplane/provider-aws/blob/8aa68a2/pkg/clients/rds/rds.go#L368) and [IsUpToDate](https://github.com/crossplane/provider-aws/blob/8aa68a2/pkg/clients/rds/rds.go#L456) as well so that update works and we get feature-parity with existing `RDSInstance`?

apis/rds/v1alpha1/generator-config.yaml
PocketMobsters: There's a compiler error about type mismatching in the autogenerated code if I leave these as they are. I don't have the exact text but I added these after seeing that the DBCluster resource does the same thing for `VpcSecurityGroupIds`

muvaf (approved):
I have one blocking comment regarding `isUpToDate` and we need an example YAML in `examples/rds`. I'd also add a comment about the test scenarios you've done.

Thank you @PocketMobsters !

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func lateInitializeHook(in *svcapitypes.DBInstanceParameters, out *svcsdk.DescribeDBInstancesOutput) error {
+	if len(out.DBInstances) == 0 {
+		return errors.New("no DBInstance in DescribeDBInstancesOutput, can not lateIntitialize")
+	}
+	return lateInitialize(in, out.DBInstances[0])
+}
+
+func lateInitialize(in *svcapitypes.DBInstanceParameters, db *svcsdk.DBInstance) error { // nolint:gocyclo
```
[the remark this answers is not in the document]
PocketMobsters: Wonderful, will adjust

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,361 @@
...
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+func (e *custom) isUpToDate(cr *svcapitypes.DBInstance, out *svcsdk.DescribeDBInstancesOutput) (bool, error) {
+	// (PocketMobsters): because we don't have a context here, we cannot check
```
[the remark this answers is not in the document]
PocketMobsters: Perfect, I'll get that added in a second here

PocketMobsters:
Just rebased on top of master

PocketMobsters:
Working through the example DBInstance w/ @muvaf

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,373 @@
...
+		obj.MasterUserPassword = aws.String(pw)
+	}
+	return nil
+}
+
+func preDelete(_ context.Context, cr *svcapitypes.DBInstance, obj *svcsdk.DeleteDBInstanceInput) (bool, error) {
+	obj.DBInstanceIdentifier = aws.String(meta.GetExternalName(cr))
+	obj.FinalDBSnapshotIdentifier = aws.String(cr.Spec.ForProvider.FinalDBSnapshotIdentifier)
+	obj.SkipFinalSnapshot = aws.Bool(cr.Spec.ForProvider.SkipFinalSnapshot)
+	return false, nil
```
muvaf, line 161: There are some workarounds/hacks in `RDSInstance` code that should be ported over so that we don't need to fix them again. For example, deletion needs to call an update for the reasons listed [here](https://github.com/crossplane/provider-aws/blob/5a380ad/pkg/controller/database/rdsinstance.go#L240) and should skip making the deletion call if it's already in `deleting` state.

pkg/controller/rds/dbinstance/setup.go
```diff
@@ -0,0 +1,373 @@
...
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+func (e *custom) isUpToDate(cr *svcapitypes.DBInstance, out *svcsdk.DescribeDBInstancesOutput) (bool, error) {
+	db := out.DBInstances[0]
```
muvaf, line 264: One of the things we need to port over from `RDSInstance` is to [skip updating](https://github.com/crossplane/provider-aws/blob/5a380ad/pkg/controller/database/rdsinstance.go#L189) if a modification is on its way and it seems like this function could be the best place for now to do it.

muvaf (commented):
As we had chatted, there are some problems regarding update whose fix seems to exist in `RDSInstance` code already. There are a few other workarounds/hacks in place in `RDSInstance` code that we developed over time, some of them noted below. So, I'd suggest scanning the [controller code](https://github.com/crossplane/provider-aws/blob/5a380ad/pkg/controller/database/rdsinstance.go) for such things and add them as necessary. We can merge it once that update problem is solved (or if update is disabled for now by returning `true` in `isUpToDate` all the time since this is an alpha resource)

PR #681, opened 2021-05-28
Fix VPC CIDR Block Deletion

### LINKED ISSUE #677: Unable to delete the CIDR Block after VPC is deleted
### What happened?
CIDR blocks are attached to VPCs so deleting a VPC in Crossplane automatically deletes the CIDR block in AWS. However the CIDR resource does not get deleted in Crossplane. The deletion gets stuck in a loop since AWS first tries to get the VPC and when the VPC doesn't exist it sends a VPC not found error while the code is waiting for a CIDR not found error. 
We just need to account for the VPC not found error while deleting a CIDR block.

### How can we reproduce it?
1. Create a VPC in Crossplane
2. Create a CIDR block for that VPC
3. Delete the VPC in Crossplane (it wont ask you to delete the CIDR block)
4. Delete the CIDR block in Crossplane (it will give an error that the VPC is not found)

Crossplane version: 
def0bab5cb02c8c538abb40c6d0437419bf399a4

AmnaIrfan:
Signed-off-by: Amna Irfan <amirfan@ea.com>

CIDR blocks are attached to VPCs so deleting a VPC in Crossplane automatically deletes the CIDR block in AWS. However the CIDR resource does not get deleted in Crossplane. The deletion gets stuck in a loop since AWS first tries to get the VPC and when the VPC doesn't exist it sends a VPC not found error while the code is waiting for a CIDR not found error.
We just need to account for the VPC not found error while deleting a CIDR block.

Fixes #677

I have:

1. Create a VPC in Crossplane
2. Create a CIDR block for that VPC
3. Delete the VPC in Crossplane (it wont ask you to delete the CIDR block)
4. Delete the CIDR block in Crossplane (it will no longer give an error that vpc doesn't exist)

### DIFF
--- a/pkg/controller/ec2/vpccidrblock/controller.go
+++ b/pkg/controller/ec2/vpccidrblock/controller.go
@@ -104,8 +104,11 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	response, err := e.client.DescribeVpcsRequest(&awsec2.DescribeVpcsInput{
 		VpcIds: []string{aws.StringValue(cr.Spec.ForProvider.VPCID)},
 	}).Send(ctx)
-	if err != nil && !ec2.IsCIDRNotFound(err) {
-		return managed.ExternalObservation{}, awsclient.Wrap(err, errDescribe)
+
+	if err != nil {
+		return managed.ExternalObservation{
+			ResourceExists: false,
+		}, awsclient.Wrap(resource.Ignore(ec2.IsVPCNotFoundErr, err), errDescribe)
 	}
 
 	// in a successful response, there should be one and only one object


pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -104,8 +104,11 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	response, err := e.client.DescribeVpcsRequest(&awsec2.DescribeVpcsInput{
 		VpcIds: []string{aws.StringValue(cr.Spec.ForProvider.VPCID)},
 	}).Send(ctx)
-	if err != nil && !ec2.IsCIDRNotFound(err) {
-		return managed.ExternalObservation{}, awsclient.Wrap(err, errDescribe)
+
+	if err != nil {
+		return managed.ExternalObservation{
+			ResourceExists: false,
+		}, awsclient.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
```
muvaf, line 111:
```suggestion
		}, awsclient.Wrap(resource.Ignore(ec2.IsVPCNotFound, err), errDescribe)
```

I think that's probably the only error we should take into consideration here since the call is `DescribeVpc`.

muvaf (approved):
Thanks @AmnaIrfan !

PR #683, opened 2021-05-31
Fixes #682 draft of potential compatibility field

### LINKED ISSUE #682: HostedZoneStatus (atProvider.hostedZone.id) returning a prefix along with HostedZoneId
while working on the creation of the hosted zone, was trying to get hosted zone id HostedZoneStatus (atProvider.hostedZone.id), 
found that it is returning a prefix along with the id  something like below, now this makes it less useful for further services which only requires the ID and currently transforms are limited hence string transformations are not supported hence removing this is not possible to fetch the id alone.

Hosted Zone Id:         /hostedzone/Z13322DRRSS

i see in the code below 
https://github.com/crossplane/provider-aws/blob/master/pkg/clients/hostedzone/hostedzone.go

// IDPrefix is the prefix of the actual ID that's returned from GET call.const IDPrefix = "/hostedzone/"

can this be removed from HostedZoneStatus (as currently adding some prefix using transform is still possible)  or at least made "hostedzone/" thereby would solve the issue of using hosted zone temp as seen below extra / is making problem.

below "/" infront of hosted zone needs to be removed or preferably /hostedzone/ should be removed from status 
"Resource": [                                                                                                               
         "arn:aws:route53:::/hostedzone/Z00151212Z97ZJCPKUPPH"                                                                     
      ]

martin-ducar-gd:

Fixes #682 draft of potential compatibility field

I have:

TODO

### DIFF
--- a/apis/route53/v1alpha1/hostedzone_types.go
+++ b/apis/route53/v1alpha1/hostedzone_types.go
@@ -147,6 +147,9 @@ type HostedZoneResponse struct {
 	// it.
 	ID string `json:"id,omitempty"`
 
+	// CompatibiltyID with zone ID only for other resources like for example IAM.
+	CompatID string `json:"compatId,omitempty"`
+
 	// LinkedService is the service that created the hosted zone.
 	LinkedService LinkedService `json:"linkedService,omitempty"`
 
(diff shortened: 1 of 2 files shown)


AaronME:
Closing this as stale. Please re-open if this is still an issue.

PR #687, opened 2021-05-31
Update code generation guide with make target

muvaf:

I updated the guide to use the new `make services SERVICES=rds,kms` target and also added `make services.all` target to run generator for all existing services. I plan to add `make services.all` to `make generate` but that will come in with another PR as the diff is quite large. Once that's merged, we'll make sure all generated code is kept up-to-date by `make reviewable` and `make check-diff`

I have:

Manually.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ PROJECT_REPO := github.com/crossplane/$(PROJECT_NAME)
 
 PLATFORMS ?= linux_amd64 linux_arm64
 
-CODE_GENERATOR_COMMIT ?= v0.1.0
+CODE_GENERATOR_COMMIT ?= v0.2.1
+GENERATED_SERVICES="apigatewayv2,cloudfront,dynamodb,efs,kms,lambda,rds,secretsmanager,servicediscovery,sfn"
 
 # -include will silently skip missing files, which allows us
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -115,20 +116,26 @@ run: go.build
 
 .PHONY: cobertura reviewable manifests submodules fallthrough test-integration run crds.clean
 
-service: 
-	@if [ "$(ServiceID)" = "" ]; then \
-		echo "Error: Please specify value of 'ServiceID'."; \
+# NOTE(muvaf): ACK Code Generator is a separate Go module, hence we need to
```
hasheddan, line 119: @muvaf can we pin this in our `go.mod` and invoke it with `go generate` like we do for other tools? (i.e. `crossplane-tools`)

muvaf: The code generator needs non-Go files (templates) that exist in its repo to work. But when we run `go run` without cloning, those files do not come in. We get the following error:
```
Error: template crossplane/apis/doc.go.tpl not found in any base search path: Template Not Found
exit status 1
```

Seems like this can be achieved with the new [`go:embed`](https://tip.golang.org/pkg/embed/) directive, but that's not used in ACK yet.

hasheddan (approved):

PR #698, opened 2021-06-03
Regenerate APIs with required spec, optional status.atProvider

### LINKED ISSUE #697: All provider-aws CRDs should be conformant
### What happened?

We'd like to certify provider-aws as a [conformant Crossplane provider](https://github.com/cncf/crossplane-conformance). My first pass at running the conformance tests failed due to some non-conformant CRD schemas, which you can see at https://github.com/cncf/crossplane-conformance/issues/4. Specifically it looks like:

* A whole bunch of resources mark `status.atProvider` as a required field. We prefer not to mark any status fields as "required", because we don't expect humans (or any entity other than the managed resource controller) to write those statuses.
* Several managed resources don't require a `spec`. I imagine there are some edge cases where a `spec` really is optional, but for the sake of being able to test that the broad majority of resources are doing the right thing I think we should keep requiring `spec`. I believe if `spec` is required but none of the `spec` fields are required it's possible to write `spec: {}` to indicate this.

### How can we reproduce it?

Follow the instructions at https://github.com/cncf/crossplane-conformance/blob/main/instructions.md for Crossplane 1.2 and provider-aws v0.18.1.

Crossplane version: 1.2
Provider version: 0.18.1

negz:

Fixes #697
Depends on aws-controllers-k8s/code-generator#81

In the vast majority of resources a spec is required, so I'd like to mark it as such. I believe that in the edge cases where the spec really contains only optional fields it's possible to write spec: {}. Note that technically making `spec` required is a breaking change.

We prefer not to mark status fields as required, since resources will always exist at some point without their status and we consider 'requireness' to be more of a human-facing constraint than a software-facing constraint that our managed resource controllers should be subject to.

I have:

I don't imagine the `atProvider` change will affect anyone. I'd like to assess how many (if any) of the affected resources _really_ could have had an empty spec and still functioned. If any exist, I'd like to confirm my belief that folks can write `spec: {}` to indicate that their managed resource doesn't need a spec.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,8 @@ crds.clean:
 	@find package/crds -name '*.yaml.sed' -delete || $(FAIL)
 	@$(OK) cleaned generated CRDs
 
-generate.run: go.generate crds.clean
+generate.init: services.all
+generate.done: crds.clean
 
 manifests:
 	@$(WARN) Deprecated. Please run make generate instead.
--- a/Makefile
+++ b/Makefile
@@ -79,16 +79,6 @@ crds.clean:
 
 generate.run: go.generate crds.clean
 
-# Ensure a PR is ready for review.
-reviewable: services.all generate lint
-	@go mod tidy
-
-# Ensure branch is clean.
-check-diff: reviewable
-	@$(INFO) checking that branch is clean
-	@test -z "$$(git status --porcelain)" || $(FAIL)
-	@$(OK) branch is clean
-
 manifests:
 	@$(WARN) Deprecated. Please run make generate instead.
 
@@ -114,7 +104,7 @@ run: go.build
 	@# To see other arguments that can be provided, run the command with --help instead
 	$(GO_OUT_DIR)/provider --debug
 
-.PHONY: cobertura reviewable manifests submodules fallthrough test-integration run crds.clean
+.PHONY: cobertura manifests submodules fallthrough test-integration run crds.clean
 
 # NOTE(muvaf): ACK Code Generator is a separate Go module, hence we need to
 # be in its root directory to call "go run" properly.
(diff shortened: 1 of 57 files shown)


negz (commented):
> I'd like to assess how many (if any) of the affected resources really could have had an empty spec and still functioned.

It appears that all generated resources require `spec.forProvider.region`, which implies that none of these would have worked with an empty spec. I think in practice that means this change will not be breaking for anyone.

Makefile
```diff
@@ -6,7 +6,7 @@ PROJECT_REPO := github.com/crossplane/$(PROJECT_NAME)
 
 PLATFORMS ?= linux_amd64 linux_arm64
 
-CODE_GENERATOR_COMMIT ?= v0.2.1
+CODE_GENERATOR_COMMIT ?= c4592885b843b2f0a316718a15349ea07b3a50b1
```
negz, line 9: How do we feel about pinning to a commit rather than a specific release? It looks we just missed out on a code-generator release yesterday.

muvaf: I think that's fine.

Makefile
```diff
@@ -79,16 +79,6 @@ crds.clean:
 
 generate.run: go.generate crds.clean
```
muvaf, line 81:
```suggestion
generate.run: services.all go.generate crds.clean

```

We intentionally override `reviewable` here to make sure generated code is up to date. Maybe add `services.all` to `generate.init` or `generate.run`?

muvaf (changes requested):

Makefile
negz: Ah right - will fix.

negz: I took a stab at this, but it's not quite working how I'd like. For whatever reason `go generate` runs _before_ `services.all` - rather than after possibly because it's part of the Go make libraries?

negz: Found a way to fix this.

Makefile
A-Hilaly: `v0.2.2` has been released and includes the recent crossplane changes :) https://github.com/aws-controllers-k8s/code-generator/releases/tag/v0.2.2

muvaf (approved):
Thanks @negz !

PR #701, opened 2021-06-07
Fix ECR Repo policy eternally updating

### LINKED ISSUE #700: ECR Policy interaction with AWS causes update loops
### What happened?
Implementing ECR Repository Policies atop some existing ECR Repo definitions, I was using the `accountId` principal to set permissions for a repo from particular AWS accounts.

I was notified this morning of excess AWS Config charges from calls to `SetRepositoryPolicy`: 700,000 over the course of 10 days per Repo 😄 

Digging around, it looks like there's 2 cases where the ECR Repository Policy does not correctly detect the resource as up to date, and repeatedly triggers an Update call:

1. When using the `awsAccountId` principal, AWS documentation says you can specify either the plain account ID, OR the ARN of the root of the account, i.e. the below, but AWS converts the raw account ID to an ARN string internally and appears to _always_ return the ARN. `SerializeAWSPrincipal` just returns the plain user input as a string, so if the user specifies a plain account ID rather than an ARN, these will never be detected as up to date.
    - `"Principal": { "AWS": "arn:aws:iam::123456789012:root" }` or
    - `"Principal": { "AWS": "123456789012" }` are functionally equal.
2. Principals are not sorted before or during comparison, so if AWS returns the principals in a different order to the user input they will also never be detected as up to date.

### How can we reproduce it?
Create an ECR repo (settings unimportant), and then create a `RepositoryPolicy` referring to it with the following:
```yaml
apiVersion: ecr.aws.crossplane.io/v1alpha1
kind: RepositoryPolicy
metadata:
  name: my-never-uptodate-policy
spec:
  forProvider:
    policy:
      id: AllowAccounts
      statements:
      - action:
        - ecr:BatchCheckLayerAvailability
        - ecr:BatchGetImage
        - ecr:GetDownloadUrlForLayer
        effect: Allow
        principal:
          awsPrincipals:
          - awsAccountId: 123456789012
          - awsAccountId: arn:aws:iam::123456789013:root
        sid: AllowPull
      version: "2008-10-17"
    region: us-east-1
    repositoryName: <name-of-your-ecr-repo>
```

The status object of the resource will look something like this:

```yaml
status:
  atProvider:
    policyText: |-
      {
        "Version" : "2008-10-17",
        "Id" : "AllowAccounts",
        "Statement" : [ {
          "Sid" : "AllowPull",
          "Effect" : "Allow",
          "Principal" : {
            "AWS" : [ "arn:aws:iam::123456789013:root", "arn:aws:iam::123456789012:root" ]
          },
          "Action" : [ "ecr:BatchCheckLayerAvailability", "ecr:BatchGetImage", "ecr:GetDownloadUrlForLayer" ]
        } ]
      }
```

Note that the principals are always an ARN and they're not in the same order as specified by the resource.

AWS Provider version: v0.18.1

### Suggested fix
For issue 1, maybe we can convert the raw account ID into an ARN within crossplane (`arn:aws:iam::%s:root`) as I doubt the format will ever change, or if we leave as-is, make sure the behaviour is documented to avoid unintended update behaviour.

For issue 2, we should supply a `cmpopts.SortSlices` function to `IsRepositoryPolicyUpToDate()`, to make sure that principal slices are sorted before comparison - something like the below:

```go
sortSlicesOpt := cmpopts.SortSlices(func(x, y interface{}) bool {
	return x.(string) < y.(string)
})
```

Happy to submit a PR for these if approach makes sense.

benagricola:
Resolves a couple of issues around detecting changes on the ECR
`RepositoryPolicy` type:
  1. Converts `awsAccountId` principals that were not specified as an
     ARN into the ARN format, as AWS always returns in this format from
     the API.
  2. Makes sure to sort string slices within the policy document during
     comparison, to avoid issues where AWS returns values in a different
     order.

Fixes #700

Signed-off-by: Ben Agricola <bagricola@squiz.co.uk>
I have:

Additional test added to compare policies with defined principals as a string slice, that tests both of these cases.
That test fails without the fixes applied.

### DIFF
--- a/pkg/clients/ecr/repository_policy.go
+++ b/pkg/clients/ecr/repository_policy.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
+	"strconv"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
@@ -151,9 +151,10 @@ func SerializeAWSPrincipal(p v1alpha1.AWSPrincipal) *string {
 		// Note: AWS Docs say you can specify the account ID either
 		// raw or as an ARN, but AWS actually converts internally to
 		// the ARN format, which is problematic for checking if we're
-		// up to date. So here we just do the conversion ourselves to
-		// avoid unintended behaviour if user specifies raw account ID.
-		if !strings.HasPrefix(*p.AWSAccountID, "arn:aws:iam::") {
+		// up to date. So here we just do the conversion ourselves if
+		// we were given a string containing a number that looks like an
+		// AWS account ID (looks like a 12-digit integer).
+		if _, err := strconv.ParseInt(*p.AWSAccountID, 10, 64); err == nil {
 			s := fmt.Sprintf("arn:aws:iam::%s:root", *p.AWSAccountID)
 			return &s
 		}
--- a/pkg/clients/ecr/repository_policy.go
+++ b/pkg/clients/ecr/repository_policy.go
@@ -221,7 +221,16 @@ func IsRepositoryPolicyUpToDate(local, remote *string) bool {
 	}
 
 	sortSlicesOpt := cmpopts.SortSlices(func(x, y interface{}) bool {
-		return x.(string) < y.(string)
+		if a, ok := x.(string); ok {
+			if b, ok := y.(string); ok {
+				return a < b
+			}
+		}
+		// Note: Unknown types in slices will not cause a panic, but
+		// may not be sorted correctly. Depending on how AWS handles
+		// these, it may cause constant updates - but better this than
+		// panicing.
+		return false
 	})
 	return cmp.Equal(localUnmarshalled, remoteUnmarshalled, cmpopts.EquateEmpty(), sortSlicesOpt)
 }
--- a/pkg/clients/ecr/repository_policy.go
+++ b/pkg/clients/ecr/repository_policy.go
@@ -4,11 +4,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/awserr"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 
 	"github.com/crossplane/provider-aws/apis/ecr/v1alpha1"
 	awsclient "github.com/crossplane/provider-aws/pkg/clients"
@@ -146,6 +148,15 @@ func SerializeRepositoryPrincipal(p *v1alpha1.RepositoryPrincipal) (interface{},
 func SerializeAWSPrincipal(p v1alpha1.AWSPrincipal) *string {
 	switch {
 	case p.AWSAccountID != nil:
+		// Note: AWS Docs say you can specify the account ID either
+		// raw or as an ARN, but AWS actually converts internally to
+		// the ARN format, which is problematic for checking if we're
+		// up to date. So here we just do the conversion ourselves to
+		// avoid unintended behaviour if user specifies raw account ID.
+		if !strings.HasPrefix(*p.AWSAccountID, "arn:aws:iam::") {
+			s := fmt.Sprintf("arn:aws:iam::%s:root", *p.AWSAccountID)
+			return &s
+		}
 		return p.AWSAccountID
 	case p.IAMRoleARN != nil:
 		return p.IAMRoleARN
@@ -209,7 +220,10 @@ func IsRepositoryPolicyUpToDate(local, remote *string) bool {
 		return false
 	}
 
-	return cmp.Equal(localUnmarshalled, remoteUnmarshalled)
+	sortSlicesOpt := cmpopts.SortSlices(func(x, y interface{}) bool {
+		return x.(string) < y.(string)
+	})
+	return cmp.Equal(localUnmarshalled, remoteUnmarshalled, cmpopts.EquateEmpty(), sortSlicesOpt)
 }
 
 // RawPolicyData parses and formats the RepositoryPolicy struct
--- a/pkg/clients/ecr/repository_policy_test.go
+++ b/pkg/clients/ecr/repository_policy_test.go
@@ -20,7 +20,10 @@ var (
 	boolCheck   = true
 	testID      = "id"
 	policy      = `{"Statement":[{"Action":"ecr:ListImages","Effect":"Allow","Principal":"*"}],"Version":"2012-10-17"}`
-	params      = v1alpha1.RepositoryPolicyParameters{
+	cpxPolicy   = `{"Statement":[{"Action":"ecr:ListImages","Effect":"Allow","Principal":{"AWS":["arn:aws:iam::111122223333:userARN","111122223334","arn:aws:iam::111122223333:roleARN"]}}],"Version":"2012-10-17"}`
+	// Note: different sort order of principals than input above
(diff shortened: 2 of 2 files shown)


pkg/clients/ecr/repository_policy.go
```diff
@@ -209,7 +220,10 @@ func IsRepositoryPolicyUpToDate(local, remote *string) bool {
 		return false
 	}
 
-	return cmp.Equal(localUnmarshalled, remoteUnmarshalled)
+	sortSlicesOpt := cmpopts.SortSlices(func(x, y interface{}) bool {
+		return x.(string) < y.(string)
```
benagricola, line 224: Not sure we can get away with this, as anything in a slice that isn't a string will cause a panic. Every example I can find of an AWS IAM policy has all values as strings and I've not found any other documentation to suggest you can pass non-strings to a policy (even when comparing numbers, they're sent as quoted strings).

Feedback appreciated.

pkg/clients/ecr/repository_policy.go
```diff
@@ -146,6 +148,15 @@ func SerializeRepositoryPrincipal(p *v1alpha1.RepositoryPrincipal) (interface{},
 func SerializeAWSPrincipal(p v1alpha1.AWSPrincipal) *string {
 	switch {
 	case p.AWSAccountID != nil:
+		// Note: AWS Docs say you can specify the account ID either
+		// raw or as an ARN, but AWS actually converts internally to
+		// the ARN format, which is problematic for checking if we're
+		// up to date. So here we just do the conversion ourselves to
+		// avoid unintended behaviour if user specifies raw account ID.
+		if !strings.HasPrefix(*p.AWSAccountID, "arn:aws:iam::") {
+			s := fmt.Sprintf("arn:aws:iam::%s:root", *p.AWSAccountID)
```
muvaf, line 157: Is this always `root`?

pkg/clients/ecr/repository_policy.go
muvaf: We can handle the input types we know and if an unexpected type came in, returning `false` would be better than panic IMO.

muvaf (commented):
Thanks for looking into this @benagricola !

pkg/clients/ecr/repository_policy.go
benagricola: I think when specifying only the account ID, yes - `root` is assumed (from the IAM docs) and means any role in the account in question:

```
For example, given an account ID of 123456789012, you can use either of the following methods to specify that account in the Principal element:

"Principal": { "AWS": "arn:aws:iam::123456789012:root" }

"Principal": { "AWS": "123456789012" }
```

benagricola:
@muvaf latest commit conditionally sorts or returns false based on whether the interfaces can be converted to string or not. I've only implemented string as I _believe_ it is the only valid format, but again - this is an assumption made based on the doco, not knowledge 🙂

pkg/clients/ecr/repository_policy.go
```diff
@@ -209,7 +220,19 @@ func IsRepositoryPolicyUpToDate(local, remote *string) bool {
...
+			if b, ok := y.(string); ok {
+				return a < b
+			}
+		}
+		// Note: Unknown types in slices will not cause a panic, but
+		// may not be sorted correctly. Depending on how AWS handles
+		// these, it may cause constant updates - but better this than
+		// panicing.
+		return false
+	})
```
chlunde, line 234:
```suggestion
	sortSlicesOpt := cmpopts.SortSlices(func(x, y string) bool { return x < y })
```

I think cmpopts checks the signature and types here. https://play.golang.org/p/RlhDKTVo0-T

benagricola: Mentioned in thread in slack but this behaves differently, looks like it aborts sorting the entire slice as soon as it comes across a value that isn't a string.

Don't know what the correct behaviour here is, if we're happy that slices containing non-strings will be left (partially?) unsorted then I'm happy to change this, it just means in cases where unknown types were already in the right position in the slice we might end up returning `false` instead of `true` and causing an update.

chlunde: I'm happy with the code as it is, I assumed it would be string only because the schema is string only. As long as we end up with one implementation of "compare policy" in the code base I think it's OK if it is very "defensively coded". But if we get 10 copies I think it would be best to handle only cases we know can happen :)

pkg/clients/ecr/repository_policy.go
chlunde: I think it _might_ be a good idea to check if the account id is numeric only before doing this re-format. It will be more future-compatible and if someone makes mistake, like copy the output from "get-caller-identity", it will not format it, for example `arn:aws:sts::123:root` as `arn:aws:iam::arn:aws:sts:123:root:root`

benagricola: @chlunde Yeah that's a fair point. Given that AWS account numbers are 12 numeric digits I think we can just do the following:

```go
if _, err := strconv.ParseInt(*p.AWSAccountID, 10, 64); err == nil {
	s := fmt.Sprintf("arn:aws:iam::%s:root", *p.AWSAccountID)
	return &s
}
```

rather than checking for a string prefix. Anything with non-numerics should error and fall back to no re-format.

pkg/clients/ecr/repository_policy.go
muvaf: If we expect non-string types, would it make sense to supply a sort option for each type we expect? Something like the following:
```go
sortSlicesOpts := cmp.Options{
		cmpopts.SortSlices(func(x, y string) bool { return x < y }),
		cmpopts.SortSlices(func(x, y int) bool { return x < y }),
		cmpopts.SortSlices(func(x, y bool) bool { return x != y }),
	}
return cmp.Equal(localUnmarshalled, remoteUnmarshalled, cmpopts.EquateEmpty(), sortSlicesOpts)
```

I don't know how `cmp` behaves with JSON though, so might be totally off base here.

benagricola: @muvaf According to all of the doco I've seen, and the policy grammar that @chlunde found (https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_grammar.html) it seems that _everything_ should be a string, including policy conditions that contain numbers or booleans - so I don't think we necessarily have to sort non-string types like that, as they may well be completely rejected by AWS anyway.

We do however need to _accept_ them without error, as a user can supply whatever JSON document they want.

Rather than adding additional types which I think will probably be rejected by the API anyway, would it be better if we simply provide an `interface{}` sorter that always returns false, alongside the basic string sorter? That way, we should always complete sorting the input _strings_ even if we encounter other types, and let the API deal with the fallout.

```go
sortSlicesOpts := cmp.Options{
	cmpopts.SortSlices(func(x, y string) bool { return x < y }),
	cmpopts.SortSlices(func(x, y interface{}) bool { return false }),
}
return cmp.Equal(localUnmarshalled, remoteUnmarshalled, cmpopts.EquateEmpty(), sortSlicesOpts)
```

muvaf: Sounds good to me. Let me know when it's ready to merge.

benagricola: I've done a load of testing on this and the above code sample doesn't work - and neither does @chlunde's original suggestion. I suspect this is because the values as seen by `cmp.Equal` are never directly strings, but strings inside interfaces, because they're generically unmarshalled - so we can't detect the variable type as part of the options sort.

The existing code in this PR _does_ do what we want but is a little more verbose because of this limitation.

chlunde: @benagricola sorry for the extra work, I think you're right. `cmpopts` doesn't check the types of each element when we have `[]interface{}`: https://play.golang.org/p/XdzfTqK_idZ - so most likely we have a `[]interface{}` and not `[]string`

pkg/clients/ecr/repository_policy.go
benagricola: Resolved by attempting to convert `x` and `y` to strings and returning `false` if either variable is an unknown type.

benagricola:
@muvaf I think this is good to go - I've left all the changes in, as I think @chlunde planned on using this as a base for merging the policy code from ECR and S3 (and maybe others?).

Any chance we can merge this before #704 is implemented as it should at least fix the underlying issue for ECR repos 😃

muvaf (approved):
Thanks @benagricola !

PR #702, opened 2021-06-08
Modifications in example resources for conformance tests

ulucinar:
Signed-off-by: Alper Rifat Ulucinar <ulucinar@users.noreply.github.com>

Changes done in some example resources for conformance tests. Some had syntax errors (like fields nested under wrong elements), or API versioning errors, and some changes were done for ease of operation (like using ARNs instead of selectors).

I have:

Resources have been applied against a cluster where provider-aws from commit 4dafd3fcace2328d8280a78cb9c0227f41eeeb9b is running. Successfully observed transition to `Ready` state. Also successfully observed deletion of the managed resources except for `Key.kms`. The problem with `Key.kms` is that although the corresponding AWS KMS CMK is successfully scheduled for deletion, we have to wait for a mandatory interval in the order of days for the CMK to be actually deleted:
https://docs.aws.amazon.com/kms/latest/developerguide/deleting-keys.html#deleting-keys-how-it-works

### DIFF
--- a/examples/ec2/natgateway.yaml
+++ b/examples/ec2/natgateway.yaml
@@ -1,14 +1,16 @@
-apiVersion: ec2.aws.crossplane.io/v1alpha1
+apiVersion: ec2.aws.crossplane.io/v1beta1
 kind: NATGateway
 metadata:
   name: sample-natgateway
 spec:
   forProvider:
     region: us-east-1
-    allocationIdRef:
-      name: sample-eip
-    subnetIdRef:
-      name: sample-subnet1
+    allocationId: "eipalloc-08f9f35936f4e4fc2"
+#   allocationIdRef:
+#     name: sample-eip
+    subnetId: "subnet-0b9b78fa81d457876"
+#   subnetIdRef:
+#     name: sample-subnet1
     tags:
       - key: Name
         value: sample-natgateway
--- a/examples/ec2/natgateway.yaml
+++ b/examples/ec2/natgateway.yaml
@@ -5,12 +5,10 @@ metadata:
 spec:
   forProvider:
     region: us-east-1
-    allocationId: "eipalloc-08f9f35936f4e4fc2"
-#   allocationIdRef:
-#     name: sample-eip
-    subnetId: "subnet-0b9b78fa81d457876"
-#   subnetIdRef:
-#     name: sample-subnet1
+    allocationIdRef:
+      name: sample-eip
+    subnetIdRef:
+      name: sample-subnet1
     tags:
       - key: Name
         value: sample-natgateway
--- a/examples/s3/bucket.yaml
+++ b/examples/s3/bucket.yaml
@@ -1,7 +1,7 @@
 apiVersion: s3.aws.crossplane.io/v1beta1
 kind: Bucket
 metadata:
-  name: conformancetest2
+  name: test-bucket
 spec:
   forProvider:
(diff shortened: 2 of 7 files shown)


examples/ec2/natgateway.yaml
```diff
@@ -1,14 +1,16 @@
...
-    allocationIdRef:
-      name: sample-eip
-    subnetIdRef:
-      name: sample-subnet1
+    allocationId: "eipalloc-08f9f35936f4e4fc2"
+#   allocationIdRef:
+#     name: sample-eip
+    subnetId: "subnet-0b9b78fa81d457876"
+#   subnetIdRef:
+#     name: sample-subnet1
```
negz, line 13: What's the thinking behind replacing these references? I think @muvaf mentioned a preference for having the references point to resources with corresponding examples, which sounds like a good approach to me.

examples/s3/bucket.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: s3.aws.crossplane.io/v1beta1
 kind: Bucket
 metadata:
-  name: test-bucket
+  name: conformancetest2
```
negz, line 4: Perhaps we should keep `test-bucket` as the `metadata.name` but set a likely-to-be-unique `crossplane.io/external-name` annotation?

examples/ec2/natgateway.yaml
ulucinar: I also agree indirect references make better examples. For full transparency into how I got things working, I opened this PR with the exact changes I made. One exception is the change in `examples/kms/key.yaml`, where you need to grant the proper access rights to the IAM principal in the key policy, which in the context of these tests were my own principal. Instead of my user's IAM, I went with the example "root" user.

So in certain places, using plain names/ids of AWS resources instead of references to MRs bought me some time (I did not have to provision them via managed resources, I could reference already existing ones in the playground project or lazily provisioned from the UI). 

Having said this, I think we could devise a more formal process for running these tests, where the env is standardized:
We provision MRs in dependency order, e.g., in order to test acm-pca or a cloudfront distribution, or a bucket policy, you need a bucket first. For some dependents the constraints on a common bucket could be conflicting (and need different parent in dep. graph) but for some others a common bucket could be leveraged. I think we already have a vague standardization in the examples (like most, if not all, refer to a providerconfig named `example`) but we may consider documenting or even scripting the provisioning of a "conformant" environment (set of parent resources either provisioned as MRs or if we have missing types provisioned through cloud platform tools). 

Having documented these tested versions, I will revert such transformations, and only keep "real" changes.

examples/s3/bucket.yaml
ulucinar: Yes, sure. As explained above, I just wanted to document how I tested for full transparency. I could not use the "test-bucket" in the tests as bucket names are globally shared with other users. 

One improvement could be to use unique (generated) names for certain resources when provisioning a "conformant testing/example environment" discussed in the above comment. Because we are using shared projects or even some names being globally-scoped, it might be a good idea to generate names not to interfere with each other. I think this is not relevant for examples. Should we have a separate set of manifests/scripts for conformance tests in the future?

ulucinar:
Thanks @negz for the review. I have added another commit that only leaves example manifest fixes.

muvaf (approved):
Thanks @ulucinar !

PR #703, opened 2021-06-08
Add Examples for Service Discovery

turkenh:

This PR ensures working examples for the resources mentioned [here](https://github.com/cncf/crossplane-conformance/issues/4#issuecomment-856232337).
It also marks HostedZone `status.atProvider` as optional for Crossplane Conformance.

I have:

1. Deploy crossplane 1.2.2 and provider-aws with these changes.
1. Setup a ProviderConfig
1. Create a VPC (e.g. using `examples/ec2/vpc.yaml`) and provide its `id` in `examples/servicediscovery/privatednsnamespace.yaml`
1. Create example resources under `examples/servicediscovery` and ensure they are reporting ready/synced.

### DIFF
--- a/apis/route53/v1alpha1/hostedzone_types.go
+++ b/apis/route53/v1alpha1/hostedzone_types.go
@@ -49,7 +49,7 @@ type HostedZoneSpec struct {
 // HostedZoneStatus represents the observed state of a HostedZone.
 type HostedZoneStatus struct {
 	xpv1.ResourceStatus `json:",inline"`
-	AtProvider          HostedZoneObservation `json:"atProvider"`
+	AtProvider          HostedZoneObservation `json:"atProvider,omitempty"`
 }
 
 // HostedZoneParameters define the desired state of an AWS Route53 Hosted HostedZone.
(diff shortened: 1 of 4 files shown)


examples/servicediscovery/privatednsnamespace.yaml
```diff
@@ -3,11 +3,14 @@ kind: PrivateDNSNamespace
...
+  providerConfigRef:
+    name: example
   forProvider:
-    region: eu-central-1
+    region: us-east-1
     description: "test"
     name: "test"
-    vpc: default
+    # The ID of the Amazon VPC that you want to associate the namespace with
+    vpc: vpc-02504719d1cb7a9aa
```
muvaf, line 13:
```suggestion
    vpcIdRef:
      name: sample-vpc
```

Can we reference to VPC example YAML here so that we're closer to getting `kubectl apply -f examples` to work?

muvaf (approved):
LGTM after one small comment.

examples/servicediscovery/privatednsnamespace.yaml
turkenh: This was also something that I considered, however, it looks like this is not supported by `PrivateDNSNamespace` API: https://github.com/crossplane/provider-aws/blob/72d273177575641f74fe5d225848f2d3dd09e2fe/apis/servicediscovery/v1alpha1/zz_private_dns_namespace.go#L49

negz: Interesting. If we weren't thinking of making references more generic (per https://github.com/crossplane/crossplane/issues/1770) it would probably make sense to require the existence of such references part of the conformance test suite.

PR #707, opened 2021-06-09
Conformance test fixes

muvaf:

Small fixes along YAML ones.

I have:

Manually.


apis/lambda/v1alpha1/custom_types.go
```diff
@@ -43,11 +47,39 @@ type CustomFunctionParameters struct {
...
+	// see VPC Settings (https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html).
+	CustomFunctionVPCConfigParameters *CustomFunctionVPCConfigParameters `json:"vpcConfig,omitempty"`
+
+	// The code for the function.
+	// +kubebuilder:validation:Required
+	CustomFunctionCodeParameters *CustomFunctionCodeParameters `json:"code"`
+}
+
+// CustomFunctionCodeParameters includes custom fields for FunctionCode struct.
+type CustomFunctionCodeParameters struct {
```
muvaf, line 62: @smcavallo That was the only way I could get `Function` to see the references. Is there something I'm missing that made it work the last time?

muvaf:
I have added notes to the top of the YAMLs that require manual intervention, like disabling or preparing an image to work. Once we have a script for testing, we can do those individually and test automatically.

apis/lambda/v1alpha1/custom_types.go
```diff
@@ -43,11 +47,39 @@ type CustomFunctionParameters struct {
...
-	CustomFunctionVPCConfigParameters CustomFunctionVPCConfigParameters `json:"vpcConfig,omitempty"`
+	// For network connectivity to AWS resources in a VPC, specify a list of security
+	// groups and subnets in the VPC. When you connect a function to a VPC, it can
+	// only access resources and the internet through that VPC. For more information,
+	// see VPC Settings (https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html).
+	CustomFunctionVPCConfigParameters *CustomFunctionVPCConfigParameters `json:"vpcConfig,omitempty"`
+
+	// The code for the function.
+	// +kubebuilder:validation:Required
+	CustomFunctionCodeParameters *CustomFunctionCodeParameters `json:"code"`
```
hasheddan, line 58: Should this be a pointer if required?

ulucinar (approved):

apis/lambda/v1alpha1/custom_types.go
muvaf: I made it so just to be inline with the rest of the CRD. You feel like it should be value?

muvaf:
@hasheddan e2e test seems to fail because provider-aws takes longer than expected to come up. I ran the test locally and saw bunch of permission errors:
```
E0609 17:41:56.831017       1 reflector.go:138] github.com/crossplane/provider-aws/.work/pkg/pkg/mod/k8s.io/client-go@v0.20.1/tools/cache/reflector.go:167: Failed to watch *v1alpha1.API: failed to list *v1alpha1.API: apis.apigatewayv2.aws.crossplane.io is forbidden: User "system:serviceaccount:crossplane-system:provider-aws-provider-aws" cannot list resource "apis" in API group "apigatewayv2.aws.crossplane.io" at the cluster scope
```

Then looking at rbac-manager, I saw this: 
```
I0609 17:41:35.878400       1 leaderelection.go:243] attempting to acquire leader lease crossplane-system/crossplane-leader-election-rbac...
E0609 17:41:35.922482       1 event.go:329] Could not construct reference to: '&v1.Lease{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"crossplane-leader-election-rbac", GenerateName:"", Namespace:"crossplane-system", SelfLink:"/apis/coordination.k8s.io/v1/namespaces/crossplane-system/leases/crossplane-leader-election-rbac", UID:"bf867ddf-d131-4244-88f5-474dc719a107", ResourceVersion:"722", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:63758857295, loc:(*time.Location)(0x283dc20)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry{v1.ManagedFieldsEntry{Manager:"crossplane", Operation:"Update", APIVersion:"coordination.k8s.io/v1", Time:(*v1.Time)(0xc00074bfa0), FieldsType:"FieldsV1", FieldsV1:(*v1.FieldsV1)(0xc00074bfc0)}}}, Spec:v1.LeaseSpec{HolderIdentity:(*string)(nil), LeaseDurationSeconds:(*int32)(nil), AcquireTime:(*v1.MicroTime)(nil), RenewTime:(*v1.MicroTime)(nil), LeaseTransitions:(*int32)(nil)}}' due to: 'no kind is registered for the type v1.Lease in scheme "pkg/runtime/scheme.go:100"'. Will not report event: 'Normal' 'LeaderElection' 'crossplane-rbac-manager-5cbdc6568f-tgkbn_fefca87e-7fce-4ae6-8fd0-92a28a53bb31 became leader'
I0609 17:41:35.922611       1 leaderelection.go:253] successfully acquired lease crossplane-system/crossplane-leader-election-rbac
```

But it doesn't happen every time. Sometimes provider-aws starts later and log no problem about permissions.

apis/lambda/v1alpha1/custom_types.go
negz: @muvaf I think typically we make `Required` values, right?

muvaf: > I think typically we make Required values, right?

@negz Yes, we do but ACK always uses pointers, no matter required or not. Because the SDK types are pointers and the generator of the conversion function does not have a logic to account for type differences of the fields, hence all fields on CRDs are also pointers. In this case, the field is not used in generated code but I just made it pointer to be inline with other fields in the same CRD but let me change it to a value type

negz: Got it - thanks for the context!

negz (approved):

PR #711, opened 2021-06-11
e2e: replace sleep commands with wait directives of the related tools

muvaf:

We're seeing increase in e2e test failures due to conditions being not met after a specific duration. Increasing the timeout would just make it take longer because the checks were not done periodically. So, I changed it to run periodically and use the related tools' waiting mechanisms wherever possible, like `helm`'s `--wait` flag and `kubectl wait --timeout=60s` command.

I have:

Manually with `make e2e`

### DIFF
--- a/cluster/local/integration_tests.sh
+++ b/cluster/local/integration_tests.sh
@@ -32,103 +32,6 @@ echo_error(){
     exit 1
 }
 
-# k8s watchers
-
-wait_for_pods_in_namespace(){
-    local timeout=$1
-    shift
-    namespace=$1
-    shift
-    arr=("$@")
-    local counter=0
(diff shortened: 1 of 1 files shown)


negz (commented):
LGTM at a glance, but I'll defer to Dan who wrote most of this.

I would like to ensure we propagate these changes to the other providers that were using these tests.

hasheddan (approved):
@muvaf this definitely seems like an improvement here, thanks! These tests were originally set up quickly and I think there are plenty of optimizations that can be made as part of the conformance effort. I'd also like for us to nail down what is causing the increased run time here, which may be related to package manager having more "resource has been modified" errors when installing CRDs. I'll investigate as a follow up 👍🏻

PR #713, opened 2021-06-12
Further fixes to get provider-aws to pass conformance tests

negz:

This fixes the same issue as https://github.com/crossplane/provider-aws/pull/698 (`status.atProvider` being required and `spec` not being required) for hand curated, non-generated resources.

Edit: I previously had included a commit that corrected a misnamed ReplicationGroup reference field; that is going to be a larger issue so I'll tackle it elsewhere.

I have:

I've run this in combination with https://github.com/crossplane/conformance/pull/24 and verified that all CRDs are considered conformant.

### DIFF
--- a/apis/acm/v1alpha1/certificate_types.go
+++ b/apis/acm/v1alpha1/certificate_types.go
@@ -138,7 +138,7 @@ type Certificate struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   CertificateSpec   `json:"spec,omitempty"`
+	Spec   CertificateSpec   `json:"spec"`
 	Status CertificateStatus `json:"status,omitempty"`
 }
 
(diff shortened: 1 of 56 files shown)


hasheddan (commented):
@negz it seems like we could add a check to the resolver that would use the singular if both were populated and the plural if only it was provided. However, do you think the potential confusion for new users would be worth supporting both? I typically would like to preserve beta APIs at all costs, but I wonder if this could fall into a unique category where it is considered more of a bug fix because their is a mismatch in the stated semantics of plural reference fields and the behavior of this resource currently. Either way, we should reach out to the community and try to get feedback from any users of this resource 👍🏻

negz:
> Either way, we should reach out to the community and try to get feedback from any users of this resource 

My thinking is we need to move fast here so if it's concerning enough to solicit feedback we should probably just try support both paths and mark the buggy old one as deprecated. I'll try take a look on Monday.

muvaf:
It seems like the type of the field is not array, only the name is plural. So TBH, I'd just leave it as is and open an issue to be addressed with conversion webhooks. If conformance enforces an array type if it sees that the name is plural, then supporting two fields sounds good to me. I don't think we should break the API.

negz:
I'd prefer to try fix it sooner if possible rather than just leaving it, but you're right that it's not strictly an issue for conformance. I've removed the relevant commit from this PR and will open a different one to address the misnamed field if it turns out to be an easy fix; PTAL.

hasheddan (approved):

muvaf (approved):
Thanks @negz !

PR #716, opened 2021-06-14
Use singular, not plural cacheSubnetGroupNameRef in ReplicaGroup

negz:

This PR adds a `cacheSubnetGroupNameRef` (singular) reference. The existing (accidentally) plural `cacheSubnetGroupNameRefs` field is maintained for backward compatibility.

I have:

I've created a `ReplicaGroup` using the examples under`examples/cache` and validated that that:

* If both reference fields are set we get an error.
* If the deprecated plural field is set, we use that.
* Otherwise we use the new singular field.

### DIFF
--- a/apis/cache/v1beta1/replication_group_types.go
+++ b/apis/cache/v1beta1/replication_group_types.go
@@ -346,14 +346,14 @@ type ReplicationGroupParameters struct {
 	// +optional
 	CacheSubnetGroupName *string `json:"cacheSubnetGroupName,omitempty"`
 
-	// CacheSubnetGroupNameRef is a reference to a Subnet Group
+	// CacheSubnetGroupNameRef is a reference to a Subnet Group used to set
 	// the CacheSubnetGroupName.
 	// +immutable
 	// +optional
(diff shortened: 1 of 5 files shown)


apis/cache/v1beta1/replication_group_types.go
```diff
@@ -350,7 +350,18 @@ type ReplicationGroupParameters struct {
 	// the CacheSubnetGroupName.
 	// +immutable
 	// +optional
-	CacheSubnetGroupNameRef *xpv1.Reference `json:"cacheSubnetGroupNameRefs,omitempty"`
+	CacheSubnetGroupNameRef *xpv1.Reference `json:"cacheSubnetGroupNameRef,omitempty"`
+
+	// DeprecatedCacheSubnetGroupNameRef is a reference to a Subnet Group
```
hasheddan, line 355: Nitpick: This isn't specific to this PR, but this doc string does read a little choppy.

hasheddan (approved):

apis/cache/v1beta1/replication_group_types.go
negz: Ah, so it does. I'll push a commit that fixes that grammar.

PR #721, opened 2021-06-15
add repo not exist to ignore

### LINKED ISSUE #714: Can't delete ECR RepositoryPolicy for non-existent repository
### What happened?

I'm spinning up and down all of our examples in order to run some Crossplane conformance tests. My suspicion is that deleting the repository automatically deletes the policy, but Crossplane's logic for deleting the policy requires the repository to exist. Perhaps if we can't find the repository the policy should apply to when deleting we should just consider it gone? That would be around this area I think - https://github.com/crossplane/provider-aws/blob/84acbf60/pkg/controller/ecr/repositorypolicy/controller.go#L100

Disclaimer: I haven't dug deep here and am making some assumptions about how ECR repos and policies work.

### How can we reproduce it?

I notice that I can `kubectl apply -f examples/ecr` at  successfully, but I can't `kubectl delete -f` the same directory. 

Crossplane version :1.2.2
provider-aws commit: 84acbf60bd68f2f7527c36dc3d23e955261bd932

enderv:

Fixes #714 

I have:

Created a repo and policy for it, deleted repo first and made sure could still then delete policy

### DIFF
--- a/pkg/controller/ecr/repositorypolicy/controller.go
+++ b/pkg/controller/ecr/repositorypolicy/controller.go
@@ -98,7 +98,7 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}).Send(ctx)
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


negz (approved):
Thanks @enderv!

PR #727, opened 2021-06-16
Ensure all managed resources have external names

negz:

In attempting to make provider-aws pass the https://github.com/crossplane/conformance test suite I noticed that three of our managed resources don't expose an external name annotation, which we require. These resources are all somewhat tricky in that they don't have their own obvious identity but are instead bindings between other resources (e.g. an IAM user and a group), so I've set external names that combine the identity of the two bound resources - e.g. `group/user`.

I have:

I've built the provider, applied the relevant examples from the `examples/` directory and ensured they resources become healthy and set their external names appropriately.


pkg/controller/identity/iamgroupusermembership/controller.go
```diff
@@ -117,11 +118,17 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+	// names of the group and user that are bound.
+	meta.SetExternalName(cr, cr.Spec.ForProvider.GroupName+"/"+cr.Spec.ForProvider.UserName)
+
 	cr.SetConditions(xpv1.Available())
 
 	return managed.ExternalObservation{
-		ResourceExists:   true,
-		ResourceUpToDate: true,
+		ResourceExists:          true,
+		ResourceLateInitialized: true,
```
muvaf, line 130: Even if it's no-op, the spec update call will override status from what I can tell. We usually handle this operation either in the [initializer itself](https://github.com/crossplane/crossplane-runtime/blob/406fe0b/pkg/reconciler/managed/api.go#L56) or in `Create` by returning [`ExternalNameAssigned: true`](https://github.com/crossplane/crossplane-runtime/blob/f2418dc/pkg/reconciler/managed/reconciler.go#L722).

Given that we'd like to cut a release urgently with these fixes, I'll open another PR with these commits and add a commit changing this to assign in `Create`.

negz: I forgot about ExternalNameAssigned! I may be able to tackle that during a meeting today haha.

muvaf: Ugh I was planning to cut the release before you wake up. Anyway I had opened the PR already here with some other small fixes https://github.com/crossplane/provider-aws/pull/728 . Manually tested all affected resources.

negz:
This is superceded by https://github.com/crossplane/provider-aws/pull/728.

Issue #726, opened 2021-06-16
CertificateAuthorityPermission should not use a constant principal


negz:
### What happened?

https://docs.aws.amazon.com/acm-pca/latest/APIReference/API_CreatePermission.html
https://github.com/crossplane/provider-aws/blob/ea83766c5476eb136ddae6378901c3b2934ca64c/pkg/controller/acmpca/certificateauthoritypermission/controller.go#L48

The ACM `CertificateAuthorityPermission` API requires that a principal be specified, but the only valid principal is `acm.amazonaws.com`. Our `CertificateAuthorityPermission` managed resource exposes an optional `spec.forProvider.prinicipal` field but does not honor it and instead always uses a constant principal. This goes against our "high fidelity principal". The field should be required and not default to a constant.

### How can we reproduce it?

Apply the example from `examples/acmpca`.

### What environment did it happen in?
Crossplane version: provider-aws ea83766c5476eb136ddae6378901c3b2934ca64c

PR #728, opened 2021-06-16
Ensure all managed resources have external names

muvaf:

Quick fixes for https://github.com/crossplane/provider-aws/pull/727 added on top of the base commits.

Also fixes https://github.com/crossplane/provider-aws/issues/726.

I have:

Manually by creating all related IAM resources and both `acmpca` resources.

### DIFF
--- a/apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
+++ b/apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
@@ -60,6 +60,7 @@ type CertificateAuthorityPermissionParameters struct {
 	// The AWS service or identity that receives the permission. At this
 	// time, the only valid principal is acm.amazonaws.com.
 	// +immutable
+	// +kubebuilder:default:=acm.amazonaws.com
 	Principal string `json:"principal"`
 
 	// Calling Account ID
--- a/apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
+++ b/apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
@@ -57,10 +57,10 @@ type CertificateAuthorityPermissionParameters struct {
 	// +immutable
 	Actions []string `json:"actions,omitempty"`
 
-	// The AWS Service or identity
-	// +optional
+	// The AWS service or identity that receives the permission. At this
+	// time, the only valid principal is acm.amazonaws.com.
 	// +immutable
-	Principal *string `json:"principal,omitempty"`
+	Principal string `json:"principal"`
 
 	// Calling Account ID
 	// +optional
--- a/pkg/controller/acmpca/certificateauthoritypermission/controller.go
+++ b/pkg/controller/acmpca/certificateauthoritypermission/controller.go
@@ -30,6 +30,7 @@ import (
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/logging"
+	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/ratelimiter"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
@@ -44,8 +45,6 @@ const (
 	errGet              = "failed to get ACMPCA with name"
 	errCreate           = "failed to create the ACMPCA resource"
 	errDelete           = "failed to delete the ACMPCA resource"
-
-	principal = "acm.amazonaws.com"
 )
 
 // SetupCertificateAuthorityPermission adds a controller that reconciles ACMPCA.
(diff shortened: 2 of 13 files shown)


turkenh (approved):

apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
```diff
@@ -57,10 +57,10 @@ type CertificateAuthorityPermissionParameters struct {
 	// +immutable
 	Actions []string `json:"actions,omitempty"`
 
-	// The AWS Service or identity
-	// +optional
+	// The AWS service or identity that receives the permission. At this
+	// time, the only valid principal is acm.amazonaws.com.
 	// +immutable
-	Principal *string `json:"principal,omitempty"`
+	Principal string `json:"principal"`
```
hasheddan, line 63: Should we bump this to `v1alpha2`? Or was this previously implicitly required by AWS rejecting if not supplied?

muvaf: It seems like we were using a constant for the value of this field and the field was no-op. So, making it required now would break people who didn't supply a value for this field. I made it optional again but added this as default value. I think we can fix the schema when we bump to v1beta1.

muvaf:
@hasheddan should be ready for another look. We might be able to get https://github.com/crossplane/provider-aws/pull/680 in as well for the v0.19.0 release if we wait a bit. cc @PocketMobsters

pkg/controller/identity/iamgroupusermembership/controller.go
```diff
@@ -93,16 +94,27 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+	if meta.GetExternalName(cr) == "" {
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
+
+	// We need to use external name if it exists in order to support resource import.
+	nn := strings.Split(meta.GetExternalName(cr), "/")
+	if len(nn) != 2 {
+		return managed.ExternalObservation{}, errors.New("external name has to be in the following format <group-name>/<user-name>")
+	}
+	groupName, userName := nn[0], nn[1]
```
negz, line 106: My choice not to support using the external name for importing resources was conscious - I'm not sure this change makes sense. Aren't we now duplicating information from the spec (i.e. the group and user name) in the external name rather than _moving_ those fields into the external name? I don't think there's an obvious way to move them into the EN wholesale since they're both set by references.

apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
```diff
@@ -57,9 +57,10 @@ type CertificateAuthorityPermissionParameters struct {
 	// +immutable
 	Actions []string `json:"actions,omitempty"`
 
-	// The AWS Service or identity
-	// +optional
+	// The AWS service or identity that receives the permission. At this
+	// time, the only valid principal is acm.amazonaws.com.
 	// +immutable
+	// +kubebuilder:default=acm.amazonaws.com
 	Principal *string `json:"principal,omitempty"`
```
negz, line 64: Good idea to default this. Given we're doing that, would it hurt to make this required? Presumably anyone who had omitted it would have it defaulted when they updated (plus this is a v1alpha1 API). I'd prefer to match the fact that this is required by the upstream AWS API.

pkg/controller/acmpca/certificateauthoritypermission/controller.go
```diff
@@ -95,17 +95,39 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	if !ok {
 		return managed.ExternalObservation{}, errors.New(errUnexpectedObject)
 	}
+
+	if meta.GetExternalName(cr) == "" {
+		return managed.ExternalObservation{ResourceExists: false}, nil
+	}
```
negz, line 101: These checks that assume the resource doesn't exist if we don't have the external name feel a little odd given that we're just deriving our external name from our spec.

apis/acmpca/v1alpha1/certificateauthoritypermission_types.go
muvaf: I [tried](https://github.com/crossplane/provider-aws/pull/728#discussion_r652804560) creation with required+defaulting and it didn't work but I just tested your assumption and it seems like update does go through and adds this value to the missing ones. I'll make it required again cc @hasheddan

pkg/controller/acmpca/certificateauthoritypermission/controller.go
muvaf: We could do it without checking for empty string here but then we'd have to use spec fields for identification, which means resource import also has to be done through spec which is not inline with others or handle that in a separate branch.

negz: I guess the difference with other resources is that (iirc) we derive the EN from the name automatically before the original observe. 🤔 Would it make sense for us to derive a "temporary" external name from the spec here that we'd use just for observe? e.g. fall back to setting the `principal` and `caARN` to their spec vars?

No need to spend too much time on this though - what we have here is an improvement.

negz (approved):

PR #741, opened 2021-06-24
Add global poll interval flag

hasheddan:

Adds a `--poll-interval` flag to provider, which is passed to all controllers to set their polling interval in the reconciliation happy path.

*Note to reviewers: these changes are almost entirely find and replace.*

xref https://github.com/crossplane/crossplane-runtime/issues/253

I have:

I ran this build of `provider-aws` with `-d` and `--poll-interval 10m` and observed the following logs and metrics using the process described in https://danielmangum.com/posts/controller-runtime-scrape-prometheus-local/.

buckets2

```
2021-06-24T17:44:04.420-0400	DEBUG	provider-aws	Reconciling	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket"}
2021-06-24T17:44:05.519-0400	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket", "uid": "d2f1e730-41e2-4627-bd84-5f23130f8de5", "version": "1167", "external-name": "blah-crossplane-example-bucket", "requeue-after": "2021-06-24T17:54:05.519-0400"}
2021-06-24T17:44:05.524-0400	DEBUG	provider-aws	Reconciling	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket"}
2021-06-24T17:44:06.747-0400	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket", "uid": "d2f1e730-41e2-4627-bd84-5f23130f8de5", "version": "1167", "external-name": "blah-crossplane-example-bucket", "requeue-after": "2021-06-24T17:54:06.747-0400"}
2021-06-24T17:54:05.525-0400	DEBUG	provider-aws	Reconciling	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket"}
2021-06-24T17:54:06.566-0400	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket", "uid": "d2f1e730-41e2-4627-bd84-5f23130f8de5", "version": "1167", "external-name": "blah-crossplane-example-bucket", "requeue-after": "2021-06-24T18:04:06.566-0400"}
2021-06-24T18:04:06.583-0400	DEBUG	provider-aws	Reconciling	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket"}
2021-06-24T18:04:07.589-0400	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/bucket.s3.aws.crossplane.io", "request": "/test-bucket", "uid": "d2f1e730-41e2-4627-bd84-5f23130f8de5", "version": "1167", "external-name": "blah-crossplane-example-bucket", "requeue-after": "2021-06-24T18:14:07.589-0400"}
```

### DIFF
--- a/CODE_GENERATION.md
+++ b/CODE_GENERATION.md
@@ -92,7 +92,7 @@ of the provider. Create a file called `pkg/controller/<serviceid>/setup.go` and
 add the setup function like the following:
 ```golang
 // SetupStage adds a controller that reconciles Stage.
-func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
+func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, poll time.Duration) error {
 	name := managed.ControllerName(svcapitypes.StageGroupKind)
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
@@ -103,6 +103,7 @@ func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) er
 		Complete(managed.NewReconciler(mgr,
 			resource.ManagedKind(svcapitypes.StageGroupVersionKind),
 			managed.WithExternalConnecter(&connector{kube: mgr.GetClient()}),
+			managed.WithPollInterval(poll),
 			managed.WithLogger(l.WithValues("controller", name)),
 			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
 }
@@ -165,7 +166,7 @@ Likely, we need to do this injection before every SDK call. The following is an
 example for hook functions injected:
 ```golang
 // SetupStage adds a controller that reconciles Stage.
-func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
+func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, poll time.Duration) error {
 	name := managed.ControllerName(svcapitypes.StageGroupKind)
 	opts := []option{
 		func(e *external) {
@@ -185,6 +186,7 @@ func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) er
 			resource.ManagedKind(svcapitypes.StageGroupVersionKind),
 			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
 			managed.WithInitializers(managed.NewDefaultProviderConfig(mgr.GetClient())),
+			managed.WithPollInterval(poll),
 			managed.WithLogger(l.WithValues("controller", name)),
 			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
 }
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -37,6 +37,7 @@ func main() {
 		app            = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
 		syncPeriod     = app.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").Duration()
+		pollInterval   = app.Flag("poll-interval", "Default poll interval for all manage resource controllers.").Default("1m").Duration()
 		leaderElection = app.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
 	)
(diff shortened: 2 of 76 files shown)


CODE_GENERATION.md
```diff
@@ -92,7 +92,7 @@ of the provider. Create a file called `pkg/controller/<serviceid>/setup.go` and
 add the setup function like the following:
 ```golang
 // SetupStage adds a controller that reconciles Stage.
-func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter) error {
+func SetupStage(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, poll time.Duration) error {
```
negz, line 95: We're getting pretty close to the point that I'd prefer to pass through a struct of setup dependencies rather than keep tacking on arguments to this function signature. Something like:

```go
func SetupStage(mgr ctrl.Manager, d setup.Dependencies) error { ... }
```

I could be convinced to punt on that until we find ourselves adding the 5th argument though.

cmd/provider/main.go
```diff
@@ -37,6 +37,7 @@ func main() {
 		app            = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
 		syncPeriod     = app.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").Duration()
+		pollInterval   = app.Flag("poll-interval", "Default poll interval for all manage resource controllers.").Default("1m").Duration()
```
negz, line 40:
```suggestion
		syncInterval     = app.Flag("sync", "Controller manager sync interval.").Short('s').Default("1h").Duration()
		pollInterval   = app.Flag("poll", "Default poll interval for all manage resource controllers.").Default("1m").Duration()
```

What do you think about this pattern? This way we're consistent with the existing `--sync` flag that serves a similar purpose. Speaking of which, I could easily see readers having a hard time distinguishing between those two flags. It would be ideal to try to describe what the difference is between the sync period and the poll interval in the help text, though it might be tough to do succinctly.

CODE_GENERATION.md
hasheddan: I think that would be a good idea, I almost wonder if we could encode some of this in `crossplane-runtime`. I want to think about this a bit so I prefer to defer it until later for now :+1:

cmd/provider/main.go
hasheddan: I like it! Let me take a pass at updating the help text here and let you take another look :+1:

hasheddan (commented):
@negz made some updates if you want to take another look :+1:

cmd/provider/main.go
```diff
@@ -36,7 +36,8 @@ func main() {
 	var (
 		app            = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
 		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncPeriod     = app.Flag("sync", "Controller manager sync period duration such as 300ms, 1.5h or 2h45m").Short('s').Default("1h").Duration()
+		syncInterval   = app.Flag("sync", "Sync interval controls the minimum frequency at which all watched resources are reconciled.").Short('s').Default("1h").Duration()
+		pollInterval   = app.Flag("poll", "Poll interval controls how long after a successful reconcile an individual resource is reconciled.").Default("1m").Duration()
```
negz, line 40: Nit: This is an improvement, but it's still not super clear to the uninitiated reader when the minimum vs the other interval would come into play. Maybe an update to the poll interval to mention something about it being the goal or the soonest interval would help?

negz (approved):
One nit that you can feel free to ignore. Thanks for getting this in place!

cmd/provider/main.go
hasheddan: @negz what would you think about something more abstract like "Controls how often an individual resource is checked for drift."? It feels a little less accurate, but it is more accessible to someone who isn't familiar with the underlying machinery. I could go either way.

negz: I think that sounds good for poll interval, and I'd maybe try frame sync interval as kind of a backup/fallback. Maybe:

* Poll: How often each individual resource should be checked for drift.
* Sync: How often all resources will be double checked for drift.

psinghal20: Hi @hasheddan @negz, I looked at the updated help message for the `Sync` flag. As a newcomer to Crossplane or Kubernetes internals, the `double check` is still a little bit confusing as to why it is required. Looking up to controller-runtime docs helped clear the reasons for double-checking for drift. 

This change is definitely better than the previous message, and it is tough to sum up the official documentation in a one-line command, but can we add a reference to the official docs here? I feel it would help a lot to figure out the need for this sync interval. Thanks for the amazing work! :smile:

Issue #495, opened 2021-01-15
Force Delete Flag for some resources 


enderv:
### What problem are you facing?
Some resources do not allow deletion if they have dependent resources (s3 bucket with items, vpc with resources, etc). This can cause the provider reconciler to fail to delete resources. Our use case revolves around s3 and vpc's. We want users to be able to create and cleanup buckets and vpcs even if they have items created in them from outside crossplane. 

### How could Crossplane help solve your problem?
Was thinking for these resources add a force delete flag in the spec and logic to controller to do the cleanup. Similar to [force_destroy option for s3 buckets in terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#force_destroy).

negz:
Thanks for this feature request @enderv! I agree this would be good functionality to have, but we need to think about how it should work.

Architecturally we try to ensure that our managed resources map cleanly to a cloud provider API object or at least concept to keep our implementation granular and predictable. Terraform et al achieve this "force delete" functionality by listing and deleting all objects in the bucket being deleted, which blends concepts a little more than I'd like - i.e. ideally when you operate on a Crossplane "bucket" you could could be confident that you're only ever exercising code that operates on AWS buckets, and not objects.

muvaf:
I think I'm in the same boat with @negz . It's a very useful feature for sure, but with the high fidelity goal, I'm not sure if it fits in the bucket controller. Maybe a small webhook or controller could do what you want; `if <bucket cr>.deletion { list and delete every object }` or maybe you could incorporate this functionality into your uninstall/removal process of the application if possible.

enderv:
Agree not really a fan of flag but wanted to throw it out there. Something we are thinking about doing is using a helm release that just uses pre-delete hook to run a job that does the cleanup and just include that in the composition.

Which makes me think a handy pre-delete job hook somewhere would be nice.

chlunde:
@ulucinar / @muvaf asked me on slack to list some examples of stuck-on-deletion issues I had seen. As there already is an issue I'll write them here:

# Patched invalid external-name
```yaml
apiVersion: ecr.aws.crossplane.io/v1alpha1
kind: Repository
metadata:
  name: delete-demo1
spec:
  forProvider:
    imageScanningConfiguration:
      scanOnPush: true
    imageTagMutability: MUTABLE
    region: eu-north-1

```

```
Status:
  At Provider:
[...]

PR #743, opened 2021-06-25
ecr.repository: Add forceDelete to delete images

chlunde:
Add flag forceDelete to the CRD directly under spec. This enabled the
force flag on the DeleteRepository AWS API call, which will also delete
images if the repository contains any.

Without this option, AWS blocks deletion of repositories without images.

Updates #495

Implementation notes:
- I'm testing the proper flags in the mock. Not sure if you think this is abuse of the current mocks, but I think it is important to test, and without a different mock approach I don't know of any other way than this.
- We should consider the name `forceDelete` carefully, does it make sense for recursive delete on S3 too? I guess it's pretty generic (unlike `recursiveDelete` which makes sense for S3 and ECR but not "delete an active bundle" as discussed in #495)

I have:

- [x] Unit tests
- [x] Test IRL with a non-empty repo and ensure it is not deleted without the flag
- [x] Set the flag and ensure it is deleted


ulucinar:
Hi @chlunde,
Do I understand correctly that in your case involving ECR, AWS already provides the force-delete option as part of its API?

It may make sense to put such a flag under `spec.forProvider` because it’s a parameter that’s part of the AWS API (it's for the provider). In the case mentioned [here](https://github.com/crossplane/provider-aws/issues/495#issuecomment-867666748), AWS does not include a similar parameter in its API. @hasheddan, what's your opinion here?

chlunde:
@ulucinar you're right, in this case the provider does support the option. The name is simply `force` under the `delete` API request, so if I should move it I would like to defer to you and @hasheddan for the name :) 

I notice secrets manager has `forProvider.forceDeleteWithoutRecovery` - but this name matches the provider option and `forProvider.force` is surely not the correct name :D

hasheddan (changes requested):
Apologies for the delayed response here folks! @chlunde I think putting this under `forProvider` as @ulucinar suggests is the right move, and I also think calling it something like `forceDelete` is appropriate here. The only other route I would consider going is an embedded struct under `forProvider` with something like `delete.force`, but I agree that just using `forceDelete` is more aligned with fields like `skipFinalSnapshotBeforeDeletion`, etc.

hasheddan (approved):
LGTM! Thanks @chlunde! Did you test this against an actual ECR repository?

chlunde:
Yeah, I tested on an actual cluster. I have not re-tested yet after moving the field.

chlunde:
re-tested OK:

```json
2021-08-02T20:20:31.613+0200	DEBUG	controller-runtime.manager.events	Warning	{"object": {"kind":"Repository","name":"example","uid":"70fd7f05-cd0e-4267-9ff0-21b326c8972d","apiVersion":"ecr.aws.crossplane.io/v1alpha1","resourceVersion":"4837"}, "reason": "CannotDeleteExternalResource", "message": "failed to delete the repository resource: RepositoryNotEmptyException: The repository with name 'example' in registry with id '21913760....' cannot be deleted because it still contains images"}
```
```bash
$ kubectl patch repository.ecr example --type=merge --patch '{"spec": {"forProvider": {"forceDelete": true}}}'
repository.ecr.aws.crossplane.io/example patched
```

```json
2021-08-02T20:25:19.267+0200	DEBUG	provider-aws	Reconciling	{"controller": "managed/repository.ecr.aws.crossplane.io", "request": "/example"}
chlunde@fedora ~/src/provider-aws$ 2021-08-02T20:25:19.534+0200	DEBUG	provider-aws	Successfully requested deletion of external resource	{"controller": "managed/repository.ecr.aws.crossplane.io", "request": "/example", "uid": "70fd7f05-cd0e-4267-9ff0-21b326c8972d", "version": "5338", "external-name": "example", "deletion-timestamp": "2021-08-02 20:20:21 +0200 CEST"}
2021-08-02T20:25:19.535+0200	DEBUG	controller-runtime.manager.events	Normal	{"object": {"kind":"Repository","name":"example","uid":"70fd7f05-cd0e-4267-9ff0-21b326c8972d","apiVersion":"ecr.aws.crossplane.io/v1alpha1","resourceVersion":"5338"}, "reason": "DeletedExternalResource", "message": "Successfully requested deletion of external resource"}
2021-08-02T20:25:19.555+0200	DEBUG	provider-aws	Reconciling	{"controller": "managed/repository.ecr.aws.crossplane.io", "request": "/example"}
```

AaronME (approved):
This works in testing. Confirmed that `forceDelete: false` results in an error on delete. `forceDelete: true` successfully forces delete.

Thanks, @chlunde!

PR #689, opened 2021-06-01
Feature for creation of VPC Peerings

Dkaykay:
### Description of your changes

Create VPC Peering Connections and approve the connection on AWS

Fixes #325

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

- `make reviewable test`
- create VPC Peering connection
- delete VPC Peering connection

PR #751, opened 2021-06-29
 version: bump ec2/vpccidrblock package to manualv1alpha1

haarchri:
Signed-off-by: Christopher Haar <chhaar30@googlemail.com>

bump package version for ec2/vpccidrblock to manualv1alpha1 to use v1alpha1 for code-generator stuff 
in #689 we see the problem between custom and code-generator stuff in one - for example:  of zz_types.go 
think it is better not mix-up the both approaches

what did you guys think ?

I have:

```
➜  provider-aws git:(ec2-vpccidrblock) ✗ kubectl apply -f examples/ec2/vpccidrblock.yaml
vpccidrblock.ec2.aws.crossplane.io/sample-vpc-cidr-block created
```

```
➜  provider-aws git:(ec2-vpccidrblock) ✗ kubectl delete -f examples/ec2/vpccidrblock.yaml
vpccidrblock.ec2.aws.crossplane.io "sample-vpc-cidr-block" deleted
```


AaronME:
I see what you are saying. And I agree that it would be nice to have a distinction between the generated and non-generated types.  Could using `alpha2` for non-generated create confusion? Someone could infer that `alpha2` indicates a more recent, and preferred, approach. When really we prefer people to follow practices in `alpha1.`

haarchri:
so what is your prefered way ? at the moment we need to modify in ec2 zz_ files if we want integrate more services with code-generator

hasheddan (changes requested):
@haarchri we don't need to actually change the API version, we can just change the package name so that the generated code lives in a different package. Users of this API should have the same experience as they do currently 👍🏻

haarchri:
@hasheddan  can you provide a example ? At the moment the code-generator will add in ec2 v1alpha1

apis/ec2/v1alpha2/referencers.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v1alpha1
+package v1alpha2
```
hasheddan, line 17: @haarchri you can just name this package and directory something like `manualv1alpha1` and then the generated one can use `v1alpha1`

apis/ec2/v1alpha2/register.go
```diff
@@ -15,9 +15,9 @@ limitations under the License.
 */
 // +kubebuilder:object:generate=true
 // +groupName=ec2.aws.crossplane.io
-// +versionName=v1alpha1
+// +versionName=v1alpha2
```
hasheddan, line 18: Then you would keep the version inside of `manualv1alpha1` as `v1alpha1`. CRDs should not change for existing `VPCCIDRBlock`, you will just need to change imports in the controller / client / etc.

hasheddan (commented):
@haarchri left some guiding comments below, let me know if you want to sync up about them if they aren't illustrating my suggestions well 👍🏻

haarchri:
@hasheddan i changed the packag to manualv1alpha1 and the generated crds using v1alpha1 - think this was this what you mentioned ;)

tested via apply and delete

for delete i added a fix that the ressource is removed when vpccidrblock is in status VpcCidrBlockStateCodeDisassociated
you can have a look

pkg/controller/ec2/vpccidrblock/controller.go
```diff
@@ -121,17 +121,17 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	observed := response.Vpcs[0]
 
 	currentStatusCode, err := ec2.FindVPCCIDRBlockStatus(meta.GetExternalName(cr), observed)
-	if err != nil {
-		return managed.ExternalObservation{}, errors.Wrap(resource.Ignore(ec2.IsCIDRNotFound, err), errDescribe)
+
+	// vpccidrblock is in status disassociated after deletion
+	if currentStatusCode == awsec2.VpcCidrBlockStateCodeDisassociated {
+		return managed.ExternalObservation{}, errors.Wrap(err, errDescribe)
 	}
```
hasheddan, line 128: @haarchri it appears we are now not handling the error from `ec2.FindVPCCIDRBlockStatus` unless the status code is `awsec2.VpcCidrBlockStateCodeDisassociated`, which I don't think will work. Feel free to break this out into a separate PR so we can discuss without blocking the PRs which depend on this package change 👍🏻

hasheddan (changes requested):

pkg/controller/ec2/vpccidrblock/controller.go
haarchri: done - i will open a followup pr for this

hasheddan (approved):
Thank you @haarchri !

Issue #414, opened 2020-11-04
Implement Simple Email Service (SES) resources


marcinc:
### What problem are you facing?

AWS Provider does not have support for SES resources.

### How could Crossplane help solve your problem?

Implement SES resources.

ikarldasan:
Any progress on this issue? I would like to help with this resource

kelvinwijaya:
Hi there,

I would like to contribute on this resource if no one has started on the controllers logic yet. I have some working code that I have developed for `EmailIdentity` and `ConfigurationSet` Resources using AWS SESv2 Service API, thinking adding another functionality for `ConfigurationSetEventDestination` Resource and probably can be a good start for v1alpha1 release

kelvinwijaya:
Hi @luebken @ikarldasan ,

Would you mind if i help on the development of the SES resource as my project needs to leverage on and have short timeline for this? 

Apologise if i have repeated the same request, as i have already done some wireframe of the custom resource and written the controller logic . So thinking of contributing them since I am leveraging heavily on crossplane for other managed resources

ikarldasan:
Hi @kelvinwijaya 
Yes definitely. Would be great to get your help 

Would like to mention that I am currently working on improving the PR for SES Template (https://github.com/crossplane/provider-aws/pull/755).

Thank you

kelvinwijaya:
Thanks @ikarldasan 

What I am planning as the first cut (v1alpha1) would be the 3 main resources:
- EmailIdentity
- ConfigurationSet
- ConfigurationSetEventDestination

Yes thanks for highlighting and I do notice your PR on the SES Template (which is based on SESv1 API). This might potentially be helpful when rendering `EmailTemplate` resource for SESv2 which i hit into error: `pkg conversion - executing "sdk_find_read_many"`

Just to highlight that I will be using the SESv2 API which I think is more simplified. Both version has slightly differences in the resource names [link](https://github.com/aws/aws-sdk-go/tree/v1.37.4/models/apis) but functionality wise are quite similar as i hv switched my AWS Web Console to v2 and my data is transitioning smoothly.

kelvinwijaya:
Thanks @jbw976 for assigning this ticket, let me know if any further help or needs for amending the respective PR

PR #755, opened 2021-07-01
Implement AWS SES Template

ikarldasan:
- Adds Support for AWS SES Template

"Fixes a part of  #414  "

I have:

Manually tested the creation, updation and deletion of Template


examples/ses/template.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: ses.aws.crossplane.io/v1alpha1
+kind: Template
+metadata:
+    name: example
+spec:
+    forProvider:
+    region: us-east-2
```
chlunde, line 8: indent this section under `forProvider`, and check that it works

pkg/controller/ses/template/setup.go
```diff
@@ -0,0 +1,108 @@
...
+			managed.WithInitializers(managed.NewDefaultProviderConfig(mgr.GetClient())),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
+			managed.WithPollInterval(poll),
+			managed.WithLogger(l.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name)))))
+}
+
+func preObserve(_ context.Context, cr *svcapitypes.Template, obj *svcsdk.GetTemplateInput) error {
+	obj.TemplateName = aws.String(meta.GetExternalName(cr))
+	cr.Spec.ForProvider.Template.TemplateName = obj.TemplateName
```
chlunde, line 68: I *think* you should add this field to ignore in code gen. Look at lambda for example, the field `FunctionName` is ignored, so it's only part of external-name and not from the cr.
https://github.com/crossplane/provider-aws/blob/36e425acdcb52a670f78ab5e37b8f81864d3f393/pkg/controller/lambda/function/setup.go#L87-L90

https://github.com/crossplane/provider-aws/blob/36e425acdcb52a670f78ab5e37b8f81864d3f393/apis/lambda/v1alpha1/generator-config.yaml#L1-L3

examples/ses/template.yaml
```diff
@@ -0,0 +1,15 @@
...
+    template:
+        templateName: example
+        htmlPart: "<h1>sample html part</h1>"
+        subjectPart: "sample subject part"
+        textPart: "sample text part"
+    providerConfigRef:
+    name: example
```
chlunde, line 15: nitpick: if all other files have a final newline, it would be nice to add it to this file and `generator-config.yaml` too.

chlunde (changes requested):
Nice work! I'm not a maintainer and I haven't used code gen, but I think I found two small issues you should look at, hope this helps.

examples/ses/template.yaml
ikarldasan: Indented the section under `forProvider` . 
Checked for Create, Update and Delete operations - Works successfully.

examples/ses/template.yaml
ikarldasan: Added newline to the examples.

pkg/controller/ses/template/setup.go
ikarldasan: The `CreateTemplateInput` struct contains a `*Template` which contains the `TemplateName`.

Do you suggest adding TemplateName to ignore field_paths like this?
```
ignore:
  field_paths:
    - CreateTemplateInput.Template.TemplateName
```


I tried adding the same. The resultant code generated after re running make generate does not have the Template Parameters in the ForProvider i.e the CRD does not accept the `Template_SDK struct` that contains `HTMLPart, SubjectPart, TemplateName, TextPart`

chlunde: I think there is a need to update the code-generator, it does not support three levels like `Foo.Bar.Baz`.

I made a (draft) PR here: https://github.com/aws-controllers-k8s/code-generator/pull/119

examples/ses/template.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: ses.aws.crossplane.io/v1alpha1
+kind: Template
+metadata:
+    name: example
+spec:
+    forProvider:
+        region: us-east-2
+        template:
+            templateName: example
```
AaronME, line 10: I tested updating `spec.forProvider.template.templateName` on an existing Template resource and received the following error:

```
2021-07-23T15:24:19.702-0400    DEBUG   provider-aws    Cannot update external resource {"controller": "managed/template.ses.aws.crossplane.io", "request": "/example", "uid": "b15f8ce9-c765-4543-bf1c-435caaaa8bc6", "version": "50082", "external-name": "example"}
2021-07-23T15:24:19.702-0400    DEBUG   controller-runtime.manager.events       Warning {"object": {"kind":"Template","name":"example","uid":"b15f8ce9-c765-4543-bf1c-435caaaa8bc6","apiVersion":"ses.aws.crossplane.io/v1alpha1","resourceVersion":"50082"}, "reason": "CannotUpdateExternalResource", "message": "cannot update Template in AWS: TemplateDoesNotExist: Template example2 does not exist."}
```

AaronME (changes requested):
Works well to create, delete.

Update works for changes to SubjectPart, TextPart, and HtmlPart.

I've included some thoughts on how to handle an update to TemplateName.

pkg/controller/ses/template/setup.go
chlunde: @muvaf points out that this can be achieved already using

```yaml
ignore:
  field_paths:
    - Template.TemplateName
```

AaronME:
@ikarldasan are you still working on this PR?

haarchri:
@ikarldasan any chance to see this finalized ?

MisterMX:
Closing as this was implemented in #791. Feel free to reopen if that is not the case.

PR #757, opened 2021-07-05
Fix Address descriptions

dweebo:

Update a few of the descriptions for the `addresses.ec2.aws.crossplane.io` CRD.
Refer to EIPs instead of VPC.  

I have:

Since it is just documentation I have not done any testing

### DIFF
--- a/apis/ec2/v1beta1/address_types.go
+++ b/apis/ec2/v1beta1/address_types.go
@@ -24,7 +24,7 @@ import (
 
 // AddressParameters define the desired state of an AWS Elastic IP
 type AddressParameters struct {
-	// Region is the region you'd like your VPC to be created in.
+	// Region is the region you'd like your Address to be created in.
 	Region string `json:"region"`
 
 	// [EC2-VPC] The Elastic IP address to recover or an IPv4 address from an address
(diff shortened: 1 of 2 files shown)


hasheddan (changes requested):
Hey @dweebo, thanks for the PR! We actually generate these description fields from the doc strings on the structs, so they will need to be updated there, then you can run `make generate` to reflect them here. For example, you would update this doc string for `Region`: https://github.com/crossplane/provider-aws/blob/36e425acdcb52a670f78ab5e37b8f81864d3f393/apis/ec2/v1beta1/address_types.go#L27

dweebo:
> Hey @dweebo, thanks for the PR! We actually generate these description fields from the doc strings on the structs, so they will need to be updated there, then you can run `make generate` to reflect them here. For example, you would update this doc string for `Region`:
> 
> https://github.com/crossplane/provider-aws/blob/36e425acdcb52a670f78ab5e37b8f81864d3f393/apis/ec2/v1beta1/address_types.go#L27

Ah yes of course.  Fixed now.

apis/ec2/v1beta1/address_types.go
```diff
@@ -123,16 +123,15 @@ type AddressObservation struct {
...
 type AddressStatus struct {
 	xpv1.ResourceStatus `json:",inline"`
 	AtProvider          AddressObservation `json:"atProvider,omitempty"`
 }
 
 // +kubebuilder:object:root=true
 
-// A Address is a managed resource that represents an AWS VPC Security
-// Group.
+// A Address is a managed resource that represents an AWS Elastic IP Address.
```
hasheddan, line 134: nitpick:
```suggestion
// An Address is a managed resource that represents an AWS Elastic IP Address.
```

hasheddan (approved):
Thanks @dweebo!

PR #765, opened 2021-07-07
s3: replication configuration - allow user to drop Filter similar AWS documentation

chlunde:
The filter part of a replication rule is as far as I can tell optional in the documentation. However, it was not possible to create a filter-less replication rule similar to the XML examples.

Using cloud trail and the console noticed that if you create a rule in the s3 console, cloud trail says `filter: ""`. With `aws s3api get-bucket-replicaton` I see `"Filter": {},`. However, in the XML examples, there is no filter:

https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-add-config.html#replication-config-min-rule-config

It turns out that if you don't want a filter, you still have to include the filter element to the Go SDK.

Question: I'm not sure if this is an issue with the Go SDK, or if the documentation is simply wrong, that you need a filter XML tag? If we can dump the XML perhaps we can know if this is a good idea or not.

I have:

Created replication rule and check the console.

### DIFF
--- a/pkg/controller/s3/bucket/replicationConfig.go
+++ b/pkg/controller/s3/bucket/replicationConfig.go
@@ -293,6 +293,8 @@ func createRule(input v1beta1.ReplicationRule) awss3.ReplicationRule {
 		if Rule.Filter.Tag != nil {
 			newRule.Filter.Tag = &awss3.Tag{Key: awsclient.String(Rule.Filter.Tag.Key), Value: awsclient.String(Rule.Filter.Tag.Value)}
 		}
+	} else {
+		newRule.Filter = &awss3.ReplicationRuleFilter{}
 	}
 	if Rule.SourceSelectionCriteria != nil {
 		newRule.SourceSelectionCriteria = &awss3.SourceSelectionCriteria{


AaronME (approved):
Passes testing with the following manifest:

```
---
kind: Bucket
metadata:
  name: repl-src
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: us-west-2
    paymentConfiguration:
      payer: BucketOwner
      rules:
          sseAlgorithm: AES256
    versioningConfiguration:
      status: Enabled
    replicationConfiguration:
      role: arn:aws:iam::...:role/role-name # Requires external role creation
      rules:
      - destination:
          storageClass: STANDARD_IA
          bucket: arn:aws:s3:::repl-dest
        deleteMarkerReplication:
          Status: Disabled
        priority: 0
        id: rule-1
        status: Enabled
  providerConfigRef:
    name: example
---
kind: Bucket
metadata:
  name: repl-dest
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: us-west-2
    paymentConfiguration:
      payer: BucketOwner
      rules:
          sseAlgorithm: AES256
    versioningConfiguration:
      status: Enabled
  providerConfigRef:
    name: example
```

Issue #657, opened 2021-05-07
duplication: RouteTable is also used to create Routes


Dkaykay:
### What happened?
As discussed with @muvaf via Slack:

**RouteTable** encapsulates **Route** creation ([source](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/ec2/routetable/controller.go#L299)), which is not a wanted feature. 
It should be possible to create RouteTables and Routes independently. 

This also seems to conflict with CPs aim for [high fidelity](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity) representation of the provider API.

References:

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateRouteTable.html
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateRoute.html

### How can we reproduce it?
Try to create a route in an existing routetable

### What environment did it happen in?
Crossplane version: 1.2.0
provider-aws: 0.18

muvaf:
I think we had had this discussion with @hasheddan but I couldn't find it now. I agree that having [`Route`](https://github.com/crossplane/provider-aws/blob/b8f8842/pkg/controller/ec2/routetable/controller.go#L337), [`RouteTable`](https://github.com/crossplane/provider-aws/blob/b8f8842/pkg/controller/ec2/routetable/controller.go#L165) and [`RouteTableAssociation`](https://github.com/crossplane/provider-aws/blob/b8f8842/pkg/controller/ec2/routetable/controller.go#L391) CRDs would be better rather than having it all in `RouteTable`.

It seems Terraform has `aws_route`, `aws_route_table` and `aws_route_table_association` but `aws_route_table` has optional list of `aws_route`s as well. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route_table

Dkaykay:
If you want, we could try to split up the resources.
Would this introduce a breaking change? How do you usually handle such changes? @muvaf

muvaf:
@Dkaykay I think we'd keep `RouteTable` as is for a while until it's `v1` and implement the additional resources.

@hasheddan thoughts about this issue?

hasheddan:
[...]

PR #768, opened 2021-07-13
Fix routetable isuptodate logic for multiple cidr and associations

enderv:

Fixes isUpToDate logic for routetables with multiple associations and in VPC's with multiple CIDR Blocks. IsRtUpTodate was always returning false due to the additional subnet association and default routes added. Also changed the resolvereferences logic so that it will only be attempted if there is something to resolve.

Just saw https://github.com/crossplane/provider-aws/issues/657 so ideally this logic won't be needed forever but in the meantime. 

I have:

Tested by scaling down deployed provider and running locally to ensure the isUpToDate was returning the expected values on routetables that exist in VPC with multiple CIDR blocks.

### DIFF
--- a/apis/ec2/v1beta1/referencers.go
+++ b/apis/ec2/v1beta1/referencers.go
@@ -121,7 +121,7 @@ func (mg *Subnet) ResolveReferences(ctx context.Context, c client.Reader) error 
 }
 
 // ResolveReferences of this RouteTable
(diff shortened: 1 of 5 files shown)


haarchri:
with #1084 we added ignoreRoutes in RouteTable and implemented Route resource

@enderv thanks for your PR - think we can close it because of #1084

Issue #227, opened 2020-05-25
Restore RDS Instance from a backup


rberrelleza:
# What problem are you facing?
Every night, a new snapshot with the latest test data is generated and stored in S3. I'd like to be able to use that data when creating an RDS instance with Crossplane. 

AWS supports this natively: https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceFromS3.html

negz:
Do you have any thoughts on how this might be modelled @rberrelleza? It seems like a fairly imperative API operation. One option would be to mirror Terraform's approach and essentially make it a subobject of our spec, maybe `spec.forProvider.createFromBackup` or similar?

rberrelleza:
That's what I was thinking @negz .  The [AWS command](https://docs.aws.amazon.com/cli/latest/reference/rds/restore-db-instance-from-s3.html) receives the location of the backup as a parameter. I think the same could be done here on a sub-object, to keep it clean, as you suggest:
```
spec:
  forProvider:
    fromBackup:
     s3: 
       bucket: mybucket
       prefix: prefix
      role: ....
```

Rhe 'recover from s3' functionality is aws-only, but cloudsql also has a similar mechanism to [create a DB from a backup](https://cloud.google.com/sql/docs/mysql/backup-recovery/backups#gcloud_1).  Not sure how much this would factor into the design, but long term it would be great to be able to support this across different cloud providers.

jbhennin:
Curious if this has been looked at recently?  Being able to create an instance from a snapshot and point in time snapshot would be a great feature.

juv:
Another feature that is being implicitly enabled by this is to allow developers to spin up a new database from a previous snapshot and then test new functionality on that database, then throw it away afterwards.

matthewhembree:
Without this feature, if the database were to become corrupted, the recovery process would be:

AWS CLI, Console, Terraform, Ansible, etc. (not Crossplane):
1. Restore (point in time) a DB snapshot to a new instance.
2. pg_dump the database(s) from the new instance.
3. drop all databases in the original Crossplane instance.
4. pg_restore the database(s) to the original Crossplane instance.

This enhancement would be very helpful and provide peace of mind --skipping the intermediary instance creation.

I personally would like to see snapshots (all engines supported) as a source over S3. S3 restore is MySQL only.

danports:
[...]

PR #770, opened 2021-07-15
Restore RDS instance from a MySQL backup

danports:
Adds support for restoring an RDS instance from a MySQL backup stored in an S3 bucket. Basic version of the functionality discussed in #227.

Fixes https://github.com/crossplane/provider-aws/issues/227

I have:

- [x] Read and followed Crossplane's contribution process.
- [x] Written tests
- [x] Tested on AWS by deploying a small (<1GB) MySQL backup from a bucket
- [x] Tested on AWS by deploying a medium (~35GB) MySQL backup from a bucket

Open to feedback on the API shape...just wanted to get something working for now. This my first time writing Go code, so this code probably needs a bit of work...some of it is not very DRY.


apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -368,6 +389,10 @@ type RDSInstanceParameters struct {
 	// +optional
 	EngineVersion *string `json:"engineVersion,omitempty"`
 
+	// FromBackup specifies the details of the backup to restore when creating a new RDS instance.
+	// +optional
+	FromBackup *RestoreBackupConfiguration `json:"fromBackup"`
```
chlunde, line 394: I think this name should take into consideration what the name/structure for restoring from DB snapshots should be too. If the structure will be specific to S3, maybe S3 should be part of the name?

`RestoreDBInstanceFromS3`
`RestoreDBInstanceFromDBSnapshot`
`RestoreDBInstanceToPointInTime`

negz: (Disclaimer: I'm not familiar with RDS restores apart from a cursory glance at the API).

If there are in fact multiple different sources you can restore from (e.g. snapshot, S3) perhaps using a type distinction here would make sense. This is a pattern we frequently use in our APIs, e.g.:

```yaml
spec:
  forProvider:
    fromBackUp:
      source: S3
      s3:
        s3bucketName: ...
```

I don't love that we're having to do so much interpretation of the APIs. Ideally our Crossplane managed resource would map directly to one API type (e.g. `CreateDBInstance`), while here we're decorating that API type with another. Given how imperative this call is and that it in fact results in the creation of a DB instance (AFAICT) I can't think of any great alternatives though.

pkg/controller/database/rdsinstance.go
```diff
@@ -169,11 +170,20 @@ func (e *external) Create(ctx context.Context, mg resource.Managed) (managed.Ext
...
 	}
 
-	req := e.client.CreateDBInstanceRequest(rds.GenerateCreateDBInstanceInput(meta.GetExternalName(cr), pw, &cr.Spec.ForProvider))
-	_, err = req.Send(ctx)
-	if err != nil {
-		return managed.ExternalCreation{}, awsclient.Wrap(err, errCreateFailed)
+	if cr.Spec.ForProvider.FromBackup == nil {
+		req := e.client.CreateDBInstanceRequest(rds.GenerateCreateDBInstanceInput(meta.GetExternalName(cr), pw, &cr.Spec.ForProvider))
+		_, err = req.Send(ctx)
+		if err != nil {
```
negz, line 176:
```suggestion
		if _, err = req.Send(ctx); err != nil {
```

Just a little style nit. Feel free to ignore, but the above suggestion is arguably more idiomatic in Go.

apis/database/v1beta1/rdsinstance_types.go
danports: Thanks, this feedback was exactly what I was looking for - I am not familiar enough with Crossplane's API patterns to understand how you guys would model a subclass/discriminator case like this. I'll update the model. Yeah, I don't really see any other options either given that the restore operation is a separate API call - it would be nicer if we could pass the restore parameters to the `CreateDBInstance` call, but it doesn't support that.

pkg/controller/database/rdsinstance.go
danports: Sure, thanks for the feedback - this was some copypasta from existing code but I'll update it.

stephanedicesare:
We are also interested in this functionality.
Is there a timeline when this can be merged?

jbhennin:
@danports have you considered also adding this functionality to DBCluster and DBInstance?  This would allow for feature parity and then the restore from snapshot and Pitr can be fast followers.

haarchri:
@chlunde  @muvaf  what are the open points here ? so rebase is open but in general the test is working

chlunde:
@haarchri I think this is the latest comment: https://github.com/crossplane/provider-aws/pull/770#discussion_r685420404 - so maybe the API should be written as negz shows to make it easier to implement the other types of restores (and easier to use).

I think the next step should be to first list all the unique parameters for each restore type, and then decide here in this issue if it makes sense to have the union of all restore options in the same struct (which is the approach the PR currently leads to) or one struct/type for each type of restore.

jbhennin:
@chlunde @haarchri @muvaf Just curious on how this PR stands.  We are really interested in the ability to restore from a snapshot.

danports:
This PR has been on the back burner for me since I needed the feature for a one-time migration that has long since been completed - I don't really have a use case for it anymore. But I'd be happy to address the feedback I received earlier and then leave it to others to take it a step further by adding support for other restore types, etc.

danports:
> I think the next step should be to first list all the unique parameters for each restore type, and then decide here in this issue if it makes sense to have the union of all restore options in the same struct (which is the approach the PR currently leads to) or one struct/type for each type of restore.

@chlunde I took a brief look at the parameters for the other restore types (DB snapshot, point in time) and there isn't really any overlap between the parameters for the different restore types, so the type distinction structure that @negz suggested earlier would probably make the most sense here. I will push a new version of the PR with these changes soon.

pkg/controller/database/rdsinstance.go
danports: Resolved by prior changes to the code due to AWS SDK updates.

apis/database/v1beta1/rdsinstance_types.go
danports: Introduced type distinction in the latest version of the PR.

danports:
I need to test the updated code on AWS, but otherwise it should be good to go...please let me know if you have more feedback @chlunde @negz. Happy 2022! 🥂

danports:
OK, lint issues should be gone now. Had a problem running `make lint` on my machine earlier, but was able to [fix](https://github.com/golangci/golangci-lint/issues/1219#issuecomment-656793602) that today.

haarchri:
@chlunde can you check ?

danports:
Tested my updated code on AWS and everything is looking good! 🥳 Let me know if there are any last nits to address here @chlunde.

haarchri:
I tested as well without issues LGTM @chlunde please do final review

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -106,6 +106,37 @@ type ScalingConfiguration struct {
 	SecondsUntilAutoPause *int `json:"secondsUntilAutoPause,omitempty"`
 }
 
+// S3RestoreBackupConfiguration defines the details of the S3 backup to restore from.
+type S3RestoreBackupConfiguration struct {
+	// BucketName is the name of the S3 bucket containing the backup to restore.
+	BucketName *string `json:"bucketName"`
```
chlunde, line 112: ideally, this and below would be referencers, but personally I don't mind this as a followup-PR at this point, since you've had to wait so long.

Here's an example: https://github.com/crossplane/provider-aws/commit/c9a8b38ae93a01cff5dd1ca7367f226d1eafecda
Docs: https://github.com/crossplane/crossplane-tools/pull/35

pkg/clients/rds/rds.go
```diff
@@ -131,6 +132,71 @@ func GenerateCreateDBInstanceInput(name, password string, p *v1beta1.RDSInstance
 	return c
 }
 
+// GenerateRestoreDBInstanceInput from RDSInstanceSpec
+func GenerateRestoreDBInstanceInput(name, password string, p *v1beta1.RDSInstanceParameters) *rds.RestoreDBInstanceFromS3Input {
+	c := &rds.RestoreDBInstanceFromS3Input{
```
chlunde, line 137: maybe leave a short comment/note inside `GenerateCreateDBInstanceInput` that any changes there should be applied in `GenerateRestoreDBInstanceInput`. It makes it more likely to pick up a bug either during implementation or code review.

chlunde (changes requested):
Looks very good, thanks for your contribution and patience! Due to the necessary duplication I think a friendly note for the future reviewers would be nice, except for that it looks good to me.

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -368,6 +399,10 @@ type RDSInstanceParameters struct {
 	// +optional
 	EngineVersion *string `json:"engineVersion,omitempty"`
 
+	// RestoreFrom specifies the details of the backup to restore when creating a new RDS instance. (If the RDS instance already exists, this property will be ignored.)
+	// +optional
+	RestoreFrom *RestoreBackupConfiguration `json:"restoreFrom"`
```
muvaf, line 404:
```suggestion
	RestoreFrom *RestoreBackupConfiguration `json:"restoreFrom,omitempty"`
```

The optional fields should [have](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) `omitempty` tag.

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -106,6 +106,37 @@ type ScalingConfiguration struct {
...
+type S3RestoreBackupConfiguration struct {
+	// BucketName is the name of the S3 bucket containing the backup to restore.
+	BucketName *string `json:"bucketName"`
+
+	// IngestionRoleARN is the IAM role RDS can assume that will allow it to access the contents of the S3 bucket.
+	IngestionRoleARN *string `json:"ingestionRoleARN"`
+
+	// Prefix is the path prefix of the S3 bucket within which the backup to restore is located.
+	// +optional
+	Prefix *string `json:"prefix"`
```
muvaf, line 119:
```suggestion
	Prefix *string `json:"prefix,omitempty"`
```

muvaf (approved):
Sorry this was assigned to me for review and has been open for so long and thanks for your patience. It looks good to me in general with a few small notes. @chlunde feel free to merge it once your concerns are addressed.

muvaf:
Also feel free to implement it for `DBInstance` in `rds` group as well, which would be mostly copy-paste, in another PR.

pkg/clients/rds/rds.go
danports: Added a comment to both functions.

apis/database/v1beta1/rdsinstance_types.go
danports: Yeah, let's leave that for a later time - I would like to get this out the door without changes that would require additional testing, etc. Referencers look pretty neat though!

haarchri (approved):
all concerns are addressed - did a final test - everything is working as expected

thanks @danports for implementation

chlunde (approved):
Looks good to me, thank you for your contribution.

PR #781, opened 2021-07-28
Remove ECR policyText observation

### LINKED ISSUE #780: ECR RepositoryPolicy constantly reconciles
### What happened?
Seeing a very high rate (upwards of 10 per minute per resource) of reconciles for `repositorypolicy.ecr.aws.crossplane.io` against `provider-aws` master, even though all policies exist, are up to date and not late-initialized.

Given that there can only be one policy per repository, I would expect to see a similar reconcile rate as seen with repositories but this is not the case.

It looks like this is caused by the AWS API constantly returning array items inside my repository policy (in this case AWS Principals) in a different order. I've confirmed this locally - no two subsequent calls to `aws ecr get-repository-policy` for a particular repository returns the same policy order.

This is correctly ignored by `provider-aws` and marked as up to date (as we ignore sort order when comparing the policy as of #701), but it looks like when the controller persists the 'updated' resource `status` back to the Kubernetes API, it causes the requeue period that `provider-aws` requested to be ignored, and the resource is requeued straight away.

This ends up in an endless loop of updates that has some other nasty side effects like blowing out the AWS SDK connection pool.

### How can we reproduce it?

Apply an ECR `repositorypolicy` with the following spec, with many different AWS account ID's:

```yaml
spec:
   deletionPolicy: Delete
   forProvider:
     policy:
       id: AllowAccounts
       statements:
       - action:
         - ecr:BatchCheckLayerAvailability
         - ecr:BatchGetImage
         - ecr:GetDownloadUrlForLayer
         effect: Allow
         principal:
           awsPrincipals:
           - awsAccountId: arn:aws:iam::3398....7229:root
           - awsAccountId: arn:aws:iam::9718....6735:root
           - awsAccountId: arn:aws:iam::8433....0030:root
           - awsAccountId: arn:aws:iam::4883....5958:root
           - awsAccountId: arn:aws:iam::9418....5648:root
           - awsAccountId: arn:aws:iam::3468....4853:root
         sid: AllowPullNonProdAccounts
       version: "2008-10-17"
     region: us-east-1
```

And then monitor the number of reconciles. I worked out what was going on by dumping the policy inside kubernetes at set intervals and then diffing the yaml output.

Crossplane version: v1.3
Provider-AWS version: 009f048d509ba924ce6c456740ab946666a49bb6

### Fix
I wonder if `policyText` should even be available on this resource, as this is just a different representation of a user-defined setting and the resource should be updated if these don't match.

benagricola:

Amazon ECR's `GetRepositoryPolicyRequest` API appears to return policy
statement entries in any order it feels like - this is properly handled
by Crossplane when identifying if a policy needs an update, but we fall
foul of Kubernetes' handling of resource watches, where the changing
`policyText` contents of the resource trigger a new reconcile straight
after the previous one.

This causes constant API observations, and with enough ECR policies and
a short enough poll interval, blows out the connection pool of the AWS
SDK.

Removing the `policyText` observation field and using the data available
from the API response call to detect updates avoids this situation from 
occurring.

Signed-off-by: Ben Agricola <bagricola@squiz.co.uk>

Fixes #780 

I have:

Extra test added to make sure the resource correctly returns up- and not-up-to-date on appropriate policies. Have tested on a real cluster that was experiencing the issue in question to confirm that it resolves the original problem.

### DIFF
--- a/apis/ecr/v1alpha1/policy_types.go
+++ b/apis/ecr/v1alpha1/policy_types.go
@@ -235,10 +235,7 @@ type RepositoryPolicySpec struct {
 }
 
 // RepositoryPolicyObservation keeps the state for the external resource
-type RepositoryPolicyObservation struct {
-	// The JSON repository policy text associated with the repository.
-	PolicyText string `json:"policyText,omitempty"`
-}
+type RepositoryPolicyObservation struct{}
 
 // A RepositoryPolicyStatus represents the observed state of a repository policy
 type RepositoryPolicyStatus struct {
--- a/package/crds/ecr.aws.crossplane.io_repositorypolicies.yaml
+++ b/package/crds/ecr.aws.crossplane.io_repositorypolicies.yaml
@@ -485,11 +485,6 @@ spec:
               atProvider:
                 description: RepositoryPolicyObservation keeps the state for the external
                   resource
-                properties:
-                  policyText:
-                    description: The JSON repository policy text associated with the
-                      repository.
-                    type: string
                 type: object
               conditions:
                 description: Conditions of the resource.
--- a/pkg/controller/ecr/repositorypolicy/controller.go
+++ b/pkg/controller/ecr/repositorypolicy/controller.go
@@ -109,10 +109,6 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		return managed.ExternalObservation{}, errors.Wrap(err, errGet)
 	}
 
-	if response.PolicyText != nil {
-		cr.Status.AtProvider.PolicyText = *response.PolicyText
-	}
-
 	current := cr.Spec.ForProvider.DeepCopy()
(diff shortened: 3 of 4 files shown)


benagricola:
`make test` passes locally, but can't get the `e2e` tests to work 🤔

benagricola:
Post-discussion in Slack around whether removing this field requires an API version bump, my read was that this would not be necessary, as the field in question is a mirror of the `rawPolicy` input spec field.

Removing this field _would_ mean if a user provided the complex `policy` object instead, then they would no longer have access to the policy text generated from that, but IMHO it would not make any sense to use the resource in this manner and it is unlikely that it is already being used that way.

If we have agreement on that front, then I think this PR is good to go 👍

chlunde (approved):
I think this is a good idea for the reasons you say, and also because other APIs like `BucketPolicy` does not have the policy in the status.

chlunde (approved):

chlunde:
Thanks! Tested, works nicely.

PR #807, opened 2021-08-17
Implements ACM ResourceRecord for dns validation in status field

### LINKED ISSUE #804: Provide Certificate dns info in status

It's impossible to fully automate ACM certs using DNS validation when using Crossplane resources for AWS. I am trying to automate the certs through a Composition but there is still the manual step of adding the needed CNAME record to Route53 for the DNS validation of a Certificate to be complete. 

To create a DNS validated cert use the following template: 
```yaml
apiVersion: acm.aws.crossplane.io/v1alpha1
kind: Certificate
metadata:
  name: foo.example.com
spec:
  forProvider:
    domainName: foo.example.com
    region: us-east-1
    validationMethod: DNS
    tags:
    - key: Name
      value: foo.example.com
```
Now to fully validate the certificate:
You must manually log into the AWS console and navigate to the created resource.
There is an option to create a CNAME record in Route53 and you click to create it. 
Only now will the Crossplane resource say the Certificate is ready. 

Or you can get the information for needed CNAME like so with cli:
```
AWS_REGION=us-east-1 aws acm describe-certificate \
  --certificate-arn <arn from certificate> \
  --query 'Certificate.DomainValidationOptions'
```

My Issue is I want this to be automated in a Crossplane `Composition`, but I can't because the `Certificate` resource within kubernetes does not have information about the needed CNAME record for DNS validation. 

In the status of the `Certificate`, show information about the needed CNAME record to properly finish DNS validation. 

It could look something like this:
```yaml
status:
  atProvider:
    dnsValidation:
      recordName: _dsj24j034j80f34j0adpsw.foo.example.com
      recordValue: _dfijodfsope9023r.dfsijqwiu.acm-validations.aws
      type: CNAME
      policy: simple
```

Now to fully automate ACM certs with some patch magic in the `Composition`, create a `ResourceRecordSet` using the values from the certs status so you would end up with:
```yaml
apiVersion: route53.aws.crossplane.io/v1alpha1
kind: ResourceRecordSet
metadata:
  name: _dsj24j034j80f34j0adpsw.foo.example.com
spec:
  forProvider:
    type: CNAME
    resourceRecords:
    - value: _dfijodfsope9023r.dfsijqwiu.acm-validations.aws
    zoneIdRef:
      name: example.com
```

Automation!

haarchri:
Signed-off-by: Christopher Haar <chhaar30@googlemail.com>

Implements ACM ResourceRecord for dns validation in status field

Fixes #804

```
  At Provider:
    Certificate ARN:      arn:aws:acm:eu-central-1:255932642927:certificate/3eb5bed4-6539-4e3d-8bd4-425810f89d60
    Renewal Eligibility:  INELIGIBLE
    Resource Record:
      Name:   _ff74bcdb41a2f939c93d449daad4565a.example.com.
      Type:   CNAME
      Value:  _03494a5eb651455affe9cb5fa3bc0e75.bwlshdtstt.acm-validations.aws.
    Status:   PENDING_VALIDATION
    Type:     AMAZON_ISSUED
```

I have:


kferrone:
Looks good thank you. 
I have only one concern though, mostly curious, what is the value of `ResourceRecord` when the cert is email validated?

pkg/clients/acm/certificate.go
```diff
@@ -74,6 +74,21 @@ func GenerateCreateCertificateInput(name string, p *v1alpha1.CertificateParamete
 
 // GenerateCertificateStatus is used to produce CertificateExternalStatus from acm.certificateStatus
 func GenerateCertificateStatus(certificate acm.CertificateDetail) v1alpha1.CertificateExternalStatus {
+
+	if certificate.Type == acm.CertificateTypeAmazonIssued && certificate.DomainValidationOptions[0].ResourceRecord != nil {
```
chlunde, line 78: Did you consider adding another condition here to ensure that the whole provider will not panic if `DomainValidationOptions` is not present or zero length? normally I would check `len(certificate.DomainValidationOptions) > 0` before accessing `certificate.DomainValidationOptions[0]`

pkg/clients/acm/certificate.go
```diff
@@ -74,6 +74,21 @@ func GenerateCreateCertificateInput(name string, p *v1alpha1.CertificateParamete
 
 // GenerateCertificateStatus is used to produce CertificateExternalStatus from acm.certificateStatus
 func GenerateCertificateStatus(certificate acm.CertificateDetail) v1alpha1.CertificateExternalStatus {
+
```
chlunde, line 77: There is a test for this function: https://github.com/crossplane/provider-aws/blob/6dce815528e3df171052a7638935662ac6082ae9/pkg/clients/acm/certificate_test.go#L139-L174

Could you please add coverage for this branch? Thanks!

chlunde (changes requested):

pkg/clients/acm/certificate.go
haarchri: @chlunde  thanks for review - i will check it

pkg/clients/acm/certificate.go
haarchri: @chlunde  thanks for review - i will check it

chlunde (approved):
LGTM

conclusionlogic:
@haarchri
What about hostnames that are supplied on the [subjectAlternativeNames](https://doc.crds.dev/github.com/crossplane/provider-aws/acm.aws.crossplane.io/Certificate/v1alpha1@v0.21.0#spec-forProvider-subjectAlternativeNames) field?
These, too, need validating but currently only the main hostname requested on [domainName](https://doc.crds.dev/github.com/crossplane/provider-aws/acm.aws.crossplane.io/Certificate/v1alpha1@v0.21.0#spec-forProvider-domainName) field is parsed back into [status.atProvider.resourceRecord](https://doc.crds.dev/github.com/crossplane/provider-aws/acm.aws.crossplane.io/Certificate/v1alpha1@v0.21.0#status-atProvider-resourceRecord) object, unless I missed something?

Based on my observation, each listed SAN would need its own ResourceRecord entry because their validation record (`CNAME name` and `CNAME value`) are all different; and validating only the primary domainName will unfortunately not validate all other (SAN) hostnames, effectively breaking the Certificate issuing process if one wants to have the whole thing automated end to end.

In fact, [status.atProvider.resourceRecord](https://doc.crds.dev/github.com/crossplane/provider-aws/acm.aws.crossplane.io/Certificate/v1alpha1@v0.21.0#status-atProvider-resourceRecord) should ideally be a list of objects, making it simple to patch the whole thing into ResourceRecordSet's [resourceRecords](https://doc.crds.dev/github.com/crossplane/provider-aws/route53.aws.crossplane.io/ResourceRecordSet/v1alpha1@v0.21.0#spec-forProvider-resourceRecords) field (which expects a list of objects already) and, after a small revamp on [resourceRecords](https://doc.crds.dev/github.com/crossplane/provider-aws/route53.aws.crossplane.io/ResourceRecordSet/v1alpha1@v0.21.0#spec-forProvider-resourceRecords) side making in accept not just `value` but `name` and `type` as well, get all the DNS validation records automatically created into Route53 in one swift go without any extra gymnastics or hocus-pocus needed from the endusers.

I mean, wouldn't that be wonderful?

haarchri:
@conclusionlogic can you open a new Issue for that ? with information how the san validation entries looks like ? In general https://github.com/aws/aws-sdk-go/blob/v1.42.8/service/acm/api.go#L2336-L2372 we used the go sdk struct for resourceRecord

PR #825, opened 2021-09-08
Give all EC2 managed resources a 3 minute creation grace period

### LINKED ISSUE #802: Leaked `RouteTable` due to eventually consistent `DescribeRouteTables` API
### What happened?
After deleting a composite and waiting, with kubectl wait,  for all of the referenced api objects to be deleted, managed resources still existed in AWS.
This is a problem for us because this wait logic is a gate used to delay deletion of our k8s cluster.  If the managed resources in AWS aren't cleaned up by the time we are done waiting they will most likely end up orphaned.  ( We are seeing this often )

### How can we reproduce it?
Here is the clean up logic we are performing before allowing the cluster to be deleted

```
Get all CompositeResourceDefinition
For each CompositeResourceDefinition
  If .spec.claimNames.kind
    Get all claims of .spec.claimNames.kind
    For claim
      Get resource refs
      Delete claim
    Wait for all resources to be deleted
  For each composition of .spec.names.kind
    Get resource refs
    Delete composition
    Wait for all resources to be deleted
Get all managed resources ( kubectl get managed )
  For each managed resource
    Delete managed resource
```

We expect that once we have run through the above steps all managed resources have been deleted from AWS.
However this doesn't seem to be the case.

In the attached zip are example 
* xrd.yaml
* composition.yaml
* claim.yaml
used to create the resources in AWS.

Also in the zip is events-for-vpc-14449-d66-dmi.xlsx which shows the events around one of the leaked VPC.
* Events at top are most recent.
* Events with white background are from CloudTrails
* Events with light blue background are from my team's clean up code.
* Events with green background are from Gardener

At line 81 you can see the first event from my team's clean up code.
By line 31, all of the resources related to the VPC have been cleaned up from k8s.   There is nothing left for our clean up code to delete or wait for.
Our last attempt to wait for a resources to be cleaned up is about 9 seconds after the last event for the VPC in CloudTrails.   Given all of the deletion attempts that failed with Client.DependencyViolation I am wondering if Crossplane has given up.   ( And why are all those failures there in the first place? )
Anyhoo once our cleanup code has nothing left to wait for cluster Gardener is free to begin with the cluster deletion.   You can see at line 28, about 45 seconds after my team's clean up code
finished, where it begins taking destructive action.   It starts with  simply going through the api server and deleting all CRD that it can find.
Since Gardener is just blindly deleting things if the AWS resources haven't been cleaned up at this point it is highly likely they are going to be orphaned.
[14449-d66-dmi-leak.zip](https://github.com/crossplane/crossplane/files/6812143/14449-d66-dmi-leak.zip)

crossplane:v1.2.2
aws provider provider-aws:v0.18.1
kubernetes 1.19.10

negz:

Fixes #802

This means we'll wait three minutes from the time a resource was most recently created before we trust the EC2 API if/when it indicates the resource doesn't exist. The crossplane-runtime bump should also make our controllers more resilient to stale controller-runtime cache reads.

I have:

I tested this change extensively as part of crossplane/crossplane-runtime#283. I've run a cursory version of the same test on this specific PR (i.e. spun a composition of EC2 resources up and down once).


chlunde (approved):

go.mod
```diff
@@ -5,19 +5,19 @@ go 1.16
...
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
-	k8s.io/api v0.21.2
-	k8s.io/apimachinery v0.21.2
-	k8s.io/client-go v0.21.2
-	sigs.k8s.io/controller-runtime v0.9.2
-	sigs.k8s.io/controller-tools v0.4.0
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
+	k8s.io/client-go v0.21.3
+	sigs.k8s.io/controller-runtime v0.9.6
```
chlunde, line 21: I suspect this will require rebasing lots of the other "new resource" PRs in this repo, and possibly the two large bumps of AWS SDK PRs? If so, I'm a bit worried about contributor "churn" due to this. The "solution" is probably to reduce the number of WIP PRs, but it's also a question of what goes in first of these changes (which ones are mechanical and easier to rebase).

negz: Fair point. We already released patches with this fix, so I'm happy to hold off and take the rebase hit after we get the backlog down a bit.

AaronME:
@negz Looks like this is ready to merge?

PR #860, opened 2021-10-05
elasticache: Scale replicationgroup shards

### LINKED ISSUE #531: ReplicationGroup: Add support for shards (node groups) scaling
It is impossible to change the number of node groups for a ReplicationGroup after deployment. Redis (cluster mode enabled) replication group does allow this, but Crossplane does not.

Probably related to https://github.com/crossplane/provider-aws/blob/master/pkg/clients/elasticache/elasticache.go#L193-L212 and https://github.com/crossplane/provider-aws/blob/f600ff88ad64df73295c96e4446be83b608ddfa4/apis/cache/v1beta1/replication_group_types.go#L424 (immutable)

This is probably a common use case which Crossplane should be able to handle.

chlunde:
Scale up and down replication groups as requested by the user.

On scale down, the AWS API requires us to pick the nodes to remove. I have not considered what the best option is, I picked the first node in the list.

Co-authored-by: sb1-oscar <sb1-oscar@users.noreply.github.com>

Fixes #531

I have:

Tested changing the CR to scale up/down. I will have to retest since this PR has been rebased and it's a while since I wrote it.


JMoserCricut (approved):

sqerison:
Hi team, we are looking forward to this feature being delivered to use in our project. But, unfortunately, before that, we couldn't use Crossplane for Elasticache.

haarchri (approved):
tested the implementation - scaling replicationgroup shards is working thanks for this enhancement @chlunde ;)

Issue #846, opened 2021-09-28
Support for localstack


samidbb:
### What problem are you facing?
Developers in our company use [LocalStack](https://localstack.cloud/) to run various AWS services in their local environment. This enable them to verify their code and configuration changes before pushing them to the repo. The benefits from that are 1) they save time by not having to wait for the pipelines to finish to waiting for the changes to take effect and 2) Saving cost by reducing the need of having to spin up additional environments for verifying their code changes.

In addition to these things, we will also provide a great value for us as we are working on providing a fully functioning local development experience Kubernetes setup using Docker setup and Kind. 

As platform engineers we will greatly be benefiting from the possibility to our crossplane packages without having to spin up resources on AWS and paying a fortune for spinning up new test resources.

### How could Crossplane help solve your problem?

I usually spin up LocalStack environment using Terraform provider for AWS by providing a custom endpoints for each service I want to provision and skipping credential validation as described here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/custom-service-endpoints

So I believe the set of parameters can be enabled in the AWS Provider for Crossplane to interface with the localstack. 

I think there is an already opened issue that has something to do with the above here #811

chlunde:
@iubog2008 @samidbb I just noticed this code:

https://github.com/crossplane/provider-aws/blob/ff17f842b9d8a2d9e1e73b22b99d182725b792ef/pkg/clients/aws.go#L114-L115

Could you please try with adding these annotations to the managed object and see if you have any luck with localstack?

```
aws.alpha.crossplane.io/endpointServiceID
aws.alpha.crossplane.io/endpointURL
```

liubog2008:
I add some hostAliases manually. It's not good enough, but ok for me.

ZaradarDFDS:
Does anyone know what value to supply for endpointServiceID?

PR #869, opened 2021-10-08
clients: use HostnameImmutable property in new AWS SDK v2

muvaf:

In some services like S3, the SDK changes the request URL to `http://bucketname.endpointURL` which is not easily supported by alternative endpoints like localstack. This PR makes use of new `HostnameImmutable` property in AWS SDK v2 and removes `endpointServiceID` annotation because these configurations are already per-CRD, hence there isn't really a case where you need to choose a set of services to allow/deny.

Fixes https://github.com/crossplane/provider-aws/issues/846

I have:

Manually with localstack.


apis/v1beta1/providerconfig_types.go
```diff
@@ -26,6 +26,10 @@ import (
 type ProviderConfigSpec struct {
 	// Credentials required to authenticate to this provider.
 	Credentials ProviderCredentials `json:"credentials"`
+
+	// Endpoint is where you can override the default endpoint configuration
+	// of AWS calls made by the provider.
+	Endpoint *EndpointConfig `json:"endpoint,omitempty"`
```
hasheddan, line 32: nitpick: (and on other `omitempty` fields below)
```suggestion
	// of AWS calls made by the provider.
	// +optional
	Endpoint *EndpointConfig `json:"endpoint,omitempty"`
```

pkg/clients/aws.go
```diff
@@ -97,39 +96,48 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
...
-				}
-
-				return endpoint, &aws.EndpointNotFoundError{}
+func SetResolver(pc *v1beta1.ProviderConfig, cfg *aws.Config) *aws.Config {
+	if pc.Spec.Endpoint == nil {
+		return cfg
+	}
+	cfg.EndpointResolver = aws.EndpointResolverFunc(func(service, region string) (aws.Endpoint, error) {
+		e := aws.Endpoint{
+			URL:               pc.Spec.Endpoint.URL,
```
muvaf, line 124: I went back to initial [implementation PR](https://github.com/crossplane/provider-aws/pull/382#discussion_r504922347) and it seems like we need to construct the URL using `service` and `region` here.

muvaf: @hasheddan @chlunde Alright, I have changed the struct a little bit to account for dynamic and static URL resolving, which should cover all cases SDK allows. Tested the static config with localstack and dynamic config with `amazonaws.com`, added examples for both. It should be ready for another review.

muvaf:
Labelled as breaking-change to not forget adding instructions for migration in the release notes.

PR #870, opened 2021-10-08
Add LifecyclePolicies for AWS ECR Repositories

### LINKED ISSUE #864: LifecyclePolicy For ECR Repositories
We are using crossplane to create ECR Repositories for CICD pipelines, but while we are able to create the repositories, we are unable to add a lifecycle policy to the specific repository through crossplane.

> A lifecycle policy contains one or more rules, where each rule defines an action for Amazon ECR. This provides a way to automate the cleaning up of unused images, for example expiring images based on age or count

We would like a way of applying Lifecycle policies to ECR repositories.

Input object (a string of the json policy rules): https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecr#PutLifecyclePolicyInput
Rule syntax: https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lifecycle_policy_syntax

larhauga:
This PR includes the addition of lifecycle policies for ECR Repositories.
Fixes #864

I have:

Updated go tests, and run locally against a kind cluster that points to AWS with the following example

```example.yaml
apiVersion: ecr.aws.crossplane.io/v1alpha1
kind: Repository
metadata:
  name: example
  annotations:
    crossplane.io/external-name: example
spec:
  forProvider:
    region: eu-north-1
    imageTagMutability: IMMUTABLE
    lifecyclePolicy:
      rules:
      - rulePriority: 1
        description: Expire images older than 14 days
        selection:
          tagStatus: untagged
          countType: sinceImagePushed
          countUnit: days
          countNumber: 14
        action:
          type: "expire"

```


pkg/controller/ecr/repository/controller.go
```diff
@@ -143,17 +153,37 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		}
 	}
 
+	for _, awsClient := range e.subresourceClients {
```
muvaf, line 156: We used this pattern of subresources for S3 because it has a lot of them but I don't think we should do it for other resources unless we really have to. Do you think `LifecyclePolicy` can become a resource of its own?

And have you considered using [code generation](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md) to generate it?

larhauga: I think that an ECR Repository has some similarities with S3 with that it has additional resources for each repository. SubResources like `RepositoryPolicy` and `LifecyclePolicy` only make sense in the context of the Repository, the way the api is structured. The intent of reusing the pattern from s3 were to reduce the complexity of the lifecycle functions, and handle some of the gocyclo warnings.

If the subresource pattern is not to be used here; do you have any input on how this should be implemented in a better way?

For Lifecycle policy at least, the API is a 1-1 relationship (https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#PutLifecyclePolicyInput). With a new resource specifically for LifecyclePolicy, it would always require a selector or ref to find another object. Information in the policy is also directly related to the repository (eg. the tag prefix).

TLDR; I dont think LifecyclePolicy should be a seperate resource. I started out with trying to use code gen for this, but ended up with this implementation.

muvaf: > For Lifecycle policy at least, the API is a 1-1 relationship (https://docs.aws.amazon.com/sdk-for-go/api/service/ecr/#PutLifecyclePolicyInput). With a new resource specifically for LifecyclePolicy, it would always require a selector or ref to find another object. Information in the policy is also directly related to the repository (eg. the tag prefix).

This is indeed not a rare pattern. For example, we have `IAMRolePolicyAttachment` where it always refers to a single `IAMRole` and a single `IAMPolicy` where it could've been that policy attachment could be a part of `IAMRole`. Another example could be [`VPCCIDRBlock`](https://doc.crds.dev/github.com/crossplane/provider-aws/ec2.aws.crossplane.io/VPCCIDRBlock/v1alpha1@v0.20.2) where it always targets a specific `VPC`. There are similar cases in other providers, too.

The way we usually think about such separations is not very absolute and you'll see exceptions here and there but having it as separate is preferred most of the time. One of the main reasons for this is the applicability of RBAC; if `LifecyclePolicy` is part of `Repository` then the restrictions you put in place cannot be as granular, everyone who can create `Repository` can create its `LifecyclePolicy` which might be fine but for example AWS lets you have that granularity and we're [trying to represent](https://github.com/crossplane/crossplane/blob/master/design/one-pager-managed-resource-api-design.md#high-fidelity) the API as close as possible.

larhauga:
@muvaf this PR is replaced by #1181. Added LifecyclePolicy as a new resource with codegen as you suggested.

PR #871, opened 2021-10-12
#868 Fix broken handling of errors after aws-sdk-v2 migration

### LINKED ISSUE #868: [showstopper] Some error checks are broken with v2 migration
### What happened?

Some of the `IsNotFound` checks are reporting false negative, such as this one, because the underlying error types have changed with the new v2 SDK. The issue is seen in checks with type checks like [S3 Bucket](https://github.com/crossplane/provider-aws/blob/a6b9ff1/pkg/clients/s3/bucket.go#L121) or [EC2 Instance](https://github.com/crossplane/provider-aws/blob/a6b9ff1/pkg/clients/ec2/instance.go#L54).

This is a showstopper for the next release.

### How can we reproduce it?

Create an S3 bucket or Ec2 instance.

Crossplane version: v1.4.1
AWS version: `4339c3a`

larhauga:
Fixes #868

I have:

Tested ec2.SecurityGroup, ECR.Repository, ECR.RepositoryPolicy and S3.Bucket.

Tested ok in our environment with a fork, where it previously failed.
We are not running all the affected types, so not every change has been thoroughly tested.

### DIFF
--- a/pkg/clients/ec2/address.go
+++ b/pkg/clients/ec2/address.go
@@ -2,6 +2,7 @@ package ec2
 
 import (
 	"context"
+	"errors"
 	"sort"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -30,7 +31,8 @@ type AddressClient interface {
 
 // IsAddressNotFoundErr returns true if the error is because the address doesn't exist
 func IsAddressNotFoundErr(err error) bool {
-	if awsErr, ok := err.(smithy.APIError); ok {
+	var awsErr smithy.APIError
+	if errors.As(err, &awsErr) {
 		if awsErr.ErrorCode() == AddressAddressNotFound || awsErr.ErrorCode() == AddressAllocationNotFound {
 			return true
 		}
--- a/pkg/clients/ec2/vpccidrblock.go
+++ b/pkg/clients/ec2/vpccidrblock.go
@@ -2,12 +2,12 @@ package ec2
 
 import (
 	"context"
+	"errors"
 
 	awsgo "github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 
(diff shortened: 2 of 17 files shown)


pkg/clients/ec2/address.go
```diff
@@ -30,7 +31,8 @@ type AddressClient interface {
 
 // IsAddressNotFoundErr returns true if the error is because the address doesn't exist
 func IsAddressNotFoundErr(err error) bool {
-	if awsErr, ok := err.(smithy.APIError); ok {
+	var awsErr smithy.APIError
+	if errors.As(err, &awsErr) {
 		if awsErr.ErrorCode() == AddressAddressNotFound || awsErr.ErrorCode() == AddressAllocationNotFound {
```
muvaf, line 36: Have you validated that `errors.As` assign the actual struct into the provided pointer, i.e. `awsErr`? I remember looking at the source code of that function and it was doing only a type check.

chlunde: That is `errors.Is`, not `.As`?
https://pkg.go.dev/errors#pkg-functions
https://play.golang.org/p/CVmMXLy6_Bb

pkg/clients/ec2/vpccidrblock.go
```diff
@@ -39,16 +39,8 @@ func (r *CIDRNotFoundError) Error() string {
 
 // IsCIDRNotFound returns true if the error code indicates that the CIDR Block Association was not found
 func IsCIDRNotFound(err error) bool {
-	if _, ok := err.(*CIDRNotFoundError); ok {
```
chlunde, line 42: should this still look for both "cidr not found" and "cidr association not found", or is that one error? cc @enderv

pkg/clients/s3/bucket.go
```diff
@@ -118,14 +119,14 @@ func NewClient(cfg aws.Config) BucketClient {
...
-	return ok
+	var notFoundError *s3types.NoSuchBucket
+	return errors.As(err, &notFoundError)
 }
 
 // IsAlreadyExists helper function to test for ErrCodeBucketAlreadyOwnedByYou error
 func IsAlreadyExists(err error) bool {
-	_, ok := err.(*s3types.BucketAlreadyOwnedByYou)
-	return ok
+	var notFoundError *s3types.BucketAlreadyOwnedByYou
```
chlunde, line 128: rename variable

Issue #684, opened 2021-05-30
Not able to create natgateway ( 1.2.1)


ranjan68668:
Hi , while creating natgateway with below manifest, I am getting error. Can someone help me please ? All ther resource is there , it's just natgateway not getting created.

---
apiVersion: ec2.aws.crossplane.io/v1beta1
kind: NATGateway
metadata:
  name: natgw-1
spec:
  providerConfigRef:
    name: ${provider_name}
  forProvider:
    region: ${aws_region}
    allocationIdRef:
      name: ${EIP}
    subnetIdRef:
      name: ${pub-sub}

Error:
failed to create the NATGateway resource: InvalidRequest: The request received was invalid.

haarchri:
@ranjan68668  can you provide more information ? 
i tested it on my side - all natgateways created without problems

```
NAME                      INSTALLED   HEALTHY   PACKAGE                               AGE
crossplane-provider-aws   True        True     crossplane/provider-aws:v0.19.0
```

```
kubectl get natgateways.ec2.aws.crossplane.io
NAME                  READY   SYNCED   ID                      VPC            SUBNET            ALLOCATION ID       AGE
sample-natgateway     True    True     nat-08c65e403bb12d60c   vpc-aab73bc0   subnet-f2ccb898   eipalloc-062b020d   12m
sample-natgateway-2   True    True     nat-0c8ac0e49c6f41224   vpc-aab73bc0   subnet-f2ccb898   eipalloc-386b4233   5m45s
```

etcshad0vv:
Hello, i have similar error when i am using `NATGateway` as a managed resource within a Composition on version `v0.19.0`. If i will create this resource as primitive, works fine! Not sure if it's related to @ranjan68668 issue, although the error is the same.

Crossplane: `v1.4.1`

To me it looks like it cannot reference vpc by subnet.

```
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: iscp-eksnetwork
  labels:
    provider: aws
spec:
  compositeTypeRef:
    apiVersion: aws.cloud.iscp.io/v1
    kind: CompositeEKSNetwork
  patchSets:
  - name: common
    patches:
    - type: FromCompositeFieldPath
      fromFieldPath: "spec.parameters.providerConfigRef.name"
      toFieldPath: "spec.providerConfigRef.name"
    - type: FromCompositeFieldPath
      fromFieldPath: "spec.parameters.deletionPolicy"
[...]

PR #899, opened 2021-10-30
fix(nat): make tagSpecification optional to fits nat-gateway without tags

haarchri:

fixed 
` failed to create the NATGateway resource: api error InvalidParameter: Tag specification must have at least one tag`
- skip tagSpecification if tags nil - then nat-gateway is created without error message

Fixes https://github.com/crossplane/provider-aws/issues/684

I have:

kubectl apply -f examples/ec2/natgateway-without-tags.yaml 

```
[...]
Spec:
  Deletion Policy:  Delete
  For Provider:
    Allocation Id:  eipalloc-0386dc8235394d295
    Allocation Id Ref:
      Name:     sample-eip
    Region:     us-east-1
    Subnet Id:  subnet-0bea8bdb2151f4fbe
    Subnet Id Ref:
      Name:  sample-subnet1
  Provider Config Ref:
    Name:  example
Status:
  At Provider:
    Create Time:  2021-11-24T07:49:32Z
    Nat Gateway Addresses:
      Allocation Id:         eipalloc-0386dc8235394d295
      Network Interface Id:  eni-04b3f0dd378d5b51d
      Private Ip:            10.0.1.150
      Public Ip:             34.225.0.40
    Nat Gateway Id:          nat-0cdc37e208c2f4c36
    State:                   available
    Vpc Id:                  vpc-0557e8b3073ddddad
  Conditions:
    Last Transition Time:  2021-11-24T07:51:34Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2021-11-24T07:49:32Z
    Status:                True
    Type:                  Synced
```

kubectl apply -f examples/ec2/natgateway.yaml 

```
[...]
Spec:
  Deletion Policy:  Delete
  For Provider:
    Allocation Id:  eipalloc-0386dc8235394d295
    Allocation Id Ref:
      Name:     sample-eip
    Region:     us-east-1
    Subnet Id:  subnet-0bea8bdb2151f4fbe
    Subnet Id Ref:
      Name:  sample-subnet1
    Tags:
      Key:    Name
      Value:  sample-natgateway
  Provider Config Ref:
    Name:  example
Status:
  At Provider:
    Create Time:  2021-11-24T09:09:38Z
    Nat Gateway Addresses:
      Allocation Id:         eipalloc-0386dc8235394d295
      Network Interface Id:  eni-07bc918765cf65ba0
      Private Ip:            10.0.1.182
      Public Ip:             34.225.0.40
    Nat Gateway Id:          nat-0194775b0544d0162
    State:                   available
    Vpc Id:                  vpc-0557e8b3073ddddad
  Conditions:
    Last Transition Time:  2021-11-24T09:11:41Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2021-11-24T09:09:38Z
    Status:                True
    Type:                  Synced

```

### DIFF
--- a/pkg/controller/ec2/natgateway/controller.go
+++ b/pkg/controller/ec2/natgateway/controller.go
@@ -135,17 +135,22 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
 		return managed.ExternalCreation{}, errors.New(errUnexpectedObject)
 	}
 
-	nat, err := e.client.CreateNatGateway(ctx, &awsec2.CreateNatGatewayInput{
+	// Create an input without tags.
+	input := &awsec2.CreateNatGatewayInput{
 		ConnectivityType: awsec2types.ConnectivityType(cr.Spec.ForProvider.ConnectivityType),
 		AllocationId:     cr.Spec.ForProvider.AllocationID,
 		SubnetId:         cr.Spec.ForProvider.SubnetID,
-		TagSpecifications: []awsec2types.TagSpecification{
-			{
-				ResourceType: "natgateway",
-				Tags:         v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags),
-			},
-		},
-	})
+	}
+
+	// If we specified tags, update the above input.
+	if cr.Spec.ForProvider.Tags != nil {
+		input.TagSpecifications = []awsec2types.TagSpecification{{
+			ResourceType: "natgateway",
+			Tags:         v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags),
+		}}
+	}
+
+	nat, err := e.client.CreateNatGateway(ctx, input)
 	if err != nil {
 		return managed.ExternalCreation{}, awsclient.Wrap(err, errCreate)
 	}


apis/ec2/v1beta1/natgateway_types.go
```diff
@@ -57,8 +57,7 @@ type NATGatewayParameters struct {
 	SubnetIDSelector *xpv1.Selector `json:"subnetIdSelector,omitempty"`
 
 	// Tags represents to current ec2 tags.
-	// +optional
-	Tags []Tag `json:"tags,omitempty"`
+	Tags []Tag `json:"tags"`
```
negz, line 60: Nit: Should we update the description to explicitly mention that at least one tag appear to be required by the API.

negz (commented):
> other option could be that we inject a name-tag during creation - what you guys think ?

I think I prefer requiring tags, presuming what the API actually wants is "at least one tag, regardless of its key and value". Is that the case, or does the API require a tag with key `name`?

How sure are we that tags are always required? I don't have a ton of faith in the AWS docs, but I do note that they say tags are not required and the given example includes no tags. 🤔 Are they perhaps only required under certain circumstances?

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateNatGateway.html

negz:
P.S. This is a 'breaking change' at the API level because it makes a previously optional field required, but I'm okay with that if the optional field (tags) was actually required in practice - i.e. if omitting tags results in a broken NAT gateway 100% of the time.

haarchri:
@negz  i checked again the whole code and i found the problem is that we use tagSpecification which needs one tag - if we skip it if tags are not set then the process is working

pkg/controller/ec2/natgateway/controller.go
```diff
@@ -135,17 +135,28 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
...
-	})
+		}
+	} else {
+		input = &awsec2.CreateNatGatewayInput{
+			ConnectivityType: awsec2types.ConnectivityType(cr.Spec.ForProvider.ConnectivityType),
+			AllocationId:     cr.Spec.ForProvider.AllocationID,
+			SubnetId:         cr.Spec.ForProvider.SubnetID,
+		}
+	}
+
```
negz, line 158: Style nit: I would prefer:

```go
// Create an input without tags.
input := &awsec2.CreateNatGatewayInput{
	ConnectivityType: awsec2types.ConnectivityType(cr.Spec.ForProvider.ConnectivityType),
	AllocationId:     cr.Spec.ForProvider.AllocationID,
	SubnetId:         cr.Spec.ForProvider.SubnetID,
}

// If we specified tags, update the above input.
if cr.Spec.ForProvider.Tags != nil {
	input.TagSpecifications = []awsec2types.TagSpecification{{
		ResourceType: "natgateway",
		Tags:         v1beta1.GenerateEC2Tags(cr.Spec.ForProvider.Tags),
	}}
}
```

At a glance it seems like this should work.

haarchri: thanks for review - tested again and changed the code - works like expected

negz (approved):
Thanks @haarchri!

PR #623, opened 2021-04-12
Provider config partition

smcavallo:
Signed-off-by: smcavallo <smcavallo@hotmail.com>

### Description of your changes
Expose optional partition on iam and route53 resources.  Can be used in non aws partitions (gov|cn) to override the global region default which only applies to aws partition.

Since [global resources can't be shared across roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html), setting the partition on the providerConfig was preferred over the following:
https://github.com/crossplane/provider-aws/pull/605
https://github.com/crossplane/provider-aws/pull/608

Fixes #
https://github.com/crossplane/provider-aws/issues/596

I have:

- [X] Read and followed Crossplane's [contribution process].
- [X] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested
* Unit tests
* Testing IAMRole in aws partition without providerConfig partition set - success
* Testing IAMRole in aws partition with `aws` providerConfig partition set - success
* Testing IAMRole in aws partition with `aws-test` providerConfig partition set - fail
* Testing IAMRole in aws-us-gov partition without providerConfig partition set - fail
* Testing IAMRole in aws-us-gov partition with `aws` providerConfig partition set - fail
* Testing IAMRole in aws-us-gov partition with `aws-us-gov` providerConfig partition set - success

[contribution process]: https://git.io/fj2m9

PR #900, opened 2021-10-30
add support for per-service endpoints in providerConfig

smcavallo:

See - https://github.com/crossplane/provider-aws/pull/869/files
this PR was a breaking change which broke 2 pieces of critical functionality for us:
* support for IAM and API gateway in aws gov cloud
* FIPS endpoint support 
We are unable to use crossplane without this functionality.
The localstack implementation means that we can set the endpoint for ALL services, but our use case is that we want to be able to set endpoints on a per-service basis.
(See examples with implementation which will work in gov cloud and uses FIPS.)

The functionality introduced here is - if there is a service defined in the providerConfig which matches the service a reconciler is attempting to retrieve an awsConfig for, it will use the defined service endpoint.  Otherwise it will fall back to the default resolver.

We like this approach used in the newer endpoint implementation of being able to set endpoints in the providerConfig as opposed to the individual resource level.

Our set up - which I believe mirrors most - is that an operation/IT group manages crossplane and the provider config and individual teams implement crossplane resources.  
Being able to set endpoints at the providerConfig level means that we can administratively control the endpoints and the various teams only need to worry about creating resources.
This PR provides much more control of all the endpoints and puts that control in the hands of administrators.

Fixes #
https://github.com/crossplane/provider-aws/issues/596

Replaces #
https://github.com/crossplane/provider-aws/pull/605
https://github.com/crossplane/provider-aws/pull/608
https://github.com/crossplane/provider-aws/pull/623

I have:

* Unit tests
* Testing IAMRole in aws-us-gov partition without partition and endpoint - `failure`
* Testing IAMRole in aws-us-gov partition with aws-us-gov partition and iam endpoint set - `success`
* Testing Non-IAMRole in aws-us-gov partition with aws-us-gov partition and iam endpoint set - `success`


muvaf:
It sounds like what we need here is a way for users to override dynamic calculation of the URL specific to the service. At the same time, we already have an option to override all any calculated URLs by using `Static` type. Have you thought of having multiple `ProviderConfig`s for different services? For example, one `ProviderConfig` for IAM where you use `Static` to override the URL, one for others that can use `Dynamic` and separate ones as needed. I'm feeling hesitant to introduce another way of giving a static URL if there is a way we can avoid it.

smcavallo:
@muvaf it makes sense what you're are saying and yes it is possible to achieve with multiple `ProviderConfig`.
In the case of FIPS endpoints it will end up requiring at least 15 `ProviderConfig`, which is a lot to maintain.
Also - that shifts the burden back to the developers as to which `ProviderConfig` they need to use when they construct a resource which we would like to avoid.

muvaf:
@smcavallo Are you using `Composition` and let developers create claims? If so, I think you should be able to put which `ProviderConfig` you'd like that specific managed resource to use in its template. If you'd like them to have a choice, do you think something like the following could help?

```yaml
- type: CombineFromComposite
  combine:
    variables:
    - fromFieldPath: spec.claimRef.namespace
    # You can add more "fromFieldPath"s and construct the name
    # of the targeted ProviderConfig as you'd like.
    strategy: string
    string:
      fmt: "iam-config-%s" # fmt base could differ for each managed resource.
  toFieldPath: spec.providerConfigRef.name
  policy:
    fromFieldPath: Required
```

pkg/clients/aws.go
```diff
@@ -161,7 +179,12 @@ func SetResolver(pc *v1beta1.ProviderConfig, cfg *aws.Config) *aws.Config { // n
 		// SDK setup. However, signing region has to be us-east-1 and it needs
 		// to be set.
 		if region == "aws-global" {
-			e.SigningRegion = "us-east-1"
+			switch StringValue(pc.Spec.Endpoint.PartitionID) {
```
muvaf, line 182: I think we'll need this regardless of whether new type is introduced.

xchan:
@smcavallo do we have any update on this MR ? I believe for aws-cn partition and iamrole for serviceaccount combination, we do need a customized way to make crossplane work.

Furthur according to 
https://github.com/crossplane/provider-aws/pull/623#issuecomment-830040938
sts domain name in aws-cn parition is incorrect and the correct one may be "sts.amazonaws.com"
(See my MR for [terrraform-aws-eks](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/2026/files))

I would like to help to test iam and route in aws-cn partition.

smcavallo:
closing -
I believe this was partially fixed by https://github.com/crossplane-contrib/provider-aws/pull/1329
and the rest was fixed by the aws-sdk-go v2 upgrade in https://github.com/crossplane-contrib/provider-aws/pull/1020

PR #911, opened 2021-11-05
support for s3 replicationConfiguration with delete marker enabled (and fixes)

smcavallo:

See - https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-add-config.html
Adds support for `DeleteMarkerReplication` with status of `Enabled`.  Previously only `Disabled` was supported but `Enabled` is a valid value.

Fixes #
As part of the aws sdk go v2 bucket it appears replicationConfiguration without specifying a filter no longer results in valid XML.  
Previously when a filter was defined as `filter: {}` the provider would send XML similar to 
```
<ReplicationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
	<Rule>
		<ID>1</ID>
		<Priority>0</Priority>
		<Status>Enabled</Status>
		<DeleteMarkerReplication>
			<Status>Enabled</Status>
		</DeleteMarkerReplication>
		<Filter/>
		<Destination>
			<Bucket>arn:aws:s3:::my-s3-bucket</Bucket>
		</Destination>
	</Rule>
	<Role>arn:aws:iam::0123456789:role/my-bucket-role</Role>
</ReplicationConfiguration>
```
However with the new changes the XML sent when the filter is `filter: {}` is this 
(it's missing `<Filter/>` and violates aws XML schema):
```
<ReplicationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
	<Rule>
		<ID>1</ID>
		<Priority>0</Priority>
		<Status>Enabled</Status>
		<DeleteMarkerReplication>
			<Status>Enabled</Status>
		</DeleteMarkerReplication>
		<Destination>
			<Bucket>arn:aws:s3:::my-s3-bucket</Bucket>
		</Destination>
	</Rule>
	<Role>arn:aws:iam::0123456789:role/my-bucket-role</Role>
</ReplicationConfiguration>
```
With this PR, if no filter rules are defined (ex. `filter: {}`) an empty filter and prefix will be sent:
```
<ReplicationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
	<Rule>
		<ID>1</ID>
		<Priority>0</Priority>
		<Status>Enabled</Status>
		<DeleteMarkerReplication>
			<Status>Enabled</Status>
		</DeleteMarkerReplication>
		<Destination>
			<Bucket>arn:aws:s3:::my-s3-bucket</Bucket>
		</Destination>
		<Filter>
			<Prefix/>
		</Filter>
	</Rule>
	<Role>arn:aws:iam::0123456789:role/my-bucket-role</Role>
</ReplicationConfiguration>
```

Also - added `IsUpToDate` function for ReplicationConfiguration, added unit tests, and fixed issue where out-of-order rules caused update of external resource.

I have:

* Unit tests
* Enable aws debug logging with http requests to ensure the `ReplicationConfiguration` XML (specifically filter settings) are sent as part of the payload (above) and that a bucket without any filters defined can be created successfully.
* Create a bucket with replication disabled or enabled.  Bucket is created with the correct replication settings
* Toggle an existing bucket with replication configuration to enabled/disabled.  The change is detected during `Observe` and aws is updated accordingly
* Ensure that bucket with replicationConfiguration does not constantly reconcile with these settings.

### DIFF
--- a/apis/s3/v1beta1/replicationConfiguration_types.go
+++ b/apis/s3/v1beta1/replicationConfiguration_types.go
@@ -123,9 +123,8 @@ type ReplicationRule struct {
 // see Backward Compatibility (https://docs.aws.amazon.com/AmazonS3/latest/dev/replication-add-config.html#replication-backward-compat-considerations).
 type DeleteMarkerReplication struct {
 	// Indicates whether to replicate delete markers.
-	// In the current implementation, Amazon S3 doesn't replicate the delete markers.
-	// The status must be "Disabled".
-	// +kubebuilder:validation:Enum=Disabled
+	// Valid values are "Enabled" or "Disabled"
+	// +kubebuilder:validation:Enum=Enabled;Disabled
 	Status string `json:"status"`
 }
 
--- a/pkg/controller/s3/bucket/replicationConfig_test.go
+++ b/pkg/controller/s3/bucket/replicationConfig_test.go
@@ -459,3 +459,455 @@ func TestReplicationLateInit(t *testing.T) {
 		})
 	}
 }
+
+func TestIsUpToDate(t *testing.T) {
+	type args struct {
+		cr *v1beta1.ReplicationConfiguration
+		b  *s3types.ReplicationConfiguration
+	}
+
+	type want struct {
+		isUpToDate ResourceStatus
+		err        error
+	}
+	cases := map[string]struct {
+		args
+		want
+	}{
+		"IsUpToDate": {
+			args: args{
+				cr: &v1beta1.ReplicationConfiguration{
+					Role: &role,
+					Rules: []v1beta1.ReplicationRule{{
+						DeleteMarkerReplication: &v1beta1.DeleteMarkerReplication{Status: enabled},
+						Destination: v1beta1.Destination{
+							AccessControlTranslation: &v1beta1.AccessControlTranslation{Owner: owner},
+							Account:                  &accountID,
+							Bucket:                   &bucketName,
+							EncryptionConfiguration:  &v1beta1.EncryptionConfiguration{ReplicaKmsKeyID: &kmsID},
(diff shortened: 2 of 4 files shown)


apis/s3/v1beta1/replicationConfiguration_types.go
```diff
@@ -124,8 +124,8 @@ type ReplicationRule struct {
 type DeleteMarkerReplication struct {
 	// Indicates whether to replicate delete markers.
 	// In the current implementation, Amazon S3 doesn't replicate the delete markers.
```
negz, line 126: Is this comment still valid?

pkg/controller/s3/bucket/replicationConfig_test.go
```diff
@@ -459,3 +460,453 @@ func TestReplicationLateInit(t *testing.T) {
...
+			},
+			want: want{
+				isUpToDate: 1,
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
```
negz, line 902: Would you mind updating this to remove gomega? Should be a small diff - same approach as https://github.com/crossplane/provider-aws/pull/963/commits/aaaa66b9725b523d2bc779119ec3885ce030de2e

negz (commented):
This looks good to me, apart from my two comments. Please add a little info on how you tested this and I'll merge and backport.

apis/s3/v1beta1/replicationConfiguration_types.go
smcavallo: no - it is not valid.  This setting can be used to successfully enable the replication of delete markers when set to `Enabled`

negz (approved):
Thanks for all these fixes!

Issue #597, opened 2021-03-15
Allow assuming multiple web identity roles


hasheddan:
### What problem are you facing?

A number of `provider-aws` users make use of the [IRSA](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/) authentication [flow](https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md) to allow the provider to authenticate to AWS. This is a useful pattern, but restricts the provider `Pod` to only assuming one role because the [EKS Pod Identity Webhook](https://github.com/aws/amazon-eks-pod-identity-webhook) injects this role based on the `Pod` `ServiceAccount`, and [hardcodes the environment variables](https://github.com/aws/amazon-eks-pod-identity-webhook/blob/599cfe7cfc8c72d48a8870e675e19a8169b3dafe/pkg/handler/handler.go#L226) that point to the volume mount, and we [hardcode](https://github.com/crossplane/provider-aws/blob/8767b03cf54e916dbd9b6ff81b259bbeee20d41d/pkg/clients/aws.go#L236) accessing them.

### How could Crossplane help solve your problem?

Users _can_ choose to [manually specify](https://github.com/aws/amazon-eks-pod-identity-webhook/issues/32#issuecomment-588113740) environment variables with multiple ARNs and specific mount paths. In the linked example, two different containers are making use of the same SA token mount, but using different ARNs. Crossplane users may define additional environment variables on their provider containers using the [`ControllerConfig`](https://doc.crds.dev/github.com/crossplane/crossplane/pkg.crossplane.io/ControllerConfig/v1alpha1@v1.1.0). This means that if we added a field to the `provider-aws` `ProviderConfig` that allowed for the `InjectedIdentity` source to be customized with something like `overrideARNEnvVar`, users could have a single provider `Pod` assume multiple roles using a single SA.

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: example
spec:
  credentials:
    source: InjectedIdentity
    overrideARNEnvVar: AWS_ROLE_ARN_2
```

lewismarshall:
As an alternative to creating multiple web identities to associate with the controller process POD, a new "assume role" capability could be added, see #606.

This would present the advantage that new roles can be created as new resources are enabled without any re-deploy or update to the provider controller pod.

E.g. (repeated here for clarity) a provider config with specific roles:
```
[...]

PR #912, opened 2021-11-06
assumeRoleARN for ProviderConfig

### LINKED ISSUE #606: Enable specific role assumption in ProviderConfig

I would like to use specific and least privileged access roles for provisioning AWS resources without creating additional credentials wherever possible.

Adding a field e.g. `assumeRoleARN` to the `ProviderConfig` would allow AWS STS role assumption (after initial auth) for any specific AWS resource provisioning in conjunction with `providerConfigRef` on the associated resource instances.

This would work when either the `InjectedIdentity` ([IRSA](https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/)) option is used OR with credentials and would allow all permissions to be granted to only the specific roles for specific provider resources.

As only a single (IRSA) service account or credential would be required, extra permissions can be added as and when an organisation starts using new resources and crossplane would not need redeploying to use them (just deployment of an additional ProviderConfig).

This could be achieved by simply adding a new field `assumeRoleARN` in the ProviderConfig spec e.g.:

```
apiVersion: aws.crossplane.io/v1beta1
  kind: ProviderConfig
  metadata:
    name: my-specific-resource
  spec:
    credentials:
      source: InjectedIdentity
      assumeRoleARN: "arn:aws:iam::xyz:role/my-specific-resource" 
```
This provides an alternative to associating multiple IRSA AWS roles to a single pod with multiple environment variables as described in this different approach: #597.

haarchri:

added option assumeRoleARN for ProviderConfig #912

1.) In the absence of `assumeRoleARN`, we simply call `config.LoadDefaultConfig(ctx)`, as this will automatically grant us access to the `current` AWS account / configured `WebIdentity` / `Credentials`

2.) `assumeRoleARN` allows us to assume a role via `sts:AssumeRole` via our configured `WebIdentity` / `Credentials`
- default role is assumed by `config.LoadDefaultConfig(ctx)` - `WebIdentity` /  `Credentials`
- `sts:AssumeRole` is done by `stscreds.NewAssumeRoleProvider(assumeRoleARN)`

Fixes #606 
Implemented Parts of: #597 

I have:

for go-sdk-v1 ressources (code-generator)
for go-sdk-v2 ressources (manual implemented ressources)

create providers:
local:
```
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: aws-provider
spec:
  credentials:
    source: InjectedIdentity
```

or

```
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: aws-provider
spec:
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: creds
```

remote:
```
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: aws-provider-999999999999
spec:
  assumeRoleARN: "arn:aws:iam::999999999999:role/crossplane_deploy"
  credentials:
    source: InjectedIdentity
```

or

```
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: aws-provider-999999999999
spec:
  assumeRoleARN: "arn:aws:iam::999999999999:role/crossplane_deploy"
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: creds
```

efs:
local:
```
# aws go-sdk v1
apiVersion: efs.aws.crossplane.io/v1alpha1
kind: FileSystem
metadata:
  name: local-efs
spec:
  forProvider:
    region: eu-central-1
  providerConfigRef:
    name: aws-provider
```

remote:
```
# aws go-sdk v1
apiVersion: efs.aws.crossplane.io/v1alpha1
kind: FileSystem
metadata:
  name: remote-efs
spec:
  forProvider:
    region: eu-central-1
  providerConfigRef:
    name: aws-provider-999999999999
```

iam-role:
local:
```
# aws go-sdk v2
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRole
metadata:
  name: test-local-role
spec:
  forProvider:
    assumeRolePolicyDocument: |
      {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Principal": {
                    "Service": [
                        "ec2.amazonaws.com",
                        "eks.amazonaws.com",
                        "eks-fargate-pods.amazonaws.com",
                        "lambda.amazonaws.com"
                    ]
                },
                "Action": [
                    "sts:AssumeRole"
                ]
            }
        ]
      }
    tags:
      - key: k1
        value: v1
  providerConfigRef:
    name: aws-provider
```

remote:
```
# aws go-sdk v2
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRole
metadata:
  name: test-remote-role
spec:
  forProvider:
    assumeRolePolicyDocument: |
      {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Principal": {
                    "Service": [
                        "ec2.amazonaws.com",
                        "eks.amazonaws.com",
                        "eks-fargate-pods.amazonaws.com",
                        "lambda.amazonaws.com"
                    ]
                },
                "Action": [
                    "sts:AssumeRole"
                ]
            }
        ]
      }
    tags:
      - key: k1
        value: v1
  providerConfigRef:
    name: aws-provider-999999999999
```

### DIFF
--- a/apis/v1beta1/providerconfig_types.go
+++ b/apis/v1beta1/providerconfig_types.go
@@ -27,6 +27,10 @@ type ProviderConfigSpec struct {
 	// Credentials required to authenticate to this provider.
 	Credentials ProviderCredentials `json:"credentials"`
 
+	// AssumeRoleARN to assume with provider credentials
+	// +optional
+	AssumeRoleARN *string `json:"assumeRoleARN,omitempty"`
+
 	// Endpoint is where you can override the default endpoint configuration
 	// of AWS calls made by the provider.
 	// +optional
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -28,7 +28,9 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
+	stscreds "github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	ec2type "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
 	awsv1 "github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	credentialsv1 "github.com/aws/aws-sdk-go/aws/credentials"
@@ -92,7 +94,7 @@ func GetConfig(ctx context.Context, c client.Client, mg resource.Managed, region
 }
 
 // UseProviderConfig to produce a config that can be used to authenticate to AWS.
-func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed, region string) (*aws.Config, error) {
+func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed, region string) (*aws.Config, error) { // nolint:gocyclo
 	pc := &v1beta1.ProviderConfig{}
 	if err := c.Get(ctx, types.NamespacedName{Name: mg.GetProviderConfigReference().Name}, pc); err != nil {
 		return nil, errors.Wrap(err, "cannot get referenced Provider")
@@ -105,6 +107,13 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 
 	switch s := pc.Spec.Credentials.Source; s { //nolint:exhaustive
 	case xpv1.CredentialsSourceInjectedIdentity:
+		if pc.Spec.AssumeRoleARN != nil {
+			cfg, err := UsePodServiceAccountAssumeRole(ctx, []byte{}, DefaultSection, region, pc)
+			if err != nil {
+				return nil, err
+			}
+			return SetResolver(pc, cfg), nil
+		}
 		cfg, err := UsePodServiceAccount(ctx, []byte{}, DefaultSection, region)
 		if err != nil {
 			return nil, err
@@ -114,6 +123,13 @@ func UseProviderConfig(ctx context.Context, c client.Client, mg resource.Managed
 		data, err := resource.CommonCredentialExtractor(ctx, s, c, pc.Spec.Credentials.CommonCredentialSelectors)
 		if err != nil {
 			return nil, errors.Wrap(err, "cannot get credentials")
+		}
+		if pc.Spec.AssumeRoleARN != nil {
+			cfg, err := UseProviderSecretAssumeRole(ctx, data, DefaultSection, region, pc)
+			if err != nil {
+				return nil, err
+			}
+			return SetResolver(pc, cfg), nil
 		}
 		cfg, err := UseProviderSecret(ctx, data, DefaultSection, region)
 		if err != nil {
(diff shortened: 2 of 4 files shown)


lewismarshall:
Looks exactly right from my perspective 👍 .

https://github.com/crossplane/provider-aws/issues/606#issuecomment-962454276

pkg/clients/aws.go
```diff
@@ -80,14 +80,10 @@ const (
...
-	switch {
-	case mg.GetProviderConfigReference() != nil:
+	if mg.GetProviderConfigReference() != nil {
 		return UseProviderConfig(ctx, c, mg, region)
-	case mg.GetProviderReference() != nil:
-		return UseProvider(ctx, c, mg, region)
-	default:
-		return nil, errors.New("neither providerConfigRef nor providerRef is given")
 	}
+	return nil, errors.New("neither providerConfigRef is given")
```
hasheddan, line 86: @haarchri if we drop the `Provider` type then we likely should be removing `providerRef` as well, which is a `crossplane-runtime` change. Have there been any discussions / consensus reached on doing so?

cc @muvaf @negz

haarchri: This function was deprecated so i decided to do so - tell me i we want to drop

mmiranda:
It might be a good idea to update the docs here as well: https://github.com/crossplane/provider-aws/blob/master/AUTHENTICATION.md?plain=1#L70-L77

pkg/clients/aws.go
hasheddan: @haarchri we may be getting close to the point that we want to drop it, but that is technically a breaking change and we may need to bump API versions because of that -- it might be worth decoupling that change from this PR so that your progress is not impeded by that discussion 👍🏻

haarchri: decoupling done - i will start a e2e test in our environment and will ping you if we are ready

ddanijel:
Any progress with this one? This is a crucial component for using Crossplane at scale on AWS.

haarchri:
@hasheddan  you can start review - we tested after cleanup again - everything is working as expected to ressources with go-sdk-v1 and go-sdk-v2

apis/v1beta1/providerconfig_types.go
```diff
@@ -39,6 +39,11 @@ type ProviderCredentials struct {
 	// +kubebuilder:validation:Enum=None;Secret;InjectedIdentity;Environment;Filesystem
 	Source xpv1.CredentialsSource `json:"source"`
 
+	// AssumeRoleARN to assume with provider credentials
+	// +immutable
+	// +optional
+	AssumeRoleARN string `json:"assumeRoleARN,omitempty"`
```
muvaf, line 45: It seems like this option is a sub-configuration of `InjectedIdentity` option. Would it make sense to have a `InjectedIdentity *InjectedIdentity` field, similar to `SecretRef *SecretKeySelector` and have this option under that struct?

pkg/clients/aws.go
```diff
@@ -264,7 +266,31 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
...
+			),
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to load assumed role AWS config")
+		}
+		return &cnf, err
+	}
+
+	// serviceAccountRole
 	cfg, err := config.LoadDefaultConfig(
```
muvaf, line 294: Can we re-use the `cfg` initialized here in both cases?

pkg/clients/aws.go
```diff
@@ -264,7 +266,31 @@ func UseProviderSecret(ctx context.Context, data []byte, profile, region string)
 
 // UsePodServiceAccount assumes an IAM role configured via a ServiceAccount.
 // https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
-func UsePodServiceAccount(ctx context.Context, _ []byte, _, region string) (*aws.Config, error) {
+func UsePodServiceAccount(ctx context.Context, _ []byte, rolearn string, _, region string) (*aws.Config, error) {
+	// assumeRole
+	if rolearn != "" {
```
muvaf, line 271: Having `if rolearn == ""` and returning early feels more in-line with our [conventions](https://github.com/crossplane/crossplane/blob/master/CONTRIBUTING.md#return-early).

pkg/clients/aws.go
```diff
@@ -348,6 +374,38 @@ func UseProviderSecretV1(_ context.Context, data []byte, pc *v1beta1.ProviderCon
 // UsePodServiceAccountV1 assumes an IAM role configured via a ServiceAccount.
 // https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
 func UsePodServiceAccountV1(ctx context.Context, _ []byte, pc *v1beta1.ProviderConfig, _, region string) (*awsv1.Config, error) {
+	// assumeRole
+	if pc.Spec.Credentials.AssumeRoleARN != "" {
+		cfg, err := config.LoadDefaultConfig(ctx)
```
muvaf, line 379: Similar comments for this block as well.

muvaf (commented):
Thanks @haarchri ! I'd like @hasheddan to take a look before merging.

apis/v1beta1/providerconfig_types.go
haarchri: currently yes - but in future we can also use AssumeRoleARN for more - what do you think ?

muvaf: Do you mean another top level option? Right now, it is programatically enabled only if you use `InjectedIdentity`, so I'd say that it could be a different field in under those other top level options if they decide to need it.

hasheddan: @haarchri when you say it can be used for more, I assume you are referring to the fact that you can use the other credential sources to also assume a role -- is that correct?

haarchri: yes

hasheddan: @haarchri @muvaf with that being the case, I almost wonder if it should live outside of the `credentials` struct, similar to how `provider-gcp` has `projectID` (ref https://doc.crds.dev/github.com/crossplane/provider-gcp/gcp.crossplane.io/ProviderConfig/v1beta1@v0.19.0#spec-projectID). A helpful categorization might be:
- `credentials` is how you actually **acquire** the base credentials
- anything outside of that in `spec` is how you **use** the credentials you acquired

What do y'all think about that?

haarchri: for me fine - i can change the pr for it

haarchri: @hasheddan checked & changed all suggestions ;)  - now i will start testing a few cases after chage

apis/v1beta1/providerconfig_types.go
```diff
@@ -27,6 +27,12 @@ type ProviderConfigSpec struct {
 	// Credentials required to authenticate to this provider.
 	Credentials ProviderCredentials `json:"credentials"`
 
+	// AssumeRoleARN to assume with provider credentials
+	// can be used together with InjectedIdentity
```
muvaf, line 31: Now that you can assume role with all types of creds, this comment needs to be updated.

muvaf (commented):
For some reason I can't comment on the conversation that @hasheddan added to. I think it's a great idea. @haarchri is it possible today make the code actually use that `AssumeRoleARN` field no matter what credential type you use if it's given?

apis/v1beta1/providerconfig_types.go
muvaf: I submitted my comment and now I can 😒 
> For some reason I can't comment on the conversation that @hasheddan added to. I think it's a great idea. @haarchri is it possible today make the code actually use that AssumeRoleARN field no matter what credential type you use if it's given?

haarchri:
@muvaf  @hasheddan  tested the implementation with
`source: InjectedIdentity` and  ` source: Secret`
examples updates in description of this PR

muvaf (approved):
Thanks @haarchri ! Great to see this land!

hasheddan (approved):
Really nice work here @haarchri! 🙌🏻

ddanijel:
Great job @haarchri

hanlins:
Nice work! Btw, is it possible to include `ExternalID` as a follow-up improvement?

Issue #784, opened 2021-07-29
Adopted 'Bucket' MR constantly updates when missing 'paymentConfiguration' field


nrhtr:
### What happened?

Hi,

I have an orphaned S3 bucket that I would like Crossplane to 'adopt' when reinstalling the managed resource. We're using a Composition but I can reproduce with a minimal example.

The 'paymentConfiguration' field on the bucket is left null / unspecified.

The resource never becomes 'ready' because the controller is constantly updating the 'paymentConfiguration' on the bucket. 

We can work around this by defaulting that field in our Composition, but wanted to raise this assuming it needs to be fixed in the provider.

### How can we reproduce it?

* Have an 'orphan' S3 bucket
* Create a managed resource to 'adopt' it back into Crossplane
* Observe it never reaching 'ready'

Example:
```shell
~ cat bucket.yaml
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: acme-my-test-bucket
spec:
  deletionPolicy: Orphan
  forProvider:
    acl: private
    locationConstraint: ap-southeast-2
    publicAccessBlockConfiguration:
      blockPublicAcls: true
      blockPublicPolicy: true
      ignorePublicAcls: true
      restrictPublicBuckets: true
    serverSideEncryptionConfiguration:
      rules:
      - applyServerSideEncryptionByDefault:
          sseAlgorithm: aws:kms
  providerConfigRef:
    name: default
~ kubectl apply -f bucket.yaml
bucket.s3.aws.crossplane.io/acme-my-test-bucket created
~ kubectl get bucket
NAME                                      READY   SYNCED   AGE
acme-my-test-bucket                       True    True     5s
~ kubectl delete -f bucket.yaml
bucket.s3.aws.crossplane.io "acme-my-test-bucket" deleted
~ kubectl get bucket
<bucket is deleted, only other buckets exist>
~ kubectl apply -f bucket.yaml
bucket.s3.aws.crossplane.io/acme-my-test-bucket created
~ kubectl get bucket
NAME                                      READY   SYNCED   AGE
acme-my-test-bucket                               True     4s
~ kubectl get bucket
NAME                                      READY   SYNCED   AGE
acme-my-test-bucket                               True     67s
```

I put a debug print into the provider and found that the cause appeared to be the bucket "payment configuration".

[...]

PR #916, opened 2021-11-09
fix s3 nil paymentConfiguration preventing bucket from being ready

smcavallo:

Fixes https://github.com/crossplane/provider-aws/issues/784

In aws-sdk-go-v2 the `Payer` property for `GetBucketRequestPaymentOutput` has changed.
Previously it was a string:
https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#GetBucketRequestPaymentOutput
But in v2 it is a type:
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#GetBucketRequestPaymentOutput
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3@v1.18.0/types#Payer

Perhaps it was set via `LateInitialize` previously?   Either way, the comparison functionality is broken now.
What is currently happening is if `PayerConfiguration` is not set it will ALWAYS need an update due to this code block:
```
	case config == nil && len(external.Payer) != 0:
		return NeedsUpdate, nil
```
This fixes the comparison functionality, which should allow bucket status to be ready
 

Fixes #

I have:

### DIFF
--- a/pkg/controller/s3/bucket/requestPaymentConfig.go
+++ b/pkg/controller/s3/bucket/requestPaymentConfig.go
@@ -45,22 +45,21 @@ func NewRequestPaymentConfigurationClient(client s3.BucketClient) *RequestPaymen
 
 // Observe checks if the resource exists and if it matches the local configuration
 func (in *RequestPaymentConfigurationClient) Observe(ctx context.Context, bucket *v1beta1.Bucket) (ResourceStatus, error) {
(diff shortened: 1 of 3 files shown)


haarchri:
@smcavallo thanks for your work - does your MR also solves https://github.com/crossplane/provider-aws/issues/784 ?

negz (approved):
Thank you @smcavallo!

Issue #896, opened 2021-10-28
S3 bucket would not reach Available condition when set notificationConfiguration


haarchri:
### What happened?

I created an `buckets.s3.aws.crossplane.io` object  but the CR never reached available condition and the controller was constantly updating the bucket.

```
kubectl get bucket
NAME                         READY   SYNCED   AGE
255932642927-lambda-bucket   False   True     17m
```

```
  Normal   CreatedExternalResource          7m27s                  managed/bucket.s3.aws.crossplane.io  Successfully requested creation of external resource
  Normal   UpdatedExternalResource          5m14s (x4 over 7m26s)  managed/bucket.s3.aws.crossplane.io  Successfully requested update of external resource
  Normal   UpdatedExternalResource          74s (x4 over 4m18s)    managed/bucket.s3.aws.crossplane.io  Successfully requested update of external resource
  Normal   UpdatedExternalResource          21s                    managed/bucket.s3.aws.crossplane.io  Successfully requested update of external resource
  Normal   UpdatedExternalResource          5s (x9 over 7m20s)     managed/bucket.s3.aws.crossplane.io  Successfully requested update of external resource
```

### How can we reproduce it?
Create an s3 bucket
Create an buckets.s3.aws.crossplane.io object to adopt the s3 bucket, example:
* full example available here: https://gist.github.com/haarchri/95f72bd6279f42b821082b74d2b149d1
```
---
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: 255932642927-lambda-bucket
spec:
  forProvider:
    acl: private
    locationConstraint: eu-central-1
    accelerateConfiguration:
      status: Enabled
    versioningConfiguration:
      status: Enabled
    notificationConfiguration:
      lambdaFunctionConfigurations:
        - events: ["s3:ObjectCreated:*"]
          lambdaFunctionArn: arn:aws:lambda:eu-central-1:255932642927:function:lambda
    paymentConfiguration:
      payer: BucketOwner
    tagging:
      tagSet:
        - key: s3-bucket
          value: lambda-bucket
    objectLockEnabledForBucket: false
  providerConfigRef:
    name: default
```

every run the name-id is changed in aws
via debug we can see the following:

```
        NewVersioningConfigurationClient(client),  #### no update needed
        NewAccelerateConfigurationClient(client), #### no update needed
        NewCORSConfigurationClient(client), #### no update needed
[...]

PR #917, opened 2021-11-09
fix s3 notificationConfiguration

smcavallo:

fixes s3 notificationConfiguration continually being updated.

Updates `IsUpToDate` comparisons for `notificationConfiguration` - most notably ignore the `Id` and ignore ordering when there are multiple configurations.

Adds additional unit tests.

Note: this excludes the `Id` properties during the comparison.  This *should* be fine because the actual rules are being compared. 
Per the docs https://docs.aws.amazon.com/cli/latest/reference/s3api/put-bucket-notification-configuration.html

> Id -> (string)

    An optional unique identifier for configurations in a notification configuration.  If you don't provide one, Amazon S3 will assign an ID.
 Ignoring the `Id` in the comparison is workaround since these properties are not `LateInitialized`

Fixes https://github.com/crossplane/provider-aws/issues/896

I have:

* Unit testing
* Reconcile a bucket and ensure the `ID` in the console has not changed
* force an update to a bucket and ensure the `ID` in the console has not changed
* Debug logging in the sdk to watch API calls

### DIFF
--- a/pkg/controller/s3/bucket/notificationConfig.go
+++ b/pkg/controller/s3/bucket/notificationConfig.go
@@ -18,10 +18,12 @@ package bucket
 
 import (
 	"context"
+	"sort"
 
 	"github.com/aws/smithy-go/document"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
+	"github.com/aws/aws-sdk-go-v2/aws"
 	awss3 "github.com/aws/aws-sdk-go-v2/service/s3"
 	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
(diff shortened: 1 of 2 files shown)


haarchri:
@smcavallo can you squash commits for Clean history ? Thanks

pkg/controller/s3/bucket/notificationConfig.go
```diff
@@ -54,24 +56,79 @@ func (in *NotificationConfigurationClient) Observe(ctx context.Context, bucket *
...
 }
 
+func sortLambda(configs []types.LambdaFunctionConfiguration) {
+	sort.Slice(configs, func(i, j int) bool {
+		if a, b := configs[i].LambdaFunctionArn, configs[j].LambdaFunctionArn; a != b {
+			return aws.ToString(a) < aws.ToString(b)
+		}
+		return true
+	})
+}
```
negz, line 112: This is totally fine as is, but for future reference you could also pass in a couple of`cmpopts.SortSlices` options, each with a less function for a slice of the relevant type of struct.

pkg/controller/s3/bucket/notificationConfig_test.go
```diff
@@ -378,3 +379,235 @@ func TestNotifLateInit(t *testing.T) {
...
+				isUpToDate: 2,
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+			actual, err := IsNotificationConfigurationUpToDate(tc.args.cr, tc.args.b)
+			g.Expect(err).NotTo(HaveOccurred())
```
negz, line 606: We try to avoid Gomega usage in our tests. I want to include this with the v0.21.1 release today so I plan to merge and quickly follow up to remove this.

negz (approved):
Thanks again @smcavallo!

PR #930, opened 2021-11-11
observe iampolicy which already exists

smcavallo:

When `provider-aws` attempts to observe an iam policy which already exists, it fails with this error:
```
{
  "object": {
    "kind": "IAMPolicy",
    "name": "crossplane-lateinitialize-test",
    "uid": "some-uid",
    "apiVersion": "identity.aws.crossplane.io/v1alpha1",
    "resourceVersion": "49576502"
  },
  "reason": "CannotCreateExternalResource",
  "message": "failed to create the IAM Policy: EntityAlreadyExists: A policy called my-iam-test-policy already exists. Duplicate names are not allowed."
}
```
Since the external name is not set `Observe` skips lookup of the resource and it goes to `Create`.  For other types of resources (ex. iam role) even when the external name is not set crossplane is able to detect and take ownership of the resource.   The reason being - those aws API calls are based on resource _NAME_ so it can look up the existing aws resource by name without the external resource being set.  
That is not true for iam policy where the external name is an ARN.

With this PR, if the external-name is not set, `Observe` will attempt to construct the ARN from the result of https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html
If found it will update the external-name and move on.  If not found, it will continue with the existing behavior which is to try to create.

IAM Policy has been a painpoint during `provider-aws` upgrades - somehow the external-name is being reset and it takes a lot of manually editing resources to fix.   This should solve that problem.

Fixes #

I have:

* Unit tests
* attempt to create a crossplane object where the policy name already exists
* remove the external-name property on a policy (and watch it get re-added)

### DIFF
--- a/pkg/controller/identity/iampolicy/controller.go
+++ b/pkg/controller/identity/iampolicy/controller.go
@@ -99,7 +99,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}
 
 	if meta.GetExternalName(cr) == "" {
-		return managed.ExternalObservation{}, nil
+		// If external name not set there is still a change it may already exist
+		// Try to get the policy by name
+		policyArn, _ := e.getPolicyArnByName(ctx, cr.Spec.ForProvider.Name)
+		if policyArn == nil {
+			return managed.ExternalObservation{}, nil
+		}
+		meta.SetExternalName(cr, aws.ToString(policyArn))
+		_ = e.kube.Update(ctx, cr)
 	}
 
 	policyResp, err := e.client.GetPolicy(ctx, &awsiam.GetPolicyInput{
@@ -274,4 +281,45 @@ func (e *external) deleteNonDefaultVersions(ctx context.Context, policyArn strin
 	}
 
 	return nil
+}
+
+// ListPoliciesPages paginates ListPolicies
+func (e *external) ListPoliciesPages(ctx context.Context, input *awsiam.ListPoliciesInput, fn func(*awsiam.ListPoliciesOutput, bool) bool) error {
+	for {
+		output, err := e.client.ListPolicies(ctx, input)
+		if err != nil {
+			return err
+		}
+		lastPage := awsclient.StringValue(output.Marker) == ""
+		if !fn(output, lastPage) || lastPage {
+			break
+		}
+		input.Marker = output.Marker
+	}
+	return nil
+}
+
+// getPolicyArnByName will paginate through all policies to retrieve by name instead of arn
+func (e *external) getPolicyArnByName(ctx context.Context, policyName string) (*string, error) {
+
+	input := &awsiam.ListPoliciesInput{Scope: awsiamtypes.PolicyScopeTypeLocal}
+
+	policyArn := awsclient.String("")
+	err := e.ListPoliciesPages(ctx, input, func(page *awsiam.ListPoliciesOutput, lastPage bool) bool {
+		if page == nil {
+			return !lastPage
+		}
+		// Set the ARN and return if found
+		for _, policy := range page.Policies {
+			if policyName == awsclient.StringValue(policy.PolicyName) {
+				policyArn = policy.Arn
+				return lastPage
+			}
+		}
+		return !lastPage
+	})
+	if err != nil {
+		return nil, err
+	}
+	return policyArn, nil
 }
--- a/pkg/controller/identity/iampolicy/controller.go
+++ b/pkg/controller/identity/iampolicy/controller.go
@@ -21,8 +21,10 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	awsiam "github.com/aws/aws-sdk-go-v2/service/iam"
 	awsiamtypes "github.com/aws/aws-sdk-go-v2/service/iam/types"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/pkg/errors"
 	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -66,7 +68,7 @@ func SetupIAMPolicy(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter
 		For(&v1alpha1.IAMPolicy{}).
 		Complete(managed.NewReconciler(mgr,
 			resource.ManagedKind(v1alpha1.IAMPolicyGroupVersionKind),
-			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewPolicyClient}),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewPolicyClient, newSTSClientFn: iam.NewSTSClient}),
 			managed.WithInitializers(),
 			managed.WithConnectionPublishers(),
 			managed.WithPollInterval(poll),
@@ -75,8 +77,9 @@ func SetupIAMPolicy(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter
 }
 
 type connector struct {
-	kube        client.Client
-	newClientFn func(config aws.Config) iam.PolicyClient
+	kube           client.Client
+	newClientFn    func(config aws.Config) iam.PolicyClient
+	newSTSClientFn func(config aws.Config) iam.STSClient
 }
 
 func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.ExternalClient, error) {
@@ -84,11 +87,12 @@ func (c *connector) Connect(ctx context.Context, mg resource.Managed) (managed.E
 	if err != nil {
 		return nil, err
 	}
-	return &external{client: c.newClientFn(*cfg), kube: c.kube}, nil
+	return &external{client: c.newClientFn(*cfg), sts: c.newSTSClientFn(*cfg), kube: c.kube}, nil
 }
 
 type external struct {
 	client iam.PolicyClient
+	sts    iam.STSClient
 	kube   client.Client
 }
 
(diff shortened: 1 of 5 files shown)


pkg/controller/identity/iampolicy/controller.go
```diff
@@ -275,3 +282,44 @@ func (e *external) deleteNonDefaultVersions(ctx context.Context, policyArn strin
...
+	}
+	return nil
+}
+
+// getPolicyArnByName will paginate through all policies to retrieve by name instead of arn
+func (e *external) getPolicyArnByName(ctx context.Context, policyName string) (*string, error) {
+
+	input := &awsiam.ListPoliciesInput{Scope: awsiamtypes.PolicyScopeTypeLocal}
+
+	policyArn := awsclient.String("")
```
chlunde, line 307: This is confusing to me, and it's compared against nil later. Is it simpler with this?
```suggestion
	var policyArn *string
```

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -275,3 +282,44 @@ func (e *external) deleteNonDefaultVersions(ctx context.Context, policyArn strin
...
+		}
+		input.Marker = output.Marker
+	}
+	return nil
+}
+
+// getPolicyArnByName will paginate through all policies to retrieve by name instead of arn
+func (e *external) getPolicyArnByName(ctx context.Context, policyName string) (*string, error) {
+
+	input := &awsiam.ListPoliciesInput{Scope: awsiamtypes.PolicyScopeTypeLocal}
```
chlunde, line 305: add path here, ref. performance? I think the path is immutable, so I'm not sure what's best to do if we have another policy with he same name and a different path, but we should at least think about the behavior if you create

* two policies in with different contents and same name/path
* two policies with different paths and same name

I think at least the second case might be confusing with this change?

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -99,7 +99,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}
 
 	if meta.GetExternalName(cr) == "" {
-		return managed.ExternalObservation{}, nil
+		// If external name not set there is still a change it may already exist
+		// Try to get the policy by name
+		policyArn, _ := e.getPolicyArnByName(ctx, cr.Spec.ForProvider.Name)
```
chlunde, line 104: If there are thousands of policies, this might be a bit expensive? It might be cheaper to do `arn:aws:iam::+AccountID+:policy+Path+.Name`, but maybe it's difficult because we don't have easy access to the account number?

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -99,7 +99,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+		if policyArn == nil {
+			return managed.ExternalObservation{}, nil
+		}
+		meta.SetExternalName(cr, aws.ToString(policyArn))
+		_ = e.kube.Update(ctx, cr)
```
chlunde, line 109: Often it causes issues if you do `kube.Update` inside a function like this, because the provider runtime also does it but with a different version number. Did you notice any errors in the log when you tested this?

pkg/controller/identity/iampolicy/controller.go
smcavallo: @chlunde - all good points.  I am going to see if [`GetCallerIdentity`](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html) can be used instead.  If that call is able to return enough values to construct the ARN then the ARN should be able to used to make one single api call to see if the policy exists.

pkg/controller/identity/iampolicy/controller.go
smcavallo: I did not see any errors in the logs

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -275,3 +294,22 @@ func (e *external) deleteNonDefaultVersions(ctx context.Context, policyArn strin
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Use it to construct an arn for the policy
+	policyArn := arn.ARN{Partition: identityArn.Partition,
+		Service:   "iam",
+		Region:    identityArn.Region,
+		AccountID: identityArn.AccountID,
+		Resource:  "policy/" + policyName}
```
chlunde, line 312: I think we need to handle `role.Spec.Path` as `nil` or `""` (this case, which means `"/"`), `"/"` or `"/foo/bar/"`.

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -99,7 +103,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	}
 
 	if meta.GetExternalName(cr) == "" {
-		return managed.ExternalObservation{}, nil
+		// If external name not set there is still a change it may already exist
+		// Try to get the policy by name
```
chlunde, line 107: any reason for not returning, or at least logging, unexpected errors here and in 
`kube.Update`?

smcavallo: the main reason is - if this fails it will silently fall back to the previous behavior.

`kube.Update` - IMO it would be better to not fail if kube can't be updated - and I *think* there is already logging included within that (???)

Thinking about this from the end user perspective - what you said makes sense - logging/returning a better error makes sense.  I added a new exception message specific to the external-name lookup failures.
What end users should know when they see an exception in that block is - there is something wrong with the external-name field which they have to correct. (and if the external name field is correct then there is something they need to correct as to why to why their policy can't be observed)

chlunde (approved):
Tested OK

pkg/controller/identity/iampolicy/controller.go
```diff
@@ -99,7 +104,14 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+		policyArn, policyErr := e.getPolicyArnByNameAndPath(ctx, cr.Spec.ForProvider.Name, cr.Spec.ForProvider.Path)
+		if policyArn == nil || policyErr != nil {
+			return managed.ExternalObservation{}, awsclient.Wrap(policyErr, errExternalName)
+		}
+		meta.SetExternalName(cr, aws.ToString(policyArn))
```
muvaf, line 113: If it's possible to construct the whole ARN, maybe we should consider using external-name as the name of the policy and construct ARN wherever needed?

chlunde: Do we have to make it cheaper to get the account id to do this?

chlunde:
This build was green, but master is now broken:

```
pkg/controller/identity/iampolicy/controller.go:100:1: cyclomatic complexity 12 of func `(*external).Observe` is high (> 10) (gocyclo)
```

pkg/controller/identity/iampolicy/controller.go
muvaf: What do you mean? Like storing account id in status to not query every time?

chlunde: In status, or the providerconfig(?). If we need to do an extra AWS API call for every observe.

chlunde:
> This build was green, but master is now broken:
> 
> ```
> pkg/controller/identity/iampolicy/controller.go:100:1: cyclomatic complexity 12 of func `(*external).Observe` is high (> 10) (gocyclo)
> ```

it's nothing wrong with CI or the PRs, it's just that there have been multiple PRs changing the same function, each one adding a bit of extra complexity i different parts of the same functions. After merge, the complexity is too high.

There are some checks that seem unnecessarily paranoid, but maybe they are there for a reason:

```diff
diff --git a/pkg/controller/identity/iampolicy/controller.go b/pkg/controller/identity/iampolicy/controller.go
index 1facf5d9..2a685afc 100644
--- a/pkg/controller/identity/iampolicy/controller.go
+++ b/pkg/controller/identity/iampolicy/controller.go
@@ -52,7 +52,6 @@ const (
        errDelete        = "failed to delete the IAM Policy"
        errUpdate        = "failed to update the IAM Policy"
        errExternalName  = "failed to update the IAM Policy external-name"
-       errEmptyPolicy   = "empty IAM Policy received from IAM API"
        errPolicyVersion = "No version for policy received from IAM API"
        errUpToDate      = "cannot check if policy is up to date"
 )
@@ -117,14 +116,10 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
        policyResp, err := e.client.GetPolicy(ctx, &awsiam.GetPolicyInput{
                PolicyArn: aws.String(meta.GetExternalName(cr)),
        })
-
        if err != nil {
                return managed.ExternalObservation{}, awsclient.Wrap(resource.Ignore(iam.IsErrorNotFound, err), errGet)
        }
 
-       if policyResp.Policy == nil {
-               return managed.ExternalObservation{}, errors.New(errEmptyPolicy)
-       }
        policy := policyResp.Policy
 
        cr.SetConditions(xpv1.Available())
@@ -142,13 +137,11 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
                PolicyArn: aws.String(meta.GetExternalName(cr)),
                VersionId: aws.String(cr.Status.AtProvider.DefaultVersionID),
        })
-
-       if err != nil || versionRsp.PolicyVersion == nil {
+       if err != nil {
                return managed.ExternalObservation{}, awsclient.Wrap(err, errPolicyVersion)
        }
 
        update, err := iam.IsPolicyUpToDate(cr.Spec.ForProvider, *versionRsp.PolicyVersion)
-
        if err != nil {
                return managed.ExternalObservation{}, awsclient.Wrap(err, errUpToDate)
        }
```

Should we just follow up with a nolint PR for this method @muvaf?

chlunde:
https://github.com/crossplane/provider-aws/pull/978

pkg/controller/identity/iampolicy/controller.go
muvaf: We could possibly store it in the status so that it's available everywhere and query only if that status field is empty.

Issue #599, opened 2021-03-18
No support for resource tags in IAMPolicy


mf-lit:
### What happened?

I've noticed that the IAMPolicy MR has no support for tags (unlike say, IAMRole). I couldn't see an existing issue or PR for this so thought I'd note it quickly.

### What environment did it happen in?
Crossplane version: 1.1
Provider: 0.17

ra-grover:
@mf-lit I opened a PR to add this feature.
#910

PR #931, opened 2021-11-12
Added tags for iam policy

ra-grover:

This PR adds feature to tag IamPolicy, fixes https://github.com/crossplane/provider-aws/issues/599

I have:

The changes have been tested on a local kind cluster and verified the state of IamPolicy via cli.


package/crds/identity.aws.crossplane.io_iampolicies.yaml
```diff
@@ -78,6 +78,22 @@ spec:
                   path:
                     description: The path to the policy.
                     type: string
+                  tags:
+                    items:
+                      description: Tag represents a tag attached to a v1alpha1.User
```
chlunde, line 83: `description: Tag represents a tag attached to a v1alpha1.User`. Something went wrong here due to sharing the struct. Maybe the Tag struct doc strings could be changed to be generic?

apis/identity/v1alpha1/iampolicy_types.go
```diff
@@ -37,6 +37,8 @@ type IAMPolicyParameters struct {
 
 	// The name of the policy.
 	Name string `json:"name"`
+
+	Tags []Tag `json:"tags,omitempty"`
```
chlunde, line 41: Most of the time there's a doc string. Not sure how helpful this is in this case, but still:

Maybe something like.
```
       // Tags attached to the policy. For more information about
       // tagging, see Tagging customer managed policies (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_tags.html)
       // in the IAM User Guide.
```

package/crds/identity.aws.crossplane.io_iampolicies.yaml
ra-grover: Sure @chlunde , I will make a change. I copied that bit from IAMRole crd.

apis/identity/v1alpha1/iampolicy_types.go
ra-grover: Sure thing.

ra-grover: Added @chlunde , marking this as resolved.

pkg/clients/iam/iamrole.go
```diff
@@ -198,25 +198,23 @@ func IsRoleUpToDate(in v1beta1.IAMRoleParameters, observed iamtypes.Role) (bool,
...
 
 // DiffIAMTags returns the lists of tags that need to be removed and added according
-// to current and desired states.
-func DiffIAMTags(local []v1beta1.Tag, remote []iamtypes.Tag) (add []iamtypes.Tag, remove []string) {
-	addMap := make(map[string]string, len(local))
-	for _, t := range local {
-		addMap[t.Key] = t.Value
-	}
+// to current and desired states, also returns if desired state needs to be updated
+func DiffIAMTags(local map[string]string, remote []iamtypes.Tag) (add []iamtypes.Tag, remove []string, areTagsUpToDate bool) {
```
chlunde, line 202: Sorry I missed this. This was a pure function, and now mutates the input. I don't know if crossplane has any guidelines around this but I think it is less error prone. Have you checked all callers?

cc @haarchri what do you think?

ra-grover: Hey @chlunde yeah I have checked all the callers. I modified this to actually reuse this piece. `IamRoles/IamPolicies` both are using this for diffing the tags.

@haarchri Sir, any inputs ?

package/crds/identity.aws.crossplane.io_iampolicies.yaml
ra-grover: @chlunde  My bad, `make` is dynamically generating this description. And is consistent with `IAMRole` crd. Reverting to what was already there, as `make check-diff` is breaking.

pkg/clients/iam/iamrole.go
haarchri: @chlunde is this a Blocker for you for me test was successful and tests incode looking AS expected

chlunde: @haarchri @ra-grover no, it's good. I see the reason is that local is not a map, it used to be a slice. All callers take a copy of the input before calling this function so there should not be any bugs.

I hope we can make a generic version of DiffTags when Go 1.18 is out.. :)

chlunde (approved):
Testet tagging/untagging/retagging policy and role - works :)

chlunde:
Thanks, merging. The codecov report points seems to be incorrect, it reports changes in coverage in methods you have not changed.

PR #939, opened 2021-11-14
feat(cw): added cloudwatch loggroup

haarchri:

added cloudwatch loggroup 
why we need this:
- control over eks controlplane logs as "managed" loggroup in format `/aws/eks/<cluster-name>/cluster` to set `retentionInDays` and the option to delete the loggroup when eks controlplane is removed

I have:

- create & delete

```
kubectl get loggroup
NAME                READY   SYNCED   EXTERNAL-NAME
sample-loggroup     True    True     /aws/eks/sample-cluster/cluster
```

- create & delete multiple loggroups

```
 kubectl get loggroup
NAME                READY   SYNCED   EXTERNAL-NAME
sample-loggroup     True    True     /aws/eks/sample-cluster/cluster
sample-loggroup-2   True    True     /aws/eks/sample-cluster-2/cluster
```

```
kubectl describe loggroup.cloudwatchlogs.aws.crossplane.io/sample-loggroup
Name:         sample-loggroup
Namespace:    
Labels:       <none>
Annotations:  crossplane.io/external-create-pending: 2021-11-14T22:25:03+01:00
              crossplane.io/external-create-succeeded: 2021-11-14T22:25:04+01:00
              crossplane.io/external-name: /aws/eks/sample-cluster/cluster
API Version:  cloudwatchlogs.aws.crossplane.io/v1alpha1
Kind:         LogGroup
Metadata:
  Creation Timestamp:  2021-11-14T21:25:03Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generation:  1
    Manager:      kubectl-client-side-apply
    Operation:    Update
    Time:         2021-11-14T21:25:03Z
    API Version:  cloudwatchlogs.aws.crossplane.io/v1alpha1
    Manager:         provider
    Operation:       Update
    Time:            2021-11-14T21:25:04Z
  Resource Version:  32821
  UID:               40914687-b241-40d3-bec6-925d37d37581
Spec:
  Deletion Policy:  Delete
  For Provider:
    Log Group Name:     /aws/eks/sample-cluster/cluster
    Region:             us-east-1
    Retention In Days:  1
  Provider Config Ref:
    Name:  example
Status:
  At Provider:
  Conditions:
    Last Transition Time:  2021-11-14T21:25:04Z
    Status:                True
    Type:                  Ready
    Last Transition Time:  2021-11-14T21:25:04Z
    Status:                True
    Type:                  Synced
Events:
  Type    Reason                   Age   From                                               Message
  ----    ------                   ----  ----                                               -------
  Normal  CreatedExternalResource  10s   managed/loggroup.cloudwatchlogs.aws.crossplane.io  Successfully requested creation of external resource
  Normal  UpdatedExternalResource  10s   managed/loggroup.cloudwatchlogs.aws.crossplane.io  Successfully requested update of external resource
```


apis/cloudwatchlogs/v1alpha1/custom_types.go
```diff
@@ -0,0 +1,47 @@
...
+package v1alpha1
+
+import (
+	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
+)
+
+// CustomLogGroupParameters contains the additional fields for LogGroup.
+type CustomLogGroupParameters struct {
+	// The number of days to retain the log events in the specified log group.
+	// To set a log group to never have log events expire, use DeleteRetentionPolicy
```
chlunde, line 26: see `0` is added as a special case for crossplane. Not sure if we have something similar elsewhere. I guess `nil` causes issues with late init? Another option might be to use k8s `intstr` and support `"Forever"` or `"Infinite"`?

I don't think this comment relevant in user facing docs exactly as it is written here: `// To set a log group to never have log events expire, use DeleteRetentionPolicy`, but I do think it would be good to have a note about the special case w/ `0` and that this calls `DeleteRetentionPolicy`.

pkg/controller/cloudwatchlogs/loggroup/setup.go
```diff
@@ -0,0 +1,207 @@
...
+		}); err != nil {
+			return managed.ExternalUpdate{}, awsclients.Wrap(err, errUpdate)
+		}
+	}
+
+	return managed.ExternalUpdate{}, nil
+}
+
+// diffTags returns tags that should be added or removed.
+func diffTags(spec map[string]*string, current map[string]*string) (addMap map[string]*string, remove []*string) {
```
chlunde, line 190: Looks like we have this variant another place too, consider if it would be possible to share this implementation, maybe in `pkg/clients/`: `pkg/controller/lambda/function/setup.go`

chlunde (commented):
Looked through the code, looks good to me, but I don't have time to test it at the moment. I think the special case resetting the retention policy should be double checked on slack #dev or with owners.

apis/cloudwatchlogs/v1alpha1/custom_types.go
haarchri: Fair enough i will Change the Description - my Idea was to use 0 Like Cloudwatch LogGroup in Terraform "If you select 0, the events in the log group are always retained and never expirex

pkg/controller/cloudwatchlogs/loggroup/setup.go
haarchri: moved diffTags `map[string]*string ` for ressources lambda, eks-addon and cloudwatch loggroup to aws clients and tested the creation, update , delete
thanks for pointing this out

apis/cloudwatchlogs/v1alpha1/custom_types.go
haarchri: changed the description

chlunde (approved):

Issue #932, opened 2021-11-12
IAMRole diff is to picky which results in unnecessary reconciling


eloo:
### What happened?

When a IAMRole is created with a assumePolicyDocument which is transformed by AWS crossplane will reconcile it the whole time. 

### How can we reproduce it?

Deploy a role like this
```
apiVersion: identity.aws.crossplane.io/v1beta1
kind: IAMRole
metadata:
  annotations:
    crossplane.io/external-name: test-role
  name: test-role
spec:
  deletionPolicy: Delete
  forProvider:
    assumeRolePolicyDocument: '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Federated":"arn:aws:iam::account-id:oidc-provider/oidc.eks.eu-central-1.amazonaws.com/id/oidc-id"},"Action":["sts:AssumeRoleWithWebIdentity"],"Condition":{"StringLike":{"oidc.eks.eu-central-1.amazonaws.com/id/oidc-id:sub":"system:serviceaccount:dev:config-server"}}}]}'
    maxSessionDuration: 3600
    path: /
    permissionsBoundary: arn:aws:iam::account-id:policy/acs/acs-boundary-policy
  providerConfigRef:
    name: aws-provider-config
```

Then you will see in the logs that this role is reconciled the whole time. 
If you check the diff you will see something like this

```
 assumeRolePolicyDocument: '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Federated":"arn:aws:iam::account-id:oidc-provider/oidc.eks.eu-central-1.amazonaws.com/id/oidc-id"},"Action":"sts:AssumeRoleWithWebIdentity","Condition":{"StringLike":{"oidc.eks.eu-central-1.amazonaws.com/id/oidc-id:sub":"system:serviceaccount:dev:config-server"}}}]}'
```
should be

```
 assumeRolePolicyDocument: '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Federated":"arn:aws:iam::account-id:oidc-provider/oidc.eks.eu-central-1.amazonaws.com/id/oidc-id"},"Action":["sts:AssumeRoleWithWebIdentity"],"Condition":{"StringLike":{"oidc.eks.eu-central-1.amazonaws.com/id/oidc-id:sub":"system:serviceaccount:dev:config-server"}}}]}'
```

[...]

PR #942, opened 2021-11-15
normalize iam policy actions

smcavallo:

Enhanced policy comparison.  See linked issue.
Converts `Action` with a single item from array to string for comparison purposes:
`"Action":["sts:AssumeRoleWithWebIdentity"]` to `"Action":"sts:AssumeRoleWithWebIdentity"` 

In effect this is the same permission but the current comparison interpret them as different.

Fixes #
https://github.com/crossplane/provider-aws/issues/932

I have:

* Unit tests

### DIFF
--- a/pkg/clients/iam/iampolicy.go
+++ b/pkg/clients/iam/iampolicy.go
@@ -3,6 +3,8 @@ package iam
 import (
 	"context"
 	"net/url"
+	"regexp"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/iam"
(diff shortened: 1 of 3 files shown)


chlunde (commented):
Thanks for looking at this!

I think the linked issue is for IAM Role assume policy conditions, while this change is for IAM policies. This highlights another issue, there are IAM policies everywhere, not just in the IAM API group but for many APIs there are resource based policies too.

I think it would be nice to solve the issue in the shared helper, ideally without regular expressions. Do you think it is possible to fix the shared helper and change iam policies to use that instead? Perhaps as an option to `cmp.Diff` (POC: https://play.golang.org/p/uZqVdENJcIZ - but this applies the transformation for too many JSON keys, like Condtions, which I'm not sure about. Maybe FilterPath can ensure we only transform in safe places.)

https://github.com/crossplane/provider-aws/blob/master/pkg/clients/aws.go#L739-L768

smcavallo:
@chlunde - I 100% agree.  Will propose:
- `IAMRole` and `IAMPolicy` be processed through the same function.
- Conditions/Actions/Resources - single value arrays be converted to strings for comparisons
- Ignore sort/ordering be implemented as well
- This can be achieved through custom Marshal/Unmarshal instead of regex - will solve the sort/ordering and can be applied to only the fields where both ways are supported (ex. `Actions|Resources|etc`) instead of all

I did not take that approach since it is a much larger change and I did not know how Marshaling would affect policies/roles in the wild or if there were edge cases which would break if implemented.  There aren't enough unit tests around testing here, and one positive about the change is many more tests can be added.

pkg/clients/iam/iampolicy.go
```diff
@@ -45,14 +47,44 @@ func IsPolicyUpToDate(in v1alpha1.IAMPolicyParameters, policy iamtypes.PolicyVer
...
 	}
 
 	return cmp.Equal(compactIAMPolicy, compactSpecPolicy), nil
 }
+
+// replaceActionArray converts Actions with a single item from an array to a string
+func replaceActionArray(compactJSON string) string {
+	// Ex.  Convert "Action": ["sts:AssumeRole"] -> "Action": "sts:AssumeRole"
+	// But ignore "Action": ["sts:AssumeRole", "sts:GetFederationToken" ] since there are multiple actions
+	r := regexp.MustCompile("\"Action\":\\[(\\S*?)\\]")
```
MisterMX, line 79: I don't think a regex is suited here. The parser from #1772 could be used here as well.

MisterMX:
Closing in favor of #1772.

PR #779, opened 2021-07-26
Implements kms(cmk) alias

haarchri:
### Description of your changes
Implements kms(cmk) alias

Fixes #778

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

```
➜  provider-aws git:(kms-alias) ✗ kubectl apply -f examples/kms/alias.yaml
alias.kms.aws.crossplane.io/dev-key created
```

```
➜  provider-aws git:(kms-alias) ✗ kubectl get alias
NAME      READY   SYNCED   EXTERNAL-NAME
dev-key   True    True     alias/dev-key
```

```
➜  provider-aws git:(kms-alias) ✗ kubectl describe alias dev-key
Name:         dev-key
Namespace:    
Labels:       <none>
Annotations:  crossplane.io/external-name: alias/dev-key
API Version:  kms.aws.crossplane.io/v1alpha1
Kind:         Alias
Metadata:
  Creation Timestamp:  2021-07-26T21:49:35Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generation:        2
  Resource Version:  1975464
  Self Link:         /apis/kms.aws.crossplane.io/v1alpha1/aliases/dev-key
  UID:               36159cba-4325-48f7-a3d8-964be025676c
Spec:
  Deletion Policy:  Delete
  For Provider:
    Alias Name:     alias/dev-key
    Region:         eu-central-1
    Target Key ID:  ea87e120-8417-4ade-8499-9e093d5664d7
    Target Key ID Ref:
      Name:  dev-key
  Provider Config Ref:
    Name:  default
Status:
  At Provider:
  Conditions:
    Last Transition Time:  2021-07-26T21:49:35Z
    Reason:                ReconcileSuccess
    Status:                True
    Type:                  Synced
    Last Transition Time:  2021-07-26T21:49:35Z
    Reason:                Available
    Status:                True
    Type:                  Ready
Events:
  Type    Reason                   Age   From                                 Message
  ----    ------                   ----  ----                                 -------
  Normal  CreatedExternalResource  35s   managed/alias.kms.aws.crossplane.io  Successfully requested creation of external resource
```

```
➜  provider-aws git:(kms-alias) ✗ kubectl delete -f examples/kms/alias.yaml
alias.kms.aws.crossplane.io "dev-key" deleted
```

[contribution process]: https://git.io/fj2m9

PR #950, opened 2021-11-17
Key alias fix

muvaf:

I merged https://github.com/crossplane/provider-aws/pull/779 along with a few other PRs but apparently it broke `lint` action because of usage of a deprecated function from crossplane-runtime. Since we don't require PRs to be up-to-date, it slipped through the cracks.

When I tried to fix it, I realized that we can actually use external name annotation feature instead of `spec.aliasName`. But if we remove that field from spec, then we'd have all fields in `ignore` (total of two fields) and this causes generated `for` loops to be empty, hence code won't compile since the secondary value in for loop, i.e. `_, elem := ...` is unused. It's a bug in the code generator we need to fix and there isn't much workaround. Including `TargetKeyId` field didn't help since we have to ignore it because it's marked as required otherwise but we use ref & selector for that field. So, I ended up removing this from code-gen for now until the issue is fixed.

One additional thing I had to do was to assume that if `Key` is in `PendingDeletion` state, then we should treat it as deleted, otherwise it'll be stuck for days.

I have:

Manually together with `KMS Key`.


muvaf:
Opened https://github.com/aws-controllers-k8s/community/issues/1069 to address ignoring all fields in `ReadMany` type APIs.

package/crds/kms.aws.crossplane.io_aliases.yaml
```diff
@@ -84,9 +75,9 @@ spec:
                       For example: \n    * Key ID: 1234abcd-12ab-34cd-56ef-1234567890ab
                       \n    * Key ARN: arn:aws:kms:us-east-2:111122223333:key/1234abcd-12ab-34cd-56ef-1234567890ab
                       \n To get the key ID and key ARN for a CMK, use ListKeys or
-                      DescribeKey.'
+                      DescribeKey."
                     type: string
-                  targetKeyIDRef:
+                  targetKeyIdRef:
```
negz, line 80: Is it possible to maintain the old casing here? I believe if we did this would not be a breaking change.

pkg/controller/kms/alias/controller.go
```diff
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by ack-generate. DO NOT EDIT.
-
```
negz, line 18: Could be worth a note somewhere to explain why this resource looks like generated code, but "isn't".

negz (approved):
None of my comments are blockers.

package/crds/kms.aws.crossplane.io_aliases.yaml
negz: Oh nevermind - just realised this is a new managed resource so it's not a breaking change anyhow.

pkg/controller/kms/alias/controller.go
muvaf: Added a comment to all those files pointing to description of this PR for details.

PR #959, opened 2021-11-19
Highlight issues in false value comparisons (rds)

adamkirk:
I've added some tests to highlight cases where the IsUpToDate
function in the rds client returns false, when it should be true.

This seems to be reliant on the CreatePatch and LateInitialize
function; one of which I suspect is the root cause of this issue.
So I've added some tests for the CreatePatch method which highlight
that when a boolean is false in both the CRD and AWS response, it
is still included in the patch.

This is to support issue: #960 

I've been running these tests in a docker container by running `docker run --rm -it -v "$(pwd):/srv" -w /srv golang:1.16 bash` at the root of the repository to use the golang 1.16 runtime. And then running these tests with `go test -v -run TestIsUpToDate ./pkg/clients/rds/...` and `go test -v -run TestCreatePatch ./pkg/clients/rds/...`. 

Everything was passing until I started messing around

### DIFF
--- a/pkg/clients/rds/rds_test.go
+++ b/pkg/clients/rds/rds_test.go
@@ -103,34 +103,135 @@ func TestCreatePatch(t *testing.T) {
 		args
 		want
 	}{
-		"SameFields": {
+		// "SameFields": {
+		// 	args: args{
+		// 		db: &rdstypes.DBInstance{
+		// 			AllocatedStorage: allocatedStorage,
+		// 			CharacterSetName: &characterSetName,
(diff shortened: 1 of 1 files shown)


chlunde:
@adamkirk could you check if this is relevant when #971 is merged?

adamkirk:
> @adamkirk could you check if this is relevant when #971 is merged?

@chlunde I've just been testing this out, but it seems that the issue was fixed somewhere between 0.19.1 and 0.22.0.

I'll close this off, thanks for the help! :)

PR #966, opened 2021-11-19
Expose eks cluster certificate authority data in status

### LINKED ISSUE #955: Expose CAData in EKS CRD

This is the current `EKS` object status:
```golang
// ClusterObservation is the observed state of a cluster.
type ClusterObservation struct {
	// The Amazon Resource Name (ARN) of the cluster.
	Arn string `json:"arn,omitempty"`

	// The Unix epoch timestamp in seconds for when the cluster was created.
	CreatedAt *metav1.Time `json:"createdAt,omitempty"`

	// The endpoint for your Kubernetes API server.
	Endpoint string `json:"endpoint,omitempty"`

	// The identity provider information for the cluster.
	Identity Identity `json:"identity,omitempty"`

	// The platform version of your Amazon EKS cluster. For more information, see
	// Platform Versions (https://docs.aws.amazon.com/eks/latest/userguide/platform-versions.html)
	// in the Amazon EKS User Guide .
	PlatformVersion string `json:"platformVersion,omitempty"`

	// The VPC configuration used by the cluster control plane. Amazon EKS VPC resources
	// have specific requirements to work properly with Kubernetes. For more information,
	// see Cluster VPC Considerations (https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html)
	// and Cluster Security Group Considerations (https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html)
	// in the Amazon EKS User Guide.
	ResourcesVpcConfig VpcConfigResponse `json:"resourcesVpcConfig,omitempty"`

	// The current status of the cluster.
	Status ClusterStatusType `json:"status,omitempty"`
}
```
We've generated the bearer token and we want to generate a k8s client based on the observed states in `EKS` object. Currently, `CAData` is missing, and we have to issue a dedicate request to AWS to retrieve the CA data, which is not ideal.

It would be ideal if `EKS` object can expose a field `CAData` in the struct above https://github.com/crossplane/provider-aws/blob/2fdde09cccaf56a8b0e770d92318d46989704a7b/apis/eks/v1beta1/types.go#L231-L259
, and it can be done by simply exposing the CA data from EKS `DescribeCluster` results https://github.com/crossplane/provider-aws/blob/2fdde09cccaf56a8b0e770d92318d46989704a7b/pkg/clients/eks/eks.go#L218-L223

hanlins:

Fixes #955

I have:

Added unit test for the new field introduced.

### DIFF
--- a/apis/eks/v1beta1/types.go
+++ b/apis/eks/v1beta1/types.go
@@ -239,6 +239,9 @@ type ClusterObservation struct {
 	// The endpoint for your Kubernetes API server.
 	Endpoint string `json:"endpoint,omitempty"`
 
+	// The Base64-encoded certificate data required to communicate with your cluster.
+	CertificateAuthorityData string `json:"certificateAuthorityData,omitempty"`
+
 	// The identity provider information for the cluster.
 	Identity Identity `json:"identity,omitempty"`
 
--- a/package/crds/eks.aws.crossplane.io_clusters.yaml
+++ b/package/crds/eks.aws.crossplane.io_clusters.yaml
@@ -357,6 +357,10 @@ spec:
                   arn:
                     description: The Amazon Resource Name (ARN) of the cluster.
                     type: string
+                  certificateAuthorityData:
+                    description: The Base64-encoded certificate data required to communicate
+                      with your cluster.
+                    type: string
                   createdAt:
                     description: The Unix epoch timestamp in seconds for when the
                       cluster was created.
--- a/pkg/clients/eks/eks.go
+++ b/pkg/clients/eks/eks.go
@@ -240,6 +240,10 @@ func GenerateObservation(cluster *ekstypes.Cluster) v1beta1.ClusterObservation {
 			VpcID:                  awsclients.StringValue(cluster.ResourcesVpcConfig.VpcId),
 		}
 	}
+
+	if cluster.CertificateAuthority != nil {
+		o.CertificateAuthorityData = awsclients.StringValue(cluster.CertificateAuthority.Data)
+	}
 	return o
 }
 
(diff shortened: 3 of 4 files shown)


hasheddan (commented):
@hanlins the CA data is currently [exposed in the connection secret](https://github.com/crossplane/provider-aws/blob/1f7f5842a2e616b70470f58c6fb31c5e65c44e8d/pkg/clients/eks/eks.go#L405) that gets created for the cluster -- is that suitable for your use case?

hanlins:
> @hanlins the CA data is currently [exposed in the connection secret](https://github.com/crossplane/provider-aws/blob/1f7f5842a2e616b70470f58c6fb31c5e65c44e8d/pkg/clients/eks/eks.go#L405) that gets created for the cluster -- is that suitable for your use case?

Hi @hasheddan, thanks a lot! Btw, is there a way to figure out what's inside that secret without digging into code? Currently, the secret reference is in the cluster spec, and the contents in that secret are actually controlled by the implementation of the controllers, not part of the CRD API contract. I'm fine relying on this implicit contract, but is there a better way to document this and make it more explicit? For now I think user has to either really provision a Cluster object and check what's in that referenced secret, or go through the code as you pointed above to learn what's in that secret, it would be nice to make this information explicit since it's the interface to users. If there're some documents on this (or did I miss any existing docs?) that make this API clear to the users, if so then I'm ok to close this PR :)

haarchri:
@hasheddan did we have in crossplane compositions an patch like `fromConnectionSecretKey` then it is easier to get informations from ConnectionSecretKey to other resources

muvaf:
Is certificate authority data sensitive? If so, we can't have it under `status`.

janwillies:
This is basically the question in https://github.com/crossplane/crossplane/issues/2772#issuecomment-998686039

haarchri:
I am fine to merge the Open PR to add this ca in status

hanlins:
Any further thoughts/concerns? Should we merge this change? @haarchri :)

ONordander:
Any updates on this? Would be nice to have in the next release.

haarchri (approved):
@hanlins  thanks for implementation 
@chlunde  thanks for add clarification for CAdata

Issue #625, opened 2021-04-12
CreateJSONPatch style update logic cannot delete items


chlunde:
### What happened?
The ec2.securitygroup controller uses `CreateJSONPatch` to find out what fields needs an update. The zero value means that the field does not need an update.

This means that if a user for example deletes all ingress rules or all tags, no Delete/Revoke call will be sent to AWS.

I believe the same issue exists for routetable - I have not checked the other code but I'm creating this issue to discuss if CreateJSONPatch is a good idea, given this problem. 

### How can we reproduce it?

1. Add at least one tag
2. Remove all tags

pkg/controller/dynamodb/table/hooks.go
pkg/clients/iam/iamrole.go
pkg/clients/rds/rds.go
pkg/clients/redshift/redshift.go
pkg/clients/resourcerecordset/resourcerecordset.go
pkg/clients/elasticloadbalancing/elb/elb.go
pkg/clients/eks/eks.go
pkg/clients/ec2/routetable.go
pkg/clients/ecr/repository.go
pkg/controller/cloudfront/distribution/setup.go

### What environment did it happen in?
provider-aws 0.17

PR #970, opened 2021-11-23
Late initialize SecurityGroup ingress and egress rules

negz:

This PR is a small follow up to https://github.com/crossplane/provider-aws/pull/631 that will now late init ingress and egress rules only when the desired array of rules is empty. This can be handy when importing security groups. It also adds a few more unit tests around the late init and up-to-date checking functionality.

I have:

I've tested this by:

1. Applying the SG at `examples/ec2/securitygroup.yaml`.
2. Validating that if I add an egress and update the ingress, those changes are reflected.
3. Deleting the SG with `deletionPolicy: Orphan`.
4. Validating that if I re-import the SG with empty ingress and egress arrays they will be late initialized.

### DIFF
--- a/pkg/clients/ec2/securitygroup.go
+++ b/pkg/clients/ec2/securitygroup.go
@@ -52,13 +52,58 @@ func IsRuleAlreadyExistsErr(err error) bool {
 	return errors.As(err, &awsErr) && awsErr.ErrorCode() == InvalidPermissionDuplicate
 }
 
-// GenerateEC2Permissions converts object Permissions to ec2 format
-func GenerateEC2Permissions(objectPerms []v1beta1.IPPermission) []ec2types.IpPermission {
-	if len(objectPerms) == 0 {
+// BuildFromEC2Permissions converts ec2 Permissions to our API's format.
+func BuildFromEC2Permissions(in []ec2types.IpPermission) []v1beta1.IPPermission {
+	if len(in) == 0 {
 		return nil
 	}
-	permissions := make([]ec2types.IpPermission, len(objectPerms))
-	for i, p := range objectPerms {
+	permissions := make([]v1beta1.IPPermission, len(in))
+	for i, p := range in {
+		ipPerm := v1beta1.IPPermission{
+			FromPort:   p.FromPort,
+			IPProtocol: aws.StringValue(p.IpProtocol),
+			ToPort:     p.ToPort,
+		}
+		for _, c := range p.IpRanges {
+			ipPerm.IPRanges = append(ipPerm.IPRanges, v1beta1.IPRange{
+				CIDRIP:      aws.StringValue(c.CidrIp),
+				Description: c.Description,
+			})
+		}
+		for _, c := range p.Ipv6Ranges {
+			ipPerm.IPv6Ranges = append(ipPerm.IPv6Ranges, v1beta1.IPv6Range{
+				CIDRIPv6:    aws.StringValue(c.CidrIpv6),
+				Description: c.Description,
+			})
+		}
+		for _, c := range p.PrefixListIds {
+			ipPerm.PrefixListIDs = append(ipPerm.PrefixListIDs, v1beta1.PrefixListID{
+				Description:  c.Description,
+				PrefixListID: aws.StringValue(c.PrefixListId),
+			})
+		}
+		for _, c := range p.UserIdGroupPairs {
+			ipPerm.UserIDGroupPairs = append(ipPerm.UserIDGroupPairs, v1beta1.UserIDGroupPair{
+				Description:            c.Description,
+				GroupID:                c.GroupId,
+				GroupName:              c.GroupName,
+				UserID:                 c.UserId,
+				VPCID:                  c.VpcId,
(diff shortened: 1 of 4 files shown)


negz:
Note that backport will require https://github.com/crossplane/provider-aws/pull/969/files to be merged first.

chlunde:
should changing an ingress rule from

```yaml
    ingress:
    - fromPort: 80
      ipProtocol: tcp
      ipRanges:
      - cidrIp: 10.0.0.0/8
      toPort: 80
 ```
 
 to 
 
 ```yaml
     ingress: []
 ```
 
 not delete that ingress rule? Could we check for nil vs len() == 0? I haven't tried that but I think I would like to be able to remove all rules.

**Comment by negz** [comment]
> should changing an ingress rule from

What you propose makes sense to me, but I'm not sure how commonly we're following that practice. For example here I note we were already checking the tags via len, not a check for nil. What do you think about moving forward with this PR without making that change and raising an issue to document/assess how we late init arrays?

File: pkg/clients/ec2/securitygroup.go [hunk 1]
```diff
@@ -52,13 +52,58 @@ func IsRuleAlreadyExistsErr(err error) bool {
...
+				GroupName:              c.GroupName,
+				UserID:                 c.UserId,
+				VPCID:                  c.VpcId,
+				VPCPeeringConnectionID: c.VpcPeeringConnectionId,
+			})
+		}
+		permissions[i] = ipPerm
+	}
+	return permissions
+}
```
negz, line 98: Reviewer note: This is pretty much a copy and paste of `GenerateEC2Permissions` but inverted.

pkg/clients/ec2/securitygroup.go
```diff
@@ -117,7 +162,15 @@ func LateInitializeSG(in *v1beta1.SecurityGroupParameters, sg *ec2types.Security
 	in.GroupName = awsclients.LateInitializeString(in.GroupName, sg.GroupName)
 	in.VPCID = awsclients.LateInitializeStringPtr(in.VPCID, sg.VpcId)
 
-	// We cannot safely late init egress/ingress rules because they are keyless arrays
+	// We cannot safely late init egress/ingress rules because they are
+	// keyless arrays, so we follow the typical pattern of only late
+	// initializing if our desired rules are zero length.
+	if len(in.Ingress) == 0 && len(sg.IpPermissions) != 0 {
```
ulucinar, line 168: nit: `len(sg.IpPermissions)` is already checked in `BuildFromEC2Permissions`.

pkg/clients/ec2/securitygroup.go
```diff
@@ -117,7 +162,15 @@ func LateInitializeSG(in *v1beta1.SecurityGroupParameters, sg *ec2types.Security
...
+		in.Ingress = BuildFromEC2Permissions(sg.IpPermissions)
+	}
+	if len(in.Egress) == 0 && len(sg.IpPermissionsEgress) != 0 {
```
ulucinar, line 171: nit: Same as the above comment on input parameter.

chlunde (approved):

ulucinar (approved):

chlunde:
> > should changing an ingress rule from `[...]` to `[]` work
> 
> What you propose makes sense to me, but I'm not sure how commonly we're following that practice. For example here I note we were already checking the tags via len, not a check for nil. What do you think about moving forward with this PR without making that change and raising an issue to document/assess how we late init arrays?

Sure, I don't have a big problem with this.

I created this issue earlier: https://github.com/crossplane/provider-aws/issues/625 - but do you think it should be in core crossplane instead - revisit late init?

The reason I raised it again now is that this issue has some security implications, and it's the "core feature" for _this_ managed resource. Here's another way "unhappy user story" which does not seem far fetched to me:

* You define an egress-only SG for a workload via crossplane.
* To debug something someone changes the SG in the console
* Late init adds that rule back to the CR
* 6 months later.. it's still there, nobody knows

avdhoot:
> > > should changing an ingress rule from `[...]` to `[]` work
> > 
> > 
> > What you propose makes sense to me, but I'm not sure how commonly we're following that practice. For example here I note we were already checking the tags via len, not a check for nil. What do you think about moving forward with this PR without making that change and raising an issue to document/assess how we late init arrays?
> 
> Sure, I don't have a big problem with this.
> 
> I created this issue earlier: #625 - but do you think it should be in core crossplane instead - revisit late init?
> 
> The reason I raised it again now is that this issue has some security implications, and it's the "core feature" for _this_ managed resource. Here's another way "unhappy user story" which does not seem far fetched to me:
> 
> * You define an egress-only SG for a workload via crossplane.
> * To debug something someone changes the SG in the console
> * Late init adds that rule back to the CR
> * 6 months later.. it's still there, nobody knows
This sounds scary. We follow gitops. In this case, the truth of the source should be crd checked in git which is applied by argocd/fluxcd/kubectl. Any changes in that managed recourse should get revered.

haarchri:
@negz @chlunde @ulucinar what is the status here ?

haarchri:
@negz @chlunde @ulucinar what is the status here ?

negz:
I don't think I'll have time to address this unfortunately.

PR #971, opened 2021-11-23
rds.rdsinstance: Compare bool correctly via LateInitialize

### LINKED ISSUE #960: Unnecessary update requests of rdsinstances
### What happened?

After updating to crossplane 1.5.0, and provider-aws 0.21.0, I could see in logs (when i enable --debug on this provider) that update requests were being sent every time my rdsinstances were reconciled. I looked into cloudtrail to see what was being updated and could see that the requestParameters for the ModifyDBInstance calls being made by my crossplane user looked like this:

```
"requestParameters": {
    "allowMajorVersionUpgrade": false,
    "dBInstanceIdentifier": "xxxxxxxx",
    "multiAZ": false,
    "applyImmediately": false,
    "publiclyAccessible": false
},
```

The thing that stood out to me was that the parameters being sent in the update request were all booleans that were set to false, and nothing else (other than the id). I checked these values against my rds instances using the `aws rds describe-db-instances --db-instance-identifier xxxxxxx` command and could see that all of these values in the request were already set to false on the instances in AWS; so these ModifyDBInstance requests shouldn't be getting sent at all as far i can tell.

This doesn't seem to cause any problems in the rdsinstance itself, but the constant update requests seemed a bit worrying.

### How can we reproduce it?

I think this can be recreated by creating an rdsinstance and setting publiclyAccesible or multiAZ to false. It's unclear exactly which fields this may happen with.

This is the spec i'm using for the rdsinstance (with some fields redacted):

```

spec:
  deletionPolicy: Orphan
  forProvider:
    allocatedStorage: 100
    allowMajorVersionUpgrade: false
    autoMinorVersionUpgrade: true
    availabilityZone: eu-west-2b
    backupRetentionPeriod: 1
    caCertificateIdentifier: rds-ca-2019
    dbInstanceClass: db.t2.micro
    dbName: xxxxxxxx
    dbParameterGroupName: default.mysql5.7
    dbSubnetGroupName: dbsubnet-group-name
    deletionProtection: true
    enablePerformanceInsights: false
    engine: mysql
    engineVersion: 5.7.33
    licenseModel: general-public-license
    masterUsername: xxxxxxxx
    multiAZ: false
    port: 3306
    preferredBackupWindow: 03:00-04:00
    preferredMaintenanceWindow: thu:00:26-thu:00:56
    publiclyAccessible: false
    region: eu-west-2
    skipFinalSnapshotBeforeDeletion: false
    storageEncrypted: true
    storageType: gp2
    tags:
    - key: crossplane-kind
      value: rdsinstance.database.aws.crossplane.io
    - key: crossplane-name
      value: xxxxxxxxxxx
      value: default
    - key: myorg.tag1
      value: xxxxxxxx
    - key: myorg.tag2
      value: xxxxxxxx
    - key: myorg.tag3
      value: xxxxxxxx
    vpcSecurityGroupIds:
    - sg-xxxxxxxxx
  providerConfigRef:
    name: default
  writeConnectionSecretToRef:
    name: xxxxxxxxxxxxx
    namespace: crossplane

```

Crossplane version: 1.5.0
Kubernetes v1.21.2
EKS v1.21.2-eks-06eac09

### Extra notes

I did some investigation into the provider myself, and it seems that there is a bug in the `IsUpToDate` function of the rds client. When the rdsinstance has a boolean setting that is false (only tested with the PubliclyAccessible field), and the CRD also has that setting equal to false, it returns that the external is NOT up to date when it should be. 

It looks like this may be due to how booleans can be returned as a nil pointer when they're false rather than a pointer to a false value. See func `Bool` in `pkg/clients/aws.go`.

I've opened a draft PR with some tests that seem to highlight the problems with the IsUpToDate and CreatePatch functions in the rds client. Hopefully this helps: https://github.com/crossplane/provider-aws/pull/959

chlunde:
False booleans were converted to nil after the SDK upgrade. Ensure false
is represented as a pointer to false instead by setting
awsclients.FieldRequired.

Fixes #960

I have:

Tested with existing RDS which was not reconciling properly, there was a diff for multiple false boolean fields. After the patch, the instance became ready.


pkg/clients/rds/rds.go
```diff
@@ -363,31 +363,31 @@ func LateInitialize(in *v1beta1.RDSInstanceParameters, db *rdstypes.DBInstance)
...
 
 	in.AllocatedStorage = awsclients.LateInitializeIntFrom32Ptr(in.AllocatedStorage, &db.AllocatedStorage)
-	in.AutoMinorVersionUpgrade = awsclients.LateInitializeBoolPtr(in.AutoMinorVersionUpgrade, awsclients.Bool(db.AutoMinorVersionUpgrade))
+	in.AutoMinorVersionUpgrade = awsclients.LateInitializeBoolPtr(in.AutoMinorVersionUpgrade, awsclients.Bool(db.AutoMinorVersionUpgrade, awsclients.FieldRequired))
 	in.AvailabilityZone = awsclients.LateInitializeStringPtr(in.AvailabilityZone, db.AvailabilityZone)
 	in.BackupRetentionPeriod = awsclients.LateInitializeIntFrom32Ptr(in.BackupRetentionPeriod, &db.BackupRetentionPeriod)
 	in.CACertificateIdentifier = awsclients.LateInitializeStringPtr(in.CACertificateIdentifier, db.CACertificateIdentifier)
 	in.CharacterSetName = awsclients.LateInitializeStringPtr(in.CharacterSetName, db.CharacterSetName)
-	in.CopyTagsToSnapshot = awsclients.LateInitializeBoolPtr(in.CopyTagsToSnapshot, awsclients.Bool(db.CopyTagsToSnapshot))
+	in.CopyTagsToSnapshot = awsclients.LateInitializeBoolPtr(in.CopyTagsToSnapshot, awsclients.Bool(db.CopyTagsToSnapshot, awsclients.FieldRequired))
```
chlunde, line 371: in hindsight, why not simply 
```suggestion
	in.CopyTagsToSnapshot = awsclients.LateInitializeBoolPtr(in.CopyTagsToSnapshot, &db.CopyTagsToSnapshot)
```

muvaf: Yeah I think taking the reference of those directly would work just as well.

sergenyalcin:
Hi @chlunde, I am working on the following PR: https://github.com/crossplane/provider-aws/pull/981

I realized that, the function I mentioned in the PR description, is also used for Rds instance: 

https://github.com/crossplane/provider-aws/blob/1e29728f989c14d07b5325e3cc4fa8d6d298c51a/pkg/clients/aws.go#L589

Some using references: https://github.com/crossplane/provider-aws/blob/1e29728f989c14d07b5325e3cc4fa8d6d298c51a/pkg/clients/rds/rds.go#L365
https://github.com/crossplane/provider-aws/blob/1e29728f989c14d07b5325e3cc4fa8d6d298c51a/pkg/clients/rds/rds.go#L368

The main problem about this function is that, `nil` and `0` values are considered same. So, if a cloud resource has `0` value for a field and this function is used for late initialization, there can be constant update loops. Could you please check this point in this PR for avoiding a potential issue?

haarchri:
@chlunde  can we get this PR in a state for merging ?

pkg/clients/rds/rds.go
chlunde: I kept it as is as I've already tested this way

haarchri (approved):
@chlunde thanks

Issue #953, opened 2021-11-17
DynamoDB SYNC=FALSE when spec.forProvider.streamSpecification.streamEnabled=false


RafalMaleska:
### What happened?

After creation of DynamoDB with: 

`spec.forProvider.streamSpecification.streamEnabled=false`

MR becomes out of sync:

```yaml
  - lastTransitionTime: "2021-11-17T11:01:45Z"
    message: "update failed: cannot update Table in AWS: ValidationException: Table
      has no stream to disable: TableName: cnp-example\n\tstatus code:
      400, request id: 0A8G1RPI2BN6HIEBUJEIHBUV3VVV4KQNSO5AEMVAA229ASUAAJG"
    reason: ReconcileError
    status: "False"
    type: Synced
```

### How can we reproduce it?

After creation of DynamoDB with: 

```yaml
kind: Table
apiVersion: dynamodb.aws.crossplane.io/v1alpha1
metadata:
  name: cnp-example
spec:
  deletionPolicy: Delete
  forProvider:
    attributeDefinitions:
    - attributeName: attribute1
      attributeType: S
    billingMode: PROVISIONED
    globalSecondaryIndexes:
    - indexName: some-name
      keySchema:
      - attributeName: attribute1
        keyType: HASH
      projection:
        nonKeyAttributes:
        - SomeId
        projectionType: INCLUDE
      provisionedThroughput:
        readCapacityUnits: 1
        writeCapacityUnits: 1
    keySchema:
    - attributeName: attribute1
      keyType: HASH
    provisionedThroughput:
      readCapacityUnits: 1
      writeCapacityUnits: 1
    region: eu-central-1
    sseSpecification:
      enabled: true
      kmsMasterKeyID: arn:aws:kms:eu-central-1:123456789:key/7a423d0b-1c47-47d4-a0ae-3592258a0445
      sseType: KMS
    streamSpecification:
      streamEnabled: false
    tags:
    - key: CostReference
      value: F-330951-51-05
    - key: crossplane-kind
      value: table.dynamodb.aws.crossplane.io
    - key: crossplane-name
      value: cnp-example
    - key: crossplane-providerconfig
      value: iat-123456789
  providerConfigRef:
    name: iat-123456789
  writeConnectionSecretToRef:
    name: dynamodb-cnp-example
    namespace: crossplane-system
```

got resource out of sync:

```yaml
  - lastTransitionTime: "2021-11-17T11:01:45Z"
[...]

Issue #946, opened 2021-11-16
DynamoDB Table Reconciliation Failure


jbhennin:
### What happened?

Received a reconciliation error when created a table with billingMode  PROVISIONED

```
update failed: cannot update Table in AWS: ValidationException: One or more parameter values were invalid: ProvisionedThroughput must be specified when BillingMode is PROVISIONED
status code: 400, request id: 604LFC8D4PLP4MGASL844S7VA3VV4KQNSO5AEMVJF66Q9ASUAAJG
```

### How can we reproduce it?
Create a table using the following yaml.
```yaml
apiVersion: dynamodb.aws.crossplane.io/v1alpha1
kind: Table
metadata:
  name: jillian-test-table
  namespace: warrior
spec:
  providerConfigRef:
    name: aws-provider
  forProvider:
    attributeDefinitions:
      - attributeName: "userId"
        attributeType: "S"
    provisionedThroughput:
      readCapacityUnits: 5
      writeCapacityUnits: 5
    keySchema:
      - attributeName: "userId"
        keyType: HASH
    region: us-west-2
    billingMode: PROVISIONED
```

### What environment did it happen in?
Crossplane version: 1.3
AWS Provider: 0.20.3

haarchri:
at the moment you can get it working not set `billingMode: PROVISIONED` when `provisionedThroughput` is set
we need to investigate why this happened

RafalMaleska:
tested with: 

```
kind: Table
apiVersion: dynamodb.aws.crossplane.io/v1alpha1
metadata:
  name: cnp-example
spec:
  deletionPolicy: Delete
  forProvider:
    attributeDefinitions:
    - attributeName: attribute1
      attributeType: S
    globalSecondaryIndexes:
    - indexName: some-name
      keySchema:
      - attributeName: attribute1
        keyType: HASH
      projection:
        nonKeyAttributes:
        - SomeId
        projectionType: INCLUDE
      provisionedThroughput:
        readCapacityUnits: 1
        writeCapacityUnits: 1
    keySchema:
    - attributeName: attribute1
      keyType: HASH
    provisionedThroughput:
      readCapacityUnits: 1
      writeCapacityUnits: 1
    region: eu-central-1
    sseSpecification:
      enabled: true
      kmsMasterKeyID: arn:aws:kms:eu-central-1:1234567890:key/7a423d0b-1c47-47d4-a0ae-3592258a0445
      sseType: KMS
    tags:
    - key: CostReference
[...]

PR #973, opened 2021-11-24
Accomodate DynamoDB API's implied defaults

negz:

Fixes https://github.com/crossplane/provider-aws/issues/946
Fixes https://github.com/crossplane/provider-aws/issues/953

This PR fixes a class of bugs that were causing Crossplane to think it needed to update a DynamoDB Table when it did not. Crossplane thought it needed to make updates because its desired state appeared to differ from the actual observed state of the table. In fact there was no difference - the problem was that some of the actual state was implied. For example the DynamoDB `DescribeTable` API only includes a billingMode in its response when the mode is `PAY_PER_REQUEST`. When the mode is `PROVISIONED` (the implied default) it's simply not included in the `DescribeTable` output. The same is true for `streamEnabled` - if it's not enabled there's just no `StreamSpecification` in the `DescribeTable` output.

I have:

I've confirmed that I can:

1. Create a Table with an explicit billingMode: PROVISIONED
2. Create a Table with an explicit billingMode: PAY_PER_REQUEST
3. Create a Table with no billingMode specified
4. Update a Table to change the billingMode
5. Create a Table with streamSpecification explicitly disabled

Note that there's quirk of the API that we don't have a great place to document - if you update your table to PAY_PER_REQUEST you need to specify a provisioned throughput of 0, or Crossplane will think it needs an update. I've added a TODO and a note in the example about this for lack of a better approach.

### DIFF
--- a/examples/dynamodb/table.yaml
+++ b/examples/dynamodb/table.yaml
@@ -11,6 +11,7 @@ spec:
     keySchema:
       - attributeName: attribute1
         keyType: HASH
+    billingMode: PROVISIONED
     provisionedThroughput:
       readCapacityUnits: 1
       writeCapacityUnits: 1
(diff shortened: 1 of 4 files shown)


haarchri (approved):
tested the examples - all working 
- create
- update 
- delete

veryterriblebadapis - yes :/ 
think squash commits is one open point

Issue #909, opened 2021-11-05
ReplicationGroup is modifying everytime when used in composition


eloo:
### What happened?
Hi,
we recently have updated our provider-aws and crossplane itself and we are now facing issues with our ReplicationGroups.

This means, now when we create a new claim of our composition the ReplicationGroup will be created in state "Creating" and then switch after a couple of time to "Modifying" and then to "Available". But after a few minutes its going back to "Modifying" then again "Available" and again "Modifying".. And then its stuck in this loop. 

This makes the Claim never come be ready and so its not usable. Related to https://github.com/crossplane/provider-aws/issues/622 

Further this seems NOT to happen when the ReplicationGroup is created using a MR directly. It seems to be only present when used within a Composition. 

The version when we have noticed it was the following:

crossplane 1.4.1 -> crossplane 1.5.0
provider-aws:v0.20.0-rc.0.94.g0ffd2429 -> crossplane/provider-aws:v0.20.0

Currently i'm not sure if this caused by the provider or crossplane itself. But currently we only see this for the ReplicationGroup and not our other resources.

### How can we reproduce it?

Composition: 
```
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: redis-composition-test
  labels:
    provider: aws
    crossplane.io/xrd: xredisinstancestest.test.cloud.test
spec:
  compositeTypeRef:
    apiVersion: test.cloud.test/v1
    kind: XRedisInstancetest
  writeConnectionSecretsToNamespace: crossplane-system
  resources:
    - base: 
        apiVersion: cache.aws.crossplane.io/v1beta1
        kind: ReplicationGroup
        metadata:
          name: METADATA_UID
        spec:
          deletionPolicy: Delete
          providerConfigRef:
            name: aws-provider-config
          forProvider:
            replicationGroupDescription: GROUP_DESCRIPTION
            applyModificationsImmediately: true
            region: eu-central-1
            engine: redis
            engineVersion: 6.x
            atRestEncryptionEnabled: true
            cacheNodeType: cache.t3.micro
            securityGroupIds: 
              - <securityGroupId of the vpc>
            transitEncryptionEnabled: true

            cacheSubnetGroupName: <subnetGroupName>

            numCacheClusters: 1
          writeConnectionSecretToRef:
            name: METADATA_NAME-MR-redis-managed-resource
            namespace: crossplane-system
      patches:
        - type: FromCompositeFieldPath
          fromFieldPath: metadata.name
[...]

PR #981, opened 2021-12-01
Fixing constantly update requests problem of replicationgroup

sergenyalcin:
Signed-off-by: Sergen Yalçın <yalcinsergen97@gmail.com>

Fixes: https://github.com/crossplane/provider-aws/issues/909

This PR addresses the constantly update requests problem of replicationgroup. `snapshotRetentionLimit` field set to 0 in the sdk side when it is skipped in MR manifest. During `LateInitialization`, since `nil` and `0` values are considered the same value in the function used, the value `0` is not written to CR and the value in CR remains `nil`. Therefore, the update call is repeated in each reconcile loop. Reference: https://github.com/crossplane/provider-aws/blob/1e29728f989c14d07b5325e3cc4fa8d6d298c51a/pkg/clients/aws.go#L589

I have:

A ReplicationGroup resource was provisioned without setting `snapshotRetentionLimit`. Then I noticed that, this field set to 0 after late initialization. I did not observe any continuous update call.

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -586,6 +586,10 @@ func LateInitializeIntPtr(in *int, from *int64) *int {
 
 // LateInitializeIntFrom32Ptr returns in if it's non-nil, otherwise returns from
 // which is the backup for the cases in is nil.
+// This function considered that nil and 0 values are same. However, for a *int32, nil and 0 values must be different
+// because if the external Azure resource has a field with 0 value, during late initialization setting this value
+// in CR must be allowed. Please see the LateInitializeIntFromInt32Ptr func.
+// Deprecated: Please use LateInitializeIntFromInt32Ptr instead.
 func LateInitializeIntFrom32Ptr(in *int, from *int32) *int {
 	if in != nil {
 		return in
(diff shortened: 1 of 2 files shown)


pkg/clients/aws.go
```diff
@@ -597,6 +597,21 @@ func LateInitializeIntFrom32Ptr(in *int, from *int32) *int {
 	return nil
 }
 
+// LateInitializeIntFromInt32Ptr returns in if it's non-nil, otherwise returns from
+// which is the backup for the cases in is nil.
+func LateInitializeIntFromInt32Ptr(in *int, from *int32) *int {
```
ulucinar, line 602: Let's also mark `LateInitializeIntFrom32Ptr` as deprecated so that we can later remove it.

And also please comment on the difference between two versions.

ulucinar (approved):
Thank you @sergenyalcin. Please also discuss this issue you've found in https://github.com/crossplane/provider-aws/issues/960.

pkg/clients/aws.go
```diff
@@ -586,6 +586,10 @@ func LateInitializeIntPtr(in *int, from *int64) *int {
 
 // LateInitializeIntFrom32Ptr returns in if it's non-nil, otherwise returns from
 // which is the backup for the cases in is nil.
+// Deprecated.
+// This function considered that nil and 0 values are same. However, for a *int32, nil and 0 values must be different
+// because if the external Azure resource has a field with 0 value, during late initialization setting this value
+// in CR must be allowed. Please see the LateInitializeIntFromInt32Ptr func.
```
muvaf, line 592:
```suggestion
// This function considered that nil and 0 values are same. However, for a *int32, nil and 0 values must be different
// because if the external Azure resource has a field with 0 value, during late initialization setting this value
// in CR must be allowed. Please see the LateInitializeIntFromInt32Ptr func.
// Deprecated: Please use LateInitializeIntFromInt32Ptr instead.
```

The deprecation comment format is `// Deprecated: ...`, see https://go.dev/blog/godoc for more details.

sergenyalcin: @muvaf thanks for your comment!

PR #992, opened 2021-12-07
Ignore fields in `glue.Crawler` and `lambda.Function` that block code-generator bump

muvaf:

Due to https://github.com/aws-controllers-k8s/community/issues/1078 , we cannot use the code generated to set these fields. To unblock code-generator upgrade, we can ignore them so that their getter/setter logic is not generated and handwritten by us in `setup.go` 

I have:

### DIFF
--- a/pkg/controller/glue/crawler/setup.go
+++ b/pkg/controller/glue/crawler/setup.go
@@ -42,7 +42,6 @@ func SetupCrawler(mgr ctrl.Manager, l logging.Logger, limiter workqueue.RateLimi
 			e.preObserve = preObserve
 			e.postObserve = postObserve
 			e.preDelete = preDelete
-			e.postCreate = postCreate
 			e.preCreate = preCreate
 		},
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/glue/crawler/setup.go
```diff
@@ -71,24 +70,17 @@ func preObserve(_ context.Context, cr *svcapitypes.Crawler, obj *svcsdk.GetCrawl
...
-func postObserve(_ context.Context, cr *svcapitypes.Crawler, obj *svcsdk.GetCrawlerOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+func postObserve(_ context.Context, cr *svcapitypes.Crawler, _ *svcsdk.GetCrawlerOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
 	if err != nil {
 		return managed.ExternalObservation{}, err
 	}
 	cr.SetConditions(xpv1.Available())
 	return obs, nil
 }
 
-func postCreate(_ context.Context, cr *svcapitypes.Crawler, obj *svcsdk.CreateCrawlerOutput, _ managed.ExternalCreation, err error) (managed.ExternalCreation, error) {
```
muvaf, line 82: External name feature implementation wasn't correct so I fixed that as well - we can never use `cr.Name`, the source of truth is always the annotation.

AaronME (approved):

AaronME:
Thanks, @muvaf !

Issue #908, opened 2021-11-04
Certificate SYNCED=FALSE


RafalMaleska:
### What happened?

After deploying a Certificate which is issued by a private ca (acm-pca) and using by an AWS Service (AppMesh) the Certificate stays in SYNCED=FALSE.

Error Message at Resource: 

```
update failed: failed to renew Certificate: api error ValidationException:
      Certificate ARN arn:aws:acm:eu-central-1:1234567890:certificate/1234567890-4c63-4127-a745-57a773cf0866
      has not been exported so it is not eligible for the Renew Certificate API.
```

### How can we reproduce it?

1. Deploy Certificate: 

```
apiVersion: acm.aws.crossplane.io/v1alpha1
kind: Certificate
metadata:
  annotations:
    crossplane.io/composition-resource-name: aws-certificate
    crossplane.io/external-create-pending: "2021-11-03T15:49:52Z"
    crossplane.io/external-create-succeeded: "2021-11-03T15:49:53Z"
    crossplane.io/external-name: arn:aws:acm:eu-central-1:1234567890:certificate/1234567890-4c63-4127-a745-57a773cf0866
  creationTimestamp: "2021-11-03T15:49:52Z"
  finalizers:
  - finalizer.managedresource.crossplane.io
  generateName: front.mesh-pca-one.svc.cluster.local-4xb72-
  generation: 2
  labels:
    cnp.comp.db.de/cost-reference: F-330951-51-05
    cnp.com/owner: cnp
    crossplane.io/claim-name: front.mesh-pca-one.svc.cluster.local
    crossplane.io/claim-namespace: cnp-example
    crossplane.io/composite: front.mesh-pca-one.svc.cluster.local-4xb72
  name: front.mesh-pca-one.svc.cluster.local-4xb72
  ownerReferences:
  - apiVersion: cnp.com/v1alpha1
    controller: true
    kind: CompositeCertificate
    name: front.mesh-pca-one.svc.cluster.local-4xb72
    uid: 18d68909-62ec-4554-8d47-659818b413ba
  resourceVersion: "398699841"
  uid: b9c5e325-5d45-418b-8d0a-83d452405028
spec:
  deletionPolicy: Delete
  forProvider:
    certificateAuthorityARN: arn:aws:acm-pca:eu-central-1:1234567890:certificate-authority/0a0612e8-84de-46d0-9288-c77b876da9ab
    certificateTransparencyLoggingPreference: DISABLED
[...]

Issue #747, opened 2021-06-26
acm: renewalEligibility = INELIGIBLE -> reconciler update every second


haarchri:
### What happened?

the combination for renewCertificate = true and renewalEligibility = INELIGIBLE gets the reconciler in a status try update every second - looks that the code https://github.com/crossplane/provider-aws/blob/master/pkg/controller/acm/controller.go#L224-L237 is the problem **cr.Spec.ForProvider.RenewCertificate is never set to false** 

```
API Version:  acm.aws.crossplane.io/v1alpha1
Kind:         Certificate
[...]
Spec:
[...]
  For Provider:
    Renew Certificate:    true
[...]
Status:
  At Provider:
    Renewal Eligibility:  INELIGIBLE
```

```
2021-06-26T10:50:22.316+0200	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/certificate.acm.aws.crossplane.io", "request": "/poc-certificate", "uid": "c39b89de-9141-4732-beeb-f0c693569dff", "version": "867671", "external-name": "arn:aws:acm:eu-central-1:255932642927:certificate/d0dc7e4a-67d5-4a01-863c-21874921a6cc", "requeue-after": "2021-06-26T10:51:22.316+0200"}
2021-06-26T10:50:22.337+0200	DEBUG	provider-aws	Reconciling	{"controller": "managed/certificate.acm.aws.crossplane.io", "request": "/poc-certificate"}
2021-06-26T10:50:22.570+0200	DEBUG	provider-aws	External resource is up to date	{"controller": "managed/certificate.acm.aws.crossplane.io", "request": "/poc-certificate", "uid": "c39b89de-9141-4732-beeb-f0c693569dff", "version": "868975", "external-name": "arn:aws:acm:eu-central-1:255932642927:certificate/d0dc7e4a-67d5-4a01-863c-21874921a6cc", "requeue-after": "2021-06-26T10:51:22.570+0200"}
2021-06-26T10:50:22.769+0200	DEBUG	provider-aws	Reconciling	{"controller": "managed/certificate.acm.aws.crossplane.io", "request": "/poc-certificate"}
[...]

PR #995, opened 2021-12-09
Bump `Certificate`, `CertificateAuthority` and `CertificateAuthorityPermission` to v1beta1

muvaf:

There are some changes in their code like adding a tagger and some schema fixes but nothing breaking unless you are in the middle of certificate renewal where the value of `spec.forProvider.renewCertificate` is `true`, which is also buggy behavior as reported in #747 and #908 . So, whatever value is set there, it'll be dropped and no action will be taken.

Fixes https://github.com/crossplane/provider-aws/issues/747
Fixes https://github.com/crossplane/provider-aws/issues/908

I have:

Manually validated that `Certficate` and `CertificateAuthority` upgrade smoothly.


haarchri:
@muvaf the following: `remove renew certificate action since it is imperative and hard to get right in declarative fashion` looks like a breaking change after bump to v1beta1?
https://github.com/crossplane/provider-aws/issues/747
https://github.com/crossplane/provider-aws/issues/908

in general i tested the PR everything is working as expected for ACM

muvaf:
@haarchri It is change in functionality but I'm not sure if it's breaking because once it's converted, that field will be dropped from the CR and you won't be able to make it `true` to renew the certificate - whatever state it was supposed to be in, it stays same, which is `false` since supposedly we turn it back to `false` once we renew the certificate. Though it's worth adding a note.

muvaf:
Labeled as `breaking-change` so that we don't forget to mention it in release notes. @haarchri would you mind taking another look?

haarchri (approved):
checked acm create, delete it is working - @muvaf  thanks for adding breaking-change label

LGTM!

PR #996, opened 2021-12-10
Move all IAM resources to `iam` group and bump all of them to v1beta1

muvaf:
### Description of your changes

The ones already on `v1beta1` are kept as is with deprecation notice to not break users. All `v1alpha1` ones are moved to `iam` group from `identity` and `IAM` prefix is dropped from their kind name.

User impact of this change will be that users will still have the old CRs in their clusters but there won't be a controller reconciling them. Re-creating the exact same resource with new group and kind will be the migration path.

This PR involves a lot of surgery, I tried my best to keep commits contained to one action for easier review.

Fixes https://github.com/crossplane/provider-aws/issues/989

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

By following `cluster/UPGRADE.md` instructions that are added by this PR.

[contribution process]: https://git.io/fj2m9

PR #1024, opened 2021-12-18
Role - add iamrole shortname

smcavallo:

`IAMRole` was renamed to `Role` as part of https://github.com/crossplane/provider-aws/pull/996
When you `kubectl get Role` it will default to native kubernetes Role since Role kind exists in multiple apis -
```
kubectl get Role
NAME               CREATED AT
crossplane-admin   2021-03-03T21:51:54Z
crossplane-edit    2021-03-03T21:51:54Z
crossplane-view    2021-03-03T21:51:54Z
```
_Hey but I wanted IAM Roles!!!!!_
```
kubectl api-resources | grep -e " Role$"
roles                              iamrole            iam.aws.crossplane.io/v1beta1                                            false        Role
roles                                                 rbac.authorization.k8s.io/v1                                             true         Role
```

One current workaround is `kubectl get roles.iam.aws.crossplane.io` but this is verbose.

As a convenience this PR implements short names to allow `kubectl get iamrole`
This makes it much easier for end users to get iam roles without having to worry about the api.
Plus `iamrole` is much more distinct than `role`

Users with the older crds/apis can still `kubectl get IAMRole` or `kubectl get roles.identity.aws.crossplane.io`
In fact I don't believe the `iamrole` shortname will work until the older crds have been deleted so this should not break any existing functionality, only convenience for users who have upgraded.

Fixes #

I have:

* locally with kubectl commands

### DIFF
--- a/apis/iam/v1beta1/role_types.go
+++ b/apis/iam/v1beta1/role_types.go
@@ -114,7 +114,7 @@ type RoleStatus struct {
 // +kubebuilder:printcolumn:name="SYNCED",type="string",JSONPath=".status.conditions[?(@.type=='Synced')].status"
 // +kubebuilder:printcolumn:name="AGE",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,aws}
+// +kubebuilder:resource:scope=Cluster,categories={crossplane,managed,aws},shortName=iamrole
 type Role struct {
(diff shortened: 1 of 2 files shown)


haarchri (approved):
LGTM - tested it locally good approach

muvaf:
> In fact I don't believe the iamrole shortname will work until the older crds have been deleted so this should not break any existing functionality

Have you manually tested this scenario?

Another convenience for this problem is to use `kubectl get role.iam` instead of `kubectl get role.iam.aws.crossplane.io`, so we can document this instead of adding a short name that may collide.

PR #1076, opened 2022-01-20
feat(controller): add  flag to enable specified controllers

liubog2008:
Add --enable-controllers-pattern flag

Fixes #

I have:

```
$ ./_output/bin/linux_amd64/provider --enable-controllers-pattern="ec2.*|elbv2.*"
```

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -34,11 +34,12 @@ import (
 
 func main() {
 	var (
-		app            = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
-		debug          = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
-		syncInterval   = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
-		pollInterval   = app.Flag("poll", "Poll interval controls how often an individual resource should be checked for drift.").Default("1m").Duration()
-		leaderElection = app.Flag("leader-election", "Use leader election for the conroller manager.").Short('l').Default("false").OverrideDefaultFromEnvar("LEADER_ELECTION").Bool()
+		app                      = kingpin.New(filepath.Base(os.Args[0]), "AWS support for Crossplane.").DefaultEnvars()
+		debug                    = app.Flag("debug", "Run with debug logging.").Short('d').Bool()
+		syncInterval             = app.Flag("sync", "Sync interval controls how often all resources will be double checked for drift.").Short('s').Default("1h").Duration()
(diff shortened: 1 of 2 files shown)


muvaf:
Hi @liubog2008 ! Thanks for the contribution! What are the use cases enabled by this flag?

liubog2008:
> What are the use cases enabled by this flag?
We hope to only install CRDs that we needed.

muvaf:
> We hope to only install CRDs that we needed.

This PR will only disable controllers but all CRDs will still be installed by Crossplane package manager.

What are the problems you're facing with high number of CRDs and/or controllers?

liubog2008:
> This PR will only disable controllers but all CRDs will still be installed by Crossplane package manager.

We install provider-aws directly without package manager.

> What are the problems you're facing with high number of CRDs and/or controllers?

1. Too many CRDs leads to a big swagger.json of kube-apiserver. It may have some performance problems.
2. Not all services we needs in AWS. And disable some controllers can save resource.

muvaf:
There is a long discussion in https://github.com/crossplane/crossplane/issues/2649 related to performance in the context of Jet-based providers that can install over 700 CRDs. TLDR of the discussion is that the performance problems regarding OpenAPI have been fixed in upstream Kubernetes and released via patch releases. While selective installation or breaking up the providers into smaller ones were among the options, because the performance effect on the cluster is negligible, we didn't pursue them. FWIW, the fixed OpenAPI problems used to start to appear only after ~200 CRDs, which provider-aws doesn't have yet.

Regarding controllers, they open Go routines that just wait for input but nature of Go routines allow threads to be used by others so the only additional overhead is in memory [which is quite minimal](https://tpaschalis.github.io/goroutines-size/).

If your observations are different, please share so that we can discuss and implement this flag in all other controllers as well if need be  🙂

liubog2008:
I have read the crossplane/crossplane#2649 before I submit this PR. But I don't know whether and when we can upgrade our Kubernetes version.

And I think an option to choose which services can be enabled is reasonable. As a user, I don't want to care about features I don't used. 

And minimal CRDs help me in iam and rbac creation. I don't need to add more permissions when provider upgrades.

liubog2008:
Of course, performance is my biggest problem. But kubernetes version upgrade is more serious and hard to do.

iyuroch:
Are there any plans for resolving this issue?

negz:
I don't think we should move forward with this PR until we've aligned on a design that can apply to call providers. We don't want to be adding flags to one provider and not others.

We've previously investigated something like this and found several issues that aren't immediately obvious, including how to handle references from an enabled API group to a disabled API group, and how being able to enable and disable API groups interacts with the Crossplane package manager.

While I appreciate the attempt, I'm going to close this PR. Rest assured figuring out a way around the performance issues caused by many CRDs is a top priority for us.

Issue #813, opened 2021-08-23
Ability to create RDSInstance from snapshot


mzingarelli:
### What problem are you facing?

When creating a new `RDSInstance`, we would like to create the databases from an existing snapshot.
This issue relates #227 but takes it a step further by adding the ability to restore any database type from a snapshot (aswell as from an S3 bucket)

Something along the lines of:

```
spec:
  forProvider:
    fromBackup:
     snapShotIdentifier:
       Id: my-shapshot-id
//     s3: 
//       bucket: mybucket
//       prefix: prefix
//      role: ....
```
### How could Crossplane help solve your problem?

Without this feature, we need to manually `pg_dump` and `pg_restore` the database contents manually into a Crossplane managed database.

This feature would help us quickly spin up managed `RDSInstance` from snapshot

muvaf:
https://github.com/crossplane/provider-aws/pull/770 implements restoration from S3. I think after that's merged, it'd be trivial to add other snapshot sources like [PointInTime](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceToPointInTime.html) and [DBSnapshot](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBInstanceFromDBSnapshot.html).

@mzingarelli are these the ones that'd work for your use case?

jbhennin:
@muvaf we are very interested in being able to restore from a DBSnapshot and from PointInTime.    Note, that this should be supported for Cluster and Instance

danports:
Now that #770 has been merged, I think I'll take a crack at this one next, since I could use it in my environment too.

danports:
https://github.com/crossplane/provider-aws/pull/1087 has been merged, which addresses snapshot restores for DB instances. Point in time restores and DB cluster restores are still open issues.

jbhennin:
@danports do you know the open issue for DB Cluster and DBInstance?

haarchri:
follow up in #1085 and #1214

PR #1087, opened 2022-01-23
Restore RDS instances from database snapshots

danports:

Adds support for restoring RDS instances from database snapshots.

Partially addresses #813 (doesn't handle point in time restores).

I have:

- [x] Unit tests
- [x] In-cluster tests on AWS


pkg/clients/rds/rds.go
```diff
@@ -199,6 +201,55 @@ func GenerateRestoreDBInstanceInput(name, password string, p *v1beta1.RDSInstanc
 	return c
 }
 
+// GenerateRestoreDBInstanceFromSnapshotInput from RDSInstanceSpec
+func GenerateRestoreDBInstanceFromSnapshotInput(name, password string, p *v1beta1.RDSInstanceParameters) *rds.RestoreDBInstanceFromDBSnapshotInput {
```
danports, line 205: As it stands, the provider will never set the master password for instances restored from snapshots: We don't use `password` here because the underlying `RestoreDBInstanceFromDBSnapshot` AWS API call doesn't support it, and `GenerateModifyDBInstanceInput` disregards password changes. What's the best approach here? Should we just document this limitation or attempt to set the master password after the database is restored?

danports:
@chlunde Can you review?

jbhennin:
@danports assuming approvals are given, are you open to make the same change on DBCluster and DBInstance?

pkg/clients/rds/rds.go
chlunde: But the existing password works I assume. Then I'm not sure if this should be considered an issue specific to restoring, although I guess it's more likely that the user want a password change in this case.

I'm fine with documenting this via an issue, and then we can follow up that issue when there's interest and time. Hopefully we can find a general solution, instead of a workaround here.

apis/database/v1beta1/rdsinstance_types.go
```diff
@@ -127,13 +127,24 @@ type S3RestoreBackupConfiguration struct {
...
 	S3 *S3RestoreBackupConfiguration `json:"s3,omitempty"`
 
-	// Source is the type of the backup to restore when creating a new RDS instance. Only S3 is supported at present.
+	// Snapshot specifies the details of the database snapshot to restore from.
+	// +optional
+	Snapshot *SnapshotRestoreBackupConfiguration `json:"snapshot,omitempty"`
+
+	// Source is the type of the backup to restore when creating a new RDS instance.
+	// Only S3 and Snapshot are supported at present.
 	Source *string `json:"source"`
```
chlunde, line 148:
```
// +kubebuilder:validation:Enum=S3;Snapshot
```
could we add this here? I tried `"snapshot"` on my first attempt, I think it would be more helpful if that did not validate.

chlunde:
@danports tested, works great!

apis/database/v1beta1/rdsinstance_types.go
danports: Absolutely - I was trying to figure out how to validate enums, since using the OpenAPI enum syntax in the CRD didn't work.

pkg/clients/rds/rds.go
danports: Agreed. Documented in #1121.

chlunde (approved):

chlunde:
Thanks!

PR #1106, opened 2022-01-29
fix(transfer/user): fix WithInitializers and removed postCreate

### LINKED ISSUE #1102: Transfer User can not be created: Missing username
### What happened?

The username was set to use the external name annotation in #1053. However, the annotation is only set after the successful creation of the user and no available in the preCreate hook. 

https://github.com/crossplane/provider-aws/blob/master/pkg/controller/transfer/user/setup.go#L98
https://github.com/crossplane/provider-aws/blob/master/pkg/controller/transfer/user/setup.go#L91

### How can we reproduce it?

Apply the example resource https://github.com/crossplane/provider-aws/blob/master/examples/transfer/user.yaml

provider-aws:v0.23

haarchri:

set correct `managed.WithInitializers` missed a local commit in #1053 sorry

Fixes #1102

I have:

```
kubectl get users.transfer
NAME      READY   SYNCED   EXTERNAL-NAME
example   True    True     example
```

### DIFF
--- a/pkg/controller/transfer/user/setup.go
+++ b/pkg/controller/transfer/user/setup.go
@@ -41,7 +41,6 @@ func SetupUser(mgr ctrl.Manager, l logging.Logger, rl workqueue.RateLimiter, pol
 	opts := []option{
 		func(e *external) {
 			e.postObserve = postObserve
-			e.postCreate = postCreate
 			e.preObserve = preObserve
 			e.preDelete = preDelete
 			e.preCreate = preCreate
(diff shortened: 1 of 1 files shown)


pkg/controller/transfer/user/setup.go
```diff
@@ -84,14 +83,6 @@ func postObserve(_ context.Context, cr *svcapitypes.User, obj *svcsdk.DescribeUs
...
-	if err != nil {
-		return managed.ExternalCreation{}, err
-	}
-	meta.SetExternalName(cr, awsclients.StringValue(obj.UserName))
-	return managed.ExternalCreation{}, nil
-}
-
 func preCreate(_ context.Context, cr *svcapitypes.User, obj *svcsdk.CreateUserInput) error {
 	obj.ServerId = cr.Spec.ForProvider.ServerID
 	obj.Role = cr.Spec.ForProvider.Role
```
muvaf, line 88: I wonder how that worked out before when nothing set the external name annotation

muvaf (approved):
Thanks!

PR #1123, opened 2022-02-07
Bump SNS Topic and Subscription versions to v1beta1

### LINKED ISSUE #1109: SNS Topic and Subscription to be generated and bumped to v1beta1

We have `SNSTopic` and `SNSSubscription` resources in `notification` group and they are manually implemented. We'd like to bump them to v1beta1, hence they need to be renamed https://github.com/crossplane/provider-aws/issues/340 and we should port them to be generated by copying the parts that are still relative after moving to code generation pipeline

Implement these two resources using ACK and copy the special logic parts. I think we should mark existing `SNSTopic` and `SNSSubscription` resources as deprecated with kubebuilder tags but not take action for now since we're too close to the next release and there might be schema differences which would make it hard to migrate for people.

ezgidemirel:
Signed-off-by: ezgidemirel <ezgidemirel91@gmail.com>

This PR bumps SNS Topic and Subscription resources to v1beta1. 

Fixes #1109

I have:

I created/deleted both v1alpha1 and v1beta1 resources under `examples` folder. 


muvaf (commented):
Thanks @ezgidemirel ! We talked offline about how ACK is not able to generate those resources. Could you check whether the schema of `spec.forProvider` looks same as if it was generated by ACK? i.e. the fields of `SubscribeWithContext` and `CreateTopicWithContext` calls. Otherwise, we'd have to make a schema change once ACK is able to generate it.

ezgidemirel:
@muvaf I checked the resources, and looks like we're compatible with ACK's naming convention.

muvaf (approved):
LGTM!

PR #1148, opened 2022-02-14
fix: Avoid AssumeRoleWithWebIdentity for each reconcile

### LINKED ISSUE #828: Excessive calls to AssumeRoleWithWebIdentity w/ IRSA
There are too many calls to sts for AssumeRoleWithWebIdentity from crossplane, which results in a large number of audit events in Cloud Trail. Additionally, HTTP connection reuse/keep-alive is not used, which also results in a large number of flow log entries and presumably consumes CPU for TLS handshakes. 

Reuse HTTP clients and SDK clients/credentials for STS. This work depends on the SDK upgrades. Later SDK versions contains logic for fetching new credentials before expiry.

chlunde:

Use a shared client/credentials to avoid calling AssumeRoleWithWebIdentity for each reconcile.

aws.Config (v1/v2) is thread safe and checks token expiry/handles renewal. To handle 

Fixes #828

I have:

Ran for one week. Creating PR for discussion, how can we avoid a global variable?

As you can see below we get a significant reduction in the number of API calls to AWS:

image

### DIFF
--- a/pkg/utils/connect/aws/config.go
+++ b/pkg/utils/connect/aws/config.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
@@ -31,6 +32,7 @@ import (
 	stscredstypesv2 "github.com/aws/aws-sdk-go-v2/service/sts/types"
 	awsv1 "github.com/aws/aws-sdk-go/aws"
 	credentialsv1 "github.com/aws/aws-sdk-go/aws/credentials"
+	stscredsv1 "github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	endpointsv1 "github.com/aws/aws-sdk-go/aws/endpoints"
 	requestv1 "github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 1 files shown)


haarchri:
did we need to adjust also something for `assumeRoleARN`?

chlunde:
@haarchri yeah, I _think_ we can refactor it to share a bit more code, but first I wonder if there's a place to store this struct in a "less global" variable.

chlunde:
@muvaf do you have any idea where we could store this state? Is a global variable like this OK if there are no other options? TAs you can see this should give a performance boost and cost reduction if all AWS security products are enabled (Detective/GuardDuty/CloudTrail).

haarchri:
tested your PR today in our environment looks really great - if we adopt this after we clearify to save the state for `assumeRoleARN` would be perfect ;)

muvaf:
I think this is a very good idea if we have the tools to get the caching right. A few questions:
* Does the token in the config ever expire? How do we know that so that we can request a new one without returning auth error when we shouldn't?
* Are we certain that there can ever be a single identity for a single pod, hence we don't need to use a `map`? Not really important today, but just checking.

> Is a global variable like this OK if there are no other options?

I think that's OK.

haarchri:
@chlunde  any chance to finalize this PR ? i am happy to test this in our environment =)

atarax:
I rebased changes in this pr and we're running this version now as an experiment. So far no issues and we're seing roughly a 65% drop in AssumeRole requests: 
image

tkaesserfm:
Version was running in one of our cluster for almost 10 days, looking great. Token is refreshed properly, everything works, no restarts.

@chlunde @haarchri @MisterMX What would be options to proceed here? The change would save a lot some bucks for everyone operating provider-aws on a bigger scale like we do.

I could make a fresh PR with rebased changes and even work on finalizing the PR but i don't want to steal @chlunde the show :yum:

chlunde:
We've been running with this code for 1.5 years without issues, but we only use one provider config per provider. If someone with more complex setups, assuming roles in different accounts, also could verify it would be nice.

MisterMX:
I made a little improvement by using two separate `sync.Mutex` for V1 and V2 so they can be accessed in parallel.

MisterMX (approved):
LGTM. Thank you very much @chlunde!

PR #1161, opened 2022-02-21
fix(s3): Extract QueueARN instead of external name

### LINKED ISSUE #1160: S3 Bucket NotificationConfiguration Queue referencers should extract ARN instead of external name
### What happened?

The current referencer implementation for `spec.forProvider.noficationConfiguration.queueConfigurations[].queueArn` in `Bucket` does extract the value of the `crossplane.io/external-name` annotation but the field requires the ARN from `status.atProvider.queueArn`.

### How can we reproduce it?

1. Create a queue from `examples`
2. Create a bucket from `examples` with `spec.forProvider.noficationConfiguration` set to the following:
```yaml
        queueArnRef:
          name: test-queue3
```
3. The bucket creation fails because the field does not contain a valid ARN.

`provider-aws: v0.24.1`

MisterMX:

This changes the referencer implementation for `spec.forProvider.noficationConfiguration.queueConfigurations[].queueArn` for `Bucket` so it will extract the ARN from the `Queue` resource instead of the default external name annotation.

Fixes #1160 

It will also make the field optional, because it can be filled using referencers.

Additionally, I updated the put bucket notification configuration error message to give a
hint about missing IAM credentials. (Because the AWS error message is incomplete and might be misleading).

I have:

Manually using the following YAML:

```yaml
---
kind: Bucket
metadata:
  name: repl-dest
  annotations:
    # This will be the actual bucket name. It must be globally unique, so you
    # probably want to change it before trying to apply this example.
    crossplane.io/external-name: crossplane-example-repl-dest-123455
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: eu-central-1
    notificationConfiguration:
      queueConfigurations:
      - events:
        - "s3:ObjectCreated:*"
        queueArnRef:
          name: test-queue
    paymentConfiguration:
      payer: BucketOwner
      rules:
            sseAlgorithm: AES256
    versioningConfiguration:
      status: Enabled
  providerConfigRef:
    name: example
---
apiVersion: sqs.aws.crossplane.io/v1beta1
kind: Queue
metadata:
  name: test-queue
spec:
  forProvider:
    region: eu-central-1
    delaySeconds: 4
    policy: |
      {
        "Statement": [
          {
            "Sid":"allow-s3",
            "Effect":"Allow",
            "Principal": {
              "Service": "s3.amazonaws.com"
            },
            "Action":"sqs:SendMessage",
            "Resource":"*",
            "Condition":{
              "ArnEquals":{
                "aws:SourceArn":"arn:aws:s3:::*"
              }
            }
          }
        ]
      }
  providerConfigRef:
    name: example
```

### DIFF
--- a/apis/s3/v1beta1/notificationConfiguration_types.go
+++ b/apis/s3/v1beta1/notificationConfiguration_types.go
@@ -136,7 +136,8 @@ type QueueConfiguration struct {
 	//
 	// QueueArn is a required field
 	// +crossplane:generate:reference:type=github.com/crossplane/provider-aws/apis/sqs/v1beta1.Queue
-	QueueArn string `json:"queueArn"`
+	// +crossplane:generate:reference:extractor=github.com/crossplane/provider-aws/apis/sqs/v1beta1.QueueARN()
+	QueueArn *string `json:"queueArn,omitempty"`
 
 	// QueueArnRef references an Queue to retrieve its ARN
 	// +optional
--- a/apis/s3/v1beta1/zz_generated.deepcopy.go
+++ b/apis/s3/v1beta1/zz_generated.deepcopy.go
@@ -926,6 +926,11 @@ func (in *QueueConfiguration) DeepCopyInto(out *QueueConfiguration) {
 		*out = new(string)
 		**out = **in
 	}
+	if in.QueueArn != nil {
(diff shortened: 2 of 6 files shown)


haarchri (approved):
thanks for implementing this

PR #1187, opened 2022-03-02
Update to the latest crossplane-runtime

muvaf:

The first commit is mostly non-impactful but the second one changes how `Setup` functions are designated to the new approach.

@negz could you take a look at the rate limiting stuff to see if I got them right?

I have:

`make reviewable`


Makefile
```diff
@@ -75,14 +75,7 @@ cobertura:
 		grep -v zz_generated.deepcopy | \
 		$(GOCOVER_COBERTURA) > $(GO_TEST_OUTPUT)/cobertura-coverage.xml
 
-crds.clean:
```
haarchri, line 78: why we remove crds.clean ?

haarchri (approved):
@muvaf added one comment - did we need special release notes?

Makefile
muvaf: With the new version of controller-tools, it's not necessary anymore.

muvaf:
> did we need special release notes?

I don't think so but the new `--max-reconcile-rate` flag with a default of `10` will probably affect users with thousands of MRs. @negz do you expect the effect to be big enough that we need to warn people?

PR #1191, opened 2022-03-07
First immutability webhook

muvaf:

The first webhook for `VPC` to show how the bare webhook support can be implemented in a provider. With https://github.com/crossplane/crossplane-runtime/issues/318 , the actual implementation will probably change except the assumptions around where to find the TLS cert files.

Built on https://github.com/crossplane/provider-aws/pull/1187 , only the last commit is new.

Depends on https://github.com/crossplane/crossplane/pull/2919

I have:

Manually with https://github.com/crossplane/crossplane/pull/2919

```
admission webhook "vpcs.ec2.aws.crossplane.io" denied the request: spec.forProvider.region is immutable
```


chlunde:
Nice, do I smell another opportunity for codegen here? :)

I wonder if we should only "freeze" once ready or synced? I imagine someone attempting to create a resource, fail due to an incorrect parameter, and then being unable to do it due to the field being immutable.

PR #1214, opened 2022-03-16
feat(rdsinstance): Support for point in time restore

petteja:
This adds support for point in time restore for RDS Instances

This functionality is mentioned in #813

I have:

This code has been tested by adding unit tests and manually tested against AWS using Kind running provider-aws locally out-of-cluster.


pkg/controller/database/rdsinstance.go
```diff
@@ -196,6 +198,14 @@ func (e *external) RestoreOrCreate(ctx context.Context, cr *v1beta1.RDSInstance,
 		if err != nil {
 			return awsclient.Wrap(err, errSnapshotRestoreFailed)
 		}
+	case "PointInTime":
+		if cr.Spec.ForProvider.RestoreFrom.PointInTime.SourceDBInstanceIdentifier == nil && cr.Spec.ForProvider.RestoreFrom.PointInTime.SourceDbiResourceID == nil && cr.Spec.ForProvider.RestoreFrom.PointInTime.SourceDBInstanceAutomatedBackupsArn == nil {
+			return errors.New(errPointInTimeRestoreSourceNotDefined)
+		}
```
petteja, line 204: I don't know if it is better to give this custom error message early, or send the request to the AWS API and return that error.

haarchri: i would prefer early

haarchri:
restored point in time for RDS successful - LGTM 
@petteja would you change your comment code block and return early ?

petteja:
> 

Thanks for reviewing this. It is returning early, or did I miss something?

haarchri (approved):
@petteja  thanks not checked yesterday in deep detail - because it sounds like a question ;) LGTM

PR #1234, opened 2022-03-29
Added default tags to iam.User

### LINKED ISSUE #1232: Tag changes not reflected in iam.User
### What happened?

Changing tags (both keys or values) is not reconciled with the AWS User.

### How can we reproduce it?

In a cluster where crossplane AWS is running, create a User (for example, using the yaml included in this repo: `kubectl apply -f examples/iam/user.yaml `)

With the manifest above, the user is created with a tag: "k1":"v1".
If I modify the tag described in the manifest to "k1":"v2" the change is not reconciled with the AWS User just created. 

Crossplane version: 26d6e08aed58066205836d8b82977f652b426542

### LINKED ISSUE #1233: Add default Crossplane tags to iam.User

When I create a `iam.User`, the default Crossplane tags (`crossplane-kind`, `crossplane-name`, `crossplane-providerconfig`) are not applied.

Adding default tags management to iam.User

cebernardi:
Signed-off-by: Cecilia Bernardi <cbernardi@expediagroup.com>

Fixes #1232
Fixes #1233

I have:

- Unit Tests
- Manual Tests (in a kind cluster configured with an AWS account)
    - Creation scenario: a `iam.User` was created with a "k1":"v1" tag. All the default tags were added and Crossplane created the User successfully, reconciling all the 4 tags (3 default tags + 1 custom tag)
    - Update scenario: the same `iam.User` created above was patched to:
        - change the custom tag from "k1":"v1" to "k1":"v2" (the value change was correctly applied to the AWS resource)
        - change the custom tag from "k1":"v1" to "k2":"v1" (a new tag was added to the AWS resource, and the previous one was removed)
        - add a new "k3":"v3" tag. The AWS resource was tagged with the new tag
        - remove the tag with key "k2". The AWS resource was untagged
        - change the default tag "crossplane-name": "someuser" to "crossplane-name":"someuser2", the controller reverted the change and no changes were applied to the AWS resource
        - remove the default tag "crossplane-name". The controller reverted the change and no changes were applied to the AWS resource

### DIFF
--- a/pkg/controller/iam/user/controller.go
+++ b/pkg/controller/iam/user/controller.go
@@ -23,6 +23,7 @@ import (
 	awsiam "github.com/aws/aws-sdk-go-v2/service/iam"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -49,6 +50,8 @@ const (
 	errDelete = "cannot delete the IAM User resource"
 	errUpdate = "cannot update the IAM User resource"
 	errSDK    = "empty IAM User received from IAM API"
+	errTag    = "cannot tag the IAM User resource"
+	errUntag  = "cannot remove tags from the IAM User resource"
 
 	errKubeUpdateFailed = "cannot late initialize IAM User"
 )
@@ -69,6 +72,9 @@ func SetupUser(mgr ctrl.Manager, o controller.Options) error {
 		Complete(managed.NewReconciler(mgr,
 			resource.ManagedKind(v1beta1.UserGroupVersionKind),
 			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewUserClient}),
+			managed.WithInitializers(
+				managed.NewNameAsExternalName(mgr.GetClient()),
+				&tagger{kube: mgr.GetClient()}),
 			managed.WithConnectionPublishers(),
 			managed.WithPollInterval(o.PollInterval),
 			managed.WithLogger(o.Logger.WithValues("controller", name)),
@@ -128,9 +134,16 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 		UserID: aws.ToString(user.UserId),
 	}
 
+	crTagMap := make(map[string]string, len(cr.Spec.ForProvider.Tags))
+	for _, v := range cr.Spec.ForProvider.Tags {
+		crTagMap[v.Key] = v.Value
+	}
+	_, _, areTagsUpdated := iam.DiffIAMTags(crTagMap, observed.User.Tags)
+
 	return managed.ExternalObservation{
-		ResourceExists:   true,
-		ResourceUpToDate: aws.ToString(cr.Spec.ForProvider.Path) == aws.ToString(user.Path),
+		ResourceExists: true,
+		ResourceUpToDate: aws.ToString(cr.Spec.ForProvider.Path) == aws.ToString(user.Path) &&
+			areTagsUpdated,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


cebernardi:
e2e tests are failing, maybe because of this? https://github.com/crossplane/crossplane/issues/2944

pkg/controller/iam/user/controller.go
```diff
@@ -69,6 +72,9 @@ func SetupUser(mgr ctrl.Manager, o controller.Options) error {
 		Complete(managed.NewReconciler(mgr,
 			resource.ManagedKind(v1beta1.UserGroupVersionKind),
 			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), newClientFn: iam.NewUserClient}),
+			managed.WithInitializers(
+				&externalNamer{kube: mgr.GetClient()},
```
muvaf, line 76: The default initializer list includes an external name initializer from runtime, so you can use that one here as is. See an example [here](https://github.com/crossplane/provider-aws/blob/793e588/pkg/controller/ecr/repository/controller.go#L81).

pkg/controller/iam/user/controller.go
```diff
@@ -179,3 +224,56 @@ func (e *external) Delete(ctx context.Context, mgd resource.Managed) error {
...
+func (t *tagger) Initialize(ctx context.Context, mgd resource.Managed) error {
+	cr, ok := mgd.(*v1beta1.User)
+	if !ok {
+		return errors.New(errUnexpectedObject)
+	}
+
+	added := false
+	defaultTags := resource.GetExternalTags(mgd)
+
+	for i, t := range cr.Spec.ForProvider.Tags {
```
muvaf, line 245: nitpick: Implementation [here](https://github.com/crossplane/provider-aws/blob/793e588/pkg/controller/ecr/repository/controller.go#L257) you can copy seems to be doing the same thing maybe with fewer lines but both should work well.

pkg/controller/iam/user/controller.go
```diff
@@ -162,6 +175,38 @@ func (e *external) Update(ctx context.Context, mgd resource.Managed) (managed.Ex
...
+		UserName: aws.String(meta.GetExternalName(cr)),
+	})
+
+	if err != nil {
+		return managed.ExternalUpdate{}, awsclient.Wrap(err, errGet)
+	}
+
+	add, remove, _ := iam.DiffIAMTagsWithUpdates(cr.Spec.ForProvider.Tags, observed.User.Tags)
+
+	if len(add) > 0 {
```
muvaf, line 192: Is there a possibility that both `remove` and `add` have the same entry since it needs to be updated so we end up first updating it here and then remove it in the next `if` block?

pkg/controller/iam/user/controller.go
cebernardi: ah, thank you, that's so much better, done.

pkg/controller/iam/user/controller.go
cebernardi: I had a look at the suggested code, and I found the following divergences:
- it doesn't handle tags with empty values
- it allows users to change default tags values (actually appending two Tag elements with the same key but different values to Spec.ForProvider.Tags, guessing the last one will be "winning")

is this the intended behavior?

pkg/controller/iam/user/controller.go
cebernardi: not with `DiffIAMTagsWithUpdates`. 

The use case you're mentioning is covered by [this test](https://github.com/crossplane/provider-aws/blob/master/pkg/clients/iam/tags_test.go#L163), and more generically [here](https://github.com/cebernardi/provider-aws/blob/add-default-tags-iam-user/pkg/controller/iam/user/controller_test.go#L452).

haarchri (approved):
tested the tags for iam.User is working - thanks for implementation
@muvaf did we need more code changes as you requested some ?

haarchri:
@cebernardi thanks for implementation can you rebase and check for conflicts ? After that we can merge your branch

cebernardi:
sure @haarchri , done!

haarchri:
@cebernardi  can you Check https://github.com/crossplane-contrib/provider-aws/issues/1352 ? Looks like we have one problem  that in every reconcile Loop more and more Tags added

cebernardi:
@haarchri I provisionally opened #1355 (draft), tomorrow I'll rebase it

cebernardi:
@haarchri the PR is rebased and ready to review

PR #1240, opened 2022-03-30
Add support for IAM AssumeRole tags

### LINKED ISSUE #1212: Enhance the assumeRoleARN authentication to apply session tags in AWS STS
We are creating a provider config per namespace and are trying to create a scalable way to provide the correct amount of access to the aws-provider for each configuration.  The process we use outside of crossplane is to have an IAM user (could be a role) in one account and it can assume some default roles that are in all of our other accounts.  Those roles look at session tags to make sure that what you are deploying/changing have the same tag on the resource you are creating or deploying.

If crossplane (aws-provider) was able to pass tags to the STS session, we would be able to use our standard roles in all of the accounts that we have.  It would make it so much easier to scale.

stevendborrelli:
Signed-off-by: Steven Borrelli <steve@borrelli.org>

Fixes #1212

This PR adds the ability to add session tags and transitive tag keys (see <https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html>) to the IAM Assume Role functionality settings in the ProviderConfig.

The top-level settings for `assumeRoleARN` and `axternalID` have been marked as deprecated in favor of putting all options into the `assumeRole` stanza, as per the example below:

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: aws-provider-account-b
spec:
  assumeRole:
    roleARN: "arn:aws:iam::999999999999:role/account_b"
    externalID: "my-optional-id"
    tags:
      - key: Project
        value: Crossplane
      - key: Department
        value: Infrastructure   
    transitiveTagKeys: [ "Project", "Department"]
  credentials:
    source: Secret
    secretRef:
      namespace: crossplane-system
      name: aws-creds
      key: creds
```

I have:


brianmjones:
this looks great!!

brianmjones (approved):

haarchri (approved):
@stevendborrelli thanks for this enhancement ;) cool to see more features after the initial implementation of https://github.com/crossplane/provider-aws/pull/912. 

tested it is working in our environment - can see the session tags

PR #1242, opened 2022-03-31
Add External Secret Stores support as an alpha feature

turkenh:

This PR adds External Secret Stores support as an alpha feature.

I have:

Tested by creating an IAM AccessKey and ensuring connection secrets land in Vault. Similarly, validated the cleanup with the deletion of the resource.

### DIFF
--- a/CODE_GENERATION.md
+++ b/CODE_GENERATION.md
@@ -103,6 +103,12 @@ add the setup function like the following:
 // SetupStage adds a controller that reconciles Stage.
 func SetupStage(mgr ctrl.Manager, o controller.Options) error {
 	name := managed.ControllerName(svcapitypes.StageGroupKind)
+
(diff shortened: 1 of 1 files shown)


haarchri (approved):
@turkenh  tested the implementation locally LGTM 
can we have a small part in documentation to add `External Secret Stores support` for new resources here https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md and can we have a documentation for the future flag ?

turkenh:
> @turkenh tested the implementation locally LGTM can we have a small part in documentation to add `External Secret Stores support` for new resources here https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md and can we have a documentation for the future flag ?

Thanks @haarchri for testing.
I've updated the code-gen documentation, good catch!

For documenting the feature flag, I am a bit hesitant to add per provider documentation for a common feature and would prefer to have it in one place, e.g. https://crossplane.io/docs/v1.7/guides/vault-as-secret-store.html

PR #1250, opened 2022-04-06
linter: make sure it uses the version given in GH action file

muvaf:

Otherwise it uses go1.18 which doesn't work for us yet. Here is a failed linter job https://github.com/crossplane/provider-aws/runs/5859214191?check_suite_focus=true

I have:

Through GH PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,11 @@ jobs:
         with:
           submodules: true
 
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
       - name: Find the Go Build Cache
         id: go
         run: echo "::set-output name=cache::$(make go.cachedir)"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -73,6 +73,9 @@ jobs:
       # it comments on PRs to point out linter violations).
       - name: Lint
         uses: golangci/golangci-lint-action@v2
+        strategy:
+          matrix:
+            go-version: [1.17.x]
```
turkenh, line 78: It looks like [this](https://github.com/golangci/golangci-lint-action/issues/434) is the actual problem and this is the third solution I am seeing for the same problem :)
Did you consider those as well, asking to make sure  that we are aligned as much as possible.

- Skip go installation: https://github.com/crossplane/provider-azure/blob/afeda3e320643512c266f9bce161a50052d5c0f2/.github/workflows/ci.yml#L78
- Upgrade github action to v3: https://github.com/oam-dev/kubevela/pull/3571/files

turkenh (approved):

haarchri:
Thanks for this PR

PR #1255, opened 2022-04-07
Handle NoSuchHostedZone errors on ResourceRecordSet delete

### LINKED ISSUE #1251: Delete of ResourceRecordSet fails when the HostedZone is deleted first
### What happened?
I have a composition that creates a HostedZone and a ResourceRecordSet.  They both get created fine, but on delete, if the HostedZone happens to get deleted first, the ResourceRecordSet will fail to delete with an error that the HostedZone does not exist:

```
  status:
    conditions:
    - lastTransitionTime: "2022-04-06T22:55:57Z"
      message: 'observe failed: failed to list the ResourceRecordSet resource: NoSuchHostedZone:
        The specified hosted zone does not exist.'
      reason: ReconcileError
      status: "False"
      type: Synced
    - lastTransitionTime: "2022-04-06T22:55:56Z"
      reason: Deleting
      status: "False"
      type: Ready
```

### How can we reproduce it?
Create a HostedZone and a ResourceRecordSet in a composition,, and when the claim is deleted observe that the ResourceRecordSet fails to delete with the above error

Crossplane version: 1.7
provider-aws version 0.25

* Cloud provider or hardware configuration AWS

The terraform AWS provider handles this case explicitly for the resource, essentially ignoring the r53NoHostedZoneFound error and completing the deletion process.

bobh66:
Signed-off-by: Bob Haddleton <bob.haddleton@nokia.com>

Updated the ResourceRecordSet controller to handle the case where the HostedZone has been deleted before the ResourceRecordSet, to prevent the ResourceRecordSet from being permanently stuck.

If the HostedZone does not exist then the ResourceRecordSet cannot exist and so this case can be handled as equivalent to a "NotFound".

Fixes #1251 

I have:

Tested in a local Kind environment.  Reproduced the problem locally, then applied the changes and verified that the ResourceRecordSet gets deleted when the HostedZone has already been deleted.

### DIFF
--- a/pkg/clients/resourcerecordset/resourcerecordset.go
+++ b/pkg/clients/resourcerecordset/resourcerecordset.go
@@ -29,6 +29,8 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
+	"github.com/crossplane/provider-aws/pkg/clients/hostedzone"
+
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 
 	"github.com/crossplane/provider-aws/apis/route53/v1alpha1"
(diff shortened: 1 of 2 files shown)


haarchri (approved):
thanks for this fix

PR #1261, opened 2022-04-14
Store CallerReference value for import scenarios

### LINKED ISSUE #1260: Cannot update imported Cloudfront Distribution resources
### What happened?

After importing an existing Cloudfront distribution resource, trying to updating its fields fails with following error:
```
  Warning  CannotUpdateExternalResource  1s (x5 over 14s)  managed/distribution.cloudfront.aws.crossplane.io  cannot update Distribution in AWS: IllegalUpdate: You cannot update Origin or CallerReference.
```

### How can we reproduce it?

- Create a Cloudfront Distribution from AWS console
- Import it with following manifest
```
apiVersion: cloudfront.aws.crossplane.io/v1alpha1
kind: Distribution
metadata:
  annotations:
    crossplane.io/external-name: <external-name-of-the-distribution>
  name: example-distribution
spec:
  forProvider:
    region: us-east-1
    distributionConfig: {}
```
- Update `spec.forProvider.distributionConfig.comment`

Crossplane version: v1.6.3-up.1
provider-aws version: v0.25.0

ezgidemirel:
Signed-off-by: ezgidemirel <ezgidemirel91@gmail.com>

This change lateInitialize the "CallerReference" value for import scenarios. 

Fixes #1260 

I have:

- Created a CF distribution resource with `example.yaml`
- Changed `spec.deletionPolicy` to `Orphan` and deleted the managed resource
- Imported the same resource with following manifest 
```
apiVersion: cloudfront.aws.crossplane.io/v1alpha1
kind: Distribution
metadata:
  annotations:
    crossplane.io/external-name: <external-name>
  name: ezgi
spec:
  forProvider:
    region: us-east-1
    distributionConfig: {}
```
- Updated `spec.forProvider.distributionConfig.comment`
- Applied the same steps with `callerReference` field set 

### DIFF
--- a/pkg/controller/cloudfront/distribution/lateinit.go
+++ b/pkg/controller/cloudfront/distribution/lateinit.go
@@ -48,6 +48,10 @@ func lateInitialize(dp *svcapitypes.DistributionParameters, gdo *svcsdk.GetDistr
 	in := dp.DistributionConfig
 	from := gdo.Distribution.DistributionConfig
 
+	// "CallerReference" is not exposed to users from the console, therefore we need to derive it from the AWS output
+	// during import scenarios
+	in.CallerReference = awsclients.LateInitializeStringPtr(in.CallerReference, from.CallerReference)
+
 	if from.Aliases != nil {
 		if in.Aliases == nil {
 			in.Aliases = &svcapitypes.Aliases{}
--- a/pkg/controller/cloudfront/distribution/setup.go
+++ b/pkg/controller/cloudfront/distribution/setup.go
@@ -85,151 +85,156 @@ func SetupDistribution(mgr ctrl.Manager, o controller.Options) error {
 }
 
 func preCreate(_ context.Context, cr *svcapitypes.Distribution, cdi *svcsdk.CreateDistributionInput) error {
-	cdi.DistributionConfig.CallerReference = awsclients.String(string(cr.UID))
+	if awsclients.StringValue(cr.Spec.ForProvider.DistributionConfig.CallerReference) != "" {
+		cdi.DistributionConfig.CallerReference = cr.Spec.ForProvider.DistributionConfig.CallerReference
+	} else {
+		cdi.DistributionConfig.CallerReference = awsclients.String(string(cr.UID))
+	}
+
 	// if cr.Spec.ForProvider.DistributionConfig.Origins is not nil then cdi.DistributionConfig.Origins is not nil
 	if cr.Spec.ForProvider.DistributionConfig.Origins != nil {
 		cdi.DistributionConfig.Origins.Quantity =
-			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Origins.Items))
+			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Origins.Items), 0)
 	}
 
 	if cr.Spec.ForProvider.DistributionConfig.Aliases != nil {
 		cdi.DistributionConfig.Aliases.Quantity =
-			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Aliases.Items))
+			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Aliases.Items), 0)
 	}
 
 	if cr.Spec.ForProvider.DistributionConfig.CustomErrorResponses != nil {
 		cdi.DistributionConfig.CustomErrorResponses.Quantity =
-			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.CustomErrorResponses.Items))
+			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.CustomErrorResponses.Items), 0)
 	}
 
 	if cr.Spec.ForProvider.DistributionConfig.Restrictions != nil && cr.Spec.ForProvider.DistributionConfig.Restrictions.GeoRestriction != nil {
 		cdi.DistributionConfig.Restrictions.GeoRestriction.Quantity =
-			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Restrictions.GeoRestriction.Items))
+			awsclients.Int64(len(cr.Spec.ForProvider.DistributionConfig.Restrictions.GeoRestriction.Items), 0)
 	}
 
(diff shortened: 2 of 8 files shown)


apis/cloudfront/v1alpha1/custom_types.go
```diff
@@ -17,7 +17,9 @@ limitations under the License.
 package v1alpha1
 
 // CustomDistributionParameters includes the custom fields of Distribution.
-type CustomDistributionParameters struct{}
+type CustomDistributionParameters struct {
+	CallerReference *string `json:"callerReference,omitempty"`
```
ulucinar, line 21: Should we add some documentation stating that this field is for internal use and appropriately defaulted and the `+kubebuilder:validation:Optional` marker for this field?

M understanding is that both for provisioning & import scenarios, users of this API will never have to supply a value, we are adding this to be able to support import scenarios. This could also be only part of the status.

pkg/controller/cloudfront/distribution/setup.go
```diff
@@ -323,9 +323,13 @@ func postUpdate(_ context.Context, cr *svcapitypes.Distribution, resp *svcsdk.Up
 func preUpdate(_ context.Context, cr *svcapitypes.Distribution, udi *svcsdk.UpdateDistributionInput) error {
 	udi.Id = awsclients.String(meta.GetExternalName(cr))
 	udi.SetIfMatch(awsclients.StringValue(cr.Status.AtProvider.ETag))
-	udi.DistributionConfig.CallerReference = awsclients.String(string(cr.UID))
+	if *cr.Spec.ForProvider.CallerReference != "" { // import scenario
```
ulucinar, line 326: One point to note is that because we are adding this logic to only `preUpdate`, we are not supporting the use case of explicitly specifying a caller reference by the user of the API, i.e., for newly provisioned CloudFront Distributions, caller ref. will always be the `uid`, although we have exposed an (internal) API for it. I think this is not an important use case (as long as, for instance, AWS decides to impose a (backward-incompatible) format restriction on the caller ref. string). But I think, we should at least provide field documentation mentioning this. Is it possible to introduce this field only for the `status` kind?

apis/cloudfront/v1alpha1/custom_types.go
ezgidemirel: I added the kubebuilder marker. As discussed offline, if `callerReference` field is set, I use it during `preCreate`.

pkg/controller/cloudfront/distribution/setup.go
ezgidemirel: I updated `preCreate` function to use specified `callerReference` if exists. Also, extended the documentation.

pkg/controller/cloudfront/distribution/lateinit.go
```diff
@@ -36,6 +36,11 @@ import (
 // what the AWS SDK says that some of these fields are required.
 
 func lateInitialize(dp *svcapitypes.DistributionParameters, gdo *svcsdk.GetDistributionOutput) error {
+	// during import scenarios, we need the original caller reference info
+	if gdo.Distribution != nil && gdo.Distribution.DistributionConfig != nil {
```
muvaf, line 40: Did you consider having an `Initializer` like the ones we use for [tagging](https://github.com/crossplane/provider-aws/blob/793e588/pkg/controller/dynamodb/table/hooks.go#L78)? It'd basically populate the caller reference with UID if it's not given and then we'd be able to treat that field just like other fields in all calls since managed reconciler makes sure the initializers are called [before](https://github.com/crossplane/crossplane-runtime/blob/86fb15d/pkg/reconciler/managed/reconciler.go#L704) any external operation. In fact, we could even remove its `ignore` statement [here](https://github.com/crossplane/provider-aws/blob/86b704e/apis/cloudfront/generator-config.yaml#L14).

ezgidemirel: Hi @muvaf, I tried to use an `Initializer` which sets `spec.forProvider.distributionConfig.callerReference` value with `UID` if it's not given. But, the import scenario failed again. Setting the value in `preCreate` guarantees that the resource does not exist on the cloud provider. 

Thanks for pointing out the ignore statement. I missed that one.

pkg/controller/cloudfront/distribution/setup.go
```diff
@@ -85,151 +85,156 @@ func SetupDistribution(mgr ctrl.Manager, o controller.Options) error {
 }
 
 func preCreate(_ context.Context, cr *svcapitypes.Distribution, cdi *svcsdk.CreateDistributionInput) error {
-	cdi.DistributionConfig.CallerReference = awsclients.String(string(cr.UID))
+	if cr.Spec.ForProvider.DistributionConfig.CallerReference != nil && *cr.Spec.ForProvider.DistributionConfig.CallerReference != "" {
```
muvaf, line 88:
```suggestion
	if awsclients.StringValue(cr.Spec.ForProvider.DistributionConfig.CallerReference) != "" {
```

pkg/controller/cloudfront/distribution/lateinit.go
```diff
@@ -48,6 +48,9 @@ func lateInitialize(dp *svcapitypes.DistributionParameters, gdo *svcsdk.GetDistr
 	in := dp.DistributionConfig
 	from := gdo.Distribution.DistributionConfig
 
+	// during import scenarios, we need the original caller reference info
```
muvaf, line 51: Could you add a comment here about what we discussed offline? i.e. the caller reference is not exposed to users who would like to import the resource, hence we need to derive it from the first response from AWS if user didn't give it explicitly.

muvaf (approved):
Thanks @ezgidemirel ! LGTM!

pkg/controller/cloudfront/distribution/lateinit.go
ezgidemirel: done

pkg/controller/cloudfront/distribution/setup.go
ezgidemirel: done

pkg/controller/cloudfront/distribution/setup.go
ulucinar: Thanks @ezgidemirel!

ulucinar (approved):

PR #1275, opened 2022-04-27
fix(permission): Fix dealing with Pricipal being just a string

schroeder-paul:

On reconciliation the AWS API returns the resulting policy document for the lambda. This includes the Principal and it can be in two forms: A plain string or an object with a `Service` key. I.e.:

```json
"Principal": "foo"
```
vs.
```json
"Prinicpal": {
  "Service": "foo"
}
```

The current code breaks on the first one. This PR fixes that.

I have:

Unit tests are included, code changes are very local.

### DIFF
--- a/pkg/controller/lambda/permission/policy.go
+++ b/pkg/controller/lambda/permission/policy.go
@@ -20,17 +20,23 @@ import (
 	"encoding/json"
 )
 
-// type statementAction []string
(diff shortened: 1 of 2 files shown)


haarchri (approved):
thanks for taking care and add also unit-tests LGTM

PR #1285, opened 2022-05-03
DBInstance: Late Initialize all fields that cannot be set when connecting to a DBCluster

### LINKED ISSUE #1286: DBInstance: Late Initialize all fields that cannot be set when connecting to a DBCluster
### What happened?
Late init causes an update on not allowed fields of DBInstances which are member of a DBCluster.

The result is that all DBInstances in a DBCluster are going out of sync and not be synced again.

A part of this issue is already fixed in https://github.com/crossplane/provider-aws/issues/895
But only a subset of disallowed values was fixed here so the bug is still present and prevents DBInstances from being Healthy.

Thanks to @iAnomaly he found the proper resource of which fields are disallowed in DBCluster instances and was so kind to provide a PR. (https://github.com/crossplane/provider-aws/pull/1285/files)

> You can visit the AWS CLI docs for [modify-db-instance](https://docs.aws.amazon.com/cli/latest/reference/rds/modify-db-instance.html) and search for the string "managed by the DB cluster" to find all fields that cannot be set when associating a DBInstance with a DBCluster.

Together with the last fix of https://github.com/crossplane/provider-aws/pull/1197 we are currently running into heavy issues as it seems that the password in the connection is not going to match anymore to the password set in AWS as the sync is not possible right now because of this bug.

### How can we reproduce it?
Setup a DBCluster with DBInstances member and try to update the DBInstance afterwards

Crossplane version: 0.26.1

iAnomaly:

Fixes #1286 

I have:


haarchri (approved):
Tested in our Environment thanks for taking care

PR #1297, opened 2022-05-10
Observe existing OpenIDConnectProvider

### LINKED ISSUE #1280: Enable transferring ownership of OpenIDConnectProvider

Currently it's not possible to transfer ownership of an `OpenIDConnectProvider` due to the fact that the external-name is configured by AWS.

Allowing the transfer would make it easier to bootstrap a Crossplane cluster without manual intervention to "clean up".
Here's an example of how it's done in the `Observe` stage with another resource that has an external-name managed by AWS https://github.com/crossplane/provider-aws/blob/master/pkg/controller/iam/policy/controller.go#L115 where the AWS resource is inferred by the name and path.

However in the case with `OpenIDConnectProvider` we cant infer which resource that maps to a Crossplane definiton since AWS doesn't use the name. My initial thought would be to tag the object during creation and when transferring the ownership use the tags to select the correct `OpenIDConnectProvider`.
If this approach sounds acceptable I could start working on a PR for this.

ONordander:
Allow Crossplane to transfer ownership of previously
created OpenIDConnectProvider. The provider instance is derived from
the Crossplane tags.

Signed-off-by: Otto Nordander <otto.nordander@resurs.se>

Check for existing OpenIDConnectProviders with the same `crossplane-name` tag during `Observe` to allow transferring ownership of the provider.

Fixes #1280 

I have:

Added unittests and manual testing using:
```yaml
kind: OpenIDConnectProvider
metadata:
  name: test-oidc-provider
spec:
  deletionPolicy: Orphan
  providerConfigRef:
    name: example
  forProvider:
    thumbprintList:
      - 9e99a48a9960b14926bb7f3b02e22da2b0ab7280
    url: https://mytesturl.provider.aws.crossplane
 ```
```bash
kubectl apply -f oidc-provider.yaml
kubectl get openidconnectprovider.iam.aws.crossplane.io
kubectl delete -f oidc-provider.yaml # Orphaned
kubectl apply -f oidc-provider.yaml # Takes ownership of previously created provider
```

### DIFF
--- a/pkg/clients/iam/fake/openidconnectprovider.go
+++ b/pkg/clients/iam/fake/openidconnectprovider.go
@@ -29,9 +29,11 @@ var _ clientset.OpenIDConnectProviderClient = (*MockOpenIDConnectProviderClient)
 
 // MockOpenIDConnectProviderInput holds the input structures for future inspections
 type MockOpenIDConnectProviderInput struct {
-	CreateOIDCProviderInput         *iam.CreateOpenIDConnectProviderInput
-	TagOpenIDConnectProviderInput   *iam.TagOpenIDConnectProviderInput
-	UntagOpenIDConnectProviderInput *iam.UntagOpenIDConnectProviderInput
+	CreateOIDCProviderInput            *iam.CreateOpenIDConnectProviderInput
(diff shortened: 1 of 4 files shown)


haarchri (approved):
tested the change - it is working thanks for implemetation

Issue #1023, opened 2021-12-18
Elasticache resource continuously 'modifying' when securityGroup is specified


robertgates55:
When specifying a security group on a `ReplicationGroup` (either via `spec.forProvider.securityGroupIds` or `spec.forProvider.securityGroupIdRefs` - referencing a crossplane-created sg or a manually created one) the created replicationgroup in AWS is constantly reverting to `modifying`.

Example:
```
apiVersion: cache.aws.crossplane.io/v1beta1
kind: ReplicationGroup
metadata:
  name: rg-test-without-sg
  namespace: dev
spec:
  deletionPolicy: Delete
  forProvider:
    applyModificationsImmediately: true
    atRestEncryptionEnabled: true
    authEnabled: false
    automaticFailoverEnabled: true
    cacheNodeType: cache.t3.micro
    cacheParameterGroupName: default.redis5.0
    cacheSubnetGroupName: dev-dev1-cache-subnet-group
    engine: redis
    engineVersion: 5.0.6
    numCacheClusters: 3
    port: 6379
    preferredMaintenanceWindow: tue:02:00-tue:03:00
    region: eu-west-1
    replicationGroupDescription: dev/robgates-redis-no-sg
    snapshotWindow: 03:00-04:00
    tags:
    - key: Owner
      value: crossplane
    transitEncryptionEnabled: false
---
apiVersion: cache.aws.crossplane.io/v1beta1
kind: ReplicationGroup
metadata:
  name: rg-test-with-sg
  namespace: dev
spec:
  deletionPolicy: Delete
  forProvider:
    applyModificationsImmediately: true
    atRestEncryptionEnabled: true
    authEnabled: false
    automaticFailoverEnabled: true
    cacheNodeType: cache.t3.micro
    cacheParameterGroupName: default.redis5.0
    cacheSubnetGroupName: dev-dev1-cache-subnet-group
    engine: redis
    engineVersion: 5.0.6
    numCacheClusters: 3
    port: 6379
    preferredMaintenanceWindow: tue:02:00-tue:03:00
    region: eu-west-1
    replicationGroupDescription: dev/robgates-redis-with-sg
    securityGroupIds:
    - sg-0d3a26d0467c76eed
    snapshotWindow: 03:00-04:00
    tags:
    - key: Owner
      value: crossplane
    transitEncryptionEnabled: false
```

[...]

PR #1303, opened 2022-05-11
fix(elasticache) normalize preferred maintenance window

eaceaser:

Relates to https://github.com/crossplane/provider-aws/issues/1023

We discovered a state where `provider-aws` was repeatedly modifying the elasticache cluster without it becoming available, as described in #1023. Tracked the issue down to the use of a PreferredMaintenanceWindow argument that included capitalized day names: `Fri:04:20-Fri:16:20`. 

This is invalid according to [aws docs](https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html), as day names are supposed to be lowercased, however the AWS API accepts uppercased day names, and normalizes them behind the scenes to lowercased. This ends up breaking the `cacheClusterNeedsUpdate` check, which tests for exact equality between the declared maintenance window in the resource spec (which will still be uppercased), and the maintenance window returned from the AWS API, which will be lowercased.

This PR offers a potential fix, by normalizing the casing of the maintenance window when performing this check. However, it suffers from the following issues:

* we don't know the actual normalization strategy used by AWS here, so this is just a best guess.
* there are likely other cases impacted by this.

Overall it might be better to try to validate this string at resource-create time, but that would likely require a webhook, etc.

I have:

adds a unit test, also tested against the AWS API with ReplicationGroup resource w/ capitalized day names.

### DIFF
--- a/pkg/clients/elasticache/elasticache_test.go
+++ b/pkg/clients/elasticache/elasticache_test.go
@@ -26,12 +26,9 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/service/elasticache"
 	elasticachetypes "github.com/aws/aws-sdk-go-v2/service/elasticache/types"
-	"github.com/google/go-cmp/cmp"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
-
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
+	"github.com/google/go-cmp/cmp"
 
 	cachev1alpha1 "github.com/crossplane/provider-aws/apis/cache/v1alpha1"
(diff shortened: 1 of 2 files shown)


pkg/clients/elasticache/elasticache_test.go
```diff
@@ -988,11 +986,39 @@ func TestCacheClusterNeedsUpdate(t *testing.T) {
 			},
 			want: false,
 		},
+		{
+			name: "NeedsNoUpdateNormalizedMaintenanceWindow",
+			kube: replicationGroupParams(func(p *v1beta1.ReplicationGroupParameters) {
+				p.PreferredMaintenanceWindow = aws.String("Mon:00:00-Fri:23:59")
+
+			}),
+			cc: elasticachetypes.CacheCluster{
```
eaceaser, line 995: was looking at building a helper for this, but `elasticachecluster_test.go` already defines helpers for the CacheCluster struct in the same package, which would potentially collide with the needs of these tests. Would like some guidance on the best way to deal with this

haarchri (approved):
LGTM

PR #1307, opened 2022-05-13
fix(rds): include the port number in the connection details

### LINKED ISSUE #1296: DBCluster port number missing from the connection details secret
### What happened?

The connection details secret is missing the port.
Some other Crossplane providers expect the port to be in the secret, like for example, the Provider SQL.
This relates to: https://github.com/crossplane-contrib/provider-sql/issues/76

### How can we reproduce it?

Just create a DBCluster and observe the the generated connection details secret doesn't contain the port.

Crossplane version: v1.7.1
Provider AWS version: v0.26.1

I've checked the code and I think it should be added to https://github.com/crossplane/provider-aws/blob/master/pkg/controller/rds/dbcluster/setup.go#L83 would a PR to add it there be acceped or is there another way of making the port available in the secret? 

I've checked https://crossplane.io/docs/v1.7/reference/composition.html#connection-details and tried to add the port using `FromConnectionSecretKey`; `FromFieldPath` and `FromValue` and couldn't find a way to get it working...

luisdavim:

The connection details secret is missing the port.
Some other Crossplane providers expect the port to be in the secret, like, for example, the Provider SQL.
This relates to: https://github.com/crossplane-contrib/provider-sql/issues/76

Fixes #1296

I have:

I had some issues running the `make` command:
- initially, the `make` commands were mangling my working copy and I had to `export ROOT_DIR=$(pwd)` to get it to work properly.
- running `make reviewable` resulted in an error:
   ```
   Error: open /Users/luisdavim/.cache/ack-generate/src/aws-sdk-go/models/apis/servicecatalog/2015-12-10/api-2.json: too many open files
   ```
   I tried running `make -j1 reviewable` and it still failed.
- Running `make test` worked.
- When the `make` execution stopped with the error mentioned above, `goimports` had updated the imports of all the generated files, removing redundant imports and adding a space between the `stdlib` imports and the rest of the imports; I ran `go generate ./...` to revert those changes

Ran `make run` whilst connected to a real EKS cluster

### DIFF
--- a/pkg/controller/rds/dbcluster/setup.go
+++ b/pkg/controller/rds/dbcluster/setup.go
@@ -2,6 +2,7 @@ package dbcluster
 
 import (
 	"context"
+	"strconv"
 
 	svcsdk "github.com/aws/aws-sdk-go/service/rds"
 	svcsdkapi "github.com/aws/aws-sdk-go/service/rds/rdsiface"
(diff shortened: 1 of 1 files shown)


haarchri (approved):
LGTM

liciomatos:
Hi, should be great to have the reader endpoint in the connection details too. 
Could this be possible? 
tks.

PR #1322, opened 2022-05-25
GitHub-1286 fix late init dbinstance

### LINKED ISSUE #1286: DBInstance: Late Initialize all fields that cannot be set when connecting to a DBCluster

eloo-abi:

Exclude deletion protection if dbinstance is a cluster member

Fixes #1286

I have:

- [/] Read and followed Crossplane's [contribution process].

`make test` results in `package github.com/crossplane/provider-aws/pkg/clients/mock: build constraints exclude all Go files in /Users/josephweigl/git/github/crossplane/provider-aws/pkg/clients/mock`

`make reviewable` results in 
```
18:54:58 [ .. ] Generating sfn controllers and CRDs
Error: open /Users/josephweigl/.cache/ack-generate/src/aws-sdk-go/models/apis/servicecatalog-appregistry/2020-06-24/api-2.json: too many open files
```

Build the provider and deployed it in our test environment.
This changes has fixed the deletionProtection issue which cases the dbinstance to be synced false


haarchri (approved):
LGTM we will backport this fix to 0.27 and will cut shortly 0.27.1 with a other PRs

haarchri:
@eloo-abi are you working with Mac ? You can try ulimit -n 10240

eloo:
@haarchri 
thanks for the hint.. with the ulimit at least reviewable is working.

test is still failing:
```
package github.com/crossplane/provider-aws/pkg/clients/mock: build constraints exclude all Go files in /Users/josephweigl/git/github/crossplane/provider-aws/pkg/clients/mock
```

PR #1336, opened 2022-06-01
fix(LogGroup): Fix observe in case of more than 50 loggroups

### LINKED ISSUE #1188: LogGroup: Importing existing log group does not work
### What happened?
Wanted to take ownership over an existing log group using the crossplane.io/external-name annotation, but Crossplane tries to create a new LogGroup instead. This is failing because the LogGroup already exists.

### How can we reproduce it?
Create a log group manually in AWS called `/not/working`
Create a managed resource in kubernetes with matching logGroupName and external name annotation:
```
apiVersion: cloudwatchlogs.aws.crossplane.io/v1alpha1
kind: LogGroup
metadata:
  annotations:
    crossplane.io/external-name: /not/working
  name: not-working
spec:
  deletionPolicy: Orphan
  forProvider:
    logGroupName: /not/working
    region: eu-north-1
    retentionInDays: 7
  providerConfigRef:
    name: default
```

Will result in:
`create failed: cannot create LogGroup in AWS: ResourceAlreadyExistsException: The specified log group already exists`

Crossplane version: v1.6.1
EKS v1.21.9

petteja:
DescribeLogGroup returns maximum 50 loggroups pr default, which can lead to cases where loggroup exists but doesn't get returned and Crossplane will then attempt to create the resource, rather than taking ownership of the loggroup.
 
This adds the external name as LogGroupNamePrefix to the Describe call which will return loggroups that has external name as prefix.

Fixes #1188 

I have:

This has been tested out of cluster and also in our test- and production environments.

### DIFF
--- a/pkg/controller/cloudwatchlogs/loggroup/setup.go
+++ b/pkg/controller/cloudwatchlogs/loggroup/setup.go
@@ -53,6 +53,7 @@ func SetupLogGroup(mgr ctrl.Manager, o controller.Options) error {
 			u := &updater{client: e.client}
 			e.isUpToDate = u.isUpToDate
 			e.update = u.update
+			e.preObserve = preObserve
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


haarchri (approved):
Thanks tested ist working

PR #1342, opened 2022-06-07
split GenerateUpdateClusterConfigInput in 2 methods: Logging and VPC

### LINKED ISSUE #1014: EKS logging flag only works at cluster creation and does not apply for cluster update
### What happened?

I have detailed in [Slack](https://crossplane.slack.com/archives/C01718T2476/p1639442487481800), but as it looks like a bug that's impacting some other users as well, I'm cross-posting it here.

--

I have been trying to control EKS logging flag based on some XR attribute, but when I do update the flag, I get the following error from `eks.aws.crossplane.io/v1beta1/Cluster` resource:

```
CannotUpdateExternalResource: cannot update EKS cluster configuration: InvalidParameterException: Only one type of update can be allowed.
```
 
I don’t think I’m updating multiple items at once - here is the snippet of Composition:

```yaml
      base:
        apiVersion: eks.aws.crossplane.io/v1beta1
        kind: Cluster
        spec:
          forProvider:
            # ... snipped ...
            logging:
              clusterLogging:
                - enabled: false
                  types:
                    - api
                    # - audit
                    # - authenticator
                    # - controllerManager
                    # - scheduler
      patches:
        - fromFieldPath: spec.parameters.enableLogging
          toFieldPath: spec.forProvider.logging.clusterLogging[0].enabled
        # ... snipped ...
```

When creating a cluster, this setup seems to work fine for both enabled / disabled. But updating the cluster with toggling `spec.parameters.enableLogging` doesn’t seem to work. Is this some known limitation for existing cluster update?

### How can we reproduce it?

1. Create a cluster with `logging.clusterLogging` enabled
2. Update the spec with `logging.clusterLogging` disabled

(or vice versa, both seem to have some update issue.)

Crossplane version: v1.5.1
provider-aws: v0.21.2 (I'll test with v0.22.0 in the coming days, but I suspect the same issue persists with it, as I see no release note relating to the cluster setup)

adrienzieba:
Signed-off-by: Adrien Zieba <adrien.zieba@splashthat.com>

AWS CLI prevents us from updating the VPC config and the Logging in the same call. This PR is splitting the method `GenerateUpdateClusterConfigInput` in 2 distinct ones: `GenerateUpdateClusterConfigInputForLogging`, `GenerateUpdateClusterConfigInputForVPC`

Fixes #1014

I have:

- [x] Run `make reviewable test` to ensure this PR is ready for review. The test steps is failing on my laptop with the following error `package github.com/crossplane-contrib/provider-aws/pkg/clients/mock: build constraints exclude all Go`

Create a cluster with Logging enabled
```
...
        forProvider:
          logging:
            clusterLogging:
              - enabled: true
                types:
                  - api
                  - audit
                  - authenticator
                  - controllerManager
                  - scheduler
...
```

Then change the configuration to disable one of the type
```
...
        forProvider:
          logging:
            clusterLogging:
              - enabled: true
                types:
                  - api
                  - audit
                  - authenticator
                  - controllerManager
                  - scheduler
              - enabled: false
                types:
                  - controllerManager
...
```

*NOTE*: I noticed that the types needs to be in alphabetical order or the provider detects a constant drift with the aws resource.

### DIFF
--- a/pkg/clients/eks/eks.go
+++ b/pkg/clients/eks/eks.go
@@ -176,26 +176,32 @@ func CreatePatch(in *ekstypes.Cluster, target *v1beta1.ClusterParameters) (*v1be
 	return patch, nil
 }
 
-// GenerateUpdateClusterConfigInput from ClusterParameters.
-func GenerateUpdateClusterConfigInput(name string, p *v1beta1.ClusterParameters) *eks.UpdateClusterConfigInput {
+// GenerateUpdateClusterConfigInputForLogging from ClusterParameters.
+func GenerateUpdateClusterConfigInputForLogging(name string, p *v1beta1.ClusterParameters) *eks.UpdateClusterConfigInput {
 	u := &eks.UpdateClusterConfigInput{
 		Name: awsclients.String(name),
 	}
 
-	if p.Logging != nil {
-		u.Logging = &ekstypes.Logging{
-			ClusterLogging: make([]ekstypes.LogSetup, len(p.Logging.ClusterLogging)),
+	u.Logging = &ekstypes.Logging{
+		ClusterLogging: make([]ekstypes.LogSetup, len(p.Logging.ClusterLogging)),
+	}
+	for i, cl := range p.Logging.ClusterLogging {
+		types := make([]ekstypes.LogType, len(cl.Types))
+		for j, t := range cl.Types {
+			types[j] = ekstypes.LogType(t)
 		}
-		for i, cl := range p.Logging.ClusterLogging {
-			types := make([]ekstypes.LogType, len(cl.Types))
-			for j, t := range cl.Types {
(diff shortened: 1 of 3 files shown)


haarchri (approved):
LGTM - tested in our environment 
@adrienzieba did we want to do the ordering stuff also - what do you think ?

adrienzieba:
The ordering would be a nice to have. When you say at the provider level, do you mean adding some logic in [CreatePatch](https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/clients/eks/eks.go#L164) to sort `types` before comparison?
Maybe we would have to sort also 1 level above (ClusterLogging items) -> To be tested

haarchri:
would be cool to see a PR for the sorting stuff ;)

PR #1364, opened 2022-06-17
let aws go sdk detect partition and region in irsa

nabuskey:
Although changes introduced in https://github.com/crossplane-contrib/provider-aws/pull/1329 allow for IAM and Route53 resources to be provisioned in non `aws` partitions, it requires specific endpoint configurations in `ProviderConfig` to do so. 

This PR aims to allow for non `aws` partition users who run this provider in EKS with IRSA to provision global resources without endpoint configuration. 

Some background:
1. Cross partition assume role is not possible. e.g. assume role from IAM principals in `aws` partition to `aws-cn` is not possible.
2. If IRSA is used, EKS injects token and some [AWS specific environment variables into pod](https://github.com/aws/amazon-eks-pod-identity-webhook#eks-walkthrough)

Given these, I think it makes sense to let AWS SDK detect region and correct partition if reconciliation request is for a global resource. 

I have:

Created EKS clusters in China and AWS partitions with IRSA configuration enabled. Tested global and regional resource creations using these clusters. 

Prior to this change, reconciliation fails with messages similar to the following:
```
Warning  CannotCreateExternalResource  118s (x25 over 4m4s)  managed/hostedzone.route53.aws.crossplane.io  failed to create the Hosted Zone resource: InvalidIdentityToken: No OpenIDConnect provider found in your account for https://oidc.eks.cn-north-1.amazonaws.com.cn/id/ABCDEFGHIJKLMN

```

After this change, reconciliation works as expected:
```
Normal   CreatedExternalResource      53s   managed/hostedzone.route53.aws.crossplane.io  Successfully requested creation of external resource
```

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -605,6 +605,9 @@ func UsePodServiceAccountV1AssumeRole(ctx context.Context, _ []byte, pc *v1beta1
 	}
 	stsclient := sts.NewFromConfig(cfg)
 	stsAssumeRoleOptions := SetAssumeRoleOptions(pc)
+	if region == GlobalRegion {
+		region = cfg.Region
+	}
 	cnf, err := config.LoadDefaultConfig(
 		ctx,
 		userAgentV2,
@@ -681,7 +684,6 @@ func UsePodServiceAccountV1(ctx context.Context, _ []byte, pc *v1beta1.ProviderC
 	cfg, err := config.LoadDefaultConfig(
 		ctx,
 		userAgentV2,
-		config.WithRegion(region),
 	)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to load default AWS config")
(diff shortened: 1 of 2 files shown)


haarchri:
Can you Check for go-sdk-v1 (ack-generated) resources we have functions with V1 in name

nabuskey:
There is currently one resource type that uses go-sdk-v1 affected by incorrect pseudo region being passed. It's the `InstanceProfile` kind. 
Creation of this kind fails prior to this change with message: `InvalidClientTokenId: The security token included in the request is invalid`.

After the change, the creation request is successfully sent and it reconciles correctly. I've also tested other regional resources which uses aws go sdk v1 like dynamodb table. 

Unfortunately, there is an issue with deletion of this kind that's unrelated to code changes being introduced here. Attempting to delete `InstanceProfile` results in an error with:
```
managed/instanceprofile.iam.aws.crossplane.io  pre-delete failed: InvalidParameter: 1 validation error(s) found.
- missing required field, RemoveRoleFromInstanceProfileInput.RoleName.
```
 This happens without my changes. So I cannot validate the deletion part.

haarchri (approved):
LGTM - thanks for this enhancement

xchan:
@nabuskey Thanks a lot for this PR.
I've tested route53 and iam services in cn-northwest-1 with IRSA and it worked perfectly.
I can finally align my code for global region and china region.

PR #1370, opened 2022-06-24
fix route53 wildcard handling

### LINKED ISSUE #1115: route53 resource does not work well with wildcard '*'
### What happened?
I am trying to create CNAME with wildcard so I do not have to explicitly create each route53 entry. E.g. I have DNS entry `myproject.mycompany.com` which is `A` type of record. I'd like to have wildcard DNS records (type of CNAME ) for apps (app1, app2, ...), e.g. `*.myproject.mycompany.com` however when I create such record via `route53.aws.crossplane.io/v1alpha1` and kind `ResourceRecordSet` its state is `READY: False`. The same behaviour is for CNAME and A record.
```
test-dns                                 False   True     *.myproject.mycompany.com   CNAME      11m
```
However DNS entry is create in AWS but It's not deleted when I delete `ResourceRecordSet` resource.

### How can we reproduce it?
You can use following manifest 
```
apiVersion: route53.aws.crossplane.io/v1alpha1
kind: ResourceRecordSet
metadata:
  name: 'test-dns'
  annotations:
    crossplane.io/external-name: "*.myproject.mycompany.com"
spec:
  forProvider:
    type: A
    ttl: 300
    zoneId: <zoneId>
    resourceRecords:
    - value: 1.2.3.4
  providerConfigRef:
    name: default
```

Crossplane version: crossplane/provider-aws-controller:v0.22.0
Aws provider version: crossplane/crossplane:v1.5.1

nabuskey:
Signed-off-by: Manabu Mccloskey <manabu.mccloskey@gmail.com>

Route53 does not return the `*` character if it sees it as a wildcard. e.g. For a record `*.crossplane.io`, the string for its name returned by Route53 API is `\052.crossplane.io`. 

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DomainNameFormat.html?shortFooter=true#domain-name-format-asterisk

Fixes #1115

I have:

Ran this code against my private Route53 zone with the following yaml:
```yaml
apiVersion: route53.aws.crossplane.io/v1alpha1
kind: ResourceRecordSet
metadata:
  name: 'test-dns'
  annotations:
    crossplane.io/external-name: "*.crossplane.manabu.com"
spec:
  forProvider:
    type: A
    ttl: 300
    zoneId: <ZONEID>
    resourceRecords:
    - value: 1.2.3.4
  providerConfigRef:
    name: aws-provider-config
```
It used to never reach the ready state. 
```
Normal   PendingExternalResource      24m (x3 over 24m)  managed/resourcerecordset.route53.aws.crossplane.io  Waiting for external resource existence to be confirmed
```
After the change, it correctly reaches the ready state.

### DIFF
--- a/pkg/clients/resourcerecordset/resourcerecordset.go
+++ b/pkg/clients/resourcerecordset/resourcerecordset.go
@@ -39,6 +39,7 @@ import (
 
 const (
 	errResourceRecordSetNotFound = "ResourceRecordSet.NotFound"
+	wildCardCharacters           = "\\052"
 )
 
 // Client defines ResourceRecordSet operations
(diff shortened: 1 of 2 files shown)


haarchri (approved):
LGTM thanks for taking care

Issue #1306, opened 2022-05-12
Specify default tags for each resource in ProviderConfig


MisterMX:
### What problem are you facing?

To achieve a more fine-grained access control we want to tag each of our AWS resources with the ID of the tenant who owns it.

We could achieve this by patching all MRs through Crossplane compositions. However, it is a huge effort and quite complex to patch every resource in all of our compositions of which we have multiple dozens.

It would make things easier if we could set the list of tags to add in each `ProviderConfig` of which we have one per tenant and let the controller handle it.

### How could Crossplane help solve your problem?

1. Add a new field `spec.resource.tags` to `ProviderConfig` where one can specify a set of default tags that are set to every resource:

```yaml
apiVersion: aws.crossplane.io/v1beta1
kind: ProviderConfig
metadata:
  name: default
spec:
  credentials:
    # ...
  resource:
    tags:
      tenant: <id>
```

2. Add these tags to the MR during the controllers reconcile.. Maybe `lateInitialize` might be suited for that.

ksawerykarwacki:
Shouldn't it be on the provider config level? I assume that you may have different configs for different accounts/roles with different rules for default tags.

MisterMX:
@ksawerykarwacki yes, that was my intention. I somehow mixed up those resource. Thanks for pointing that out.

Issue #1244, opened 2022-04-01
Tags definition is inconsistent across resources


bobh66:
### What happened?
There are approximately 78 resources that support a "tags" parameter in the spec.forProvider section.  Of these, 60 define tags as an array of objects with key and value properties, while the remaining 18 define tags as an object with additionalProperties.type = string.

This makes it impossible to consistently define a set of tags across all resources in a composition since they are different data types.

Given the difference, I'd like to  propose that the 18 object types be converted to arrays of objects for consistency.

package/crds/apigatewayv2.aws.crossplane.io_apis.yaml
package/crds/apigatewayv2.aws.crossplane.io_domainnames.yaml
package/crds/apigatewayv2.aws.crossplane.io_stages.yaml
package/crds/apigatewayv2.aws.crossplane.io_vpclinks.yaml
package/crds/cloudwatchlogs.aws.crossplane.io_loggroups.yaml
package/crds/eks.aws.crossplane.io_addons.yaml
package/crds/eks.aws.crossplane.io_clusters.yaml
package/crds/eks.aws.crossplane.io_fargateprofiles.yaml
package/crds/eks.aws.crossplane.io_identityproviderconfigs.yaml
package/crds/eks.aws.crossplane.io_nodegroups.yaml
package/crds/glue.aws.crossplane.io_connections.yaml
package/crds/glue.aws.crossplane.io_crawlers.yaml
package/crds/glue.aws.crossplane.io_jobs.yaml
package/crds/kafka.aws.crossplane.io_clusters.yaml
package/crds/lambda.aws.crossplane.io_functions.yaml
package/crds/mq.aws.crossplane.io_brokers.yaml
package/crds/prometheusservice.aws.crossplane.io_workspaces.yaml
package/crds/sqs.aws.crossplane.io_queues.yaml

### How can we reproduce it?

See Vpc.ec2.aws for an example of tags defined as a list of objects

See LogGroup.cloudwatch.aws for an example of tags defined as an object

### What environment did it happen in?
Crossplane version: 1.7

haarchri:
it depends on the go-sdk version V1/V2 and the resources we using the upstream Implementation without type change

bobh66:
Maybe a workaround for this would be to add a MapToList conversion type to the Composition, which would take a map of key value pairs and convert it to a List of maps with {Key: keyname, Value: value} pairs?  That would at least allow the user to pass in a single Map of key/value pairs for tags and use it for both types of tag structures.

[...]

PR #1378, opened 2022-07-04
feat: Add code generator for generic resource tagging

MisterMX:

This adds a code generator that creates a method `AddTag` for each managed resource that has a field `Spec.ForProvider.Tag` to satisfy the generic interface `Tagged`.

It enables us to replace each specific `tagger` initializer that is currently copy-pasted for every controller with a central component reducing code redundancy.

The generated code abstracts from (most) of the more over a dozen tagging implementations that exist today. The following tag implementations are covered by this generator:

* `map[string]string`
* `map[string]*string`
* `[]Tag  {string, string}`
* `[]*Tag {string, string}`
* `[]Tag  {*string, *string}`
* `[]*Tag {*string, *string}`
* `[]Tag  (with string key and *string value)`
* `[]*Tag  (with string key and *string value)`

Note: It does _not_ fix the issue that tags are defined differently for each resource because that is an issue with the underlying ACK generator and AWS SDK (see #1244).

This is a preliminary for #1306.

I have:

Unit tests


apis/generate.go
```diff
@@ -16,21 +17,24 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// NOTE(negz): See the below link for details on what is happening here.
-// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
+// // NOTE(negz): See the below link for details on what is happening here.
```
haarchri, line 20: @MisterMX  looks like we have double `// `now ?

pkg/controller/common/doc.go
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright 2021 The Crossplane Authors.
```
chlunde, line 2: nit: use 2022 for all new files?

pkg/controller/common/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package common contains shared controller logic.
+package common
```
chlunde, line 18: Could we call this something more specific like tagger or tagutil?
https://github.com/golang/go/wiki/CodeReviewComments#package-names

chlunde:
looks awesome! I think we could add IsUpToDate helpers the same way later?

apis/kms/v1alpha1/tagged.go
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+// AddTag adds a tag to this Key. If it already exists, it will be overwritten.
+// It returns true if the tag has been added/changed. Otherwise false.
+func (mg *Key) AddTag(key string, value string) bool {
```
chlunde, line 21: is this one hand writtend? If so, add a comment why?

chlunde:
@MisterMX did you also try some of the resource types in a real cluster?

This is a huge PR, it looks good to me but there might be a "needle in the haystack" here. Is there anything in particular you think we should look at, anything that's not tested or that you feel less confident about?

artem-nefedov:
Will this fix inconsistency in tagging input? (like how most resources expect tags as an array, but some select ones, such as eks Cluster and NodeGroup, expect them to be a map)

haarchri:
@artem-nefedov no

PR #1380, opened 2022-07-06
enhance existing Glue resources

### LINKED ISSUE #1304: Glue controllers lack features
### What happened?

Regarding v0.27.0, the current controllers for Glue are not fully implemented and lack a bunch of features, most notably lacking `lateInitialize` and `isUpToDate`. This applies to all currently existing Glue controllers: `classifier`, `connection`, `crawler`, `database`, `job`, `securityConfiguration`.

Additionally, a bunch of CRD fields are missing referencers.

Provider version: v0.27.0

wotolom:

changes for the existing AWS Glue resources SecurityConfiguration, Database, Connection, Classifier, Crawler and Job

Fixes #1304

Add `lateInitialize`, `isUpToDate` and `update` functionality where applicable
Add missing fields and referencers
StyleFix! several fieldnames 

While adding missing fields and referencers/selectors, the opportunity arose to harmonize several existing **fieldnames**.
Here are these **Breaking Changes**:

- Job:
  - `roleArn` to `role` (**same** for `roleRef`, `roleSelector`)
- SecurityConfiguration:
  - for all encryptions (`cloudWatchEncryption`, `jobBookmarksEncryption`, `s3Encryption`):
    - `kmsKeyARN` to `kmsKeyArn` (add `kmsKeyArnRef`, `kmsKeyArnSelector`)
- Connection:
  - `securityGroupIDList` to `securityGroupIdList` (**same** for `securityGroupIdRefs`, `securityGroupIdSelector`)
  - `subnetID` to `subnetId` (**same** for `subnetIdRef`, `subnetIdSelector`)
- Crawler:
  - `roleArn` to `role` (**same** for `roleRef`, `roleSelector`)
  - for `s3Targets`:
    - `dlqEventQueueARN` to `dlqEventQueueArn` (add `dlqEventQueueArnRef`/`dlqEventQueueArnSelector`)
    - `eventQueueARN` to `eventQueueArn` (add `eventQueueArnRef`/`eventQueueArnSelector`)

People, who already used any of these Glue resources in an older crossplane version, would need to change the mentioned fieldnames in their Claims (XRCs), Compositions, XRDs etc.

Notes:

- Because this [ACK-Issue](https://github.com/aws-controllers-k8s/community/issues/1078) was resolved, I could reimplement the Schedule field as intended. :)
- Because of this [ACK-Issue](https://github.com/aws-controllers-k8s/community/issues/1372) I encountered, I had to set time-related fields for Status.AtProvider manually in postObserve. 

I have:

All 6 resources were tested manually (create, update, delete) with the objects in examples/glue.


haarchri:
@wotolom can you cleanup the examples ?

wotolom:
Hi @haarchri, are the examples ok now?

package/crds/glue.aws.crossplane.io_connections.yaml
```diff
@@ -99,12 +123,17 @@ spec:
                         description: Specifies the physical requirements for a connection.
                         properties:
                           availabilityZone:
+                            description: The connection's Availability Zone. This
+                              field is redundant because the specified subnet implies
+                              the Availability Zone to be used. Currently the field
+                              must be populated, but it will be removed in the future.
                             type: string
-                          securityGroupIDList:
+                          securityGroupIdList:
```
haarchri, line 131: this is a breaking change

package/crds/glue.aws.crossplane.io_connections.yaml
```diff
@@ -99,12 +123,17 @@ spec:
...
+                            description: The security group ID list used by the connection.
                             items:
                               type: string
                             type: array
-                          securityGroupIDRefs:
+                          securityGroupIdRefs:
```
haarchri, line 136: this is a breaking change

package/crds/glue.aws.crossplane.io_connections.yaml
```diff
@@ -187,10 +216,12 @@ spec:
                                     type: string
                                 type: object
                             type: object
-                          subnetID:
+                          subnetId:
```
haarchri, line 219: this is a breaking change

package/crds/glue.aws.crossplane.io_crawlers.yaml
```diff
@@ -103,13 +331,14 @@ spec:
                   region:
                     description: Region is which region the Crawler will be created.
                     type: string
-                  roleArn:
+                  role:
```
haarchri, line 334: this is a breaking change // why changed roleArn to role ?

package/crds/glue.aws.crossplane.io_jobs.yaml
```diff
@@ -80,14 +159,89 @@ spec:
                       scriptLocation:
                         type: string
                     type: object
-                  connections:
```
haarchri, line 83: why connections is removed ?

package/crds/glue.aws.crossplane.io_jobs.yaml
```diff
@@ -61,6 +61,85 @@ spec:
               forProvider:
                 description: JobParameters defines the desired state of Job
                 properties:
+                  SecurityConfiguration:
```
haarchri, line 64: SecurityConfiguration vs. securityConfiguration ?

package/crds/glue.aws.crossplane.io_securityconfigurations.yaml
```diff
@@ -72,11 +72,19 @@ spec:
                           encrypted.
                         properties:
                           cloudWatchEncryptionMode:
+                            description: The encryption mode to use for CloudWatch
+                              data.
+                            enum:
+                            - DISABLED
+                            - SSE-KMS
                             type: string
-                          kmsKeyARN:
```
haarchri, line 76: This is a breaking change

package/crds/glue.aws.crossplane.io_securityconfigurations.yaml
```diff
@@ -158,11 +167,19 @@ spec:
                         description: Specifies how job bookmark data should be encrypted.
                         properties:
                           jobBookmarksEncryptionMode:
+                            description: The encryption mode to use for job bookmarks
+                              data.
+                            enum:
+                            - DISABLED
+                            - CSE-KMS
                             type: string
-                          kmsKeyARN:
```
haarchri, line 162: This is a breaking change

haarchri (commented):
Thanks for this enhancement - the only open part are the fields with breaking changes for folks using glue stuff - for now it would be betternot changing so many existing fields - what do you think ? We can do this with migration hook support in the future

package/crds/glue.aws.crossplane.io_crawlers.yaml
wotolom: I just see that I would also want to change the equivalent fieldname in the Job resource from "roleArn" to "role". (I did forget it there somehow)

In both cases (Crawler/Job) AWS Doc says Name and ARN of role are accepted and AWS API does so (which is the important part). So a fieldname "role" would make more sense than "roleName" or "roleArn".

However the two cases differ in what the AWS API returns as the content of the "role"-field.
- Job: here the AWS API gives as the content of the "role"-field in the same way we (or the user) provide it. If we provide the name of a role, we get just the name back. If we provide an ARN, we get an ARN back. That's nice -> easy isUpToDate-check ;)
- Crawler: here the AWS API is not so nice. We always get the role name back. (which is funny, because the AWS Doc hints we should only get the ARN back ...) In this case we need an extra check in isUpToDate to be able to provide the "Name or ARN"-functionality to the user. (see crawler/setup.go:203) 

In summary I would say as long as we can maintain that "Name or ARN"-functionality without any or too much extra complexity, we should do so and keep the AWS API given fieldname, which reflects that. 

How do you think about this?

package/crds/glue.aws.crossplane.io_jobs.yaml
wotolom: Yes, good catch! I will fix this in the next commit!

wotolom:
Thanks for the review!

About the breaking changes:
The prior implementations of the Glue resources were not really complete, but of course already productively usable.
I still think now (with this merge) would be a good time to make these changes as the number of people who use the Glue service will only grow. And for the versioning it may be simpler to have these changes coupled together with this enhancement.

However I'm of course a bit biased. (contributing and future use) ;)

haarchri:
@wotolom  sounds great - can you add all breaking changes for release notes that would be great

haarchri (approved):
thanks for Implementation - tested all glue resources

PR #1387, opened 2022-07-15
fix(s3.bucket): Fixes #1165, discrepancy in QueueConfiguration.Filter.Key.FilterRules.Name

### LINKED ISSUE #1165: S3 Bucket constantly calls PutBucketNotification
### What happened?

The issue is similar to this issue, however, I've hardcoded the ID in the notifications queue configuration and we're using version 0.21.2 which should have this fix in place:
https://github.com/crossplane/provider-aws/issues/476

I've created a queue and it continuously tries to put in the notifications configuration, generating a test message every time.  

### How can we reproduce it?

Create a bucket with a notifications queue, check cloudtrail logs to see that it's making an api call every minute.

Crossplane version:

schroeder-paul:
The AWS API returns QueueConfiguration.Filter.Key.FilterRules.Name as "Prefix"/"Suffix" but expects "prefix"/"suffix", this leads to inconsistency and a constant diff.

Fixes #1165

I have:

Test were added, see diff.

### DIFF
--- a/pkg/controller/s3/bucket/notificationConfig.go
+++ b/pkg/controller/s3/bucket/notificationConfig.go
@@ -19,6 +19,7 @@ package bucket
 import (
 	"context"
 	"sort"
+	"strings"
 
 	"github.com/aws/smithy-go/document"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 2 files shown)


haarchri (approved):
@schroeder-paul thanks for fixing this

PR #1399, opened 2022-08-04
Feature/securitygrouprule

akesser:
Support for Security Group rules was added to manage security group rules individuals for security groups.

Security groups got two new flags, ignoreIngress and ingoreEgress to be able to manage security group rules for security groups managed by crossplane

We need to manage security group rules at security groups that are not managed by crossplane, therefore we need this resource.

I have:

Tested the implementation by hand (create, update delete for single and multiple resources) and the controller in controller_test.go file


haarchri (approved):
LGTM - thanks for implementation

PR #1418, opened 2022-08-11
fix: Unblock master

akesser:
Signed-off-by: André Kesser <andre.kesser@dkb.de>

Run make generate to unblock master


haarchri (approved):
thanks for taking care

PR #1443, opened 2022-08-22
fix(eks/identityproviderconfig): Reset status field to retry create

### LINKED ISSUE #1422: EKS IdentifyProviderConfig: Associate will never be called again when failing the first time
### What happened?

The EKS IdentityProviderConfig controller checks if `status.atProvider.status != "CREATING"` to determine whether it should call `client.AssociateIdentityProviderConfig`. This field is set during observe *only* when the external resource exists.

Now, if the creation of the external resource fails it will be garbage collected and removed so `DescribeIdentityProviderConfig` will result in a `not found` which will trigger a creation. But since `status.atProvider.status` is still set to `CREATING` and is never updated, nothing will happen in create and the association will never be tried again.

### How can we reproduce it?

Create an `IdentityProviderConfig` for a cluster that is not ready.6

`provider-aws`: `v0.29.0`

MisterMX:

Fixes #1422

I have:

Unit testing

### DIFF
--- a/pkg/controller/eks/identityproviderconfig/controller.go
+++ b/pkg/controller/eks/identityproviderconfig/controller.go
@@ -111,7 +111,17 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		},
 		ClusterName: &cr.Spec.ForProvider.ClusterName})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


MisterMX:
@haarchri can you take a look at this?

haarchri (approved):
LGTM

Issue #1456, opened 2022-08-28
Doc: updating `GENERATED_SERVICES` might no longer be necessary


bigmikes:
Documentation of CODE_GENERATION.md might be outdated.

[Here](https://github.com/crossplane-contrib/provider-aws/blob/master/CODE_GENERATION.md?plain=1#L66) we say that [GENERATED_SERVICES](https://github.com/crossplane-contrib/provider-aws/blob/master/Makefile#L11) need to be updated whenever we add support for a new service. However, it looks like the script takes care of finding all supported services automatically.

Should we update the doc?

PR #1465, opened 2022-09-01
doc: removed comment about updating GENERATED_SERVICES manually

bigmikes:

The GENERATED_SERVICES variable is automatically set in Makefile
(https://github.com/crossplane-contrib/provider-aws/blob/master/Makefile#L11).
So there is no need to manually update it whenever we create a new
service, as the documentation suggests.

Fixes: #1456

I have:

No testing required, it's a doc fix.


haarchri (approved):
thanks for taking care =)

PR #1468, opened 2022-09-08
Feat/manage only custom vpcpeering

Kasama:
Enable only custom controller for VPC Peering. Other resources will be
managed by the upstream crossplane provider-aws.

In the future, any other resource that needs to be customized, shall be
added to this project.


Kasama:
please disregard, this was supposed to be sent to a fork of the project :disappointed:

PR #1490, opened 2022-09-22
fix(mq/broker): missing subnets + SGs on creation

wotolom:

This fixes the controller not putting the user provided `Subnets` and `SecurityGroups` into the `CreateBrokerRequest`. 
This miss was often disguised, because AWS would fall back to the default VPCs subnets/SG (if there was one).

I came across some AWS default-set fields, which got implemented to `LateInitialize`.

Also I used the opportunity and added a bunch of fields to `BrokerObservation` for a better Broker state/status overview.

Note:

- Because of this [ACK-Issue](https://github.com/aws-controllers-k8s/community/issues/1372) I re-encountered, I had to set time-related field for `status.AtProvider` manually in `postObserve`.

I have:

Manually tested _Create_ and _Delete_ with the broker-examples.


haarchri (approved):
LGTM - thanks for improvement

PR #1494, opened 2022-09-26
fix(crawler): RoleRef should extract ARN

wotolom:

Adds line to extract ARN from Role to be used for `Crawler`

I miss of my part, when I enhanced the Glue service here.

Role reference for the resource `Crawler` should use the ARN:
1. for consistency -> same as the Glue resource `Job`
2. with ARN through refs, users don't have to think about possible path, if other fields only take ARN

I have:

Manually with the relating example.


haarchri:
@wotolom  can you fix lint ?

haarchri (approved):
LGTM

PR #1513, opened 2022-10-04
diff iam role tags such that order of keys does not matter.

### LINKED ISSUE #1419: Default tags cause infinite reconcile with compositions
### What happened?

Since the introduction of default tags (e.g. crossplane-kind) in some resource we see a huge increase (millions) of generations of our MRs.
We have check what is happening and found out that the default tags together with custom set tags in combination with a composition causes reconciles the whole times.

We have investigated the following resources are affected in our case (but i guess all resources with default tags are affected):
- Roles
- Policy
- ReplicationGroup

A workaround here would be to not have `tags` set to custom values..
But we want to use tags :D

Maybe this issue is related to this issues:
- https://github.com/crossplane-contrib/provider-aws/issues/1352
- https://github.com/crossplane-contrib/provider-aws/issues/1412

### How can we reproduce it?
here is the composition and xrd
```
apiVersion: apiextensions.crossplane.io/v1
kind: Composition
metadata:
  name: iamrole-composition-test
  labels:
    provider: aws
spec:
  compositeTypeRef:
    apiVersion: crossplane.test/v1
    kind: XIAMRoleInstanceTest
  resources:
    - base:
        kind: Role
        metadata:
          name: METADATA_UID
        spec:
          forProvider:
            assumeRolePolicyDocument: '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"Service":["lambda.amazonaws.com"]},"Action":["sts:AssumeRole"]}]}'
            tags:
              - key: claim-namespace
                value: CLAIM_NAMESPACE_PLACEHOLDER
          providerConfigRef:
            name: aws-provider-config
      patches:
        - type: FromCompositeFieldPath
          fromFieldPath: spec.parameters.permissionsBoundary
          toFieldPath: spec.forProvider.permissionsBoundary

--- 
apiVersion: apiextensions.crossplane.io/v1
kind: CompositeResourceDefinition
metadata:
  name: xiamrolecompositionstest.crossplane.test
spec:
  group: crossplane.test
  names:
    kind: XIAMRoleInstanceTest
    plural: xiamrolecompositionstest
  claimNames:
    kind: IAMRoleInstanceTest
    plural: iamrolecompositionstest
  defaultCompositionRef: 
    name: iamrole-composition-test
  connectionSecretKeys: []
  versions:
  - name: v1
    served: true
    referenceable: true
    schema:
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            properties:
              parameters:
                type: object
                properties:
                  deletionPolicy:
                    type: string
                    description: Deletion policy for the AWS resources
                  awsAccountId:
                    type: string
                    description: Id of the aws account
                  roleName:
                    type: string
                    description: The name of the role to be created in IAM
                  serviceAccountName:
                    type: string
                    description: Service account name for this role
                  oidcProvider:
                    type: string
                    description: Id of our oidc provider
                  permissionsBoundary:
                    type: string
                    description: PermissionsBoundary is the ARN of the policy that is used to set the permissions boundary for the role.
                required:
                - awsAccountId
                - roleName
                - serviceAccountName
                - oidcProvider
                - permissionsBoundary
            required:
            - parameters
```

and then we the MR looping between the following spec.forProvider.Tags
```
...
 forProvider:
    tags:
    - key: claim-namespace
      value: CLAIM_NAMESPACE_PLACEHOLDER

and 
 forProvider:
    tags:
    - key: claim-namespace
      value: CLAIM_NAMESPACE_PLACEHOLDER
    - key: crossplane-kind
      value: role.iam.aws.crossplane.io
    - key: crossplane-name
      value: iamrole-instance-n6xg2-nv9fz
      value: aws-provider-config
```

Crossplane version: v1.7.0
Provider version: v0.28.1

### LINKED ISSUE #1501: IAM role frequent updates when using custom tags
### What happened?

I think that this is similar to #792. When I create a role with custom tags, the ordering changes and that triggers updates to the role. If I remove the custom tags this issue does not occur. I think that the ordering is still changing but as I am not specifying `tags:` in my config it does not add it to the diff.

```diff
&types.Role{
        ... // 1 ignored and 7 identical fields
        PermissionsBoundary: &{PermissionsBoundaryArn: &\"arn:aws:iam::000000000000:policy/boundary\", PermissionsBoundaryType: \"Policy\"},
        RoleLastUsed:        &{},
        Tags: []types.Tag{
                {Key: &\"tenant\", Value: &\"tenantName\"},
                {Key: &\"cluster\", Value: &\"clusterName\"},
-               {Key: &\"crossplane-kind\", Value: &\"role.iam.aws.crossplane.io\"},
                {Key: &\"crossplane-name\", Value: &\"tenant-access\"},
                {Key: &\"crossplane-providerconfig\", Value: &\"default\"},
+               {Key: &\"crossplane-kind\", Value: &\"role.iam.aws.crossplane.io\"},
        },
        ... // 1 ignored field
  }
```

```diff
&types.Role{
        ... // 1
ignored and 7 identical fields
        PermissionsBoundary: &{PermissionsBoundaryArn: &\"arn:aws:iam::000000000000:policy/boundary\", PermissionsBoundaryType: \"Policy\"},
        RoleLastUsed:        &{LastUsedDate: &s\"2022-09-29 09:12:59 +0000 UTC\", Region: &\"us-east-1\"},
        Tags: []types.Tag{
                {Key: &\"tenant\", Value: &\"tenantName\"},
                {Key: &\"cluster\", Value: &\"clusterName\"},
                {Key: &\"crossplane-name\", Value: &\"tenant-access\"},
-               {Key: &\"crossplane-providerconfig\", Value: &\"default\"},
                {Key: &\"crossplane-kind\", Value: &\"role.iam.aws.crossplane.io\"},
+               {Key: &\"crossplane-providerconfig\", Value: &\"default\"},
        },
        ... // 1 ignored field
  }

```

### How can we reproduce it?

Create a role using the following 
```yaml
kind: Role
metadata:
  name: tenant-access
spec:
  forProvider:
    permissionsBoundary: arn:aws:iam::000000000000:policy/boundary
    assumeRolePolicyDocument: |
        {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Effect": "Allow",
                    "Principal": {
                        "AWS": "arn:aws:iam::000000000000:role/role"
                    },
                    "Action": "sts:AssumeRole"
                }
            ]
        }
    tags:
      - key: tenant
        value: tenantName
       - key: cluster
         value: clusterName
```

Crossplane version: v1.9.0
EKS version: v1.23.10-eks-15b7512
Provider version: v0.31.0

patelronak:

For IAM Roles if the order of same tags (key/value), in my case default tags, did not match between aws provided output and spec file the observe function that uses go cmp would result in a diff and considered the role not up-to-date.  The output logs showed following, note the tags/values are same but in different order.

```"Found observed difference in IAM role
  &types.Role{
  \t... // 1 ignored and 7 identical fields
  \tPermissionsBoundary: nil,
  \tRoleLastUsed:        &{LastUsedDate: &s\"2022-09-14 17:46:36 +0000 UTC\", Region: &\"us-east-1\"},
  \tTags: []types.Tag{
- \t\t{Key: &\"crossplane-providerconfig\", Value: &\"provider-aws\"},
+ \t\t{Key: &\"crossplane-name\", Value: &\"test-role-iam-gqp76-z6fvf\"},
  \t\t{Key: &\"crossplane-kind\", Value: &\"role.iam.aws.crossplane.io\"},
- \t\t{Key: &\"crossplane-name\", Value: &\"test-role-iam-gqp76-z6fvf\"},
+ \t\t{Key: &\"crossplane-providerconfig\", Value: &\"provider-aws\"},
  \t},
  \t... // 1 ignored field
  }
  ```

Initially I added logic to check the tags and append the diff of new/deleted tags to diff variable but the lint logic reported error (fix cyclomatic complexity 13) so I changed the logic to only get bool value if tags need to be updated and left the cmp diff logic in place to diff the tags if there is a difference and show the text difference and if there is no difference I added Tags to ignorefields list.

Fixes #1419
Fixes #1501

I have:

Created a kind cluster and configured it as per docs and ran `make run` and deployed a test role and manually edited the role to change the tags once applied to validate that it does not find differences if tags are moved around but it does observe diffs if new tags are added or existing tags are removed.

### CODE REVIEW

File: pkg/clients/iam/role.go [hunk 1]
```diff
@@ -175,13 +175,19 @@ func IsRoleUpToDate(in v1beta1.RoleParameters, observed iamtypes.Role) (bool, st
...
 	}
 
-	diff := cmp.Diff(desired, &observed, cmpopts.IgnoreInterfaces(struct{ resource.AttributeReferencer }{}), cmpopts.IgnoreFields(observed, "AssumeRolePolicyDocument"), cmpopts.IgnoreTypes(document.NoSerde{}))
-	if diff == "" && policyUpToDate {
+	_, _, areTagsUpToDate := DiffIAMTagsReturnStringList(desired.Tags, observed.Tags)
+	if areTagsUpToDate {
+		ignoreFields = append(ignoreFields, "Tags")
+	}
+
+	diff := cmp.Diff(desired, &observed, cmpopts.IgnoreInterfaces(struct{ resource.AttributeReferencer }{}), cmpopts.IgnoreFields(observed, ignoreFields...), cmpopts.IgnoreTypes(document.NoSerde{}))
```
MisterMX, line 189: I believe it is cleaner to use the builtin `cmpopts.SortSlices` function:

```go
cmpopts.SortSlices(lessTag)
```

and

```go
func lessTag(a, b iamtypes.Tag) bool {
	if a.Key == nil {
		return b.Key != nil
	}
	if b.Key == nil {
		return false
	}
	return *a.Key <= *b.Key
}
```

MisterMX (changes requested):

haarchri:
@patelronak any chance to check the open points ?

patelronak:
@haarchri @MisterMX sorry for the latest response but I will look at this tomorrow.

patelronak:
@MisterMX whenever you get a chance to review the PR.  cc: @haarchri

haarchri (approved):
Thanks for the fix

haarchri:
@MisterMX you need to approve as well because you requested changes

pkg/clients/iam/role.go
```diff
@@ -175,12 +175,13 @@ func IsRoleUpToDate(in v1beta1.RoleParameters, observed iamtypes.Role) (bool, st
 		return false, "", err
 	}
 
+	ignoreFields := []string{"AssumeRolePolicyDocument"}
```
MisterMX, line 178: Please inline this again or define it as `const`

MisterMX:
@patelronak one last thing. Sorry for being so picky.

pkg/clients/iam/role.go
patelronak: Updated the code as you requested.

patelronak:
> @patelronak one last thing. Sorry for being so picky.

@MisterMX let me know if there is any additional change that you think we should make.

haarchri:
@MisterMX can you Check again ? I will bring this in and will cut the next release

MisterMX (approved):
LGTM. Thanks a lot @patelronak!

PR #1541, opened 2022-10-27
RDS dbinstance: Check only major version if AutoMinorVersionUpgrade is true

### LINKED ISSUE #832: RDSInstance is in synced:false state because `Cannot upgrade postgres from 12.7 to 12.5`
### What happened?

We are using crossplane to manage our RDSInstances in AWS. We use argo-cd to get the objects synced. Now when we deployed RDSInstance in crossplane we used posgres `engineVersion: 12` and (`autoMinorVersionUpgrade: true`) we didn't mention this value because according to the documetation, its default value is true,  so the expected behaviour is, it will upgrade the RDS to the latest available minor version. But after upgrading, crossplane replaced engineVersion with 12.5 which is the version on which the RDS instance was initially on. We are unable to pinpoint why crossplane is trying to downgrade the instance version.
P.S. - I don't think crossplane should update the minor-version in spec. That might cause unnecessary reconciliations 

### How can we reproduce it?
```
kind: RDSInstance
metadata:
  name: some-rdsinstance
spec:
  deletionPolicy: Orphan
  forProvider:
    allocatedStorage: 100
    availabilityZone: us-west-2b
    backupRetentionPeriod: 7
    dbInstanceClass: db.m5.large
    dbParameterGroupName: default
    dbSubnetGroupName: XXXX-XXXX-XXXX
    deletionProtection: true
    enablePerformanceInsights: true
    engine: postgres
    engineVersion: '12'
    masterUsername: master_user
    multiAZ: false
    performanceInsightsRetentionPeriod: 7
    preferredBackupWindow: '06:21-06:51'
    preferredMaintenanceWindow: 'mon:00:00-mon:01:00'
    publiclyAccessible: true
    region: us-west-2
    storageEncrypted: true
    storageType: gp2
    tags:
      - key: crossplane-kind
        value: rdsinstance.database.aws.crossplane.io
      - key: crossplane-name
        value: master-postgres-instance
        value: default
    vpcSecurityGroupIds:
      - XXXX-XXXX
  providerConfigRef:
    name: default
  writeConnectionSecretToRef:
    name: master.rds-admin-connection
    namespace: master

```

Crossplane version:  crossplane-1.3.1
Kubernetes version: v1.19.7

akesser:
If a rds dbinstance is deployed with AutoMinorVersionUpgrade set to true, and aws updates the minor version, the controller thinks there is a change. In this pull request, the isUpToDate method is updated to check only the major version of the db in this case.

Fixes #832

I have:

Tested manually

### DIFF
--- a/pkg/controller/rds/dbinstance/setup.go
+++ b/pkg/controller/rds/dbinstance/setup.go
@@ -326,9 +326,25 @@ func (e *custom) isUpToDate(cr *svcapitypes.DBInstance, out *svcsdk.DescribeDBIn
 		return false, err
 	}
 
+	wantedVersion := *cr.Spec.ForProvider.EngineVersion
+	currentVersion := *db.EngineVersion
+
+	versionChanged := wantedVersion != currentVersion
+
+	if versionChanged && aws.BoolValue(cr.Spec.ForProvider.AutoMinorVersionUpgrade) {
+		wantedMaiorList := strings.Split(wantedVersion, ".")
+		wantedMaior := wantedMaiorList[0]
+		currentMaiorList := strings.Split(currentVersion, ".")
+		currentMaior := currentMaiorList[0]
+
+		versionChanged = wantedMaior != currentMaior
+
+	}
+
 	diff := cmp.Diff(&svcapitypes.DBInstanceParameters{}, patch, cmpopts.EquateEmpty(),
 		cmpopts.IgnoreTypes(&xpv1.Reference{}, &xpv1.Selector{}, []xpv1.Reference{}),
 		cmpopts.IgnoreFields(svcapitypes.DBInstanceParameters{}, "Region"),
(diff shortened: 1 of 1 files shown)


haarchri (approved):
LGTM and tested in our environment

andy-townsend:
I have `autoMinorVersionUpgrade: true` set on an RDSInstance that I've deployed as Postgres 15.2. If I manually upgrade the instance in the console to 15.3 via the AWS console (as AWS would presumably do in the next maintenance window), then the CR just goes into a `Synced: False` status and just see this sort of error on the CR. 
```
Message:               update failed: cannot modify RDS instance: api error InvalidParameterCombination: Cannot upgrade postgres from 15.3 to 15.2
```
Is there something I've missed here as the PR seems to suggest that this shouldn't be the case now? @haarchri

PR #1544, opened 2022-10-30
fix version string bug

condaatje:
EKS has deprecated 1.16, making this change necessary for the cluster to provision.

although 1.25 is the latest release, EKS only supports 1.20-1.23


haarchri:
thanks - can you check DCO signing ? then we can merge it

condaatje:
fixed, if force-push from gitlab codespace works to get the rewritten history properly DCO'd...

haarchri:
Think the latest possible Version in EKS is 1.24 https://aws.amazon.com/de/blogs/containers/amazon-eks-now-supports-kubernetes-version-1-24/

condaatje:
saw the upstream update to 1.21 - your call, feel free to close this if preferred.

PR #1551, opened 2022-11-02
cluster.kafka: Fix cluster creation after AWS change

chlunde:

AWS has changed DescribeCluster to return AccessDeniedException instead
of BadRequestException on an invalid ARN. This breaks cluster creation.

But we should not initialize external-name with a plain string based on
the kubernetes object name, it simply doesn't make sense as the service
requires an ARN. To avoid this, call managed.WithInitializers() to clear
the external name initializer.

This change was introduced in eu-north-1 within the last couple of
months.

I have:

* Ensure we can create a cluster.
* Delete a cluster in AWS and ensure a new one is created.

### DIFF
--- a/pkg/controller/kafka/cluster/setup.go
+++ b/pkg/controller/kafka/cluster/setup.go
@@ -64,7 +64,8 @@ func SetupCluster(mgr ctrl.Manager, o controller.Options) error {
 			managed.WithPollInterval(o.PollInterval),
 			managed.WithLogger(o.Logger.WithValues("controller", name)),
 			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
-			managed.WithConnectionPublishers(cps...)))
+			managed.WithConnectionPublishers(cps...),
+			managed.WithInitializers()))
 }
 
 func preDelete(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.DeleteClusterInput) (bool, error) {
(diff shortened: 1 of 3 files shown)


pkg/controller/kafka/cluster/setup.go
```diff
@@ -117,7 +118,7 @@ func postObserve(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.Describ
 }
 
 func preCreate(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.CreateClusterInput) error {
-	obj.ClusterName = awsclients.String(meta.GetExternalName(cr))
+	obj.ClusterName = awsclients.String(cr.GetName())
```
haarchri, line 121: In other PRs we say: we can never use cr.Name, the source of truth is always the annotation. 
https://github.com/crossplane-contrib/provider-aws/pull/992#discussion_r764408679

haarchri: @chlunde any updates ?

haarchri:
we merged https://github.com/crossplane-contrib/provider-aws/pull/1612

pkg/controller/kafka/cluster/setup.go
haorenfsa: This change is correct. The `external-name` should be the uniq id of the resource, which is always the `ARN` in AWS which is provided by AWS.  (check the `postCreate()`: `meta.SetExternalName(cr, awsclients.StringValue(obj.ClusterArn))`)

However, the ClusterName is the user defined name. Which is currently not offered in parameters. I think it's reasonable to use the name of CR

PR #1553, opened 2022-11-03
fix(function): check LastUpdateStatus before Update

schroeder-paul:

- Improves PR: https://github.com/crossplane-contrib/provider-aws/pull/1454 
- and makes it possible to change the imageURI on the fly without deleting and reapplying the function.

I have:

### DIFF
--- a/pkg/controller/lambda/function/setup.go
+++ b/pkg/controller/lambda/function/setup.go
@@ -2,6 +2,7 @@ package function
 
 import (
 	"context"
+	"time"
 
 	svcsdk "github.com/aws/aws-sdk-go/service/lambda"
 	svcsdkapi "github.com/aws/aws-sdk-go/service/lambda/lambdaiface"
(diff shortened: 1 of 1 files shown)


MisterMX (approved):
LGTM. Thanks a lot @schroeder-paul

szymonpk:
1. Putting there `30` is objectively worst case scenario, it just makes function deployment longer. From my testing those changes on API side were in sub-seconds 'realm', even `15` was too much.
2. 'Thanks' for taking my code, instead of helping to get a review :ok_hand:

MisterMX:
> Why we couldn't finish this in the previous PR?

A new PR was opened because the old one was inactive for quite a while and @schroeder-paul needed the fix.

> Putting there 30 is objectively worst case scenario, it just makes function deployment longer. From my testing those changes on API side were in sub-seconds 'realm', even 15 was too much.

I personally don't feel confident in bombarding AWS with requests just to save a few seconds of deploy time. However, I am open to another PR to specify a different wait time that everyone can agree one.

PR #1561, opened 2022-11-10
fix(batch): check-diff + SDK update

wotolom:

Quick fix for [Batch merge](https://github.com/crossplane-contrib/provider-aws/pull/1259), that was not properly rebased after SDK update.
This should fix the check-diff.
To avoid future breaking changes, I also removed the AWS deprecated marked field `ComputeResources.ImageId`.

I will follow up with another PR that handles the changes that came for - mostly - `ComputeEnvironment` with that SDK bump. 
Several children from field `ComputeResources` that were not updatable before, are so now under specific circumstances and seem to need to be manually taken care of in the controller.

I have:

only partly. Will test more in next PR.


MisterMX (approved):
LGTM. Thanks for the fix @wotolom

PR #1570, opened 2022-11-14
Fix documentation links in AUTHENTICATION.md

bobh66:
Signed-off-by: Bob Haddleton <bob.haddleton@nokia.com>

Updated links in the AUTHENTICATION.md file to point at Crossplane v1.10 docs instead of v1.6.

It would be better to point at a "latest" link rather than a specific release, but that will require updates in Crossplane docs.

I have:

Verified that the links are correct


haarchri (approved):
LGTM

PR #1585, opened 2022-11-23
feature(emr-on-eks): implement emr on eks

### LINKED ISSUE #1564: Add support for EMR on EKS
I have a customer who is interested in having EMR on EKS with Crossplane to drive their big data initiatives. EMR on EKS was introduced ~2 years ago to help AWS customers run EMR workloads on EKS. 

ACK recently added support for EMR on EKS (https://github.com/aws-controllers-k8s/emrcontainers-controller). ACK codegen should work to help generate this feature. I would be happy to implement this.

nabuskey:

This change adds [EMR on EKS](https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/emr-eks.html) feature to this provider. 
Note:
-  For this to work, the EKS cluster must be configured to support it. Steps from 1 - 8 in [this documentation](https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up.html) must be followed. 
These steps can be put into a composition; however, I believe it's out of scope for this PR. 
- AWS resources for this service are not explicitly removed. Instead, they transition into one of their end states. For example, Virtual Clusters are not removed when `deleteVirtualCluster` is called, it transitions into `TERMINATED` state. Therefore, hooks are necessary to handle them correctly. 

Fixes #1564

I have:

Tested on my EKS clusters with [required configurations](https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up.html). 

```bash
# kubectl get jobrun
NAME           READY   SYNCED   EXTERNAL-NAME
test-job-run   True    True     00000003156b3pv4acb
# kubectl get virtualCluster
NAME           READY   SYNCED   EXTERNAL-NAME
test-cluster   True    True     bx8vka0gha35o8mds75qm4891
```


vara-bonthu:
Nice one @nabuskey 🤘

haarchri:
@nabuskey  can you rebase to get rid of check-diff error ?

nabuskey:
@haarchri done. Looks like lint failed because it reached timeout. Weird though. It passed fine last time. May need to increase the limit or concurrent runs.

pkg/controller/emrcontainers/virtualcluster/setup.go
```diff
@@ -0,0 +1,162 @@
...
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
+			managed.WithPollInterval(o.PollInterval),
+			managed.WithLogger(o.Logger.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+			managed.WithConnectionPublishers(cps...)))
+}
+
+func preObserve(ctx context.Context, cr *svcapitypes.VirtualCluster, input *svcsdk.DescribeVirtualClusterInput) error {
+	externalName := meta.GetExternalName(cr)
+	if externalName == cr.Name {
```
haarchri, line 68: @nabuskey why you need this it block?

nabuskey: EMR containers does input parameter validation and return `ValidationException` if the ID does not satisfy `[0-9a-z]+`. This is a problem when observing a VirtualCluster or a JobRun for the first time because external name is set to the CR's name and CR names could potentially do not satisfy the regex expression. So this is done to ensure API returns 404 during observe for the first time.
It looks like you can't return non-404 error during observe, otherwise this is hit: https://github.com/crossplane/crossplane-runtime/blob/65044f043902ebb1ab1e1bb8685c3208f1c301d2/pkg/reconciler/managed/reconciler.go#L781
I thought about setting the generator to look for ValidationException as 404, but it didn't feel right.

haarchri (approved):
Thanks @nabuskey  i tested the resources working as expected

PR #1587, opened 2022-11-24
add change predication for all controllers

### LINKED ISSUE #1572: Backoff is not worked
### What happened?

Backoff is not worked in some status (e.g. delete bucket but the bucket is not empty).

Now crossplane use `For` with no `Options` as the event handler. It means all of the update events will let the queue add the item (by `queue.Add`) and trigger the `Reconcile`.

However, if the status is updated when reconciling, the item will be re-added into the queue directly no matter what result of reconciling is. 

For example, the bucket is failed to be deleted because it is not empty. Then the crossplane will update the error message with `aws request id ` into the condition. However, the `aws request id` is always changed, so the item is always updated and re-enqueued without any backoff.

A reasonable fix is adding the `GenerationChangedPredicate` options into `For` function.

Crossplane version:

liubog2008:

Fixes #1572 

I have:


haarchri:
@muvaf any chance to check this change ? or is this something for the runtime ?

pkg/controller/cognitoidentity/identitypool/setup.go
```diff
@@ -58,7 +58,12 @@ func SetupIdentityPool(mgr ctrl.Manager, o controller.Options) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Named(name).
 		WithOptions(o.ForControllerRuntime()).
-		For(&svcapitypes.IdentityPool{}).
+		For(&svcapitypes.IdentityPool{},
```
tehlers320, line 61: This is missing the imports.

```
        "sigs.k8s.io/controller-runtime/pkg/builder"
	"sigs.k8s.io/controller-runtime/pkg/predicate"
```

liubog2008: fixed

liubog2008:
@haarchri @muvaf any progressing?

negz:
@liubog2008 Please update this PR with more information around how this fixes the issue.

I tend to agree with @haarchri that it may make sense to try fix this in crossplane-runtime rather than updating every controller in every provider manually.

tehlers320:
I have this forked and built into v0.30.1 

Here is my result (metric is in request per second to the aws api)

crossplaneRPS

This type of problem can cost a company THOUSANDS of dollars a month in AWS billing as Cloudtrail is very un-forgiving. Even with this fix i would argue that the back-off is not high enough, i still see too many api calls.

negz (changes requested):
I'd like to hold this change until I've had time to dig deeper.

* I'm not sure updating every controller in every provider is the way to go. We might want to add something to crossplane-runtime to help.
* I'm not confident that we don't have reconcilers that are dependent on the behaviour this change would appear to prevent (i.e. requeueing even if nothing changed about the resource).

liubog2008:
> I'm not sure updating every controller in every provider is the way to go. We might want to add something to crossplane-runtime to help.
LGTM, my fix is not elegant. I think a new `For` function is more reasonable.

turkenh:
Closing in favor of https://github.com/crossplane-contrib/provider-aws/pull/1705

PR #1605, opened 2022-12-16
Update ini package version

jvitor03:
Signed-off-by: João Vitor Costa <jvitor03@gmail.com>

This PR updates the https://github.com/go-ini/ini package version used by the provider-aws.

The "default section" in the go-ini package is not a const anymore, so the I've changed to a literal in the crossplane provider-aws to keep the const usage, as recommended here: https://github.com/go-ini/ini/blob/main/ini.go#L31 

Im issuing this request, because one of our automations uses [Kops](https://github.com/kubernetes/kops) and Crossplane AWS provider as libraries and both relies in the go-ini package. To be able to update our Kops version to 1.25, we need Crossplane to be able to run in a newer version of go-ini.

I have:


MisterMX (approved):
LGTM. Thanks @jvitor03!

PR #1607, opened 2022-12-21
fix(docdb/dbcluster): remove unwanted dbsubnetgroup isuptodate-check

wotolom:

The controller for docdb/DBCluster checks if the field `DBSubnetGroup` isUpToDate. However it is not updatable. (e.g. see [AWS CLI docdb - modify-db-cluster](https://docs.aws.amazon.com/cli/latest/reference/docdb/modify-db-cluster.html) and [AWS DocDB DBCluster docs](https://docs.aws.amazon.com/documentdb/latest/developerguide/db-cluster-parameters.html))

This change removes that check (+ unnecessary tests).

I have:

only through `make reviewable test`


MisterMX (approved):
Thanks @wotolom. I think it makes sense to remove that check if updates are not supported by AWS.

PR #1619, opened 2022-12-30
feat(cache): fix version handling for #1617

### LINKED ISSUE #1617: Elasticache cluster: Specific version selection is not allowed for 'x.x.x', please use 'x.x'
### What happened?
When creating Elasticache cluster with engineVersion: "6.2.6" for example , the next error appears - `InvalidParameterValue: Specific version selection is not allowed for '6.2.6', please use '6.2'`

### How can we reproduce it?
Create a new Elasticache cluster with x.x.x enging version

haarchri:
Signed-off-by: Christopher Paul Haar <christopherpaul.haar@dkb.de>

changed version handling for cache cluster

Fixes #1617

I have:

added new example

```
kubectl get managed
NAME                                                          STATUS      READY   SYNCED   AGE
cachecluster.cache.aws.crossplane.io/aws-memcached-standard   available   True    True     19m
cachecluster.cache.aws.crossplane.io/redis-cache-cluster      available   True    True     19m
```

### DIFF
--- a/pkg/clients/elasticache/elasticache.go
+++ b/pkg/clients/elasticache/elasticache.go
@@ -40,7 +40,10 @@ import (
 	clients "github.com/crossplane-contrib/provider-aws/pkg/clients"
 )
 
-const errCheckUpToDate = "unable to determine if external resource is up to date"
+const (
+	errCheckUpToDate = "unable to determine if external resource is up to date"
+	errVersionInput  = "unable to parse version number"
+)
 
 // A Client handles CRUD operations for ElastiCache resources.
 type Client interface {
@@ -652,7 +655,7 @@ func IsSubnetGroupUpToDate(p cachev1alpha1.CacheSubnetGroupParameters, sg elasti
 }
 
 // GenerateCreateCacheClusterInput returns Cache Cluster creation input
-func GenerateCreateCacheClusterInput(p cachev1alpha1.CacheClusterParameters, id string) *elasticache.CreateCacheClusterInput {
+func GenerateCreateCacheClusterInput(p cachev1alpha1.CacheClusterParameters, id string) (*elasticache.CreateCacheClusterInput, error) {
 	c := &elasticache.CreateCacheClusterInput{
 		AZMode:                     elasticachetypes.AZMode(aws.ToString(p.AZMode)),
 		AuthToken:                  p.AZMode,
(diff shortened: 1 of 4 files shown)


ezgidemirel (approved):
thanks for the PR @haarchri! LGTM

pkg/clients/elasticache/elasticache.go
```diff
@@ -677,6 +679,16 @@ func GenerateCreateCacheClusterInput(p cachev1alpha1.CacheClusterParameters, id
...
 	}
 
+	if p.EngineVersion != nil {
+		version, err := getVersion(p.EngineVersion)
+		if err != nil {
+			return nil, err
+		}
+		c.EngineVersion = version
+	} else {
+		c.EngineVersion = p.EngineVersion
```
nabuskey, line 689: This else clause is not needed if you keep
```
EngineVersion:              p.EngineVersion,
```

pkg/clients/elasticache/elasticcachecluster_test.go
```diff
@@ -158,7 +158,7 @@ func TestGenerateCreateCacheClusterInput(t *testing.T) {
 
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
-			r := GenerateCreateCacheClusterInput(tc.in, clusterID)
+			r, _ := GenerateCreateCacheClusterInput(tc.in, clusterID)
```
nabuskey, line 161: I think there should be a test case for specifying a version.

pkg/clients/elasticache/elasticache.go
```diff
@@ -816,5 +840,40 @@ func IsClusterUpToDate(name string, in *cachev1alpha1.CacheClusterParameters, ob
...
+			}
+		}
+		// to ignore in following equal
+		desired.EngineVersion = observed.EngineVersion
+	}
+
 	return cmp.Equal(desired, observed, cmpopts.EquateEmpty(), cmpopts.IgnoreTypes(document.NoSerde{})), nil
 }
+
+func getVersion(version *string) (*string, error) {
```
nabuskey, line 864: If we want to be sure we are sending a request with a supported version, we could call `DescribeCacheEngineVersions` and store them in a map, then check against it. Might be an overkill though.

haarchri: think this is overkill :D

pkg/clients/elasticache/elasticache.go
haarchri: yes but i think its easier to see that we doing some version stuff with this variable - i think we should leave it as is

MisterMX (approved):

PR #1612, opened 2022-12-26
Fix kafka initializer

haorenfsa:
Signed-off-by: shaoyue.chen <shaoyue.chen@zilliz.com>

fix #1611

### Description of your changes

Fixes #

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

PR #1620, opened 2023-01-05
Fix kafka input clusterName

### LINKED ISSUE #1611: Cluster.kafka cannot be created
### What happened?

The resource Cluster.kafka cannot be created

### How can we reproduce it?

just use the example provided, you can change some fields to more reasonable values. but it doesn't make any difference. The reconcile failed before it can start create.
(https://github.com/crossplane-contrib/provider-aws/blob/master/examples/kafka/cluster.yaml)

Crossplane version:  (doesn't matter)

## Here is why it's going wrong:
The initializer of the managed resources will set the `external-name` annotation.
The `observe` function uses the `external-name` as the arn of the cluster to do the `DescribeCluster`. But a user is only allowed to describe the arn under his/her own useraccount.

Say a user is `670409000460`, and the region is `ap-southeast-1`.
Then the user can only describe the resources under `"arn:aws:kafka:ap-southeast-1:670409000460:cluster/bla/bla"`
So the `Observe` failed with 403.
I tried a way to set the `external-name` by my self to avoid it. But then I jump into annother error: 

```
  - lastTransitionTime: "2022-12-26T09:15:17Z"
    message: |-
      observe failed: failed to describe Cluster: NotFoundException: The requested resource doesn’t exist.
      {
        RespMetadata: {
          StatusCode: 404,
          RequestID: ""
        },
        InvalidParameter: "clusterArn",
        Message_: "The requested resource doesn’t exist."
      }
    reason: ReconcileError
    status: "False"
    type: Synced
```
So the here's the conclusion: the reconciler can only work correctly when `external-name` is not set, or set to correct value. While the initializer set it to the wrong value by default.

```go
func defaultMRManaged(m manager.Manager) mrManaged {
	return mrManaged{
		CriticalAnnotationUpdater: NewRetryingCriticalAnnotationUpdater(m.GetClient()),
		Finalizer:                 resource.NewAPIFinalizer(m.GetClient(), FinalizerName),
		Initializer:               NewNameAsExternalName(m.GetClient()),
		ReferenceResolver:         NewAPISimpleReferenceResolver(m.GetClient()),
		ConnectionPublisher: PublisherChain([]ConnectionPublisher{
			NewAPISecretPublisher(m.GetClient(), m.GetScheme()),
			&DisabledSecretStoreManager{},
		}),
	}
}

//...

// Initialize the given managed resource.
func (a *NameAsExternalName) Initialize(ctx context.Context, mg resource.Managed) error {
	if meta.GetExternalName(mg) != "" {
		return nil
	}
	meta.SetExternalName(mg, mg.GetName())
	return errors.Wrap(a.client.Update(ctx, mg), errUpdateManaged)
}

```

## Solution

We should add `WithInitializers` options to override the unintended default intializers

haorenfsa:
Signed-off-by: shaoyue.chen <shaoyue.chen@zilliz.com>

A follow-up of #1612 

The external-name should be the uniq id of the resource. Usually we use the `Arn` of the resource provided by AWS.
The ClusterName is the user defined name. Which is currently not offered in parameters. It's reasonable to use the name of CR as the ClusterName.

Fixes #1611

I have:

### DIFF
--- a/pkg/controller/kafka/cluster/setup.go
+++ b/pkg/controller/kafka/cluster/setup.go
@@ -118,7 +118,7 @@ func postObserve(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.Describ
 }
 
 func preCreate(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.CreateClusterInput) error {
-	obj.ClusterName = awsclients.String(meta.GetExternalName(cr))
+	obj.ClusterName = awsclients.String(cr.GetName())
 	obj.BrokerNodeGroupInfo = &svcsdk.BrokerNodeGroupInfo{
 		ClientSubnets:  cr.Spec.ForProvider.CustomBrokerNodeGroupInfo.ClientSubnets,
 		InstanceType:   cr.Spec.ForProvider.CustomBrokerNodeGroupInfo.InstanceType,


abrarcci:
Could we get an approval on this one please? Many Thanks

haorenfsa:
@haarchri take a look?

haarchri:
I will take a look but cr.name we normally Not using - i guess the original problem is around the initializer

haorenfsa:
The key is that we need a `clusterName` for the `createCluster` input,  and the `Parameters` struct for `Cluster.kafka` does not provide it.
So I think we've got 2 choices:
1. Use the CR name
2. Add ClusterName in `Parameters`

haarchri:
created the following PR https://github.com/crossplane-contrib/provider-aws/pull/1624 - the problem is when we using cr.name directly - people using compositions and the cr.name is generated per default - so its better to use clusterName in spec

haorenfsa:
> created the following PR https://github.com/crossplane-contrib/provider-aws/pull/1624 - the problem is when we using cr.name directly - people using compositions and the cr.name is generated per default - so its better to use clusterName in spec

Good point. I agree with u. I'll close this.

PR #1621, opened 2023-01-05
resourcerecordset: Should set AliasTarget when check if it is up to date

july2993:

When setting `AliasTarget` in `v1alpha1.ResourceRecordSetParameters`, `IsUpToDate()` will always return false since it does not fill up the field `AliasTarget` then `CreatePatch` will contain the part about `AliasTarget` when comparing with the expected state.

This will make it always try to do a useless update after observing the state.

I have:

Add unit test.
Note before fixing it, the first call of `IsUpToDate` in the added test case should be true, but it return false.


july2993:
@haarchri
Would you help take a look

haarchri (approved):
thanks for this fix

PR #1625, opened 2023-01-09
Support setting poll-interval of specify controller

july2993:
#### What is changed
Add --controller-poll to set specific controller poll interval. etc: --controller-poll managed/resourcerecordset.route53.aws.crossplane.io=1h

update file:
- pkg/controller/aws.go
- cmd/provider/main.go

all other files updated to export the controller name, we can set the poll option according the name when setting up the controller.

#### Why need this
For some resources we want to reduce the poll interval but do not want to change the global poll interval for all controllers.

For example, route53 API has limit of five requests per second (ref https://aws.amazon.com/premiumsupport/knowledge-center/route-53-avoid-throttling-errors/). The default poll interval is 60s, assuming we do not trigger an update after observing the status from aws, so when we have more than 300(5 * 60 / 1 list-resource-record-sets) `resourcerecordset`, we can not reconcile the resources per 60s actually. The workqueue depth will be high, new created resources can not be handled in time.

I have:

start provider:
```
./provider --controller-poll managed/resourcerecordset.route53.aws.crossplane.io=1h
```
check log:
```
...
{"level":"info","ts":1673263268.417485,"logger":"provider-aws","msg":"Setting controller poll interval","controller":"managed/resourcerecordset.route53.aws.crossplane.io","interval":3600}
...
```


july2993:
@haarchri Would you help take a look

pkg/controller/acm/controller.go
```diff
@@ -57,9 +57,12 @@ const (
 	errRemoveTagsFailed = "failed to remove tags for Certificate"
 )
 
+// ControllerName of this controller.
+var ControllerName = managed.ControllerName(v1beta1.CertificateGroupKind)
+
 // SetupCertificate adds a controller that reconciles Certificates.
 func SetupCertificate(mgr ctrl.Manager, o controller.Options) error {
-	name := managed.ControllerName(v1beta1.CertificateGroupKind)
```
haarchri, line 62: for what reason we need this change ?

july2993: Since I already Add `ControllerName`, so I think it is better to use that added var to avoid duplication and need to change both of these two places later if we need to change the name.

haarchri: I would prefer to remove this ,and leave it as is

july2993: reverted

haarchri: can you check again ? you have this kind of configuration in each and every resource - please cleanup =)

chlunde:
Nice! It might make sense to discuss this in a proposal to get consensus on an approach that could be adopted across all providers.

FYI:
I also have a related POC to have a different poll interval for resources that are ready and synced:
https://github.com/chlunde/crossplane-runtime/commit/42801b58ae74d60de3e785d78e2a1e839299f5b1
```go
func (r *Reconciler) getPollInterval(managed resource.Managed) time.Duration {
	if managed.GetCondition(xpv1.TypeReady).Status == v1.ConditionTrue && managed.GetCondition(xpv1.TypeSynced).Status == v1.ConditionTrue {
		return r.stablePollInterval
	}

	return r.pollInterval
}
```

But this does not provide the same flexibility, the primary idea behind this was to get new resources ready quickly and have another interval for checking for config drift to reduce CPU/audit logging/costs.

Another approach to command line options might be be to have annotations per resource.

july2993:
> Nice! It might make sense to discuss this in a proposal to get consensus on an approach that could be adopted across all providers.
> 
> FYI: I also have a related POC to have a different poll interval for resources that are ready and synced: [chlunde/crossplane-runtime@42801b5](https://github.com/chlunde/crossplane-runtime/commit/42801b58ae74d60de3e785d78e2a1e839299f5b1)
> 
> ```go
> func (r *Reconciler) getPollInterval(managed resource.Managed) time.Duration {
> 	if managed.GetCondition(xpv1.TypeReady).Status == v1.ConditionTrue && managed.GetCondition(xpv1.TypeSynced).Status == v1.ConditionTrue {
> 		return r.stablePollInterval
> 	}
> 
> 	return r.pollInterval
> }
> ```
> 
> But this does not provide the same flexibility, the primary idea behind this was to get new resources ready quickly and have another interval for checking for config drift to reduce CPU/audit logging/costs.
> 
> Another approach to command line options might be be to have annotations per resource.

I found in our env some resources keep in high workqueue depth, for example subnet:
Screen Shot 2023-01-10 at 17 28 41
It need about 1000(depth)/10(rate) = 100s to handle a newly created subnet.

This PR CAN NOT handle the resource kind if you need to observe whether it is `Ready` after creating it（enlarging the poll will make it not responsive to watch if it is Ready).

So I am also thinking about your idea too(different poll interval for Ready and not Ready, or have annotation per resource so we can control it ourselves by adding high-value poll interval after ready)

I am also considering solving it by annotation `crossplane.io/paused: "true"` which already works so I do not need to change crossplane. The basic idea is to pause it after observing Ready, and remove it once you update the spec or want it to observe again.

pkg/controller/acm/controller.go
july2993: I have checked it.
Note in `pkg/controller/aws.go` there are two duplicate setups before:
first:
https://github.com/crossplane-contrib/provider-aws/blob/ca0154a118ac2a0f377a56f8ffe9f4c447a9b20c/pkg/controller/aws.go#L239-L240
and:
https://github.com/crossplane-contrib/provider-aws/blob/ca0154a118ac2a0f377a56f8ffe9f4c447a9b20c/pkg/controller/aws.go#L332-L333

I have removed this duplicate and made sure all setup remains. 

> please cleanup =)

Do you mean to rebase the commits?

PR #1631, opened 2023-01-13
fix(rds): postUpdate for dbcluster using correct identifier

AlexLast:
Signed-off-by: Alex Last <alexrichardlast@gmail.com>

The `GenerateDescribeDBClustersInput` function is ack-generated, but unfortunately just returns an empty `DescribeDBClustersInput`. 

Other functions in this package modify the input and set the correct identifier, however this logic was missing for the `postUpdate` function, meaning post update checks were being made against the first DBCluster returned when describing DBClusters.

This change manually sets the correct identifier, so post update checks are made against the correct output.

I have:


haarchri (approved):
LGTM

PR #1645, opened 2023-01-23
Use resource.DesiredStateChanged() event filter for S3 Bucket

### LINKED ISSUE #1572: Backoff is not worked

turkenh:

Draft PR to demonstrate how this crossplane-runtime PR can be consumed to fix #1572. 

I have:

As described as Experiment 1 [here](https://github.com/crossplane/crossplane/issues/3455#issuecomment-1400230662), i.e. try to delete a non-empty S3 bucket.

Screen Shot 2023-01-23 at 16 22 53

### DIFF
--- a/pkg/controller/s3/bucket.go
+++ b/pkg/controller/s3/bucket.go
@@ -66,6 +66,7 @@ func SetupBucket(mgr ctrl.Manager, o controller.Options) error {
 		Named(name).
 		WithOptions(o.ForControllerRuntime()).
 		For(&v1beta1.Bucket{}).
+		WithEventFilter(resource.DesiredStateChanged()).
 		Complete(managed.NewReconciler(mgr,
 			resource.ManagedKind(v1beta1.BucketGroupVersionKind),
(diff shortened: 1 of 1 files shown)


chlunde:
I think this one can be closed as it was merged in #1705

MisterMX:
Closing since it is contained in #1705.

PR #1660, opened 2023-02-13
feature/opensearch-domain

akesser:
Added to code to update opensearchservice domains and added snapshotsOptions attribute as custom type to domain spec

Fixes #

I have:

Tested locally


pkg/controller/opensearchservice/domain/setup.go
```diff
@@ -334,3 +346,1125 @@ func GenerateObservation(obj *svcsdk.DomainStatus) svcapitypes.DomainObservation
...
+
+	if out.DomainStatus.VPCOptions != nil {
+		if obj.Spec.ForProvider.VPCOptions != nil {
+			if len(out.DomainStatus.VPCOptions.SecurityGroupIds) != len(obj.Spec.ForProvider.VPCOptions.SecurityGroupIDRefs) {
+				return false, nil
+			}
+			for _, objValue := range obj.Spec.ForProvider.VPCOptions.SecurityGroupIDs {
+				found := false
+				for _, valueOut := range out.DomainStatus.VPCOptions.SecurityGroupIds {
+					if *objValue == *valueOut {
```
haarchri, line 585: can we use here awsclient.string ? to check for nil ?

akesser: Updated the two places. Thanks for the hint!

haarchri (approved):
LGTM

PR #1669, opened 2023-02-21
Remove `masterUserPasswordSecretRef` as a required attribute

### LINKED ISSUE #1668: spec.forProvider.masterUserPasswordSecretRef shouldn't be required when creating a secondary Aurora cluster
I am trying to create a Global Aurora Cluster on AWS, the setup should look like this:

GlobalCluster
  - Regional Cluster 1 (Primary)
  - - Instance 1a
  - - Instance 1b
  - Regional Cluster 2 (Secondary)
  - - Instance 2a
  - - Instance 2b
 
The problem is that, when creating a new cluster `spec.forProvider.masterUserPasswordSecretRef` is mandatory for all clusters, no matter whether they are the primary or secondary cluster. It makes sense, that for the primary cluster this value is mandatory, but not for the second one. Right now I have the following problem:

1. If I don't specify a value for `spec.forProvider.masterUserPasswordSecretRef` in the secondary cluster, Crossplane won't let me create it.

2. If I do specify a value for `spec.forProvider.masterUserPasswordSecretRef`, Crossplane will let me create the CRD but AWS API will complain with the following error message:
```
create failed: cannot create DBCluster in AWS: InvalidParameterCombination: Cannot specify password for cross region replication cluster
```

So I'm afraid there is no way around that other than making `masterUserPasswordSecretRef` not mandatory but specify in the documentation that is actually mandatory if you only have one.

loalf:
Fixes #1668

When creating an Aurora Global cluster with more than 2 regional clusters, the password attribute is mandatory for the first one, but not for the second one. Right now, it's impossible to create global cluster with more than 2 regional cluster.

If you avoid setting up the `masterUserPasswordSecretRef`, Crossplane CRD will complain because this field is required and won't let you create the resource.

If you set up the `masterUserPasswordSecretRef`, the AWS API will complain that the cluster can't be created with the following error message

```
create failed: cannot create DBCluster in AWS: InvalidParameterCombination: Cannot specify password for cross region replication cluster
```

### DIFF
--- a/apis/rds/v1alpha1/custom_types.go
+++ b/apis/rds/v1alpha1/custom_types.go
@@ -193,8 +193,14 @@ type CustomDBClusterParameters struct {
 	// The password for the master database user. This password can contain any
 	// printable ASCII character except "/", """, or "@".
 	//
-	// Constraints: Must contain from 8 to 41 characters. Required.
-	MasterUserPasswordSecretRef *xpv1.SecretKeySelector `json:"masterUserPasswordSecretRef"`
+	// This parameter will be required in the following scenarios:
+	// - The first cluster for a global Aurora cluster
+	// - Any cluster as long as it doesn't belong to a global Aurora cluster
+	//
+	// This parameter is required for creation of a primary cluster. However, it is not required when attaching a secondary regional cluster to an existing global cluster.
+	//
+	// Constraints: Must contain from 8 to 41 characters.
+	MasterUserPasswordSecretRef *xpv1.SecretKeySelector `json:"masterUserPasswordSecretRef,omitempty"`
 
 	// A list of VPC security groups that the DB cluster will belong to.
 	//
--- a/package/crds/rds.aws.crossplane.io_dbclusters.yaml
+++ b/package/crds/rds.aws.crossplane.io_dbclusters.yaml
@@ -664,8 +664,13 @@ spec:
                   masterUserPasswordSecretRef:
                     description: "The password for the master database user. This
                       password can contain any printable ASCII character except \"/\",
-                      \"\"\", or \"@\". \n Constraints: Must contain from 8 to 41
-                      characters. Required."
+                      \"\"\", or \"@\". \n This parameter will be required in the
+                      following scenarios: - The first cluster for a global Aurora
+                      cluster - Any cluster as long as it doesn't belong to a global
+                      Aurora cluster \n This parameter is required for creation of
+                      a primary cluster. However, it is not required when attaching
+                      a secondary regional cluster to an existing global cluster.
+                      \n Constraints: Must contain from 8 to 41 characters."
                     properties:
                       key:
                         description: The key to select.
@@ -1090,7 +1095,6 @@ spec:
                     type: array
                 required:
                 - engine
-                - masterUserPasswordSecretRef
                 - region
                 type: object
               providerConfigRef:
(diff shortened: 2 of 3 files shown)


package/crds/rds.aws.crossplane.io_dbclusters.yaml
```diff
@@ -1090,7 +1090,6 @@ spec:
                     type: array
                 required:
                 - engine
-                - masterUserPasswordSecretRef
```
MisterMX, line 1093: This change here is going to be overwritten by kubebuilder. You have to set the json property to `omitempty` here: https://github.com/crossplane-contrib/provider-aws/blob/f0163fa10ab2856dcc9429e0c123d70b0c863e20/apis/rds/v1alpha1/custom_types.go#L197

MisterMX (changes requested):

package/crds/rds.aws.crossplane.io_dbclusters.yaml
loalf: Thanks for pointing me in the right direction!

apis/rds/v1alpha1/custom_types.go
```diff
@@ -192,9 +192,15 @@ type CustomDBClusterParameters struct {
...
+        // This parameter will be required in the following scenarios:
+        // - The first cluster for a global Aurora cluster
+        // - Any cluster as long as it doesn't belong to a global Aurora cluster
+        //
+        // It won't be required when trying to attach a secondary regional cluster to an existing global cluster.
 	//
-	// Constraints: Must contain from 8 to 41 characters. Required.
-	MasterUserPasswordSecretRef *xpv1.SecretKeySelector `json:"masterUserPasswordSecretRef"`
+	// Constraints: Must contain from 8 to 41 characters.
+	MasterUserPasswordSecretRef *xpv1.SecretKeySelector `json:"masterUserPasswordSecretRef,omitempty"`
```
MisterMX, line 203: Thanks for the change. However, you have to run `make generate` and commit the changes in `package`. For more info see the [contribution guidelines](https://github.com/crossplane-contrib/provider-aws/blob/master/CONTRIBUTING.md).

 Additionally, it would be great if you could provide an example test case.

apis/rds/v1alpha1/custom_types.go
```diff
@@ -192,9 +192,15 @@ type CustomDBClusterParameters struct {
 
 	// The password for the master database user. This password can contain any
 	// printable ASCII character except "/", """, or "@".
+        // 
+        // This parameter will be required in the following scenarios:
+        // - The first cluster for a global Aurora cluster
+        // - Any cluster as long as it doesn't belong to a global Aurora cluster
+        //
+        // It won't be required when trying to attach a secondary regional cluster to an existing global cluster.
```
MisterMX, line 200:
```suggestion
        // This parameter is required for creation of a primary cluster. However, it is not required when attaching a secondary regional cluster to an existing global cluster.
```

MisterMX (changes requested):

loalf:
@MisterMX thanks for guiding me in the right direction. I've updated the PR with the required changes (or so I think 😆 ).

apis/rds/v1alpha1/custom_types.go
loalf: How do you mean by `example test case`. Is it just providing a new example as the ones in the `examples` directory here, https://github.com/crossplane-contrib/provider-aws/tree/master/examples? If so, I'm more than happy to create a follow-up request with one 💪

MisterMX: Yes, please provide one (or more if necessary) sample MRs in `examples` with this PR.

MisterMX (approved):
LGTM now. Thanks a lot @loalf!

PR #1670, opened 2023-02-21
fix(rds): handle maximum parameters per dbclusterparametergroup update

AlexLast:

AWS only allows a `ModifyDBClusterParameterGroup` request to contain 20 parameters at a single time. Previously every `ModifyDBClusterParameterGroup` request would contain every parameter defined in the custom resource.

When defining > 20 parameters in a custom resource, it will continually fail to update with the following:
```
Warning  CannotUpdateExternalResource  1s  managed/dbclusterparametergroup.rds.aws.crossplane.io  cannot update DBClusterParameterGroup in AWS: InvalidParameterCombination: Cannot modify more than 20 parameters in a single request
```

This contribution ensures only the parameters that have changed or have not been set will be included in an update, updates are also limited to 20 parameters. If > 20 parameters have changed, the truncated parameters will be included in the next reconciliation.

I have:

- [x] Tested build locally and ensured everything is working as expected

### DIFF
--- a/pkg/controller/rds/dbclusterparametergroup/setup.go
+++ b/pkg/controller/rds/dbclusterparametergroup/setup.go
@@ -25,6 +25,10 @@ import (
 )
 
 const (
+	maxParametersPerUpdate = 20
+)
+
+const (
 	errRequireDBParameterGroupFamilyOrFromEngine = "either spec.forProvider.dbParameterGroupFamily or spec.forProvider.dbParameterGroupFamilyFromEngine is required"
 	errDetermineDBParameterGroupFamily           = "cannot determine DB parametergroup family"
 	errGetDBEngineVersion                        = "cannot decsribe DB engine versions"
@@ -39,7 +43,7 @@ func SetupDBClusterParameterGroup(mgr ctrl.Manager, o controller.Options) error 
 			c := &custom{client: e.client, kube: e.kube}
 			e.preCreate = c.preCreate
 			e.preObserve = preObserve
-			e.preUpdate = preUpdate
+			e.preUpdate = c.preUpdate
 			e.preDelete = preDelete
 			e.postObserve = postObserve
 			e.lateInitialize = lateInitialize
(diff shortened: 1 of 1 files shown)


pkg/controller/rds/dbclusterparametergroup/setup.go
```diff
@@ -202,3 +214,29 @@ func (c *custom) getDBEngineVersion(ctx context.Context, selector *svcapitypes.D
...
+	for _, v := range cr.Spec.ForProvider.Parameters {
+		existing, ok := observed[awsclients.StringValue(v.ParameterName)]
+
+		if !ok {
+			parameters = append(parameters, v)
+			continue
+		}
+
+		if awsclients.StringValue(existing.ParameterValue) != awsclients.StringValue(v.ParameterValue) ||
+			awsclients.StringValue(existing.ApplyMethod) != awsclients.StringValue(v.ApplyMethod) {
```
MisterMX, line 236: I think there are more fields to consider than just the value and the apply method: https://github.com/crossplane-contrib/provider-aws/blob/f0163fa10ab2856dcc9429e0c123d70b0c863e20/apis/rds/v1alpha1/zz_types.go#L1263

I believe `cmp.Equal` can help here.

MisterMX (changes requested):

pkg/controller/rds/dbclusterparametergroup/setup.go
AlexLast: Thanks for your review @MisterMX, this is actually just re-purposing the existing logic from `isUpToDate()`, however I can certainly expand it - will take a look

AlexLast: @MisterMX  After looking into this a little further I believe the only parameters we can actually set are `ParameterValue` and `ApplyMethod`. The other fields seem to be for describe operations only, when testing locally any change to other fields defined in the struct do nothing to the external resource.

The documentation here seems to confirm this - https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_ModifyDBClusterParameterGroup.html

I believe the reason for this may be that the `parameters` data type is shared between `ModifyDBParameterGroup` and `DescribeDBParameters`  as stated here - https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_Parameter.html

MisterMX:
LGTM. Thanks a lot @AlexLast.

Woul you mind applying the same changes to `rds.DBParameterGroup` and `docdb` as well?

MisterMX (approved):

AlexLast:
@MisterMX Sure, have raised the same for RDS `DBParameterGroup`, looks like there are some differences in DocumentDB, will raise a change for that next week 👍

PR #1675, opened 2023-02-22
mark cluster as ready during creating

### LINKED ISSUE #1674: Kafka controller is waiting for MSK creation to mark ready
Kafka controller can not determine how long it will take MSK to create. As a result, it should mark ready whenever cluster creation is successful. The crossplane credential can be used to determine when the MSK cluster is actually available.

As a result of the indeterminate install time, helm marks the chart as failed. Kubevela actually uninstalls and reinstalls it, which leads to all sorts of undesirable behavior.

drewwells:
Fixes #1674

### DIFF
--- a/pkg/controller/kafka/cluster/setup.go
+++ b/pkg/controller/kafka/cluster/setup.go
@@ -97,7 +97,7 @@ func postObserve(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.Describ
 	case string(svcapitypes.ClusterState_ACTIVE):
 		cr.SetConditions(xpv1.Available())
 	case string(svcapitypes.ClusterState_CREATING):
-		cr.SetConditions(xpv1.Creating())
+		cr.SetConditions(xpv1.Available())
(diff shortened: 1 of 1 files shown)


haarchri:
if we set the status to Available we running directly update requests if needed - so we avoid this in our resources and set Available only if the API set Available

drewwells:
I didn't understand your comment. 

MSK clusters take 30mins to provision and helm install times out in 5mins. Waiting for a completely provisioned MSK cluster is unnecessary. elasticsearch in this provider does a similar lazy provisioning to keep k8s timeouts within reason.

haarchri:
In opensearch we using the status from the API https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/controller/opensearchservice/domain/setup.go#L110

haarchri:
`helm upgrade --wait --timeout=30m`  via HelmCli you can set the timeout - is this an option for you ?

haarchri:
we using the status from the APi

PR #1681, opened 2023-02-24
fix(rds): handle maximum parameters per dbparametergroup update

AlexLast:

AWS only allows a `ModifyDBParameterGroup` request to contain 20 parameters at a single time. Previously every `ModifyDBParameterGroup` request would contain every parameter defined in the custom resource.

When defining > 20 parameters in a custom resource, it will continually fail to update with the following:
```
Warning  CannotUpdateExternalResource  1s  managed/dbparametergroup.rds.aws.crossplane.io  cannot update DBParameterGroup in AWS: InvalidParameterCombination: Cannot modify more than 20 parameters in a single request
```

This contribution ensures only the parameters that have changed or have not been set will be included in an update, updates are also limited to 20 parameters. If > 20 parameters have changed, the truncated parameters will be included in the next reconciliation.

I have:

- [x] Tested build locally and ensured everything is working as expected

cc @MisterMX

### DIFF
--- a/pkg/controller/rds/dbparametergroup/setup.go
+++ b/pkg/controller/rds/dbparametergroup/setup.go
@@ -24,6 +24,10 @@ import (
 )
 
 const (
+	maxParametersPerUpdate = 20
+)
+
+const (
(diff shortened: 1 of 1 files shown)


MisterMX (approved):
LGTM. Thanks a lot @AlexLast

PR #1685, opened 2023-03-01
feat(bucket): Merge BucketPolicy with Bucket

### LINKED ISSUE #1684: Merge `Bucket` and `BucketPolicy`

Currently, there is a separate MR offered to put a IAM resource policy on a bucket.

Since there can only be one policy per bucket, it would make sense to merge the functionalities of `BucketPolicy` into `Bucket` and deprecate the former.

See [AWS API](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-bucket-policy.html) for reference.

Add a new field `spec.forProvider.policy` to `Bucket` that provides the same functionality as in `BucketPolicy`.

MisterMX:

This adds a new field `spec.forProvider.policy` to `Bucket` to merge the functionality of `BucketPolicy` into the former.

To avoid breaking changes with existing setups, policies are never deleted by the bucket controller by default. If `spec.forProvider.policy` is nil, the controller will do nothing.

To change this behaviour the user can set the new field `spec.forProvider.policyUpdatePolicy.deletionPolicy` to `IfNull` (default is `Never`). This way, the controller will delete the external AWS policy.

This also deprecates `BucketPolicy` with a warning to use the policy field in `Bucket` instead.

Fixes #1684 

I have:

See the provided example.


examples/s3/bucket-with-policy.yaml
```diff
@@ -0,0 +1,48 @@
...
+            - s3:GetBucketLocation
+            - s3:ListBucketMultipartUploads
+            - s3:PutBucketCORS
+          effect: Allow
+          principal:
+            awsPrincipals:
+              - iamUserArnSelector:
+                  matchLabels:
+                    example: "true"
+          resource:
```
haarchri, line 27: is self-referencing here not an option ?

```
  resourceRef:
    name: test-bucket-with-policy
    policy:
      resolution: Optional
      resolve: Always
```

think since https://github.com/crossplane-contrib/provider-aws/pull/1361 self-referencing is possible

MisterMX: The policy schema is just copied from `BucketPolicy`. It does not currently support referencers.

haarchri: but we can add reference generation annotations and then it is working ? 
https://github.com/crossplane-contrib/provider-aws/pull/1685/files#diff-be80b55e7e7d430920dc7fc28366d7e7daec24b3ac55cbc127c3480b2f177e12R78

MisterMX: I don't think a referencer is necessary here since the resource can only be ARN of the bucket we are managing. I added a step in the controller that late initializes these fields if not set.

haarchri (approved):
LGTM

tcd156:
The deprecation warning log is firing in the provider-aws pod constantly, even though I have no BucketPolicies specified.

Also, given that we can't infer the ARN of the bucket, is there a suggested method to easily interpolate that value into the policy?

tcd156:
Without BucketPolicy, this creates a situation that I'm in where I want to create an S3 bucket, and a CloudFront distribution.

The Distribution needs the Bucket, but the Bucket needs the Distribution ARN in order to set the policy so that only CloudFront can communicate to the Bucket.

So neither can be created without the other.

MisterMX:
> The deprecation warning log is firing in the provider-aws pod constantly, even though I have no BucketPolicies specified.

That sounds more like an issue with the underlying `k8s.io/client-go` because its spams every log directly to stderr.

> Also, given that we can't infer the ARN of the bucket, is there a suggested method to easily interpolate that value into the policy?

When you are creating a bucket MR you are also know the external name of the bucket. You can derive the ARN of the bucket as long as you know the ID of the AWS account it is created in.

> The Distribution needs the Bucket, but the Bucket needs the Distribution ARN in order to set the policy so that only CloudFront can communicate to the Bucket.

Wouldn't it work if you create the bucket without a policy and add it later, once cloudfront has been initialized?

Issue #1673, opened 2023-02-22
Kafka is editing it's own spec


drewwells:
I noticed an odd entry in my spec which prevented MSK cluster from provisioning.

```
      configurationInfo:
        arn: arn:aws:kafka:us-east-1:575487752567:configuration/resource-controlplane/aef15868-2f1e-438f-a5fc-a7edaad68bb7-20
        arnRef:
          name: resource-controlplane
```

It appears the controller is modifying its own spec, which caused an issue here. Especially because it often leads to an issue where an invalid arn is entered in this spec that breaks kafka cluster creation. I found this entry on a cluster that was reinstalled:

```
      configurationInfo:
        arn: resource-controlplane
        arnRef:
          name: resource-controlplane
```

repo steps:
1. stand up provider-aws and provision CRs
```
kubectl apply -f examples/kafka/cluster.yaml -f examples/kafka/configuration.yaml
```
2. Inspect configurationInfo, see that arn was updated to an invalid value and cluster fails to provision
```
 k get cluster.kafka example -o yaml | grep '^    configurationInfo' -A4
    configurationInfo:
      arn: example-configuration
      arnRef:
        name: example-configuration
      revision: 1
```
3. Events show errors resolving this string since it's not an arn
```
  Warning  CannotResolveResourceReferences  6s               managed/cluster.kafka.aws.crossplane.io  cannot resolve references: mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARN: cannot get referenced resource: Configuration.kafka.aws.crossplane.io "example-configuration" not found
  Warning  CannotResolveResourceReferences  6s               managed/cluster.kafka.aws.crossplane.io  cannot resolve references: mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARN: referenced field was empty (referenced resource may not yet be ready)
  Warning  CannotCreateExternalResource     5s (x5 over 6s)  managed/cluster.kafka.aws.crossplane.io  cannot create Cluster in AWS: UnrecognizedClientException:
           status code: 403, request id:
 ```

drewwells:
I thought a bit about what's happening here. I believe this problem occurs because cluster controller is unable to find the configuration aws resource, since it does not yet exist in aws. As a result, it's falling back to a default external name instead of waiting to find the arn name. Removing this default behavior should trigger the right series of steps.

drewwells:
[...]

PR #1695, opened 2023-03-08
arn is being overridden to invalid values

drewwells:
ref #1673

### DIFF
--- a/apis/kafka/v1alpha1/zz_generated.resolvers.go
+++ b/apis/kafka/v1alpha1/zz_generated.resolvers.go
@@ -30,7 +30,6 @@ import (
 func (mg *Cluster) ResolveReferences(ctx context.Context, c client.Reader) error {
 	r := reference.NewAPIResolver(c, mg)
 
-	var rsp reference.ResolutionResponse
 	var mrsp reference.MultiResolutionResponse
 	var err error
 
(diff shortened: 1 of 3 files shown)


apis/kafka/v1alpha1/zz_generated.resolvers.go
```diff
@@ -69,24 +68,6 @@ func (mg *Cluster) ResolveReferences(ctx context.Context, c client.Reader) error
...
-			Selector:     mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARNSelector,
-			To: reference.To{
-				List:    &ConfigurationList{},
-				Managed: &Configuration{},
-			},
-		})
-		if err != nil {
-			return errors.Wrap(err, "mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARN")
-		}
-		mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARN = reference.ToPtrValue(rsp.ResolvedValue)
```
drewwells, line 87: This line is not checking than an actual arn is being resolved and instead sets the name of a resource instead of the required arn.

This would likely do the right behavior, but I see the ARN being replaced in hooks so not sure why it's even here.
```
if arn.IsARN(rsp.ResolvedValue) {
  mg.Spec.ForProvider.CustomClusterParameters.CustomConfigurationInfo.ARN = reference.ToPtrValue(rsp.ResolvedValue)
}}
```

haarchri:
We will shortly Push an PR to solve the issue in configuration resource - thanks

PR #1712, opened 2023-03-30
fix(rds): reset parameters removed from a dbclusterparametergroup

AlexLast:
The existing update logic for `DBClusterParameterGroup` will only update parameters defined in `spec.forProvider.parameters`, if a parameter is removed it will no longer be managed.

This contribution ensures parameters removed from `spec.forProvider.parameters` will be reset to their default value. Assuming this approach is ok, I'll raise a similar PR for `DBParameterGroup`.

I have:

- [x] Running a local build, testing existing and new functionality is working as expected


AlexLast:
Looks like some changes have been made in this area in https://github.com/crossplane-contrib/provider-aws/pull/1709, I'll rebase and refactor slightly.

AlexLast:
@haarchri @MisterMX would you mind reviewing?

MisterMX:
@AlexLast does this actually reset parameters that have been removed from the spec? There are changes in the update logic. Don't you have set the original values their as weel?

Also, can you implement the changes for `DBParameterGroup` as well?

AlexLast:
No problem, thanks for the review 👍🏻

Issue #1683, opened 2023-02-27
Updates for Kafka cluster and configuration are not/ not always applied


Patrick-L:
### What happened?
Updates for Kafka clusters and Kafka cluster configurations are successfully created in the cluster as CRD, but are not applied in AWS. This means for kafka-cluster they are not changed and for cluster-configurations no new revision is created.

### How can we reproduce it?
For Kafka cluster:
Add openMonitoring config to cluster crd. The changes are not visible in the aws management console. 

CRD excerpt:
```
   openMonitoring:
      prometheus:
        jmxExporter:
          enabledInBroker: true
        nodeExporter:
          enabledInBroker: true
```

For Cluster configuration:
Add new property like "delete.topic.enable=true". The changes are not visible in the aws management console. 

CRD excerpt:
```
    properties:
      - delete.topic.enable=true
```

### What environment did it happen in?
Crossplane version: v0.37.0
Cloud provider: AWS

hitman99:
We have the same issue. No updates to either Cluster of Configuration gets updated. It seems that only `Create` and `Update` hooks were implemented for these resources and it's using `nopUpdate` for updates.

PR #1714, opened 2023-03-31
feat(kafka): Add update function for cluster

akesser:
We implemented the functions to update a Kafka cluster

fixes the update of a Kafka cluster as mentioned in #1683 

I have:

Tested locally against aws account


pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,776 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+	}
+
+	if !isClientAuthenticationUpToDate(forProvider.ClientAuthentication, clusterInfo.ClientAuthentication) {
+		return false, nil
+	}
+
+	if !isTagsUpToDate(forProvider.Tags, clusterInfo.Tags) {
+		return false, nil
+	}
+	return true, nil
```
MisterMX, line 263: Packing all the `if`s in a switch to make things shorter and more concise:

```suggestion
	switch {
	// A cluster can not be updated while not in active status, therefore we consider the cluster as up to date
	case aws.StringValue(clusterInfo.State) != stateActive:
		return true, nil
	case !isInstanceTypeUpToDate(forProvider.CustomBrokerNodeGroupInfo, clusterInfo.BrokerNodeGroupInfo),
		!isStorageInfoUpToDate(forProvider.CustomBrokerNodeGroupInfo, clusterInfo.BrokerNodeGroupInfo),
		!isNumberOfBrokerNodesUpToDate(forProvider, clusterInfo),
		!isEnhancedMonitoringUpToDate(forProvider, clusterInfo),
		!isLoggingInfoUpToDate(forProvider.LoggingInfo, clusterInfo.LoggingInfo),
		!isOpenMonitoringInfoUpToDate(forProvider.OpenMonitoring, clusterInfo.OpenMonitoring),
		!isCustomConfigurationInfoUpToDate(forProvider.CustomConfigurationInfo, clusterInfo.CurrentBrokerSoftwareInfo),
		!isKafkaVersionUpToDate(forProvider.KafkaVersion, clusterInfo.CurrentBrokerSoftwareInfo),
		!isEncryptionInfoUpToDate(forProvider.EncryptionInfo, clusterInfo.EncryptionInfo),
		!isClientAuthenticationUpToDate(forProvider.ClientAuthentication, clusterInfo.ClientAuthentication),
		!isTagsUpToDate(forProvider.Tags, clusterInfo.Tags):
		return false, nil
	}
	return true, nil
```

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,776 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+	cr, ok := mg.(*svcapitypes.Cluster)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+	if aws.StringValue(cr.Status.AtProvider.State) != stateActive {
+		return managed.ExternalUpdate{}, errors.New(errStateForUpdate)
+	}
+	input := GenerateDescribeClusterInput(cr)
+
+	if cr.Status.AtProvider.ClusterARN != nil {
```
MisterMX, line 607: `meta.GetExternalName(cr)` should be used to get the cluster ARN.

MisterMX (changes requested):

pkg/controller/kafka/cluster/setup.go
akesser: Thanks for your suggestions!

haarchri:
@akesser can you Check for DCO ?

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+			} else if current.StorageInfo.EbsStorageInfo != nil {
+				return false
+			}
+		} else if current.StorageInfo != nil {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+	return true
```
MisterMX, line 277: Nested `if`s are hard to read. Could please change it into a more linear form?

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+				return false
+			}
+		} else if current.BrokerLogs != nil {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+
+	return true
```
MisterMX, line 366: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+			} else if current.Prometheus.NodeExporter != nil {
+				return false
+			}
+		} else if current.Prometheus != nil {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+	return true
```
MisterMX, line 409: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+			return false
+		}
+		if aws.Int64Value(wanted.Revision) != aws.Int64Value((current.ConfigurationRevision)) {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+
+	return true
```
MisterMX, line 428: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+		if current == nil {
+			return false
+		}
+		if aws.StringValue(wanted) != aws.StringValue(current.KafkaVersion) {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+	return true
```
MisterMX, line 442: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+			if aws.BoolValue(wanted.EncryptionInTransit.InCluster) != aws.BoolValue(current.EncryptionInTransit.InCluster) {
+				return false
+			}
+		} else if current.EncryptionInTransit != nil {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+	return true
```
MisterMX, line 485: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+				return false
+			}
+		} else if current.Unauthenticated != nil {
+			return false
+		}
+	} else if current != nil {
+		return false
+	}
+
+	return true
```
MisterMX, line 559: Same here. Please change it into a linear form.

pkg/controller/kafka/cluster/setup.go
```diff
@@ -186,3 +207,745 @@ func LateInitialize(cr *svcapitypes.ClusterParameters, obj *svcsdk.DescribeClust
...
+			unauthenticated := &svcsdk.Unauthenticated{}
+			if wanted.Unauthenticated.Enabled != nil {
+				unauthenticated.Enabled = wanted.Unauthenticated.Enabled
+			}
+			output.Unauthenticated = unauthenticated
+		}
+
+		return output
+	}
+	return nil
```
MisterMX, line 871: Same here. Please change it into a linear form.

MisterMX (changes requested):
@akesser could you unwrap the ifs into a more linear form. Otherwise the code is too hard to read. Also you might want to checkout `cmp.Equal` to see if can help you to save some of that code.

akesser:
According to its self description, cmp.Equal is not meant to be used in production:

> This package is intended to be a more powerful and safer alternative to reflect.DeepEqual for comparing whether two values are semantically equal. It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values. Its propensity towards panicking means that it's unsuitable for production environments where a spurious panic may be fatal. (from: https://pkg.go.dev/github.com/google/go-cmp/cmp#pkg-overview) 

In my opinion, the used ifs is the best way to keep the code readable, keeping it clear what property is under test right now. Using linearized forms of ifs requires many repetitions, making it hard to see the difference in the if to the previous and following test.

The code in other files also uses nested ifs, not to mention auto generated code.

akesser:
Additionally to the warning, that cmp.Equal should not be used, the different naming of some properties, like IAM vs Iam or EBSStorageInfo vs. EbsStorageInfo would lead to either  ignoring them in cmp.Equal and comparing them by hand again or writing transformers that do the same comparison the code already does.

haarchri:
think the code is Not easier to read when switching to cmp.Equal

MisterMX (approved):
Alright, then its decided. Thank you for your contribution @akesser!

PR #1728, opened 2023-04-18
remove required field interfaceType from ec2 instance networkInterface

lubbyhst:

We want to create ec2 instances with existing network interfaces. To do that we are referencing the network interface id of the existing ENI within the field "networkInterfaceId".  This is an optional filed.
But we need to set the required fields too. These include the field "interfaceType". But the AWS API does not allow specify bot "networkInterfaceId" and "interfaceType". See error message after creating the the CRD.

**Error Message from AWS API:**
```
create failed: failed to create the Instance resource: api error InvalidParameterCombination: A network interface may not specify both a network interface ID and an interface 
```

Here is the example Resource that reproduce this error.
**Invalid example**:
```yaml
kind: Instance
metadata:
  name: test-instance
spec:
  forProvider:
    imageId: ami-testid
    instanceType: t3.micro
    launchTemplate:
      launchTemplateId: ""
      launchTemplateName: metadata-options-crossplane
    monitoring:
      enabled: false
    networkInterfaces:
    - deviceIndex: 0
      groups: []
      interfaceType: interface
      networkInterfaceId: eni-0a5798665e2d4557
    region: eu-central-1
  providerConfigRef:
    name: aws-config
  writeConnectionSecretToRef:
    name: test-instance
    namespace: crossplane-system
```

To fix this issue. We removed the field "interfaceType" from required list from the CRD ec2.aws.crossplane.io_instances.yaml
After that we were able to create the resource and the provider created the ec2 instance as described.

**Valid example:**
```yaml
kind: Instance
metadata:
  name: test-instance
spec:
  forProvider:
    imageId: ami-testid
    instanceType: t3.micro
    launchTemplate:
      launchTemplateId: ""
      launchTemplateName: metadata-options-crossplane
    monitoring:
      enabled: false
    networkInterfaces:
    - deviceIndex: 0
      groups: []
      networkInterfaceId: eni-0a5798665e2d4557
    region: eu-central-1
  providerConfigRef:
    name: aws-config
  writeConnectionSecretToRef:
    name: test-instance
    namespace: crossplane-system
```

I have:

We testet this change on our crossplane cluster by modifying the CRD directly inside the cluster. It worked like charm.

### Open Question

I don't know if we need to change more than this CRD or if they are generated? Maybe a maintainer could clarify if we need to change the code at some other places too?

### DIFF
--- a/package/crds/ec2.aws.crossplane.io_instances.yaml
+++ b/package/crds/ec2.aws.crossplane.io_instances.yaml
@@ -753,7 +753,6 @@ spec:
                       required:
                       - deviceIndex
                       - groups
-                      - interfaceType
                       type: object
                     type: array
                   placement:


lubbyhst:
@dev_tem

is there a problem with this fix? Would be great to see this live :)

MisterMX:
Sorry @lubbyhst for the late response.

Removing the line in `package/crds` is not enough as it those files are generated. You have to make the change it in https://github.com/crossplane-contrib/provider-aws/blob/594d12c084d9642e7b0a91635dd4bd485edbaae8/apis/ec2/manualv1alpha1/instances_types.go#L131. (It should be converted to a pointer as well).

PR #1730, opened 2023-04-18
Handle VPCCIDRBlock delete case in Observe

### LINKED ISSUE #1729: VPCCIDRBlock resource is stuck in the deleting state
### What happened?
When trying to delete a VPCCIDRBlock resource it gets stuck in the deleting state.  The cloud resource is deleted from the VPC but the k8s MR stays in the deleting state (finalizer exists) until either the VPC is deleted or the finalizer is removed manually.

### How can we reproduce it?
- Create a VPCCIDRBlock resource on an existing VPC
- Delete the VPCCIDRBlock resource
- observe that kubectl delete does not return and the cloud resource is deleted but the k8s MR is stuck
- 
Crossplane version: 1.11.3

The problem is in the Observe() function - it is ignoring the CIDRBlockNotFound error which causes reconciliation to continue to wait, instead of returning ResourceExists: false so reconciliation will remove the finalizer.

I'll push a PR

bobh66:
The VPCCIDRBlock association is never actually removed from the VPC when the VPCCIDRBlock is deleted, it is put into the `disassociated` state where it remains until the VPC is deleted.

This change modifies the Observe function to check to see if the managed resource has been deleted when it finds the VPCCIDRBlock is in the `disassociated` state and return `ResourceExists: false` so that the reconciler will remove the finalizer and allow the MR to be deleted.

Fixes #1729 

I have:

Verified that the VPCCIDRBlock resource is deleted properly when this fix is applied.

### DIFF
--- a/pkg/controller/ec2/vpccidrblock/controller.go
+++ b/pkg/controller/ec2/vpccidrblock/controller.go
@@ -140,6 +140,12 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 	case awsec2types.VpcCidrBlockStateCodeAssociating:
 		cr.SetConditions(xpv1.Creating())
 	case awsec2types.VpcCidrBlockStateCodeDisassociated:
+		if meta.WasDeleted(mgd) {
+			return managed.ExternalObservation{
+				ResourceExists:   false,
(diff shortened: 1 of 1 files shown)


haarchri (approved):
Thanks for this

PR #1734, opened 2023-04-20
feat(database/rds): Add password constraints

MisterMX:

This adds the ability to set additional password constraints for `DBCluster.docdb`, `DBInstance.rds`, `DBCluster.rds` and `RDSInstance.database`.

I have:

Unit tests.


apis/docdb/v1alpha1/custom_types.go
```diff
@@ -141,6 +143,10 @@ type CustomDBClusterParameters struct {
 	// Constraints: Must contain from 8 to 100 characters.
 	MasterUserPasswordSecretRef *xpv1.SecretKeySelector `json:"masterUserPasswordSecretRef,omitempty"`
 
+	// MasterUserPaswordConstraints specifies additional constraints
+	// that apply to the password referenced by MasterUserPasswordSecretRef.
+	MasterUserPaswordConstraints *common.PasswordConstraints `json:"masterUserPaswordConstraints,omitempty"`
```
chlunde, line 148: typo, two "s" in password

chlunde:
I wonder if it would be better to just make the default passwords _really good_ (enough for 95% of users), and have the final 5% just pre-generate a password (provider-secret-password? :smile:) with features like this.

chlunde:
> However, we do allow our users to specify their own passwords so we need it to be checked.
[...]
> Would it still be possible to get theses changes merged?

I think this is a fairly special case, I'd leave it to @haarchri to decide. I think the best practise is to generate a random master password and never expose it to humans, only use it in other tooling to create less privileged users, so this feature is _normally_ not a requirement.
On the other hand, it's not a lot of code inside each controller.

MisterMX:
Closing it since this use case is too specific.

PR #1764, opened 2023-05-23
Handle vpc peer accept in a different region

### LINKED ISSUE #1071: AWS VPCPeeringConnection with acceptRequest=true don't work when VPC regions are different
### What happened?

We are trying to create VPC Peering connection between two VPCs using `VPCPeeringConnection` CR with the `AcceptRequest` option set to true. This works for VPCs that are in the same region, but it doesn't for VPCs in different regions. This should work according to the [documentation](https://doc.crds.dev/github.com/crossplane/provider-aws/ec2.aws.crossplane.io/VPCPeeringConnection/v1alpha1@v0.23.0#spec-forProvider-acceptRequest), since they are in the same tenant. The following error can be seen in the status when describing the CR:

```
Incorrect region (us-east-1) specified for this request. VPC peering connection pcx-xxxxxx must be accepted in region sa-east-1
```

We did a investigation, and discovered that the problem is that the controller is trying to do the peering accept in the same region as the requester, instead of using the peer region. This happens because [the same AWS EC2 client is used for both operations](https://github.com/crossplane/provider-aws/blob/master/pkg/controller/ec2/vpcpeeringconnection/setup.go#L77). 

We also made a PoC with the controller by creating a new session with the `peerRegion` from the spec, and it worked as described in the documentation.

### How can we reproduce it?

1) Create a VPCPeeringConnection CR between two VPCs in different regions, but in same AWS account with the `acceptRequest` set to true:

```
kind: VPCPeeringConnection
metadata:
  name: test-peering
spec:
  forProvider:
    acceptRequest: true
    peerOwnerID: "zzz"
    peerRegion: sa-east-1
    peerVPCID: vpc-yyy
    region: us-east-1
    vpcID: vpc-xxx
  providerConfigRef:
    name: default
  writeConnectionSecretToRef:
    name: peering-state
    namespace: crossplane-tests
```
After a while, the peering should be stuck in a `pending acceptance` state with the error described in the section above. This can be checked by doing a describe in the CR:

`kubectl describe vpcpeeringconnections.ec2.aws.crossplane.io test-peering`

* Crossplane version: v1.6.1
* Crossplane provider AWS version: v0.23.0
* Cloud provider or hardware configuration: AWS EC2 c5.2xlarge
* Kubernetes version (use `kubectl version`): 1.22.0
* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift): Vanilla
* OS (e.g. from /etc/os-release): Ubuntu 20.04.2 LTS
* Kernel (e.g. `uname -a`): 5.8.0-1041-aws

bobh66:
Modified the VPC Peering Connection postObserve method to handle the scenario where the VPC Peer (accepter)
is in a different region than the requester.

Fixes #1071 

I have:

Tested VPC Peering Connection resource with VPCs in different regions and in the same region, and verified that the
automated accept works, and the modify attributes also succeeds.

### DIFF
--- a/pkg/controller/ec2/vpcpeeringconnection/setup.go
+++ b/pkg/controller/ec2/vpcpeeringconnection/setup.go
@@ -82,17 +82,29 @@ func filterList(cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPee
 	return resp
 }
 
-func (e *custom) postObserve(_ context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPeeringConnectionsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) { // nolint:gocyclo
+func (e *custom) postObserve(ctx context.Context, cr *svcapitypes.VPCPeeringConnection, obj *svcsdk.DescribeVpcPeeringConnectionsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) { // nolint:gocyclo
 	if err != nil {
 		return managed.ExternalObservation{}, err
+	}
+
+	// The accept and modify operations for the Peer VPC have to be executed in the PeerRegion
+	var pc svcsdkapi.EC2API
+	if *cr.Spec.ForProvider.PeerRegion != cr.Spec.ForProvider.Region {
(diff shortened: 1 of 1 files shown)


haarchri (approved):
thanks for Implementation - tested the feature is working

PR #1765, opened 2023-05-24
feat(rds): Prevent update call for version downgrade

MisterMX:

RDS downgrades are not allowed by AWS.

This prevents an upgrade call when the only the major version is defined or the spec version is lower the external version.

Downgrades are silently ignored. To detect an eventual misconfiguration the current engine version is added to the status.

This apply to the following resources:
* `rds.DBCluster`
* `rds.DBInstance`
* `database.RDSInstance`

I have:

Unit tests


haarchri (approved):
LGTM - tested locally is working thanks

PR #1777, opened 2023-06-15
fix: Get web identity token filepath from environment

### LINKED ISSUE #1775: Respect web token locaton envrionment variable
### What happened?

IRSA token file location is [hardcoded in aws client](https://github.com/crossplane-contrib/provider-aws/blob/594d12c084d9642e7b0a91635dd4bd485edbaae8/pkg/clients/aws.go#L374). This makes it impossible to test it locally because this directory typically do not exist on local systems. 

IRSA uses the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable to determine the location of an OIDC ID token or oAuth access token to assume a role. We should respect it. 

Also note that in the realm of Crossplane, being able to use the token specified in `AWS_WEB_IDENTITY_TOKEN_FILE` only is likely not what we want. Currently, it is impossible for users to use a token not managed by EKS to assume a role in this provider or the upbound provider. We'll need to expand the spec and add a field to support it. 

I can work on this to remove the hardcoded value and use the environment variable value instead. 

Crossplane version: master

MisterMX:

Replaces the hardcoded web identity token filepath for with the environment variable `AWS_WEB_IDENTITY_TOKEN_FILE`.

Fixes #1775

I have:

n.a.

### DIFF
--- a/pkg/clients/aws.go
+++ b/pkg/clients/aws.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 
@@ -371,7 +372,7 @@ func UsePodServiceAccountAssumeRoleWithWebIdentity(ctx context.Context, _ []byte
 			stscreds.NewWebIdentityRoleProvider(
 				stsclient,
 				StringValue(roleArn),
-				stscreds.IdentityTokenFile("/var/run/secrets/eks.amazonaws.com/serviceaccount/token"),
+				stscreds.IdentityTokenFile(getWebidentityTokenFilePath()),
 				webIdentityRoleOptions,
 			)),
 		),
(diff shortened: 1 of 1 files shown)


pkg/clients/aws.go
```diff
@@ -382,6 +383,15 @@ func UsePodServiceAccountAssumeRoleWithWebIdentity(ctx context.Context, _ []byte
 	return &cnf, err
 }
 
+const webIdentityTokenFileDefaultPath = "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"
```
nabuskey, line 386: I think we should move this up in this file with other consts.

MisterMX: I would prefer to keep it together with code block where it is used for better readability since the definition order does not matter in Go.

haarchri (approved):
LGTM

PR #1784, opened 2023-06-19
feat(s3control): Add S3 AccessPoint and AccessPointPolicy

### LINKED ISSUE #1782: Add S3 AccessPoint and AccessPointPolicy from s3control API
### What resource do you need?
S3 AccessPoint and AccessPointPolicy from s3control API

### What is your use case?

Creating AccessPoint and AccessPointPolicy for s3 buckets.

### Would you be willing to contribute it using [code generator](https://github.com/crossplane/provider-aws/blob/master/CODE_GENERATION.md)?

Yes, I plan on generating the code and putting up a PR for this.

###  Link to Documentation:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_control_CreateAccessPoint.html
https://docs.aws.amazon.com/sdk-for-go/api/service/s3control/#AccessPoint
https://docs.aws.amazon.com/AmazonS3/latest/API/API_control_PutAccessPointPolicy.html

kbujanecki-dt:

Fixes #1782

I have:

```
apiVersion: s3control.aws.crossplane.io/v1alpha1
kind: AccessPoint
metadata:
  name: uber-access-point
spec:
  deletionPolicy: Orphan
  forProvider:
    accountID: "123456789012"
    bucket: "test-bucket"
    region: "eu-central-1"
    publicAccessBlockConfiguration:
      blockPublicACLs: true
      blockPublicPolicy: true
      ignorePublicACLs: true
      restrictPublicBuckets: true
    vpcConfiguration:
      vpcID: "vpc-XXXXXXXXXXX"
  providerConfigRef:
    name: "example"
```

```
apiVersion: s3control.aws.crossplane.io/v1alpha1
kind: AccessPointPolicy
metadata:
  name: uber-access-point-policy
spec:
  providerConfigRef:
    name: "example"
  forProvider:
    accessPointNameSelector:
      name: "uber-access-point"
    region: "eu-central-1"
    accountID: "123456789012"
    policy: >
      {
          "Version": "2012-10-17",
          "Statement": [
              {
                  "Effect": "Allow",
                  "Principal": {
                      "AWS": "arn:aws:iam::123456789012:role/sso/accessRole"
                  },
                  "Action": "s3:*",
                  "Resource": "arn:aws:s3:eu-central-1:123456789012:accesspoint/uber-access-point/object/*"
              }
          ]
      } 
```


przysiadZeSztanga:
fyi, we plan to add policy for s3 access point in another PR

przysiadZeSztanga:
S3 access poit policies and selectors were added.

przysiadZeSztanga:
@MisterMX I saw that you did several reviews, could you have a look?

apis/s3control/v1alpha1/zz_access_point_policy.go
```diff
@@ -0,0 +1,97 @@
...
+	// The Amazon Web Services account ID for owner of the bucket associated with
+	// the specified access point.
+	// +kubebuilder:validation:Required
+	AccountID *string `json:"accountID"`
+	// The policy that you want to apply to the specified access point. For more
+	// information about access point policies, see Managing data access with Amazon
+	// S3 access points (https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-points.html)
+	// in the Amazon S3 User Guide.
+	// +kubebuilder:validation:Required
+	Policy                            *string `json:"policy"`
```
MisterMX, line 41: Since there can only be one policy per AP it would be better to integrate the policy field into `AccessPoint` so users don' have to create two MRs. Much like it was for `Bucket` and `BucketPolicy` #1685.

It would also be good to provide the same structured schema for policies instead of passing a raw string.

pkg/controller/s3control/accesspoint/setup.go
```diff
@@ -0,0 +1,76 @@
...
+
+func preCreate(_ context.Context, point *svcapitypes.AccessPoint, input *svcsdk.CreateAccessPointInput) error {
+	input.Name = aws.String(meta.GetExternalName(point))
+	input.Bucket = point.Spec.ForProvider.BucketName
+	return nil
+}
+
+func preObserve(_ context.Context, point *svcapitypes.AccessPoint, input *svcsdk.GetAccessPointInput) error {
+	input.Name = aws.String(meta.GetExternalName(point))
+	return nil
```
MisterMX, line 67: Implementation for `isUpToDate` is missing.

MisterMX (changes requested):

kbujanecki-dt:
@MisterMX, thx you for the feedback. I implemented the requested changes. 

CR for `AccessPoint` after changes:

```apiVersion: s3control.aws.crossplane.io/v1alpha1
kind: AccessPoint
metadata:
  name: access-point
spec:
  deletionPolicy: Delete
  forProvider:
    accountID: "123456789012"
    bucketName: example-bucket
    region: "eu-central-1"
    publicAccessBlockConfiguration:
      blockPublicACLs: true
      blockPublicPolicy: true
      ignorePublicACLs: true
      restrictPublicBuckets: true
    policy:
      statements:
        - notAction:
            - s3:DeleteBucket
          effect: Allow
          principal:
            awsPrincipals:
              - iamRoleArn: "arn:aws:iam::123456789012:role/sso/localRoleGitOpsArgoCDDeploy"
          resource:
            - "arn:aws:s3:eu-central-1:123456789012:accesspoint/access-point/object/*"
      version: 2012-10-17
    vpcConfiguration:
      vpcID: "vpc-1234567890123456"
  providerConfigRef:
    name: "example"
``` 
The structure for the access point policy body was copied from `BucketPolicyBody`.

pkg/controller/s3control/accesspoint/hooks_test.go
MisterMX, line 1: I think it is better to test the whole controller (including the generated code) rather then the just hooks alone.

pkg/controller/s3control/accesspoint/policy.go
```diff
@@ -0,0 +1,172 @@
...
+	input.Name = aws.String(meta.GetExternalName(cr))
+	if cr.Spec.ForProvider.AccountID != nil {
+		input.SetAccountId(*cr.Spec.ForProvider.AccountID)
+	}
+
+	return input
+}
+
+// formatPolicy parses and formats the Policy struct
+func formatPolicy(cr *svcapitypes.AccessPoint) (*string, error) {
```
MisterMX, line 157: As said above: It is better to reuse the implementation in `clients/s3` rather than having duplicated code.

apis/s3control/v1alpha1/access_point_policy_types.go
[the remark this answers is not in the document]
kbujanecki-dt: > I believe it makes more sense to reuse the types in https://github.com/crossplane-contrib/provider-aws/blob/master/apis/s3/common/bucketpolicy_types.go instead of defining them again.

Agree, I did that because there are comments specific to s3 buckets. But If that is not a problem for you I will reuse it.

> Then it is also not necessary to implement the logic to compare the policies are second time.

Do you mean to reuse `ArePoliciesEqal` https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/utils/policy/compare.go#L9?

pkg/controller/s3control/accesspoint/policy.go
kbujanecki-dt: Same question as above, Should I make a common function for formatting policy or simply use `s3.Serialize` inside `formatPolicy` func?

apis/s3control/v1alpha1/access_point_policy_types.go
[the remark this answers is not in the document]
MisterMX: I think creating a common logic for all Resources makes the most sense. If there are S3-specific comments in the policy block then they can (most likely) be dropped.

pkg/controller/s3control/accesspoint/policy.go
MisterMX: Yes, a common logic would be best.

MisterMX:
@kbujanecki-dt this looks good to me now. Thank you very much for your contribution! Can you please squash your changes into a single commit? Then I can merge it.

kbujanecki-dt:
> @kbujanecki-dt this looks good to me now. Thank you very much for your contribution! Can you please squash your changes into a single commit? Then I can merge it.

I squashed commits and pushed. On my fork I see a single commit but PR view looks unrefreshed (13 commits)

MisterMX (approved):
LGTM. Thank you very much for your contribution @kbujanecki-dt!

PR #1787, opened 2023-06-21
fix(topic): Improve topic policy json diff check

x4b1:

In the observed policy on the topic, the comparison of JSON strings was changed to comparing actual JSON objects to prevent the update loop. This way it recognizes that the two objects are the same.

I have:

I have added a unit test to check this case.


MisterMX (approved):
LGTM. Thanks a lot @x4b1. Can you fix the linter issues? Then this is ready merge.

MisterMX:
Thanks, @x4b1!

PR #1796, opened 2023-06-29
feat(ecs)!: Make force deletion explicit

bmartins-unit:

Fix ECS Service deletion when the task are in error

Currently with the validation, if the ecs service creates a task that is not possible to spin up (e.g. wrong image), it is also not possible to delete the service, because the cr.Status.AtProvider.RunningCount and cr.Status.AtProvider.PendingCount won't be greater than 0.
In aws console it was also possible to delete the service (with task in error) by selecting the "force" option
 

I have:

Created a ECS Task Definition with non existing image, created a ECS Service pointing to this Task Definition, and then removed the Service CR

### DIFF
--- a/pkg/controller/ecs/service/setup.go
+++ b/pkg/controller/ecs/service/setup.go
@@ -99,13 +99,8 @@ func preCreate(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.CreateSer
 }
 
 func preDelete(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.DeleteServiceInput) (bool, error) {
-	if aws.StringValue(cr.Spec.ForProvider.SchedulingStrategy) == svcsdk.SchedulingStrategyReplica {
-		if aws.Int64Value(cr.Status.AtProvider.RunningCount) > 0 || aws.Int64Value(cr.Status.AtProvider.PendingCount) > 0 {
-			// Force deletion of services even if they are not scaled down.
-			obj.SetForce(true)
-		}
-	}
 
+	obj.SetForce(cr.Spec.ForProvider.ForceDeletion)
 	obj.Cluster = cr.Spec.ForProvider.Cluster
 	obj.SetService(meta.GetExternalName(cr))
 
--- a/apis/ecs/v1alpha1/custom_types.go
+++ b/apis/ecs/v1alpha1/custom_types.go
@@ -96,6 +96,10 @@ type CustomServiceParameters struct {
 	ClusterRef      *xpv1.Reference `json:"clusterRef,omitempty"`
 	ClusterSelector *xpv1.Selector  `json:"clusterSelector,omitempty"`
 
+	// Force Service to be deleted, even with task Running or Pending
+	// +optional
+	ForceDeletion bool `json:"forceDeletion,omitempty"`
+
 	// A load balancer object representing the load balancers to use with your service.
 	// For more information, see Service Load Balancing (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-load-balancing.html)
 	// in the Amazon Elastic Container Service Developer Guide.
--- a/package/crds/ecs.aws.crossplane.io_services.yaml
+++ b/package/crds/ecs.aws.crossplane.io_services.yaml
@@ -240,6 +240,10 @@ spec:
                       is enabled for the service. If true, this enables execute command
                       functionality on all containers in the service tasks.
                     type: boolean
+                  forceDeletion:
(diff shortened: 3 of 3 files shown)


pkg/controller/ecs/service/setup.go
```diff
@@ -100,10 +100,8 @@ func preCreate(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.CreateSer
 
 func preDelete(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.DeleteServiceInput) (bool, error) {
 	if aws.StringValue(cr.Spec.ForProvider.SchedulingStrategy) == svcsdk.SchedulingStrategyReplica {
-		if aws.Int64Value(cr.Status.AtProvider.RunningCount) > 0 || aws.Int64Value(cr.Status.AtProvider.PendingCount) > 0 {
-			// Force deletion of services even if they are not scaled down.
-			obj.SetForce(true)
```
MisterMX, line 105: What are the implications if the user does NOT want to enforce deletion on running instances? I wonder if it wouldn't be better if there's an explicit `spec.forProvider.forceDeletion` property?

bmartins-unit: Your proposal is to add the field `ForceDeletion` at `apis/ecs/v1alpha1/zz_service.go` struct `ServiceParameters` and in the `preDelete` func do the obj.SetForce with the value in CR?

MisterMX: Yes, I think that would make things more explicit and avoid a potential breaking behaviour for existing resources.

bmartins-unit: Could you please give guidance on how to add the field, I have tried to change the `generator-config.yaml`, but without success.

MisterMX: APIs that are generated with ACK have a `custom_types.go` that allow you to add additional properties to `spec.forProvider`. For `ecs` you can find a [`CustomServiceParameters`](https://github.com/crossplane-contrib/provider-aws/blob/master/apis/ecs/v1alpha1/custom_types.go#L89) struct that you can extend.

You then have to regenerate the methods and CRDs in `package` using `make generate`.

You can find more info here: https://github.com/crossplane-contrib/provider-aws/blob/master/CODE_GENERATION.md

pkg/controller/ecs/service/setup.go
```diff
@@ -99,9 +99,14 @@ func preCreate(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.CreateSer
...
+
+	obj.SetForce(cr.Spec.ForProvider.ForceDeletion)
+
 	if aws.StringValue(cr.Spec.ForProvider.SchedulingStrategy) == svcsdk.SchedulingStrategyReplica {
-		// Force deletion of services even if they are not scaled down.
-		obj.SetForce(true)
+		if aws.Int64Value(cr.Status.AtProvider.RunningCount) > 0 || aws.Int64Value(cr.Status.AtProvider.PendingCount) > 0 {
+			// Force deletion of services even if they are not scaled down.
+			obj.SetForce(true)
+		}
```
MisterMX: This change is not necessary anymore since force is now explicitly set by the user.

MisterMX:
@bmartins-unit this looks good so far. Can you remove the implicit setting of force and squash + sign your commit?

bmartins-unit (commented):
What do you mean by "implicit setting of force"?
something like this?
```
if cr.Spec.ForProvider.ForceDeletion {
  obj.SetForce(cr.Spec.ForProvider.ForceDeletion)
}
```

MisterMX:
> What do you mean by "implicit setting of force"? something like this?

I was talking about calling `obj.SetForce` if a special condition (here: number of running jobs) without the user having explicitly set the value.

bmartins-unit (commented):
I'm a bit lost:
If we have the option to forceDelete by the user, as you said
https://github.com/crossplane-contrib/provider-aws/pull/1796#discussion_r1255389582

shouldn't the "calculation" of ForceDeletion be done in my own Operator (e.g. number of running jobs), and in the `preDelete` do the `obj.SetForce(cr.Spec.ForProvider.ForceDeletion)`

MisterMX:
In https://github.com/crossplane-contrib/provider-aws/pull/1796#discussion_r1250767356 I was talking about the logic that is implemented before. For me it is good the way it is implemented in this PR.

Can you squash your changes into a single commit and sign it?

MisterMX (approved):
LGTM. Thanks a lot @bmartins-unit!

PR #1797, opened 2023-06-30
Ignore ASG desiredCapacity when it's not set to make it compatible with the cluster-autoscaler

handlerww:

As mentioned in https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md?plain=1#L24-L27 and https://github.com/crossplane-contrib/provider-aws/pull/502, there is a race condition
  between crossplane and cluster autoscaler. When we use asg as self-managed nodegroup with cluster-autoscaler, the reconciling of crossplane asg controller will revert changes from cluster-autoscaler. If `desiredCapacity` is not set, we should leave it blank, and just let the cluster-autoscaler to manage it. This PR implements a similar logic in the managed nodegroup.

I have:

I've manually test it, the test plan is here:

1. create ASG:
   a. create a ASG without `desiredCapacity`, the ASG can be created.
  ```
    maxSize: 400
    minSize: 0
  ```
  ```
   maxSize: 400
   minSize: 1
  ```
  b. create a ASG with `desiredCapacity`, the ASG configuration is as expected.
  ```
   maxSize: 400
   minSize: 0
   desiredCapacity: 1
```
3. update ASG
  a. modify ASG configuration on AWS console, modify the maxSize to 401, trigger the reconciling, the `desiredCapacity` doesn't change.
  b. add `desiredCapacity` configuration on ASG CR, the ASG configuration on AWS console changes as expected.
  c. remove `desiredCapacity` configuration from ASG CR, and modify the configuration on AWS console, it works as expected.

### DIFF
--- a/pkg/controller/autoscaling/autoscalinggroup/setup.go
+++ b/pkg/controller/autoscaling/autoscalinggroup/setup.go
@@ -216,7 +216,10 @@ func lateInitialize(in *svcapitypes.AutoScalingGroupParameters, asg *svcsdk.Desc
 	in.CapacityRebalance = awsclients.LateInitializeBoolPtr(in.CapacityRebalance, obs.CapacityRebalance)
 	in.DefaultCooldown = awsclients.LateInitializeInt64Ptr(in.DefaultCooldown, obs.DefaultCooldown)
 	in.DefaultInstanceWarmup = awsclients.LateInitializeInt64Ptr(in.DefaultInstanceWarmup, obs.DefaultInstanceWarmup)
-	in.DesiredCapacity = awsclients.LateInitializeInt64Ptr(in.DesiredCapacity, obs.DesiredCapacity)
+	// if desiredCapacity is not set, don't update the value
+	if in.DesiredCapacity != nil {
+		in.DesiredCapacity = awsclients.LateInitializeInt64Ptr(in.DesiredCapacity, obs.DesiredCapacity)
+	}
 	in.DesiredCapacityType = awsclients.LateInitializeStringPtr(in.DesiredCapacityType, obs.DesiredCapacityType)
 	in.HealthCheckGracePeriod = awsclients.LateInitializeInt64Ptr(in.HealthCheckGracePeriod, obs.HealthCheckGracePeriod)
 	in.HealthCheckType = awsclients.LateInitializeStringPtr(in.HealthCheckType, obs.HealthCheckType)

### CODE REVIEW

**Comment by csuzhangxc** [comment]
@MisterMX can you take a looK?

File: pkg/controller/autoscaling/autoscalinggroup/setup.go [hunk 1]
```diff
@@ -216,7 +216,10 @@ func lateInitialize(in *svcapitypes.AutoScalingGroupParameters, asg *svcsdk.Desc
 	in.CapacityRebalance = awsclients.LateInitializeBoolPtr(in.CapacityRebalance, obs.CapacityRebalance)
 	in.DefaultCooldown = awsclients.LateInitializeInt64Ptr(in.DefaultCooldown, obs.DefaultCooldown)
 	in.DefaultInstanceWarmup = awsclients.LateInitializeInt64Ptr(in.DefaultInstanceWarmup, obs.DefaultInstanceWarmup)
-	in.DesiredCapacity = awsclients.LateInitializeInt64Ptr(in.DesiredCapacity, obs.DesiredCapacity)
+	// if desiredCapacity is not set, don't update the value
+	if in.DesiredCapacity != nil {
+		in.DesiredCapacity = awsclients.LateInitializeInt64Ptr(in.DesiredCapacity, obs.DesiredCapacity)
```
MisterMX, line 221: While I understand the issue this change wants to tackle I don't think it makes sense the way it is implemented: `LateInitialize` is supposed to set update the spec with the observed value *if* the former is nil.

Also, `LateInitializeInt64Ptr` only updates the value if `in.DesiredCapacity == nil`. So this change would effectively cause `in.DesiredCapacity` to be never late initialized.

I think it makes more sense to explicitly specify the desired behaviour of the controller rather then doing it implicitly through a special prop value. It could by defining a new property (i.e. `scalingMode: Managed | External`) that tells the controller to not set the value in the request to AWS during `Update`.

bobh66:
This issue will also be addressed by the Observe Only feature where you will be able to set `spec.initProvider.desiredSize: X` and leave `spec.forProvider.desriedSize` empty which will cause the field to be set on create and then ignored for all subsequent reconciles.

handlerww:
> This issue will also be addressed by the Observe Only feature where you will be able to set `spec.initProvider.desiredSize: X` and leave `spec.forProvider.desriedSize` empty which will cause the field to be set on create and then ignored for all subsequent reconciles.

So it would be addressed in [Ignore Changes](https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md?plain=1)? It seems a better solution.

bobh66:
> > This issue will also be addressed by the Observe Only feature where you will be able to set `spec.initProvider.desiredSize: X` and leave `spec.forProvider.desriedSize` empty which will cause the field to be set on create and then ignored for all subsequent reconciles.
> 
> So, would it be addressed in [Ignore Changes](https://github.com/crossplane/crossplane/blob/master/design/one-pager-ignore-changes.md?plain=1)? It seems like a better solution.

Yes, my mistake.  Thanks for the correction.

PR #1802, opened 2023-07-06
feat(efs/filesystem): update tags

### LINKED ISSUE #1284: EFS File System Update Tags
### What happened?

When updating the tags on the EFS `FileSystem` managed resource, I expect the corresponding tags to be updated on the AWS resource itself, however they are not.

### How can we reproduce it?

Create an EFS `FileSystem` and once it is fully `Ready` update the `tags` on that resource. Observe that the tags on the AWS FileSystem resource are not updated.

Crossplane version: 1.6.2
Provider version: 0.23.0

wotolom:

Fixes #1284

add tag update functionality to EFS Filesystem

Similar to RDS Optiongroup, I put the tags-update-call into isUpToDate() and not into postUpdate(). (see here: https://github.com/crossplane-contrib/provider-aws/blob/9205a212d0c301cca16697c1c159d4de60cc2826/pkg/controller/rds/optiongroup/setup.go#L115-L122)

The reason is probably the same: AWS denies the "regular" UpdateFileSystem()-Call, if there is no actual change included. See the received error:

> {RespMetadata: {StatusCode: 400, RequestID: ""}, 
ErrorCode: "BadRequest", Message_: "The file system won't be updated. The requested throughput mode or provisioned throughput value are the same as the current mode and value."}

I have:

Manually with example

### DIFF
--- a/pkg/controller/efs/filesystem/setup.go
+++ b/pkg/controller/efs/filesystem/setup.go
@@ -2,10 +2,14 @@ package filesystem
 
 import (
 	"context"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/efs"
+	"github.com/aws/aws-sdk-go/service/efs/efsiface"
+	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/connection"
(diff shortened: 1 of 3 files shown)


pkg/controller/efs/filesystem/setup.go
```diff
@@ -56,12 +59,29 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
...
+		return false, err
+	}
+	// update tags here, bc AWS only accepts the regular update-request if there is an actual change included
+	if !tagsUpToDate {
+		err := svcutils.UpdateTagsForResource(e.client, cr.Spec.ForProvider.Tags, cr.Status.AtProvider.FileSystemID)
+		if err != nil {
+			return false, err
+		}
+	}
+
```
MisterMX, line 84: `Observe` and `isUpToDate` should have no side effects. Tag updates should be in `Update`.

pkg/controller/efs/tags.go
```diff
@@ -0,0 +1,159 @@
...
+		if _, exists := specMap[key]; !exists {
+			removeTags = append(removeTags, awsclient.String(key))
+		}
+	}
+
+	return addTags, removeTags
+}
+
+// AddExternalTags to spec if they don't exist
+func AddExternalTags(mg resource.Managed, spec []*svcapitypes.Tag) []*svcapitypes.Tag {
```
MisterMX, line 131: Is function called anywhere?

MisterMX (changes requested):

pkg/controller/efs/filesystem/setup.go
wotolom: As discussed, I now put the tags-update-call to postUpdate(), where we ignore the "no change"-response from AWS to the regular resource-update-call.

wotolom:
updated + new changes:

- added Initialize() to handle crossplane external tags 
- added lateInitialize() to catch `ThroughputMode` AWS default 
- moved tags-update to postUpdate() and ignore AWSs "no-change" error
- added `ThroughputMode` to isUpToDate() (parameter was not updateable yet) 
- added if-check to isUpToDate(), to avoid unnecessary update-calls while `FileSystem` is updating

MisterMX (approved):
LGTM. Thanks a lot @wotolom!

PR #1806, opened 2023-07-14
feat(servicecatalog): Add provisioned product

teeverr:
Adds new resource - ProvisionedProduct, which represents life cycle of launched product in service catalog

I have:

Many scenarios have been tested manually In development env and also I wrote some unit tests for custom controller logic


mleahu:
Hi @MisterMX! 

Can you please check this PR and give us your first impression?

Thanks a lot!

MisterMX (changes requested):
Thank you for your contribution @teeverr. I added some remarks to the code. Can you take a look at them? Thanks!

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (c *custom) isUpToDate(cr *svcapitypes.ProvisionedProduct, resp *svcsdk.DescribeProvisionedProductOutput) (bool, error) {
+	if aws.StringValue(cr.Status.AtProvider.Status) == string(svcapitypes.ProvisionedProductStatus_SDK_UNDER_CHANGE) {
+		return true, nil
+	}
+	getPPInput := &svcsdk.GetProvisionedProductOutputsInput{ProvisionedProductId: resp.ProvisionedProductDetail.Id}
+	getPPOutput, err := c.client.GetProvisionedProductOutputs(getPPInput)
+	if err != nil {
+		return true, err
```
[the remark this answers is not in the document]
teeverr: Fixed

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	ppStatus := aws.StringValue(resp.ProvisionedProductDetail.Status)
+	switch {
+	case ppStatus == string(svcapitypes.ProvisionedProductStatus_SDK_AVAILABLE):
+		cr.Status.SetConditions(xpv1.Available())
+	case ppStatus == string(svcapitypes.ProvisionedProductStatus_SDK_UNDER_CHANGE) && *describeRecordOutput.RecordDetail.RecordType == "UPDATE_PROVISIONED_PRODUCT":
+		cr.SetConditions(xpv1.Available().WithMessage("Provisioned Product is updating, availability depends on product"))
+	case ppStatus == string(svcapitypes.ProvisionedProductStatus_SDK_ERROR):
+		cr.Status.SetConditions(xpv1.ReconcileError(errors.New(errProvisionedProductStatusSdkError)))
```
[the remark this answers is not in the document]
teeverr: It is true if provisioned product has Tainted state, but if it has Error update [is not allowed](https://docs.aws.amazon.com/servicecatalog/latest/userguide/enduser-update.html), so crossplane will never reconcile that resource

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+
+	cr.Status.AtProvider.Outputs = outputs
+	cr.Status.AtProvider.ARN = resp.ProvisionedProductDetail.Arn
+	cr.Status.AtProvider.CreatedTime = &metav1.Time{Time: *resp.ProvisionedProductDetail.CreatedTime}
+	cr.Status.AtProvider.LastProvisioningRecordID = resp.ProvisionedProductDetail.LastProvisioningRecordId
+	cr.Status.AtProvider.LaunchRoleARN = resp.ProvisionedProductDetail.LaunchRoleArn
+	cr.Status.AtProvider.Status = resp.ProvisionedProductDetail.Status
+	cr.Status.AtProvider.StatusMessage = resp.ProvisionedProductDetail.StatusMessage
+	cr.Status.AtProvider.ProvisionedProductType = resp.ProvisionedProductDetail.Type
+	cr.Status.AtProvider.RecordType = describeRecordOutput.RecordDetail.RecordType
```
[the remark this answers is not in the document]
teeverr: Yes, some of them already set in zz_controller.go in Create func which use resp of [ProvisionProduct](https://github.com/aws/aws-sdk-go-v2/blob/service/servicecatalog/v1.20.1/service/servicecatalog/api_op_ProvisionProduct.go#L30) method for retrieving values for these field if its not nil. The problem it is nil, e.g. aws api doesn't return values for some fields from [ProvisionProductOutput](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/servicecatalog#ProvisionProductOutput). Is it happens everytime or time to time, can't say, I didn't dig deep into this topic. But the fact is managed resource doesn't contain this information in Status.atProvider

Thats why I set the fields in postObserve which uses another method([DescribeProvisionedProduct](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/servicecatalog#Client.DescribeProvisionedProduct)), in this case everything works fine

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+		return managed.ExternalCreation{}, err
+	}
+	meta.SetExternalName(cr, aws.StringValue(resp.RecordDetail.ProvisionedProductId))
+	return cre, nil
+}
+
+func preDelete(_ context.Context, cr *svcapitypes.ProvisionedProduct, obj *svcsdk.TerminateProvisionedProductInput) (bool, error) {
+	obj.TerminateToken = aws.String(genIdempotencyToken())
+	obj.ProvisionedProductId = aws.String(meta.GetExternalName(cr))
+	if *cr.Status.AtProvider.Status == string(svcapitypes.ProvisionedProductStatus_SDK_UNDER_CHANGE) {
```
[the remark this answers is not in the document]
teeverr: Fixed, thanks

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+			continue
+		} else {
+			return false
+		}
+	}
+	return true
+}
+
+func genIdempotencyToken() string {
+	var mutex sync.Mutex
```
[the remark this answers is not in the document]
teeverr: fixed, ty

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,265 @@
...
+			resource.ManagedKind(svcapitypes.ProvisionedProductGroupVersionKind),
+			managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
+			managed.WithPollInterval(o.PollInterval),
+			managed.WithLogger(o.Logger.WithValues("controller", name)),
+			managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
+			managed.WithConnectionPublishers(cps...)))
+}
+
+func setupExternal(customClient clientset.Client) (func(*external), error) {
+	if reflect.DeepEqual(customClient, &clientset.CustomServiceCatalogClient{}) {
```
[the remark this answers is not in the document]
teeverr: Okay, thanks. I refactored and renamed this function

teeverr:
Thanks fore review @MisterMX! I hope I fulfilled all of your remarks : )
p.s. This commit history were accidentally squashed in a wrong way makes reading history a little harder

pkg/controller/servicecatalog/provisionedproduct/setup.go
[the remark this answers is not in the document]
MisterMX: If you want to prevent an update you can look up the state during `Update` and do nothing if it is not allowed - or you return `ResourceIsUpdToDate: true` in `Observe`. Either way, setting the `ReconcileError` condition here is still not correct and you should use `Unvailable`.

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,268 @@
...
+		ProductId:                crParams.ProductID,
+		ProvisioningArtifactId:   crParams.ProvisioningArtifactID,
+		ProductName:              crParams.ProductName,
+		ProvisioningArtifactName: crParams.ProvisioningArtifactName,
+	}
+	output, err := c.client.DescribeProvisioningArtifact(&input)
+	if err != nil {
+		return "", err
+	}
+	return *output.ProvisioningArtifactDetail.Id, nil
```
MisterMX, line 263: You might run into a panic when dereferencing a nil value. Better to check if it is nil before returning it.

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func provisioningParamsAreNotChanged(cfStackParams []cfsdkv2types.Parameter, currentParams []*svcapitypes.ProvisioningParameter) bool {
+	if len(cfStackParams) != len(currentParams) {
+		return false
+	}
+
+	cfStackKeyValue := make(map[string]string)
+	for _, v := range cfStackParams {
+		cfStackKeyValue[*v.ParameterKey] = *v.ParameterValue
```
MisterMX, line 216: Possible nil deref. Better use `pointer.StringDeref`.

MisterMX (changes requested):

MisterMX:
Thanks for the changes @teeverr. There are some minor issues that need to be solved before merging. Can you take a look at them?

Also, can you squash your changes into a single commit and sign it?

threadproc:
Thanks for the review! I work on the team handling this on Swisscom's side, and we should have an updated PR ready soon that incorporates this feedback, and fixes several bugs and issues we've discovered so far.

threadproc:
@MisterMX I've pushed a new version of this work, which addresses the feedback you provided as well as resolving the following discovered issues:

- `productName` does not work past the first provision
- updates performed when a provisioned product is in the `UNDER_CHANGE` state fail with an obscure error message (replaced this with a handled error and better message)
- products that are `UNDER_CHANGE` do not reconcile with much frequency, so they are not available until they are polled next - we treat this like a ReconcileError now, since we expect them to become `AVAILABLE` and they should be re-reconciled until that is the case
- provisioned products are not deleted when they are in `ERROR` or `TAINTED` status

As well, I did some refactoring to try to clean things up a bit and keep it organized.

threadproc:
Fixed the linter errors, apologies.

pkg/controller/servicecatalog/provisionedproduct/lifecycle.go
```diff
@@ -0,0 +1,194 @@
...
+// Update replaces the ExternalClient.Update function so we can perform some custom actions before delegating back to it
+func (c *custom) Update(ctx context.Context, mg resource.Managed) (managed.ExternalUpdate, error) {
+	cr, ok := mg.(*svcapitypes.ProvisionedProduct)
+	if !ok {
+		return managed.ExternalUpdate{}, errors.New(errUnexpectedObject)
+	}
+
+	// We want to check if we are in a state where we can actually update it
+	ppStatus := pointer.StringDeref(cr.Status.AtProvider.Status, "")
+	if ppStatus == "" || ppStatus == string(svcapitypes.ProvisionedProductStatus_SDK_UNDER_CHANGE) {
```
teeverr, line 171: Why not to do it in preUpdate?

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+		} else {
+			return true
+		}
+	}
+
+	return false
+}
+
+// productOrArtifactAreChanged will attempt to determine whether or not the currently requested SC product and provisioning artifact IDs have changed when using names instead of IDs
+func (c *custom) productOrArtifactAreChanged(ds *svcapitypes.ProvisionedProductParameters, resp *svcsdk.ProvisionedProductDetail) (bool, error) { // nolint:gocyclo
```
teeverr, line 38: Could you please explain why you replaced `productOrArtifactAreNotChanged`(https://github.com/swisscom/provider-aws/pull/1/conflicts) to ` productOrArtifactAreChanged`, am asking not about `Not` : ) It's okay. But about meaning...
```suggestion
func (c *custom) productOrArtifactAreChanged(ds *svcapitypes.ProvisionedProductParameters, resp *svcsdk.ProvisionedProductDetail) (bool, error) {
	// ProvisioningArtifactID and ProvisioningArtifactName are mutual exclusive params, the same about ProductID and ProductName
	// But if describe a provisioned product aws api will return only IDs, so it's impossible to compare names with ids
	// Conditional statement below works only if desired state includes ProvisioningArtifactID and ProductID
	if ds.ProvisioningArtifactID != nil && ds.ProductID != nil &&
		(*ds.ProvisioningArtifactID != *resp.ProvisioningArtifactId ||
			*ds.ProductID != *resp.ProductId) {
		return true, nil
		// In case if desired state includes not only IDs provider runs func `getArtifactID`, which produces
		// additional request to aws api and retrieves an artifact id based on ProductId/ProductName and ProvisioningArtifactId/ProvisioningArtifactName
		// for further comparison with artifact id in the current state
	} else if ds.ProvisioningArtifactName != nil || ds.ProductName != nil {
		desiredArtifactID, err := c.getArtifactID(ds)
		if err != nil {
			return true, err
		}
		if desiredArtifactID != *resp.ProvisioningArtifactId {
			return true, nil
		}
	}
	return false, nil
```

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+	var productID, artifactID string
+
+	if ds.ProductID != nil {
```
teeverr, line 41: this if is redundant

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+		productID = pointer.StringDeref(ds.ProductID, "")
+	}
+	if ds.ProvisioningArtifactID != nil {
```
teeverr, line 44: this if is redundant

pkg/controller/servicecatalog/provisionedproduct/setup.go
teeverr: Aws sdk v1 says it can't by nil, or you recommend to deref it anyway, just in case?

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+			}
+		}
+
+		if artifactID == "" {
+			// we were unable to find an artifact that matches
+			return true, errors.New("could not find provisioning artifact")
+		}
+	}
+
+	return pointer.StringDeref(resp.ProductId, "") != productID ||
```
teeverr, line 95: These fields can't be nil, so deref is redundant. But as I understand @MisterMX recommend to deref any pointers, idk
Another question about this func in general... as I wrote above.

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+
+		if productID == "" {
+			// fill in the product ID if we do not yet have it
+			if describeProduct == nil || describeProduct.ProductViewSummary == nil {
+				return true, errors.New("could not find product")
+			}
+			productID = pointer.StringDeref(describeProduct.ProductViewSummary.ProductId, "")
+		}
+
+		if artifactID == "" {
```
teeverr, line 79: while?

pkg/controller/servicecatalog/provisionedproduct/utils.go
teeverr, line 38: `// nolint:gocyclo` I think the idea of this rule is to keep code clean and protect it from some things... : )

teeverr (commented):
Thanks for your changes @threadproc
I left some comments here, the summary(and concern) of which is to understand that all of these 'refactoring' are really necessary and improve the PR and these are not changes for changes.

Because it would be more than enough to add 10-20 simple lines of code to fix 3-4 bugs.

Ps. It would be really nice to create new PR to this branch before. Because after squashing is happend it became hardly understandable, but I restored the original branch from my local backup. So I used this link[https://github.com/swisscom/provider-aws/pull/1/conflicts](url) in my comments/remarks for comparison and clarification

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+			// find the matching artifact
+			for _, artifact := range describeProduct.ProvisioningArtifacts {
```
teeverr, line 81: O(n) here, but O(1) in original branch with another method(DescribeProvisioningArtifact). Why?

pkg/controller/servicecatalog/provisionedproduct/utils.go
```diff
@@ -0,0 +1,101 @@
...
+	if productID == "" || artifactID == "" {
+		// If we are using names for either the product or artifact, we should
+		// do a lookup using the DescribeProduct API to find the matching
+		// product (either by name or ID). In the returned output, we will have
+		// access to both the product ID and all of the available provisioning
+		// artifact IDs so we will be able to discover both IDs with a single
+		// API call.
+		dpInput := &svcsdk.DescribeProductInput{}
+
+		if productID != "" {
```
teeverr, line 57: This two field are mutualy exclusive, it means this if is redundant again and even more it follows to bugs or unpected behavior, cuz if user set both he/she don't get relevant error, cuz you use productid every time.
And again the question arises what was wrong with the original code/func?

mleahu:
Hi @MisterMX 

 Thanks a lot for your reviews and feedbacks!

 We still run internal tests with this service catalog feature of provider-aws and we might still hit some bugs.

 I propose to keep this PR open/draft until we're done with the testing phase. I'll keep you updated!

Thank you again! Wish a nice day!

Marius

teeverr:
This PR has been replaced by this - https://github.com/crossplane-contrib/provider-aws/pull/1908

PR #1811, opened 2023-07-21
fix(rds/dbinstance): remove minor engineVersion lateinit

wotolom:

Since the change from https://github.com/crossplane-contrib/provider-aws/pull/1765, the handling of major and minor engine version is now thoroughly in isUpToDate(). Also the actual `engineVersion` is now displayed in `status.atProvider`.
I noticed that for DBInstance, there was still some code in lateInit(), that also tackled that issue. This results in "unclean" display of the complete engine version also in `forProvider` even when the user only wants to set the major engine version.
This can and should now be removed.

Fixes #

I have:

Manually


MisterMX (approved):
LGTM. I think it makes sense to drop this from `LateInitialize`. Thanks a lot @wotolom!

PR #1821, opened 2023-08-01
fix: ignore returned false value when local nil

akesser:
AWS returns enabled: false for kafka cluster logging settings by default, this pull request updates the provider code to ignore this if the logging settings of the managed resource are set to nil

Fixes #

I have:

tested locally 


haarchri (approved):
LGTM

PR #1830, opened 2023-08-04
chore: Add context to ACK isUpToDate and return diff

### LINKED ISSUE #1828: Add `context.Context` to ack-generated `isUpToDate` functions

The current ACK template for Crossplane does not include a `context.Context` in the generated `isUpToDate` functions. This leads to a `context.TODO` being created in many controllers where an additional API is needed.

The generated `isUpToDate` als don't return a `diff` that can be returned with resulting `managed.ExternalObservation`.

Extend ACK to include `context.Context` and return a diff string as result. This will need every generated controller to be adjusted afterwards.

MisterMX:

Fixes #1828
Depends on https://github.com/aws-controllers-k8s/code-generator/pull/458

I have:

Most of the code remains untouched. Only function parameters were added and `context.TODO` was replaced with the real context. In some cases the controllers and tests needed to bed changed since their weren't using the `...WithContext` APIs. In some other cases `cmp.Equal` was replaced with `diff := cmp.Diff(...); diff != ""` to keep the diff and return it later.

### DIFF
--- a/pkg/clients/dax/fake/fake.go
+++ b/pkg/clients/dax/fake/fake.go
@@ -46,7 +46,8 @@ type MockDaxClient struct {
 	MockDescribeParameterGroups            func(*dax.DescribeParameterGroupsInput) (*dax.DescribeParameterGroupsOutput, error)
 	MockDescribeParameterGroupsWithContext func(context.Context, *dax.DescribeParameterGroupsInput, []request.Option) (*dax.DescribeParameterGroupsOutput, error)
 
-	MockDescribeParameters func(*dax.DescribeParametersInput) (*dax.DescribeParametersOutput, error)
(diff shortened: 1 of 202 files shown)


chlunde:
I'm not super fan about the name utils, ref. https://github.com/golang/go/wiki/CodeReviewComments#package-names - if you have an alternative that would be nice.

And as you say, we should wait with this until merged upstream to prevent depending on a unmerged PR.

maybe in a followup, it would be nice if cases like this returned a hint like "spec.forProvider.pitStatus" instead of the empty string?
 
```
	if !isPitrUpToDate(cr, pitrStatusBool) {
		return false, "", nil
```

chlunde (approved):

PR #1840, opened 2023-08-09
chore: Add ACK as go tool

MisterMX:

This adds the ACK code generate as a go module. 
~It removes the need to `git clone` and `go build` it manually in the Makefile and instead lets it be execute through `go run`.~
It removes the need to `git clone` the ack repo manually. A static binary is still built that is executed many times during code generation which is faster than executing it through `go run`.

The templates that are needed for the controller generation are moved from the code generator directly into the provider repository making changes in the generated controllers much easier.

I have:

n.a. since no code has been changed.


Makefile
```diff
@@ -141,18 +139,10 @@ AWS_SDK_GO_VERSION ?= $(shell awk '/github.com\/aws\/aws-sdk-go / { print $$2 }'
...
 # be in its root directory to call "go run" properly.
 services: $(GOIMPORTS)
-	@if [ ! -d "$(WORK_DIR)" ]; then \
-		$(INFO) creating $(WORK_DIR) folder; \
-		mkdir $(WORK_DIR); \
-	fi
-	@if [ ! -d "$(WORK_DIR)/code-generator" ]; then \
-		cd $(WORK_DIR) && git clone "$(CODE_GENERATOR_REPO)"; \
-	fi
-	@cd $(WORK_DIR)/code-generator && git fetch origin && git checkout $(CODE_GENERATOR_COMMIT) && go build -tags codegen -o ack-generate cmd/ack-generate/main.go
```
chlunde, line 151: I think I moved `go run` out of the loop because `go run` on a bigger project like this was fairly slow (linking), so try timing `make generate`. I think it might be worth to build a binary once like it used to.

I assume we can't use go run github.com/aws-controllers-k8s/code-generator/cmd/ack-generate@vX.X due to the tags.

MisterMX: I didn't notice a huge performance gain by using a prebuild binary over `go run` but I'm OK with having `go build` before that. I changed the code accordingly.

However, `go run` itself works fine if used with `-tags codegen`.

chlunde (approved):

PR #1844, opened 2023-08-18
fix(eks): Add reconciliation for addon configurationValues

smcavallo:
eks addon configurationValues is supported already in the spec but is not included in `isUpToDate` and therefore can never be changed with the provider.
https://github.com/crossplane-contrib/provider-aws/blob/master/apis/eks/v1alpha1/zz_addon.go#L40-L43

This PR adds the custom `isUpToDate` in order to update the `configurationValues`
The configurationValues is a string value but both json and yaml are supported. (See - https://aws.amazon.com/blogs/containers/amazon-eks-add-ons-advanced-configuration/)
Note the implementation for the string comparison.
Whitespace and indendation is difficult to get correct when using a string value.
This PR does UnMarshal and Marshal the json or yaml to ensure that the comparison is more accurate.  It lessens the risk that this provider will continually request an update of the external resource due to formatting, similar to the "normalization" done with iam policies.  Unit test cases have been added for testing format differences.

The addOn example has been updated to reflect yaml definition but json is supported as well.

yaml:
```
    configurationValues: |
      controller:
        podAnnotations:
          fluentbit.io/exclude: "true"
```
json:
```
      configurationValues: '{"controller": {"podAnnotations": {"fluentbit.io/exclude":
      "true"}}}'
```

What this PR does NOT do:  
* Any validation for yaml/json formatting
* Each addOn publishes a configuration schema.  This provider does not validate the cr against the provided schema.  This PR will send the update and aws will reject the change due to invalid schema

Fixes #

I have:

* define configurationValues for an addOn using the above json
* ensure provider-aws reconciles the resource and json configurationValues appear in the console
* ensure that upon the next reconcile the `External resource is up to date`
* set configurationValues to ""
* ensure provider-aws reconciles the resource and the default value of `-` appears in the console
* ensure that upon the next reconcile the `External resource is up to date`

* do the same as the above but with yaml
* ensure that no changes are made when `configurationValues` is not defined for an `addOn` - provider-aws reconciles the resource and the default value of `-` appears in the console
* ensure that the isUpToDate for configurationValues handles exceptions and returns them so they can be written as errors in the logs
* ensure that when using values which violate the schema the provider throws the exception to the logs:
```
2023-08-18T03:16:27.324Z        DEBUG   events  cannot update Addon in AWS: InvalidParameterException: ConfigurationValue provided in request is not supported: Yaml schema validation failed with error: [$.controller2: is not defined in the schema and the schema does not allow additional properties]
{
  RespMetadata: {
    StatusCode: 400,
    RequestID: ""
  },
```

### DIFF
--- a/pkg/controller/eks/addon/setup.go
+++ b/pkg/controller/eks/addon/setup.go
@@ -19,11 +19,14 @@ package addon
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go/aws"
 	awseks "github.com/aws/aws-sdk-go/service/eks"
 	"github.com/aws/aws-sdk-go/service/eks/eksiface"
+	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
 
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/connection"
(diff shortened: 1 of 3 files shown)


pkg/controller/eks/addon/setup.go
```diff
@@ -144,10 +149,69 @@ func (h *hooks) isUpToDate(cr *eksv1alpha1.Addon, resp *awseks.DescribeAddonOutp
...
+		return false, crErr
+	}
+	// Compare normalized strings
+	return cmp.Equal(objConfigurationValues, crConfigurationValues), nil
+}
+
+// convertConfigurationValuesToString will deserialize and serialize in order to normalize
+func convertConfigurationValuesToString(configurationValues string) (string, error) {
+	var jsonConfigurationValues map[string]interface{}
+	if strings.HasPrefix(configurationValues, "{") {
```
MisterMX, line 196: That doesn't work if the JSON is not starting with `{`. It would be better to just use `json.Unmarshal` and directly and use `yaml.Unmarshal` if that fails.

pkg/controller/eks/addon/setup.go
```diff
@@ -144,10 +149,69 @@ func (h *hooks) isUpToDate(cr *eksv1alpha1.Addon, resp *awseks.DescribeAddonOutp
...
+		result, marshalErr := json.Marshal(&jsonConfigurationValues)
+		return string(result), marshalErr
+	}
+	// If YAML
+	err := yaml.Unmarshal([]byte(configurationValues), &jsonConfigurationValues)
+	if err != nil {
+		return configurationValues, err
+	}
+	// Convert back to string for better comparison
+	result, marshalErr := yaml.Marshal(&jsonConfigurationValues)
```
MisterMX, line 211: I think it would be better to compare the parsed objects instead of the raw strings and use `cmp.Diff` to detect any changes.

smcavallo: will update, thanks!

pkg/controller/eks/addon/setup.go
smcavallo: I initially used the return error - if `SyntaxError` then assume it's yaml.  I didn't like that approach as it would try to assume yaml even if the json syntax was incorrect.
There is a json schema published which should enforce that the json MUST start with `{` so IMO it was safer to do it this way.   Also....the json is much more difficult to use as you have to minimize or escape it and as most k8s is defined in yaml.  I will try yaml parsing first.

MisterMX (approved):
LGTM. Thank you very much for your contribution @smcavallo!

PR #1862, opened 2023-09-05
feat: expose metrics for all AWS API calls

threadproc:

This change will add a new Prometheus metric, `aws_api_calls_total`, exposed by controller-runtime that counts all requests made to the AWS APIs. This is useful when using expensive APIs like the Service Catalog API ($0.0007/request) for accounting purposes on accounts with multiple clusters.

I have:

This has been built and deployed to a test cluster where it was successfully observed to cause no errors when calling either v1 or v2 SDK methods, and successfully included the new metric on the `/metrics` endpoint.

### DIFF
--- a/cmd/provider/main.go
+++ b/cmd/provider/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/crossplane-contrib/provider-aws/apis/v1alpha1"
 	"github.com/crossplane-contrib/provider-aws/pkg/controller"
 	"github.com/crossplane-contrib/provider-aws/pkg/features"
+	"github.com/crossplane-contrib/provider-aws/pkg/utils/metrics"
 )
 
 func main() {
@@ -123,6 +124,7 @@ func main() {
 		log.Info("Alpha feature enabled", "flag", features.EnableAlphaManagementPolicies)
 	}
 
+	kingpin.FatalIfError(metrics.SetupMetrics(), "Cannot setup AWS metrics hook")
 	kingpin.FatalIfError(controller.Setup(mgr, o), "Cannot setup AWS controllers")
 	kingpin.FatalIfError(mgr.Start(ctrl.SetupSignalHandler()), "Cannot start controller manager")
 
--- a/pkg/utils/metrics/setup.go
+++ b/pkg/utils/metrics/setup.go
@@ -1,0 +1,24 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
+	k8smetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	metricAWSAPICalls = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "aws_api_calls_total",
+		Help: "Number of API calls to the AWS API",
+	}, []string{"service", "operation", "api_version"})
+)
+
+// SetupMetrics will register the known Prometheus metrics with controller-runtime's metrics registry
+func SetupMetrics() error {
+	return k8smetrics.Registry.Register(metricAWSAPICalls)
+}
+
+// IncAWSAPICall will increment the aws_api_calls_total metric for the specified service, operation, and apiVersion tuple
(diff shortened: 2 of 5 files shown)


pkg/controller/aws.go
```diff
@@ -54,8 +54,8 @@ import (
 	"github.com/crossplane-contrib/provider-aws/pkg/controller/kinesis"
 	"github.com/crossplane-contrib/provider-aws/pkg/controller/kms"
 	"github.com/crossplane-contrib/provider-aws/pkg/controller/lambda"
-	"github.com/crossplane-contrib/provider-aws/pkg/controller/mq"
-	"github.com/crossplane-contrib/provider-aws/pkg/controller/mwaa"
+	mwaa "github.com/crossplane-contrib/provider-aws/pkg/controller/mq"
+	mq "github.com/crossplane-contrib/provider-aws/pkg/controller/mwaa"
```
MisterMX, line 58: Theses changes should not be made. A fix is provided with #1863.

pkg/controller/aws.go
```diff
@@ -73,12 +73,15 @@ import (
...
 	"github.com/crossplane-contrib/provider-aws/pkg/controller/sqs"
 	"github.com/crossplane-contrib/provider-aws/pkg/controller/transfer"
+	"github.com/crossplane-contrib/provider-aws/pkg/utils/metrics"
 	"github.com/crossplane-contrib/provider-aws/pkg/utils/setup"
 )
 
 // Setup creates all AWS controllers with the supplied logger and adds them to
 // the supplied manager.
 func Setup(mgr ctrl.Manager, o controller.Options) error {
+	metrics.SetupMetrics(mgr, o)
```
MisterMX, line 83: I wonder if it would be better to call this in `cmd/provider/main.go` since this is some sort of "meta" feature that is not directly related to the controller setup.

MisterMX (changes requested):
Thank you very much for your contribution, @threadproc! I think including AWS metric would a very meaningful enhancement.

I have made some small remarks that would need to be solved before merging. Can you take a look at them?

pkg/controller/aws.go
threadproc: Agreed, I will make these changes

threadproc:
@MisterMX I've made the requested changes and resolved some linting errors as well, thanks for the review!

chlunde:
Great, very nice that you found a way to two this via hooks in the SDK!

cmd/provider/main.go
```diff
@@ -123,6 +124,8 @@ func main() {
 		log.Info("Alpha feature enabled", "flag", features.EnableAlphaManagementPolicies)
 	}
 
+	metrics.SetupMetrics()
```
MisterMX, line 127: To be consistent with the other setups, can we use `Register` instead of `MustRegister` and pass the error into `kingpin.FataIfError`? The result (a crash) would be the same, but it is more integrated into the control flow.

```suggestion
	kingpin.FatalIfError(metrics.SetupMetrics(), "Cannot setup AWS metrics hook")
```

pkg/utils/metrics/setup.go
```diff
@@ -0,0 +1,24 @@
...
+var (
+	metricAWSAPICalls = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "aws_api_calls_total",
+		Help: "Number of API calls to the AWS API",
+	}, []string{"service", "operation", "api_version"})
+)
+
+// SetupMetrics will register the known Prometheus metrics with controller-runtime's metrics registry
+func SetupMetrics() {
+	k8smetrics.Registry.MustRegister(metricAWSAPICalls)
```
MisterMX, line 18:
```suggestion
func SetupMetrics() error {
	return k8smetrics.Registry.Register(metricAWSAPICalls)
```

MisterMX (changes requested):

threadproc:
@MisterMX I've implemented the requested changes and squashed them down again, thank you!

MisterMX (approved):
LGTM. Thank you very much for your contribution @threadproc!

PR #1869, opened 2023-09-07
 fix(autoscaling): fix diff issue of auto scaling group

### LINKED ISSUE #1868: Diff bug of AutoScalingGroup resource
### What happened?

Crossplane will always try to update the AutoScalingGroup even if it has been updated.

### How can we reproduce it?

Create an AutoScalingGroup with `spec.forProvider.launchTemplate.launchTemplateName` field.
Then `isUpToDate` will always return false.

The diff will fail because the field `launchTemplateID` is returned by `DescribeAutoScalingGroup` but is not set in spec.

Crossplane version: 0.43.1

liubog2008:

Fixes #1868 

I have:

### DIFF
--- a/pkg/controller/autoscaling/autoscalinggroup/setup.go
+++ b/pkg/controller/autoscaling/autoscalinggroup/setup.go
@@ -2,6 +2,7 @@ package autoscalinggroup
 
 import (
 	"context"
+	"fmt"
 	"sort"
 
 	svcsdk "github.com/aws/aws-sdk-go/service/autoscaling"
@@ -13,6 +14,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/google/go-cmp/cmp"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	svcapitypes "github.com/crossplane-contrib/provider-aws/apis/autoscaling/v1beta1"
@@ -71,15 +73,15 @@ func isUpToDate(_ context.Context, obj *svcapitypes.AutoScalingGroup, obs *svcsd
 	asg := obs.AutoScalingGroups[0]
 
 	if !cmp.Equal(in.CapacityRebalance, asg.CapacityRebalance) {
-		return false, "", nil
+		return false, "spec.forProvider.capacityRebalance", nil
 	}
 	// DefaultInstanceWarmup can be updated
 	if !cmp.Equal(in.DefaultInstanceWarmup, asg.DefaultInstanceWarmup) {
-		return false, "", nil
+		return false, "spec.forProvider.defaultInstanceWarmup", nil
 	}
 	// DesiredCapacityType can be updated
 	if !cmp.Equal(in.DesiredCapacityType, asg.DesiredCapacityType) {
-		return false, "", nil
+		return false, "spec.forProvider.desiredCapacityType", nil
 	}
 	// Context is reserved
 	// if !cmp.Equal(*in.Context, *asg.Context) {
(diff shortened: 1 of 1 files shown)


liubog2008:
@MisterMX PTAL

pkg/controller/autoscaling/autoscalinggroup/setup.go
```diff
@@ -219,6 +239,18 @@ func isUpToDate(_ context.Context, obj *svcapitypes.AutoScalingGroup, obs *svcsd
...
+	if a != nil && b != nil {
+		if *a != *b {
+			return false
+		}
+	}
+	if a == nil || b == nil {
+		return false
+	}
+	return true
+}
```
MisterMX, line 252: Have you considered https://pkg.go.dev/k8s.io/utils@v0.0.0-20230726121419-3b25d923346b/ptr#Equal ? I think it can be used instead of a custom implementation.

liubog2008: PTAL again

MisterMX (changes requested):

pkg/controller/autoscaling/autoscalinggroup/setup.go
```diff
@@ -72,84 +74,90 @@ func isUpToDate(_ context.Context, obj *svcapitypes.AutoScalingGroup, obs *svcsd
...
 	// VPCZoneIdentifier can be updated
 	if !cmp.Equal(in.VPCZoneIdentifier, asg.VPCZoneIdentifier) {
-		return false, "", nil
+		return false, "spec.forProvider.vpcZoneIdentifier", nil
 	}
 	// LaunchTemplate can be updated
 	if in.LaunchTemplate != nil && asg.LaunchTemplate != nil {
-		if !cmp.Equal(in.LaunchTemplate.LaunchTemplateID, asg.LaunchTemplate.LaunchTemplateId) {
-			return false, "", nil
+		if in.LaunchTemplate.LaunchTemplateID != nil {
```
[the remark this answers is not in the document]
liubog2008: if `launchTemplateID` is not set in spec, diff will failed because `launchTemplateId` is always set in describe result.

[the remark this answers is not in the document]
liubog2008: That's why I pull this request.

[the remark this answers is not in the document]
liubog2008: @MisterMX PTAL again

liubog2008:
@hasheddan @krishchow @chlunde @MisterMX anyone could give me more suggestions about this PR?

csuzhangxc:
@hasheddan @krishchow @chlunde @MisterMX anyone could give more suggestions about this PR?

pkg/controller/autoscaling/autoscalinggroup/setup.go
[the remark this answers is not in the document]
MisterMX: I think it would make more sense then to set the field in `LateInitialize` so it is reflected in spec if it is not set there. Then would not need this nil check here.

MisterMX:
@liubog2008 @csuzhangxc regarding my comment https://github.com/crossplane-contrib/provider-aws/pull/1869#discussion_r1341434584 could you add a late init step for launchtemplate? Also, could you rebase your changes on the latest master? Thanks!

liubog2008:
@MisterMX LaunchTemplateId and LanchTemplateName cannot both be set in update api call.

pkg/controller/autoscaling/autoscalinggroup/setup.go
```diff
@@ -179,8 +187,21 @@ func isUpToDate(_ context.Context, obj *svcapitypes.AutoScalingGroup, obs *svcsd
...
+		if !ptr.Equal(a.Key, b.Key) {
+			return false, fmt.Sprintf("spec.forProvider.tags[%d].key", i), nil
+		}
+		if !ptr.Equal(a.Value, b.Value) {
+			return false, fmt.Sprintf("spec.forProvider.tags[%d].value", i), nil
+		}
+		if !ptr.Equal(a.PropagateAtLaunch, b.PropagateAtLaunch) {
+			return false, fmt.Sprintf("spec.forProvider.tags[%d].propagateAtLaunch", i), nil
+		}
 	}
```
MisterMX, line 205: I think it would be simpler to just use `cmp.Diff` with a sort op:

```suggestion
        if diff := compareTags(asg.Tags, in.Tags); diff != "" {
        	return false, "spec.forProvider.tags:" + diff, nil
        }
```

```golang
func compareTags(spec, current []svcapitypes.Tag) string {
	return cmp.Diff(current, spec, cmpopts.SortSlices(func(i, j svcapitypes.Tag) bool {
		return ptr.Deref(i.Key, "") < ptr.Deref(j.Key, "")
	}))
}
```

MisterMX (changes requested):

MisterMX:
> @MisterMX LaunchTemplateId and LanchTemplateName cannot both be set in update api call.

Then there needs to be some kind of check before making the update call that only sets one of them if both are defined.

liubog2008:
> Then there needs to be some kind of check before making the update call that only sets one of them if both are defined.

That's hard to determine which one we should use in update call. If we set both of them in `LateInitialize`, we have to change them at same time. However, what I want is always only set  `LanchTemplateName` when creating and updating the asg.

And I think `spec` means what I want but not what really is. `LanchTemplateId` is not the user set but the system auto selected so I don't think it have to be initialized in `LateInitialize`.

liubog2008:
1. We set `LanchTemplateName` to xxx and create an asg.
2. `LanchTemplateId` is late initialized.
3. We change the `LanchTemplateName` to yyy
4. The provider cannot determine which one should be used in api call.

pkg/controller/autoscaling/autoscalinggroup/setup.go
liubog2008: Types of in.Tags and asg.Tags are not same

MisterMX: You are right. I did not see that.

MisterMX:
I had a bit of thinking around this issue. Usually we would late initialize properties in the spec that are set by AWS. However, AWS only allows to set one of `launchTemplateId` and `launchTemplateName` so we cannot do that.

I thought we could something similar to #1907 and perform a lookup for the ID before comparison and only let the user specify the name. But this would mean a breaking change and it also breaks with the design principle that a provider should match the external API as closely as possible. So this is not an option, either.

In conclusion I believe this PR is good as it is now and we can move on with this.

Thanks @liubog2008 for your contribution and the discussion!

MisterMX (approved):

PR #1882, opened 2023-10-04
fix(kafka/cluster): fill connectionDetails with bootstrap brokers

### LINKED ISSUE #1711: MSK: write bootstrap connection strings
When MSK is provisioned, the broker information is written to the desired connection string. To connect to MSK, it's preferred to use the bootstrap TLS and bootstrap Plaintext strings. These strings should be written out as connection information. 

I'm not a kafka expert, but I don't think it's desirable to report back the broker URLs. The bootstrap connections will allow clients to discover them.

wotolom:

use an `GetBootstrapBrokers` API call (https://docs.aws.amazon.com/cli/latest/reference/kafka/get-bootstrap-brokers.html) to correctly fill in the different endpoints of the AWS MSKafka Cluster into the `ConnectionDetails`

- add `GetBootstrapBrokers` API call in `postObserve` to insert endpoints into `ConnectionDetails`
- correct some wrongly set variable in `generateClientAuthentication`

Fixes #1711

I have:

manually

### DIFF
--- a/pkg/controller/kafka/cluster/setup.go
+++ b/pkg/controller/kafka/cluster/setup.go
@@ -40,6 +40,7 @@ import (
 )
 
 const (
+	errGetBootstrapBrokers        = "cannot get BootstrapBrokers of Cluster in AWS"
 	errUpdateBrokerType           = "cannot update BrokerType of Cluster in AWS"
 	errUpdateBrokerStorage        = "cannot update BrokerStorage of Cluster in AWS"
 	errUpdateBrokerCount          = "cannot update BrokerCount of Cluster in AWS"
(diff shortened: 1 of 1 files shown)


pkg/controller/kafka/cluster/setup.go
```diff
@@ -121,13 +122,37 @@ func preObserve(_ context.Context, cr *svcapitypes.Cluster, obj *svcsdk.Describe
...
 	switch awsclients.StringValue(obj.ClusterInfo.State) {
 	case string(svcapitypes.ClusterState_ACTIVE):
 		cr.SetConditions(xpv1.Available())
+
+		// see: https://docs.aws.amazon.com/msk/latest/developerguide/msk-get-bootstrap-brokers.html
+		// retrieve cluster bootstrap brokers (endpoints)
+		// not possible in every cluster state (e.g. "You can't get bootstrap broker nodes for a cluster in DELETING state.")
+		endpoints, err := u.client.GetBootstrapBrokersWithContext(ctx, &svcsdk.GetBootstrapBrokersInput{ClusterArn: awsclients.String(meta.GetExternalName(cr))})
+		if err != nil {
+			return obs, awsclients.Wrap(err, errBootstrapBrokers)
```
MisterMX, line 146:
```suggestion
			return obs, awsclients.Wrap(err, errGetBootstrapBrokers)
```

MisterMX (changes requested):

pkg/controller/kafka/cluster/setup.go
wotolom: Makes sense. Done.

MisterMX (approved):

PR #1886, opened 2023-10-06
feat(kafka): Support cross account access

MisterMX:

Adds cluster policy and connectivity properties that allow configuring cross account accessibility of a broker.

It is mostly copy-paste from the specialized bucket policy types into a generic `ResourcePolicy` struct and a bunch of late initialization and conversion methods.

It also fixes a bug that `spec.forProvider.encryptionInfo` was not late initialized properly which led to an error during updates.

I have:

Using the existing example.


chlunde (approved):

chlunde:
I see lint due to k8 utils pointer just "fails green" due to other issues in other builds, while this PR the linter ran to completion and reported the error.

PR #1890, opened 2023-10-09
fix(kinesis): fix bug related with the use of external-name

### LINKED ISSUE #1751: Kinesis stream resource not working as expected when using 'external-name' annotation
### What happened?
Creating a Kinesis stream object with the _crossplane.io/external-name_ annotation set with a name including underscores and capital letters doesn't work as expected. 
Instead of creating a stream object with the name provided on the _external-name_ annotation, Crossplane creates two streams: one with the external name provided and one with the external name to lowercase and scores (instead of underscores). Kubernetes stream.kinesis.aws.crossplane.io/v1alpha1 resource created previously is binded to the AWS resource on lower case and doesn't respect the _external name_ annotation. 

This issue occurs on a dev environment using Localstack, but also on a production one using AWS as a real cloud provider.

### How can we reproduce it?

1. Create a Kind cluster and install Crossplane controller and provider-aws v33 using this command: 

`helm upgrade --install crossplane crossplane -n crossplane-system --repo https://charts.crossplane.io/stable --create-namespace --set provider.packages="{xpkg.upbound.io/crossplane-contrib/provider-aws:v0.33.0}" --wait --debug`

2. Clone [my repo](https://github.com/alexfernandezv/kinesis-crossplane-bug) to reproduce it.
3. Apply the yaml manifests on the root folder by executing: 
`kubectl apply -f .`

4. Wait until Localstack is deployed and the stream is reconciled with Crossplane & Localstack. 
5. Execute `kubectl describe stream infinity-stream` and you will see that the stream ARN does not match with the external name defined on the _kinesis-stream.yaml_. 
6. Execute `kubectl exec -i -t deployment/localstack -- awslocal kinesis list-streams --region eu-central-1` two streams are created, when only requesting creation of one:

```
{
    "StreamNames": [
        "INFINITY_STREAM_DEVELOPMENT",
        "infinity-stream"
    ]
}
```

Crossplane version: crossplane/crossplane:v1.12.1
* Kubernetes version: 1.24

nilde:
When you have a different external-name compared to name and a bug occurres and generates 2 different streams one with the original name and other with the external name, also solves the problem related with getting the correct stream status.

Signed-off-by: Domene Esteban, Nil
<nildomenee@gmail.com>

Fixes #1751 

I have:

Tested locally using Localstack creating multiple Streams with different configurations.

### DIFF
--- a/pkg/controller/kinesis/stream/setup.go
+++ b/pkg/controller/kinesis/stream/setup.go
@@ -58,6 +58,10 @@ func SetupStream(mgr ctrl.Manager, o controller.Options) error {
 
 	reconcilerOpts := []managed.ReconcilerOption{
 		managed.WithExternalConnecter(&connector{kube: mgr.GetClient(), opts: opts}),
+		managed.WithReferenceResolver(managed.NewAPISimpleReferenceResolver(mgr.GetClient())),
+		managed.WithInitializers(
+			managed.NewDefaultProviderConfig(mgr.GetClient()),
+			managed.NewNameAsExternalName(mgr.GetClient())),
 		managed.WithPollInterval(o.PollInterval),
 		managed.WithLogger(o.Logger.WithValues("controller", name)),
 		managed.WithRecorder(event.NewAPIRecorder(mgr.GetEventRecorderFor(name))),
@@ -124,7 +128,6 @@ func postCreate(_ context.Context, cr *svcapitypes.Stream, obj *svcsdk.CreateStr
 	if err != nil {
 		return managed.ExternalCreation{}, err
 	}
-	meta.SetExternalName(cr, *awsclients.String(meta.GetExternalName(cr)))
 	return managed.ExternalCreation{ExternalNameAssigned: true}, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/controller/kinesis/stream/setup.go
```diff
@@ -124,7 +124,7 @@ func postCreate(_ context.Context, cr *svcapitypes.Stream, obj *svcsdk.CreateStr
 	if err != nil {
 		return managed.ExternalCreation{}, err
 	}
-	meta.SetExternalName(cr, cr.Name)
+	meta.SetExternalName(cr, *awsclients.String(meta.GetExternalName(cr)))
```
MisterMX, line 127: This would have no effect and just writes the value of the field into itself.

Given that in `preCreate` it is assumed that the external name is already set (and where it is actually needed), there is no need to set in `postCreate`.

I agree that using `cr.Name` makes no sense and is against the convention (that external name should be used). If we want to use the external name as default if external name is not set, we should use the `managed.NewNameAsExternalName` initializer as it is done in other places like https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/controller/batch/computeenvironment/setup.go#L61.

MisterMX (changes requested):

MisterMX (approved):
LGTM. Thank you very much @nilde!

PR #1893, opened 2023-10-10
fix(rds): Infer cache secret namespace from pod environment

### LINKED ISSUE #1834: [DBInstance] Needs namespace "crossplane-system" to store values in cache (version 0.42)
### What happened?
After the upgrade for version 0.42 I am not able to create new RDS due to error:

```
    "Message: create failed: pre-create failed: cannot cache password: cannot create object: namespaces "crossplane-system" not found"
```

This is probably due to the refactor of the password cache ( https://github.com/crossplane-contrib/provider-aws/pull/1756 ) cc. @schroeder-paul / @MisterMX . It seems the Cache namespace is not fetching the namespace from "masterUserPasswordSecretRef", so the fix seems to be something simple. 

I am using:

```
    masterUserPasswordSecretRef:
      key: password
      name: example-dbinstance
      namespace: default
```
So, I guess Crossplane should use that namespace for caching.

### How can we reproduce it?

Here is the manifest applied

```
apiVersion: rds.aws.crossplane.io/v1alpha1
kind: DBInstance
metadata:
  name: example-dbinstance
spec:
  forProvider:
    region: eu-central-1
    allocatedStorage: 20
    dbName: example
    engine: postgres
    engineVersion: "12.9"
    dbInstanceClass: db.t3.micro
    masterUsername: adminuser
    masterUserPasswordSecretRef:
      key: password
      name: example-dbinstance
      namespace: default
  writeConnectionSecretToRef:
    name: example-dbinstance-out
    namespace: default
  providerConfigRef:
    name: example

---
apiVersion: v1
kind: Secret
metadata:
  name: example-dbinstance
  namespace: default
type: Opaque
data:
  password: dGVzdFBhc3N3b3JkITEyMw== # testPassword!123
```

Crossplane version: 0.42

Thanks for the help! :)

MisterMX:

This adds a function that infers the namespace of the RDS password cache secret from the following locations (in order):

1. `POD_NAMESPACE` env var
2. `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file
3. fallback to `crossplane-system`

Until now `crossplane-system` was the hardcoded value.

Fixes #1834

Please note that this might cause a regeneration of passwords if the provider is running in another namespace than `crossplane-system`, but, due to the change, is now looking for the secret in another namespace where it does not exist.

I have:

n.a.

### DIFF
--- a/pkg/clients/rds/common.go
+++ b/pkg/clients/rds/common.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	svcapitypes "github.com/crossplane-contrib/provider-aws/apis/rds/v1alpha1"
+	kubeutils "github.com/crossplane-contrib/provider-aws/pkg/utils/kube"
 )
 
 // Publicly usable variables
@@ -53,8 +54,6 @@ const (
 	errGetCachedPassword    = "cannot get cached password"
 	errGetCachedRestoreInfo = "cannot get cached restore info"
 	errGetMasterPassword    = "cannot get master password"
-
-	secretNamespace = "crossplane-system"
 )
 
 type restoreSate string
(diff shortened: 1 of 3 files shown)


chlunde (approved):

flaviomoringa:
@MisterMX I was waiting for this fix, but apparently its still not in the latest version, right?

MisterMX:
> I was waiting for this fix, but apparently its still not in the latest version, right?

We haven't backported this since it is a breaking change. It is going to be in the next minor release.

PR #1897, opened 2023-10-12
fix(rds): Add missing tag updates

### LINKED ISSUE #494: RDS tags inconsistent with the resource created in the cloud during composition update
### What happened?

RDS User defined tags is not tally with the one created in the cloud 

### How can we reproduce it?
1. Existing RDSInstance objects has been created with some tags
2. Change the composition spec. Adding new tags there
3. RDSInstance objects managed to get the updated value from the composition file and status show as SYNC true
4. Verify in RDS Console, some RDS Instance still not reflecting the newly added tags (not every instances has this issue or intermittent), only 1 out of 5 RDS get proper Tags update

Crossplane version: 1.0.0-rc.0.26.gd8254833

* Cloud provider or hardware configuration: AWS
* Kubernetes version (use `kubectl version`)

Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-eks-e16311", GitCommit:"e163110a04dcb2f39c3325af96d019b492541
9eb", GitTreeState:"clean", BuildDate:"2020-03-27T22:40:13Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95
ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"} 

* Kubernetes distribution (e.g. Tectonic, GKE, OpenShift) : EKS

### Suspect:

Referencing the External Tags logic in the `Initialize` phase which includes the Tags sorting: https://github.com/crossplane/provider-aws/blob/cd5da906f3f1fc8e259d78e9351584e6e4cfddd2/pkg/controller/database/rdsinstance.go#L285

But not during the ``Update`` phase, patching of the Tags may required some kind of sorting before calling the : https://github.com/crossplane/provider-aws/blob/cd5da906f3f1fc8e259d78e9351584e6e4cfddd2/pkg/controller/database/rdsinstance.go#L223

MisterMX:

Adds missing tag comparisons and updates to

* `rds.DBClusterParameterGroup`
* `rds.DBParameterGroup`
* `rds.DBInstance`

Fixes #494

I have:

Simple tag adds and removals using the already existing examples.

### DIFF
--- a/pkg/controller/rds/dbinstance/setup.go
+++ b/pkg/controller/rds/dbinstance/setup.go
@@ -3,6 +3,7 @@ package dbinstance
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
 	"sort"
 	"strconv"
@@ -39,6 +40,8 @@ const (
 	errSnapshotRestoreFailed    = "cannot restore DB instance from snapshot"
 	errPointInTimeRestoreFailed = "cannot restore DB instance from point in time"
 	errUnknownRestoreSource     = "unknown DB Instance restore source"
+	errAddTags                  = "cannot add tags"
+	errRemoveTags               = "cannot remove tags"
 )
 
 // time formats
@@ -104,6 +107,11 @@ type custom struct {
 	kube     client.Client
 	client   svcsdkapi.RDSAPI
 	external *external
+
+	cache struct {
+		addTags    []*svcsdk.Tag
+		removeTags []*string
+	}
 }
 
 func preObserve(_ context.Context, cr *svcapitypes.DBInstance, obj *svcsdk.DescribeDBInstancesInput) error {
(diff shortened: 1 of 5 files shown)


pkg/controller/rds/dbinstance/setup.go
```diff
@@ -104,6 +107,11 @@ type custom struct {
 	kube     client.Client
 	client   svcsdkapi.RDSAPI
 	external *external
+
+	cache struct {
```
chlunde, line 111: Have we done this anywhere else? I wonder if we have any guarantees that this will not be reused for other resources.

I think it is Ok..

chlunde (approved):

pkg/controller/rds/dbinstance/setup.go
MisterMX: Yes, I have this multiple times. Since the hook is created for every reconcile, we can be sure that there is always a unique `hook` (or `external`) for every reconcile.

PR #1900, opened 2023-10-12
fix(mq): Do not set EncryptionOptions for RabbitMQ

### LINKED ISSUE #1643: AmazonMQBroker RabbitMQ encryptionOptions is not supported for RabbitMQ brokers.
### What happened?

We had applied an update to our rabbitmq broker and then the state was going to synced false and ready false.

The error is `encryptionOptions is not supported for RabbitMQ brokers.`
Which is also documented here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-amazonmq-broker-encryptionoptions.html

Looks like the encryption option needs a condition here in the late init if its a activeMQ or rabbitMQ
https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/controller/mq/broker/setup.go#L177

### How can we reproduce it?

Create rabbitmq and try to update it

Crossplane version: 
crossplanecontrib/provider-aws:v0.36.0
crossplane/crossplane:v1.10.1

MisterMX:

Adds a small check that unset `EncryptionOptions` in `preCreate` if engine type `RabbitMQ`.

Fixes #1643

I have:

n.a.

### DIFF
--- a/pkg/controller/mq/broker/setup.go
+++ b/pkg/controller/mq/broker/setup.go
@@ -2,6 +2,7 @@ package broker
 
 import (
 	"context"
+	"strings"
 	"time"
 
 	svcsdk "github.com/aws/aws-sdk-go/service/mq"
@@ -15,6 +16,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


chlunde:
I think we also could prevent this combination using CEL expressions/xvalidation? I'm not super confident that ignoring a security option set is the best approach here, but I'll approve it and leave it for you.

chlunde (approved):

MisterMX:
> I think we also could prevent this combination using CEL expressions/xvalidation? I'm not super confident that ignoring a security option set is the best approach here, but I'll approve it and leave it for you.

I also thought about this. I am not super with this solution as it adds some additional "magic" inside the controller that is not visible from the outside. However, as it is written in the AWS docs, encryption options are totally unsupported for RabbitMQ so it makes sense to just leave it out for convenience.

PR #1907, opened 2023-10-13
fix(dynamodb): Resolve KMSMasterKeyArn for comparison

### LINKED ISSUE #1440: Cannot update dynamodb table with kms encryption
### What happened?

1. I've created a `dynamodb` table using `provider-aws` without explicit encryption resulting in the default `Owned by Amazon DynamoDB` encryption as expected.

2. Updated the same table definition with a customer key by specifying `kmsMasterKeyID` with my key alias. 
I can see that the managed resource (table) has the correct updated encryption and the resource is synced and ready but the `dynamodb` table keeps the initial `Owned by Amazon DynamoDB` encryption - Drift between expected and actual state

4. Finally I've added a global secondary index to the managed resource definition and **somehow** it triggered the change on table encryption to the one I've defined before but the index wasn't created.

So in the end (after creating the index without success) I got the encryption that I wanted but I can see this error on the managed resource:

`cannot update Table in AWS: LimitExceededException: Subscriber limit exceeded: Encryption mode changes are limited in the 24h window ending at 2022-08-20T08:42:25.398Z. After the first 4 change, each subsequent change in the same window can be performed │
│  at most once every 21600 seconds. Number of updates today: 4` 

### How can we reproduce it?
Create the `dynamodb` table:

```yaml
apiVersion: dynamodb.aws.crossplane.io/v1alpha1
kind: Table
metadata:
  name: crossplane-table-test
spec:
  providerConfigRef:
    name: aws-provider
  forProvider:
    region: eu-west-1
    attributeDefinitions:
      - attributeName: Id
        attributeType: S
    billingMode: PAY_PER_REQUEST
    keySchema:
      - attributeName: Id
        keyType: HASH
```

Create a customer managed key with an alias:
```yaml
apiVersion: kms.aws.crossplane.io/v1alpha1
kind: Key
metadata:
  name: crossplane-key-test
spec:
  forProvider:
    region: eu-west-1
    enableKeyRotation: true
  providerConfigRef:
    name: aws-provider
---
apiVersion: kms.aws.crossplane.io/v1alpha1
kind: Alias
metadata:
  name: crossplane-key-alias-test
spec:
  forProvider:
    region: eu-west-1
    targetKeyIdRef:
      name: crossplane-key-test
  providerConfigRef:
    name: aws-provider
```
Update the table definition by adding customer encryption - Encryption won't change:
```yaml
#[...]
sseSpecification:
  enabled: true
  kmsMasterKeyID: alias/crossplane-key-alias-test
  sseType: "KMS"
```

Update table definition to add a `global secondary index`:
```yaml
#[...]
attributeDefinitions:
  # add the bellow attribute
  - attributeName: OtherId
    attributeType: S
globalSecondaryIndexes:
   - indexName: OtherIdIndex
        keySchema:
          - attributeName: OtherId
            keyType: HASH
        projection:
          projectionType: ALL
```

You will see the encryption changing and the error appearing on table managed resource.

Crossplane version: v1.9.0
AWS Provider version: 0.30.0

* Cloud provider or hardware configuration: M1
* Kubernetes version: v1.24.3
* Kubernetes distribution: Kind
* OS: MacOs 12.3.1
* Kernel: Darwin

MisterMX:

This adds an extra step in `isUpToDate` to resolve the ARN of the key that is referenced by whatever is given in `spec.forProvider.sseSpecification.kmsMasterKeyId`. This is necessary because AWS always returns the ARN even if the create or update call was made with an alias which triggered constant update calls.

Fixes #1440

This uses Gomock to mock the KMS client. I the same step I added a bunch of refactoring, like moving all mocks in the same package and splitting the API code generation from mock code generation.

I have:

Unit tests and existing examples


pkg/controller/dynamodb/table/hooks.go
```diff
@@ -365,11 +405,21 @@ func createPatch(in *svcsdk.DescribeTableOutput, target *svcapitypes.TableParame
 	return patch, nil
 }
 
-func isCoreResourceUpToDate(cr *svcapitypes.Table, resp *svcsdk.DescribeTableOutput) (bool, error) {
+func (e *updateClient) getMKSKeyArnFromID(ctx context.Context, keyId *string) (*string, error) {
```
chlunde, line 408: nit: KMSKey?

chlunde (approved):

pkg/controller/dynamodb/table/hooks.go
MisterMX: Thanks for pointing that one out!

PR #1908, opened 2023-10-13
feat(servicecatalog): Add provisioned product

teeverr:
Adds new resource - ProvisionedProduct, which represents life cycle of launched product in service catalog

I have:

Many scenarios have been tested manually In development env and also I wrote some unit tests for custom controller logic

It is a new PR, instead of [this one](https://github.com/crossplane-contrib/provider-aws/pull/1806)


pkg/utils/metrics/setup.go
```diff
@@ -10,11 +10,37 @@ var (
...
+	// MetricAWSAPICallsRec can be used in controllers for exposing a total amount of AWS API calls for every create/observe/update/delete operation
+	MetricAWSAPICallsRec = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "aws_api_calls_reconcile_managed_resource_total",
+		Help: "Amount of calls to the AWS API produced by controller per reconciliation for every managed resource and controller operation type",
+	}, []string{"api_version", "kind", "resource_name", "controller_operation_type"})
+	// MetricManagedResRec can be used in controllers for exposing a total amount of reconciliations for every managed resource
+	MetricManagedResRec = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "managed_resource_reconciles_total",
+		Help: "Total amount of reconciliation loops per managed resource",
+	}, []string{"api_version", "kind", "resource_name"})
```
MisterMX, line 22: I'm not so sure about these additional metrics. Are they really necessary?

If this would be applied to every controller, it would create a huge number of additional metrics: `189 CRDs x 4 x #MRs`.

If it is not actually necessary I would propose to leave it out of this PR to get it merged and maybe open a second one for the metrics.

MisterMX (changes requested):

pkg/utils/metrics/setup.go
teeverr: @MisterMX thanks for looking at the pr!

>I'm not so sure about these additional metrics. Are they really necessary?
Well, it depends. I think this question is actually for a major part of 6600+ metrics which are already exposed, including `aws_api_calls_total`.
I believe it is nice to have, because based on these metrics you can precisely observe the amount of calls for each of your managed resources, it helps find 'right' poll interval and understand your aws cost.

>If this would be applied to every controller, it would create a huge number of additional metrics: 189 CRDs x 4 x #MRs.

Not exactly: )
Unlike metricAWSAPICalls(aws_api_calls_total) these two new counters should be implemented `manually` for every controller, because there is no way(or I didn't find it) to use middleware for automated exposing. No data about managed resources and reconciles on the client scope.
So these metrics work only for ProvisioneProduct for now, and only if it is used, I mean these metrics are exposed per managed resource.

 As I texted above, the metrics exposed per managed resource(ProvisionedProduct) but not every service CRD. But if so(it is not), it is okay for prom endpoints. Current payload of the `/metrics` endpoint is around ~650K which is not small but okay and 400+ metrics doesn't make a big change. But again, it is not 400+. 

>If it is not actually necessary I would propose to leave it out of this PR to get it merged and maybe open a second one for the metrics.

I also thought about it, but these two metrics work only if implemented on the controller side, that is why I decided to put everything in this PR. Cuz I didn't touch other/existeng resources, it's a lot of work and testing.



Ps: I found inaccuracy in the description of `aws_api_calls_reconcile_managed_resource_total` - it is total counter but not `per reconciliation`. I will fix it if my arguments have dispelled your concerns.

MisterMX: Thanks for the clarification. If I understand it right if it would be implemented for every controller we would still most likely end up with at least 4 metrics per MR? Given that users could have thousands of MRs this would mean a huge number of additional metrics (on top of the already huge amount).

I understand why you want this for this particular controller. However, I would like to have the metrics part to be implemented and discussed separately from this PR. I am not opposed to have additional and more detailed metrics for controllers. But there are some more things that should be considered in this discussion, like the implementation for other controllers and the amount of metrics being created. Ideally, we should also give users the ability to control which metrics should be exposed so we are not creating to much spam.

teeverr: >least 4 metrics per MR?

Even more - 5
```
aws_api_calls_reconcile_managed_resource_total{api_version="servicecatalog.aws.crossplane.io/v1alpha1",controller_operation_type="create",kind="ProvisionedProduct",resource_name="kirills-test-2"} 1
aws_api_calls_reconcile_managed_resource_total{api_version="servicecatalog.aws.crossplane.io/v1alpha1",controller_operation_type="delete",kind="ProvisionedProduct",resource_name="kirills-test-2"} 0
aws_api_calls_reconcile_managed_resource_total{api_version="servicecatalog.aws.crossplane.io/v1alpha1",controller_operation_type="observe",kind="ProvisionedProduct",resource_name="kirills-test-2"} 27415
aws_api_calls_reconcile_managed_resource_total{api_version="servicecatalog.aws.crossplane.io/v1alpha1",controller_operation_type="update",kind="ProvisionedProduct",resource_name="kirills-test-2"} 0
managed_resource_reconciles_total{api_version="servicecatalog.aws.crossplane.io/v1alpha1",kind="ProvisionedProduct",resource_name="kirills-test-2"} 5483
```

>However, I would like to have the metrics part to be implemented and discussed separately from this PR
...
Ideally, we should also give users the ability to control which metrics should be exposed so we are not creating to much spam.

Okay, I will move metrics to dedicated branch and then create a pr, also I will think about feature flag or something. But first I'll cleanup this branch, it will take some time. Probably tomorow it will be ready.

teeverr: @MisterMX it is done - I removed metrics from this PR.

teeverr: @MisterMX could you run ci checks once again?
Idk what is wrong, [here](https://github.com/swisscom/provider-aws/actions/runs/6577103192) and localy everything is okay.

MisterMX: @teeverr some refactoring has been made in #1917. The code in `pkg/clients` has been moved to `pkg/utils`. You need to rebase on the latest master.

teeverr: Oh, I see. Thanks!
I rebased it with the latest master

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,367 @@
...
+
+func (c *custom) provisioningParamsAreChanged(ctx context.Context, ds *svcapitypes.ProvisionedProduct) (bool, error) {
+	nn := types.NamespacedName{
+		Name: ds.GetName(),
+	}
+	xr := svcapitypes.ProvisionedProduct{}
+	err := c.kube.Get(ctx, nn, &xr)
+	if err != nil {
+		return false, err
+	}
```
MisterMX, line 285: I don't think we need to call K8s a second time as we already have the latest version of the MR.

What should happen to compare MRs with the external state:

1. Observe external resource
2. LateInialize MR fields with external values if unset (to detect late initialization, you can create a deep copy)
3. Compare the late initialized MR with the observed state. There are multiple ways to do this: One would be to compare all fields manually or generate a spec from the external state and compare it with `cmp.Diff`.

MisterMX (changes requested):

pkg/controller/servicecatalog/provisionedproduct/setup.go
teeverr: If I understood you correctly, but I'm not sure, it won't work.

The context and a problem are that:
external resource(cloudformation stack which can be found by  CloudformationStackARN from a provisioned product outputs) keeps all provisioning params every time, even if some of them are not specified when a product was ran, some times value for not specified params is empty string, sometimes it is a default value, depends on product.  
E.g. we have a product with 10 params, but a user specified only two of them, in this case CloudformationStackARN will have all 10 params, 2 of them set by an user and others have default or empty values.

In this case it is possible to detect if some of parameters where deleted only if a user specified all of them, which is not necessary and usually users specified only what a part of parameters. Ss mentioned before, external resource has all parameters and if MR has only 2/10 of params, how can we be sure it was not have 3/10 before, comparison with exteranl shows nothing. 
That's why I added atProvider.status.lastProvisiongParams which are ProvisioningParams from the previous reconciliation loop, and the controller use them for comparison with current MR params.

>or generate a spec from the external state and compare it with cmp.Diff
I used cmp.Equal because it allows to sort array before comparison, the idea is do not run update if only sequence of params are changred

pkg/controller/servicecatalog/provisionedproduct/setup.go
```diff
@@ -0,0 +1,367 @@
...
+
+func (c *custom) lateInitialize(spec *svcapitypes.ProvisionedProductParameters, _ *svcsdk.DescribeProvisionedProductOutput) error {
+	acceptLanguageEnglish := acceptLanguageEnglish
+	spec.AcceptLanguage = pointer.LateInitializeStringPtr(spec.AcceptLanguage, &acceptLanguageEnglish)
+	return nil
+}
+
+func preCreate(_ context.Context, ds *svcapitypes.ProvisionedProduct, input *svcsdk.ProvisionProductInput) error {
+	input.ProvisionToken = aws.String(genIdempotencyToken())
+	if ds.GetName() != meta.GetExternalName(ds) {
```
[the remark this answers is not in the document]
teeverr: >metadata.name should not be used to refer to the external resource. The controllers should always only consider meta.GetExternalName. The external name does not belong to the spec as well.

Cool, so I can delete preCreate at all. 

>Do we need to provide support for both ProvisionedProductName and ProvisionedProductId? 

I tried to give as much functionality as possible, in major cases ProvisionedProductName should be enough, but if some want to use id, why not.

>In that case we need to find a way to store both of them in the external name.

Are you sure? AWS API allows you to use only name or id at the same time, so if external-name reflects this logic it is sounds transparent and correct for me.

pkg/controller/servicecatalog/provisionedproduct/setup.go
MisterMX: You already have the latest status of the MR during reconcile. Because it is overwritten during observe (I guess thats why you reretrieved it from the cluster) you can store a copy of it in `custom` during `preObserve`.

> I used cmp.Equal because it allows to sort the arrays before comparison, the idea is do not run update if only a sequence of params is changred

`isUpToDate` allows you to return a diff that is printed in the logs and might help users to determine issues during reconciles. It might be a good idea to use `cmp.Diff` and return its result.

teeverr: >You already have the latest status of the MR during reconcile

Exactly, *svcapitypes.ProvisionedProduct represent MR during reconcile, but for detection of removed paramers we should compare provisioning params(Spec.ForProvider.ProvisioningParameters) with params from previous reconcile(Status.AtProvider.LastProvisioningParameters) but *svcapitypes.ProvisionedProduct doesn't contian this data before we set it postObserve(this data will be used for comparison with provisioning params in the next reconciliation loop)

pkg/controller/servicecatalog/provisionedproduct/setup.go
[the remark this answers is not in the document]
teeverr: Okay, copy that. I think we can remove support of ProvisionedProductID as external-name.

pkg/controller/servicecatalog/provisionedproduct/setup.go
MisterMX: By "latest status of the MR" I meant the params from the previous reconcile. There is no need to call `kube.Get` a second time because you can just store `cr.DeepCopy()` in `custom.cache` during `preObserve` before `cr` gets updated with the external values.

teeverr: Sorry, I didn't understand the idea the first time(and second too : ) ), now I got it. Thanks!
Fixed this. I also tried cmp.Diff and it is too humanreadable for one line logging, \n and spaces looks nor realy nice. So I   left cmp.Equal.

MisterMX: > I also tried cmp.Diff and it is too humanreadable for one line logging

Yeah, it is not the ideal thing, but better than nothing. However, `cmp.Diff` helps a lot to debug reconcile bugs so I would prefer it instead of `cmp.Equal`.

teeverr: Done

teeverr: @MisterMX is it okay, could you check?

MisterMX:
LGTM. @teeverr can you rebase it on the latest `master` and fix the conflicts?

teeverr:
> LGTM. @teeverr can you rebase it on the latest `master`, fix the conflicts and squash your changes?

@MisterMX Done, I rebased it on the lastest master and squashed my commits.

teeverr:
@MisterMX a final build, which included changes from the master, didn't pass tests.
Only after reverting of this [commit](https://github.com/swisscom/provider-aws/commit/3693b47a1b4580bc7af38541447243f3b381aa67) everything where stabilized

Context: 
   - ProvisionedProduct has mutual exclusive params: productID and productName, provisioningArtifactID and provisioningArtifactName
   - ProvisionedProduct uses ack-generated create/delete/observe/update funcs

Expected behaviour: ProvisionedProduct custom resource has absolutely the same set of `spec.forProvider` parameters that set in the manifest which was applied, plus lately initialized params in this case it is only one - `acceptLanguage`
But if apply a manifest:

```
---
apiVersion: servicecatalog.aws.crossplane.io/v1alpha1
kind: ProvisionedProduct
metadata:
  name: awesome-s3-bucket
spec:
  forProvider:
    region: us-east-1
    productName: s3
    provisioningArtifactName: v1.0.0
    provisioningParameters:
      - key: BucketName
        value: "awesome-s3-bucket"
      - key: BucketVersioning
        value: "enabled"
      - key: TransferAcceleration
        value: "suspended"
      - key: EncryptionType
        value: "SSE-KMS"
      - key: KMSArn
        value: "none"
  providerConfigRef:
    name: provider-aws
```

The custom resource will look liks this:

```
apiVersion: servicecatalog.aws.crossplane.io/v1alpha1
kind: ProvisionedProduct
...
spec:
  forProvider:
    acceptLanguage: en # it is okay, added by late init 
    region: us-east-1
    productID: prod-bz6oapwxnu67u # it is not okay, this param not specified in the manifest
    productName: s3
    provisioningArtifactID: pa-7lzz6kt44qevo # it is not okay, this param not specified in the manifest
    provisioningArtifactName: v1.0.0
    provisioningParameters:
      - key: BucketName
        value: "awesome-s3-bucket"
      - key: BucketVersioning
        value: "enabled"
      - key: TransferAcceleration
        value: "suspended"
      - key: EncryptionType
        value: "SSE-KMS"
      - key: KMSArn
        value: "none"
  providerConfigRef:
    name: provider-aws
```

It happens only if the controller creates a new resource, if it takes control over an existing resource(by external name) everything will be ok. And it started after crossplane-runtime had been updated to 1.14.1.
So it looks like Create func of controller enriches current/desired state by all params from resp, as result custom resource has mutual exclusive params and it breaks isUpTodate func.

Probably this crossplane-runtime update also affected other resources in the provider

MisterMX:
@teeverr the effects you see have nothing to do with the crossplane-runtime as it does not modify your spec. The fields are modified in the `Create` method which is generated by ACK: https://github.com/crossplane-contrib/provider-aws/pull/1908/files#diff-e76b8787ede5bd7470b087cadc20b14a091d05b51bc9d2a02b4bd7a8d9dc7a65R140

teeverr:
@MisterMX I didn't say crossplay-runtime modifies spec, but probably it is indirectly affects `Create` method which is uses this package. I didn't research this problem deeply, but I found that the same zz_controller.go(which inlcludes Create method) has different behaviour with packages from [commit](https://github.com/swisscom/provider-aws/commit/3693b47a1b4580bc7af38541447243f3b381aa67) and without. Which is not okay, I assume.
Probably another package from the commit above affects custom resource specs. Maybe github.com/crossplane/crossplane-tools?

MisterMX:
https://github.com/crossplane/crossplane-runtime only implements the underlying, fundemental controller logic. The way managed resources are reconciled hasn't changed afaik.

https://github.com/crossplane/crossplane-tools is a code generator (angryjet) that generates resolver functions that are necessary to satisfy the `resource.Managed` interface. But it does not cover external API specific fields that inside `spec.forProvider`.

As I said, the fields you mentioned are set in the ACK generated `Create` in `zz_controller.go` (https://github.com/crossplane-contrib/provider-aws/pull/1908/files#diff-e76b8787ede5bd7470b087cadc20b14a091d05b51bc9d2a02b4bd7a8d9dc7a65R156). This also explains why the behaviour only occurs for new resources and not during imports of existing once (fields are not overwritten in `Update`).

To solve this I guess it should suffice to ignore the respective fields in the `generator-config` from AWS response. Something like:

```yaml
ignore:
  field_paths:
    - ProvisionedProductOutput.RecordDetail.ProvisioningArtifactId
```

teeverr:
@MisterMX  yes, `ignore.filed_paths` helped, custom resource doesn't have these fields anymore if they're not explicitly set. Thanks!
But in the same time I had to delete
```
  fields:
      LastPathID:
        is_read_only: true
        from:
          operation: DescribeRecord
          path: RecordDetail.PathId
```
from generator-config, because it used the same type, it's not critical but nice to know what pathId your resource used. Looks like there is only one option to return it back - write custom `Create` method

It's really weird, because `Create` had if statements like this before the [commit](https://github.com/swisscom/provider-aws/commit/3693b47a1b4580bc7af38541447243f3b381aa67)
```
       if resp.RecordDetail.PathId != nil {
               cr.Spec.ForProvider.LastPathID = resp.RecordDetail.PathId
       } else {
               cr.Spec.ForProvider.LastPathID = nil
       }
```
but fields in spec.forProvider wheren't filled implicitly. So as I already texted, probably it affected, something else.

Anyway, it looks like service catalog is good for now. Thanks for your help : ) I will squash a commit history tomorrow, when tests be successfully passed

teeverr:
@MisterMX 
Done, commits are squashed

MisterMX:
@teeverr there is some diff generated because of #1924 and #1943. Can you rebase on the latest master?

teeverr:
@MisterMX done!

MisterMX (approved):
LGTM. Thank you very much for all the hard work @teeverr!

PR #1888, opened 2023-10-06
chore: Bump submodule

MisterMX:
### Description of your changes

Bump golangci-lint to v1.54.0

Superseedes #1859

I have:

- [ ] Read and followed Crossplane's [contribution process].
- [ ] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

[contribution process]: https://git.io/fj2m9

PR #1910, opened 2023-10-13
chore(ci): Disable codeql step

MisterMX:

Removes the `codeql` check that was introduced with #1888 because it takes a very long time to run and we are not using the results anywhere.

I have:

n.a.


chlunde:
I think we could run it scheduled instead, or as part of the (pre)release-process

MisterMX:
> I think we could run it scheduled instead, or as part of the (pre)release-process

This would be an option, but I am not sure, what we would currently do with the results... 😆

PR #1917, opened 2023-10-17
refactor: Move methods in `pkg/clients/` into invidual util packages

MisterMX:

This moves the individual `pkg/clients/*.go` files into their own `pkg/utils/` packages to provide better modularization. It also makes imports a bit cleaner as the `pkg/clients` was imported in the code with many different names (`awsclients`, `awsclient`, `clients`, `aws` etc.).

There are no functional changes to the code. Only the imports have changed. In a very small amount of places the pointer function has been replaced with the respective one from `k8s.io/utils/ptr`.

I have:

n.a.


chlunde:
I think we should ensure that the replaced functions have identical behaviour, it will be very hard to test if any of these changes have negative side effects?

```
        controller_test.go:239: r: -want, +got:
              &v1beta1.Topic{
              	TypeMeta:   {},
              	ObjectMeta: {Annotations: {"crossplane.io/external-name": "arn:aws:sns:ap-south-1:862356124505:some-topic"}},
              	Spec: v1beta1.TopicSpec{
              		ResourceSpec: {},
              		ForProvider: v1beta1.TopicParameters{
              			Region:         "",
              			Name:           "some-topic",
              			DisplayName:    &"some-topic-01",
            - 			KMSMasterKeyID: &"",
            + 			KMSMasterKeyID: nil,
            - 			Policy:         &"",
            + 			Policy:         nil,
            - 			DeliveryPolicy: &"",
            + 			DeliveryPolicy: nil,
              			FifoTopic:      nil,
              			Tags:           nil,
              			```

chlunde (changes requested):
Enssure we have identical behaviour for empty string vs nil

MisterMX:
Sorry, I think I messed something up while search-replacing the pointer methods. It should now work without any behavioral changes.

chlunde (approved):

PR #1885, opened 2023-10-05
feat(firehose): Add DevileryStream controller

nilde:
### Description of your changes

Fixes #1314 

I have:

- [x] Read and followed Crossplane's [contribution process].
- [x] Run `make reviewable test` to ensure this PR is ready for review.

### How has this code been tested

DeliveryStream created and deleted successfully both locally and in a real AWS environment.

[contribution process]: https://git.io/fj2m9

PR #1919, opened 2023-10-17
(fix): Firehose external name 

nilde:

Fixes #1890  but for firehose.

I have:

Tested locally


MisterMX:
For the record: This actually contains the implementation for Firehose that was supposed to bemerged with #1885 but for due to some mistake was not merged (only minor refactorings got merged appearently).

@nilde could you squash all changes into a single commit and sign them? Please also make sure that the commit message follows the [conventional commits standard](https://www.conventionalcommits.org/en/v1.0.0/). Since you are actually adding a feature here it might make sense to change it into something like `feat(firehose): Add DevileryStream controller`. I think you also need to rebase your changes on the latest master, as some refactoring has happened in the mean time.

apis/kafka/generator-config.yaml
```diff
@@ -18,9 +18,3 @@ resources:
       errors:
         404:
           code: NotFoundException
-    fields:
-      ClusterPolicyVersion:
-        is_read_only: true
-        from:
-          operation: GetClusterPolicy
-          path: CurrentVersion
```
MisterMX, line 26: Why did you removed these lines?

nilde: Sorry about that, some code left when I made the rebase. Should be solved on my last commit. Thanks for reviewing.

MisterMX (approved):
LGTM. Thank you very much @nilde!

PR #1923, opened 2023-10-19
refactor(pointer): Use generic pointer functions

MisterMX:

Some more refactoring for pointer functions:

* Add generic functions for late initialize and pointer refs and derefs.
* Replace calls that are using the `pointer.FieldRequired` option with ptr.To which as the same effect but is much simpler.
* Remove pointer functions that are not used anymore.
* Change pointer functions so they don't depend on the AWS sdk anymore but only `k8s.io/utils/ptr`.

No behavioral changes to the code have been made.

I have:

n.a.


pkg/utils/pointer/pointer.go
```diff
@@ -17,327 +17,75 @@ limitations under the License.
...
+	return &val
 }
 
-// IntFrom32Address converts the supplied int32 pointer to an int pointer, returning nil if
-// the pointer is nil.
-func IntFrom32Address(i *int32) *int {
-	if i == nil {
+// ToIfNotDefault returns a pointer to val if it does NOT match the default
+// value of T. Otherwise it returns nil.
+func ToIfNotDefault[T comparable](val T) *T {
```
chlunde, line 85: I find it a bit hard to remember what exactly this does. Not sure about a better name.

I think "zero value", not default value, is the terminology used in the go documentation, for example: https://tip.golang.org/doc/effective_go


Not sure if `pointer.NilIfZeroValue()` is better :)

chlunde (approved):

pkg/utils/pointer/pointer.go
MisterMX: How about `pointer.ToOrNilIfZeroValue()`?

MisterMX: Done

PR #1924, opened 2023-10-20
feat(ack): Do not call isUpToDate if resource is deleted

MisterMX:

Modifies the ACK controller template to only call `isUpToDate` if not `meta.WasDeleted(cr)`. This may safe some time and unnecessary AWS calls.

Only affects controllers that are generated with ACK.

I have:

n.a.


chlunde:
have you done any manual testing? I think this looks sane, but it is a type of change that could have unintended consequences

chlunde (approved):

MisterMX:
> have you done any manual testing? I think this looks sane, but it is a type of change that could have unintended consequences

I only tested this once with `rds.DBInstance` to check whether the implementation works fine. I did not test the other resources. However, I looked through all `isUpToDate` functions to see if there is something umcommon (like anything besides read operations) but that does not seem to be the case.

PR #1933, opened 2023-10-30
feat(certificate): add key algorithm

wotolom:

for resource `acm.Certficate`

- update github.com/aws/aws-sdk-go-v2/service/acm to v1.16.0
- add optional, immutable field `KeyAlgorithm`

This allows users to set a different encryption algorithm than RSA_2048 for ACM certificates on creation.

I have:

manually


apis/acm/v1beta1/certificate_types.go
```diff
@@ -124,6 +125,24 @@ type CertificateParameters struct {
...
+	// Check the requirements for the AWS service where you plan to deploy your
+	// certificate.
+	// Default: RSA_2048
+	// Note: ACM can request the issue of new certificates using the following algorithms:
+	// "RSA_2048", "EC_prime256v1" and "EC_secp384r1".
+	// The remaining algorithms are supported only for imported certificates.
+	// See also AWS docs: https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html#algorithms.title
+	// +optional
+	// +immutable
+	KeyAlgorithm types.KeyAlgorithm `json:"keyAlgorithm,omitempty"`
```
MisterMX, line 144: I think it would be better to have a dedicated type `KeyAlgorithm` in `apis` to not be dependend on the AWS SDK.

I also think it should be a pointer because it is optional?

MisterMX (changes requested):

apis/acm/v1beta1/certificate_types.go
wotolom: Type: I was inspired by `CertificateAuthority` https://github.com/crossplane-contrib/provider-aws/blob/1712285bd6460d056c3157a01344328d6d0bb648/apis/acmpca/v1beta1/certificateauthority_types.go#L90

This way we do not have to keep a dedicated type uptodate over sdk-upgrades?

Pointer: hmm probably got confused by inconsistency :https://github.com/crossplane-contrib/provider-aws/blob/1712285bd6460d056c3157a01344328d6d0bb648/apis/acm/v1beta1/certificate_types.go#L166

MisterMX: > This way we do not have to keep a dedicated type uptodate over sdk-upgrades?

Unfortunately we do, because it is only an alias for string. Kubebuilder does not detect enums that way. Also it is not good design to have the API types depend on what is used in the implementation. They should be separated.

wotolom: Ok. Thanks for the explanations.
Updated the first commit `feat(certificate): add key algorithm` with the suggestions.

package/crds/acm.aws.crossplane.io_certificates.yaml
```diff
@@ -678,6 +678,31 @@ spec:
...
+                      are supported only for imported certificates. See also AWS docs:
+                      https://docs.aws.amazon.com/acm/latest/userguide/acm-certificate.html#algorithms.title'
+                    enum:
+                    - RSA_1024
+                    - RSA_2048
+                    - RSA_3072
+                    - RSA_4096
+                    - EC_prime256v1
+                    - EC_secp384r1
+                    - EC_secp521r1
```
MisterMX, line 704: Just a thought, but do we want to create an enum here?
If AWS ever adds new values, we will need to adjust this here. 

Maybe it makes more sense to just add it to the description?

wotolom: Removed enums. Should be enough information n the description.
Also there will be an AWS Error if the string in `KeyAlgorithm` is not valid

MisterMX (approved):

Issue #1934, opened 2023-11-02
Pod restart due to "fatal error: concurrent map iteration and map write"


Technovation22:
### What happened?
Recently, we observed pod restarts due to a concurrent map iteration and writes to the same map in some of our production environments,
```
2023-10-16T03:29:57Z    2023-10-16T03:29:57.143Z    DEBUG    events    cannot get cache cluster list: api error Throttling: Rate exceeded    {"type": "Warning", "object": {"kind":"ReplicationGroup","name":"r-acf54be9-dc8c-445d-a712-1507bfd0660b-krb4x-gg48h","uid":"a0489de6-3fb1-4690-b248-12c482a3c175","apiVersion":"cache.aws.crossplane.io/v1beta1","resourceVersion":"647156184"}, "reason": "CannotObserveExternalResource"}
2023-10-16T03:29:57Z    2023-10-16T03:29:57.151Z    DEBUG    provider-aws    Reconciling    {"controller": "managed/replicationgroup.cache.aws.crossplane.io", "request": {"name":"r-acf54be9-dc8c-445d-a712-1507bfd0660b-krb4x-gg48h"}}
2023-10-16T03:29:57Z    fatal error: concurrent map iteration and map write
```
after investigating this issue, it looks like mutation on the cached object in `crossplane-runtime` caused this issue (based on the result we see in the race detector). This bug is already fixed in `crossplane-runtime` in https://github.com/crossplane/crossplane-runtime/pull/561 and it is expected from V1.14.0 we don't see this issue again but I see the latest `crossplane-runtime` upgrade in `provider-aws` is v1.13.0. So it looks like `provider-aws` is using the old version of `crossplane-runtime` and we can not get the benefit of this bug fix.

Since we are facing this in the production environment, is there any chance we get the new version upgraded?
### How can we reproduce it?
running race detector shows the result. going to attach the result we got.

### What environment did it happen in?
Crossplane version: 1.13.2
Cloud provider: AWS
Provider-aws: v0.43.0
Kubernetes version v1.25.10
Os: Unavailable at this time
Kernel: Unavailable at this time
[race.txt](https://github.com/crossplane-contrib/provider-aws/files/13243772/race.txt)

PR #1935, opened 2023-11-03
feat!: Update crossplane-runtime to 1.14.1

MisterMX:

* Update controller-runtime and other libs as well.
* Remove deprecated property `spec.providerRef` from all CRDs

Fixes: #1934

I have:

n.a.


chlunde (approved):

chlunde:
> How has this code been tested
> n.a.

I wonder if we could have a basic e2e/integration test that installs the previous version in a `kind` cluster, add some resources connected to local stack, upgrade and do some basic changes

MisterMX:
> I wonder if we could have a basic e2e/integration test that installs the previous version in a `kind` cluster, add some resources connected to local stack, upgrade and do some basic changes

I know what you mean. We are thinking about a solution for this issue for quite a while already. Thing is that local stack only provides a subset of AWS resources and some of them only in the premium tier.

We have this integration test kinda thing in `cluster/local/integration_tests.sh` but it only tries to install the provider in a kind cluster and does not test the individual controllers.

There is a proposal in https://github.com/crossplane-contrib/provider-argocd/pull/89 that uses 
https://github.com/maximilianbraun/xp-testing and https://github.com/kubernetes-sigs/e2e-framework/ to run E2E tests. However, it is much simpler there because you only need an ArgoCD instance instead of a whole AWS environment.

chlunde:
@MisterMX I was just thinking about testing a couple of services to get coverage for crossplane-runtime etc. Probably at least one service "hand written" and one generated, not full e2e coverage.

Issue #1930, opened 2023-10-24
ACM Certificate - Tags flapping


jlpedrosa:
### What happened?
Resource acm.Certificate is constantly reconciling, Tags are being set and removed.

goes from tags to without tags constantly:
```
 "spec": {
      
        "forProvider": {
          ...
            "region": "us-east-1",            
            "tags": [],
            "validationMethod": "EMAIL"
        },
        "providerConfigRef": {
            "name": "aws"
        }
    },

```

to:
```
 "forProvider": {
       ...
            "tags": [
                {
                    "key": "crossplane-kind",
                    "value": "certificate.acm.aws.crossplane.io"
                },
                {
                    "key": "crossplane-name",
                    "value": "private-ingress"
                },
                {
                    "key": "crossplane-providerconfig",
                    "value": "aws"
                }
            ],
            "validationMethod": "EMAIL"
        },     

```

### How can we reproduce it?
Creating a Private CA, and linking it to an object similar to:
```
 &acmv1alpha.Certificate{
		ObjectMeta: metav1.ObjectMeta{
			// TODO:
			Name: a.Component.GetName(),
		},
		Spec: acmv1alpha.CertificateSpec{
			ResourceSpec: xpv1.ResourceSpec{
				ProviderConfigReference: &xpv1.Reference{Name: awsEnv.Spec.XPProviderConfigName},
			},
			ForProvider: acmv1alpha.CertificateParameters{
				Region:                  awsEnv.Spec.Region,
				CertificateAuthorityARN: &CAArn,
				DomainName:              certificateCommonName,
				DomainValidationOptions: []*acmv1alpha.DomainValidationOption{
					{
						DomainName:       certificateCommonName,
						ValidationDomain: certificateCommonName,
					},
				},
				Options: &acmv1alpha.CertificateOptions{
					CertificateTransparencyLoggingPreference: "ENABLED",
				},
				SubjectAlternativeNames: []*string{&certificateCommonName},
				Tags:                    []acmv1alpha.Tag{
					//{Key: "cluster", Value: string(cluster.GetClusterID())},
					//{Key: "component", Value: "ingress-controller"},
					//{Key: "ingress", Value: a.Component.GetName()},
				},
				ValidationMethod: "EMAIL",
			},
		},
	}
```

### What environment did it happen in?
0.38, but there are not significant changes in ACM since then.

MisterMX:
[...]

Issue #985, opened 2021-12-03
ReplicationGroup stuck in modifying loop


JaimeHolland:
### What happened?

When I create a ReplicationGroup, it is never marked as `Ready` - it seems to get in a loop of trying to modify something on the AWS side.  Below is the output of a `kubectl describe` on a test ReplicationGroup - I don't know if the `Pending Modified Values` section is hinting at the cause?:

<details>
  <summary>output</summary>
  
  ```txt
Name:         test-replication-group-qwdll
Namespace:
Labels:       <none>
Annotations:  crossplane.io/external-create-pending: 2021-12-03T09:51:27Z
              crossplane.io/external-create-succeeded: 2021-12-03T09:51:28Z
              crossplane.io/external-name: test-replication-group-qwdll
API Version:  cache.aws.crossplane.io/v1beta1
Kind:         ReplicationGroup
Metadata:
  Creation Timestamp:  2021-12-03T09:51:27Z
  Finalizers:
    finalizer.managedresource.crossplane.io
  Generate Name:  test-replication-group-
  Generation:     3
  Managed Fields:
    API Version:  cache.aws.crossplane.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:generateName:
      f:spec:
        .:
        f:deletionPolicy:
        f:forProvider:
          .:
          f:applyModificationsImmediately:
          f:atRestEncryptionEnabled:
          f:authEnabled:
          f:automaticFailoverEnabled:
          f:cacheNodeType:
          f:cacheParameterGroupName:
          f:cacheSubnetGroupName:
          f:engine:
          f:engineVersion:
          f:numNodeGroups:
          f:region:
          f:replicasPerNodeGroup:
          f:replicationGroupDescription:
          f:securityGroupIds:
          f:transitEncryptionEnabled:
        f:providerConfigRef:
          .:
          f:name:
    Manager:      kubectl-create
    Operation:    Update
    Time:         2021-12-03T09:51:27Z
    API Version:  cache.aws.crossplane.io/v1beta1
    Fields Type:  FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .:
          f:crossplane.io/external-create-pending:
          f:crossplane.io/external-create-succeeded:
          f:crossplane.io/external-name:
        f:finalizers:
          .:
          v:"finalizer.managedresource.crossplane.io":
      f:spec:
        f:forProvider:
          f:preferredMaintenanceWindow:
[...]

PR #1938, opened 2023-11-07
feat!: Remove Crossplane standard tags

### LINKED ISSUE #1498: Problem when setting explicit tags for some AWS resources
Hey guys, I have recently got into problem or atleast weird functioning of some resources in AWS provider, I will try to explain

### What happened?
When I deploy some AWS resouces (for example: EC2 Address, EC2 Subnet, EC2 VPC), crossplane automatically adds by itself some tags, and that's a problem:

 Why?
 - Because I deploy this resources through argoCD (I have stored the yamls in the Git, and ArgoCD is automatically syncing it and applying, but it will always result in OutOfSync mode, because the desired state will always differ from the deployed yaml (It is explained in How to reproduce below)

PS: This is not happening for other resources like EC2 NatGateway or EC2 Subnet, because on this resources crossplane is not setting another tags by itself

### How can we reproduce it?
You can reproduce it pretty easily

1. Deploy EC2 VPC with some explicit tags set
```
forProvider:
 ...
 tags:
  - key: Name
    value: production-vpc
```
2. In the deployed yaml in k8s you will also see
```
forProvider:
 ...
 tags:
  - key: Name
    value: production-vpc
  - key: crossplane-kind
    value: production-vpc
    value: default
```

- The problem is that ArgoCD wants to have there only single tag, which is defined by me, but this never happens because crossplane will always add there the other tags...
- This is the difference which ArgoCD is always reporting and trying to fix, (it is basically in infinite loop)
image

Crossplane version: 1.9.0 deployed by original Helm Chart
CloudProvider: self hosted on AWS
Kubernetes version: 1.21

MisterMX:

This is a proposal to remove the `tagger` initializer that always adds the "standard" Crossplane tags to every managed resource:
* `crossplane-kind`
* `crossplane-name`
* `crossplane-providerconfig`

There have been numerous issues with tags that were added by default, like #1930, #1419, #985 and many more.

The behaviour of setting the tags is also not standardized: Some controllers always overwrite existing tags others are only setting it if the value does not exist. Arrays of tags don't play very nice with compositions as well.

Since there is no technical requirement for this feature, one way to overcome the issues with it, is to remove it completely.

Fixes #1498

I have:


chlunde:
FYI; we use theese tags to know which resources are managed by crossplane, which are orphan etc.
We should get some feedback before removing these as this is a breaking change.
It might need to have a conditional flag if many are using this.

I think we could have a composition function which adds (similar) tags instead.

I intended to write a function that just sorted and added back the default tags to avoid extra reconciles when the order changes or default tags are lost.

One reason why we have some of these issues is that tags are not defined as maps. Compositions work much better with patching those than the keyed arrays we have now.

MisterMX:
A few more thoughts on this:

Existing resources don't get their tags removed so it is not technically breaking. However, if some external scripts (like a resource cleanup) depend on the tags being present it requires some changes on the platform operator site.

If resources are deployed using compositions, the above mentioned tags can be added to the base manifest to have them stick around. I don't know about functions but it should be easy as well to just add the tags there, too.

I am not really a fan to hide this behavior behind a feature flag, since it only makes the code more complicated and adds another edge case to test. Also, there is still no standard way of adding default tags a resource.

tkaesserfm:
Even though it feels harsh to me to remove the tags completely, looking at the code i think it might be the best trade-off.
Adding a feature flag would make sense if there would be some centralized way of handling this. But the fact that this isn't possible in a single point just highlights what @MisterMX already said, that there is no standard which naturally creates a mess here.
In an ideal world i would love to see this handled globally and without much need to interfere from the user. But to make this happen there would need to exist a standard to begin with and a unified implementation. (in the best case across crossplane providers)
However as of today i think even if it hurts it makes sense to offload the management of these tags to the user as the issues with constant reconciliation are real and can quickly mess especially with large scale setups.

chlunde:
<details>
<summary>longer rant + code example using reflection</summary>
I don't think we could have a unified implementation with generics (which I was really hoping for), but using reflection is still an option:

```go
package main

import (
	"reflect"

	"github.com/davecgh/go-spew/spew"
)

func main() {
	current := []Tag{
		Tag{Name: "k1", Value: "v1"},
		Tag{Name: "k2", Value: "v2"},
	}
	result = updateOrAddTags(current, map[string]string{
		"k2": "new",
		"k3": "v3",
	})
	spew.Dump(current)
}

type Tag struct {
	Name  string
	Value string
}

func newTag(k, v string, typ reflect.Type) any {
	tag := reflect.Indirect(reflect.New(typ))

	tag.FieldByName("Name").SetString(k)
	tag.FieldByName("Value").SetString(v)

	return tag.Interface()
}

func updateOrAddTags[T any](current []T, add map[string]string) []T {
	updated := make(map[string]bool)
	items := reflect.ValueOf(current)
	if items.Kind() == reflect.Slice {
		for i := 0; i < items.Len(); i++ {
			item := items.Index(i)
			k := item.FieldByName("Name").String()
			// v := item.FieldByName("Value").String()
			if v2, ok := add[k]; ok {
				updated[k] = true

				item.FieldByName("Value").SetString(v2)
			}
		}
	}

	for k, v := range add {
		if !updated[k] {
			items = reflect.Append(items, reflect.ValueOf(newTag(k, v, items.Type().Elem())))
		}
	}

	return items.Interface().([]T)
}
```

But it still leaves the flapping issue. I think that's a problem which would require fixes in crossplane (or requiring using composition functions), or changing to map[string]string which would be breaking.

If we would want a user toggle, we could wrap managed.GetExternalTags I think
</details>

So yeah, in the end I guess it would be easier to suggest that everyone manually tag all resources they want tagged.

chlunde (approved):

chlunde:
I think it's important to flag this in the release notes

PR #1939, opened 2023-11-08
fix: tags change if added or removed

akesser:

In isUpToDate for rds dbinstance it is checked if tags need to be added or removed. Currently the property "tagsChanged", that is amongst other checks used to determine if the dbinstance has changed, is only set to true if both tags get added AND removed. I changed this to set the value to true if tags need to be added OR removed.

Fixes #

I have:

Tested locally


akesser:
Thanks form merging so quickly!

PR #1942, opened 2023-11-09
fix: Readd empty initializers after runtime update

MisterMX:

The default behaviour of a controller is to use
`managed.NewNameWithExternalName`. Controllers need to pass `managed.WithInitializers()` to change the behaviour.

After the runtime upgrade and tagging removal, some of the empty `managed.WithInitializers()` got removed resulting in a different behaviour.

Readd them restore the previous behaviour.

Partially reverts #1935.

I have:

No, testing was performed. Just comparison to the previous version.


chlunde:
I think this CI will be green if rebased?

chlunde (approved):

PR #1966, opened 2023-12-04
fix(database): tag-only-change triggers update + add RemoveTag call

### LINKED ISSUE #1962: database.RDSInstance does not update tags
### What happened?

`database.RDSInstance` seems to not trigger an update, when only the `spec.forProvider.tags` parameter is updated. (ignored in patch diff and no extra logic)
Only through changing another field, we can trigger an update, which then also changes a tag in AWS.

https://github.com/crossplane-contrib/provider-aws/blob/4124fb7ccb72938dcfe5ada372d910f81884a270/pkg/clients/database/rds.go#L722

Also the usage of  `RemoveTagsFromResource()`  seems to be missing.  Only `AddTagsToResource()` is called in `Update()`

https://github.com/crossplane-contrib/provider-aws/blob/4124fb7ccb72938dcfe5ada372d910f81884a270/pkg/controller/database/rdsinstance/rdsinstance.go#L281

I will try to provide a PR.

### How can we reproduce it?

Create `RDSInstance` and only change tags value afterwards. 

Crossplane version: 1.14.3
provider-aws: v0.45.2 + master

wotolom:

for `database.RDSInstance`:

- fix `isUpToDate()` logic to trigger update on tag-only-change (used cache scheme/solution from `rds.DBInstance`)
- add `RemoveTagsFromResource()` from AWS API to `update()` (removing tags was apparently not possible before...)
- adapt unit-tests

Fixes #1962

I have:

manually (update tag-only, update other db parameter)

### DIFF
--- a/pkg/clients/database/rds.go
+++ b/pkg/clients/database/rds.go
@@ -49,11 +49,6 @@ const (
 	errGetPasswordSecretFailed = "cannot get password secret"
 )
 
-type Cache struct {
-	AddTags    []rdstypes.Tag
-	RemoveTags []string
-}
-
 // Client defines RDS RDSClient operations
 type Client interface {
 	CreateDBInstance(context.Context, *rds.CreateDBInstanceInput, ...func(*rds.Options)) (*rds.CreateDBInstanceOutput, error)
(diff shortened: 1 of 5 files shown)


wotolom:
I do not really understand, why `e.cache` behaves here differently then `err`:
```
upToDate, diff, e.cache, err := rds.IsUpToDate(ctx, e.kube, cr, instance)
```
is not possible:
> non-name e.cache on left side of :=compiler[(BadDecl)](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#BadDecl)

therefore I did this:
https://github.com/crossplane-contrib/provider-aws/blob/f7297fa0d479708d3d6b5e41c11d4d721b8d5072/pkg/controller/database/rdsinstance/rdsinstance.go#L158-L162

pkg/clients/database/rds.go
```diff
@@ -891,3 +891,42 @@ func areSameElements(a1, a2 []string) bool {
...
+
+	specMap := make(map[string]string, len(spec))
+	for _, t := range spec {
+		key := t.Key
+		val := t.Value
+		specMap[key] = t.Value
+
+		if currentVal, exists := currentMap[key]; exists {
+			if currentVal != val {
+				removeTags = append(removeTags, t.Key)
```
MisterMX, line 910: I don't think this is necessary, since `AddTagsToResource` will overwrite the value of the existing tags?

pkg/clients/database/rds.go
```diff
@@ -49,6 +49,11 @@ const (
 	errGetPasswordSecretFailed = "cannot get password secret"
 )
 
+type Cache struct {
+	AddTags    []rdstypes.Tag
+	RemoveTags []string
+}
+
```
MisterMX, line 56: The `Cache` struct is controller specific and should be in `/pkg/controller`.

MisterMX (changes requested):

pkg/clients/database/rds.go
wotolom: Yes you're right. For the case "same key, changed value" `AddTagsToResource` is enough, no `RemoveTagsFromResource` needed. I will remove the line here.

For the resources in the `rds` APIGroup this could/should be checked too in the future:
https://github.com/crossplane-contrib/provider-aws/blob/4124fb7ccb72938dcfe5ada372d910f81884a270/pkg/controller/rds/utils/tags.go#L107

pkg/clients/database/rds.go
wotolom: I put the `Cache` struct into `pkg/clients/rds.go` so it could be used in `IsUpToDate()`. Due to the controller and client being in different packages and the controller needing to import the client ("import cycle not allowed"), this was the only way I found for this.

But I will try to find another solution.

wotolom: Ok, simply moved the `Cache` struct now into the controller.

Maybe the "cache" solution works/writes best in ACK controller setups/resources.
A non-cache solution, I found in other "hand-written" controller setups, is to call `DiffTag` again when in `Update()`:
https://github.com/crossplane-contrib/provider-aws/blob/4124fb7ccb72938dcfe5ada372d910f81884a270/pkg/controller/acm/controller.go#L194

MisterMX (approved):

PR #1979, opened 2023-12-15
fix(ack): controllers only skip isUpToDate on deletion

### LINKED ISSUE #1978: Secret MR does not get deleted directly anymore
### What happened?

Due to this change to the ACK generated controllers https://github.com/crossplane-contrib/provider-aws/pull/1924,
MR `Secrets` stay in cluster after deletion, until `recoveryWindowInDays` (default 30 days) is passed. Also during this time the controller seems to keep sending deletion-calls.

Previously this was handled by this code block in `postObserve()` by looking if the secret is delete in AWS "frontend":
https://github.com/crossplane-contrib/provider-aws/blob/6e24fa15610961378e22aeddd0dbd148909e0c39/pkg/controller/secretsmanager/secret/setup.go#L125-L130

But now `postObserve()` is skipped through this change in the generated controller:
https://github.com/crossplane-contrib/provider-aws/blob/6e24fa15610961378e22aeddd0dbd148909e0c39/pkg/controller/secretsmanager/secret/zz_controller.go#L92-L97

### How can we reproduce it?
```yaml
apiVersion: secretsmanager.aws.crossplane.io/v1beta1
kind: Secret
metadata:
  name: example-secret
spec:
  forProvider:
    region: eu-central-1
    # forceDeleteWithoutRecovery: true # default is false
    # recoveryWindowInDays: 7 # AWS default is apparently 30 (https://doc.crds.dev/github.com/crossplane-contrib/provider-aws/secretsmanager.aws.crossplane.io/Secret/v1beta1@v0.45.2)
    stringSecretRef:
      key: password
      name: example-secret-manager
      namespace: crossplane-system
    tags:
      - key: secret
        value: "secret"
  providerConfigRef:
    name: example
---
apiVersion: v1
kind: Secret
metadata:
  name: example-secret-manager
  namespace: crossplane-system
type: Opaque
data:
  password: dGVzdFBhc3N3b3JkITEyMw== # testPassword!123
```

Crossplane version: v1.14.4
provider-aws v0.45.2 + master

wotolom:

- adapts the ACK controller template to only skip `isUpToDate()` on resource deletion (and not also `postObserve()`)
- for `docdb.Cluster` 
  - adapt unit-test
  - in `postObserve()` add `if meta.WasDeleted(cr) {...}` around secretRef accessing to avoid MR deletion stuck on missing secret
- also cleaned-up the no longer needed `if meta.WasDeleted(cr) {return}` in some `isUpToDate()` 

Fixes #1978

I have:

unit-tests, manually (`secretsmanager.Secret` + `docdb.Cluster`)

### DIFF
--- a/pkg/controller/apigateway/apikey/zz_controller.go
+++ b/pkg/controller/apigateway/apikey/zz_controller.go
@@ -89,15 +89,13 @@ func (e *external) Observe(ctx context.Context, mg cpresource.Managed) (managed.
 		return managed.ExternalObservation{}, errors.Wrap(err, "late-init failed")
 	}
 	GenerateAPIKey(resp).Status.AtProvider.DeepCopyInto(&cr.Status.AtProvider)
-	if meta.WasDeleted(cr) { // There is no need to run isUpToDate if the resource is deleted
-		return managed.ExternalObservation{
-			ResourceExists:   true,
-			ResourceUpToDate: true,
-		}, nil
-	}
-	upToDate, diff, err := e.isUpToDate(ctx, cr, resp)
-	if err != nil {
-		return managed.ExternalObservation{}, errors.Wrap(err, "isUpToDate check failed")
+	upToDate := true
+	diff := ""
+	if !meta.WasDeleted(cr) { // There is no need to run isUpToDate if the resource is deleted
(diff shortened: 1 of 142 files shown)


MisterMX (approved):
LGTM. Thank you very much @wotolom for pointing out this issue and for providing the fix!

PR #2001, opened 2024-02-15
fix(database): skip isUpToDate on deletion

wotolom:

for `database.rdsinstance`:
- implement the same logic to skip the isUpToDate()-check on deletion as https://github.com/crossplane-contrib/provider-aws/pull/1979 did for the ACK controllers

This fixes e.g. a common case of a missing k8s secret blocking the deletion, if it was deleted from the cluster before the MR.

I have:

manually

### DIFF
--- a/pkg/controller/database/rdsinstance/rdsinstance.go
+++ b/pkg/controller/database/rdsinstance/rdsinstance.go
@@ -161,11 +161,16 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		cr.Status.SetConditions(xpv1.Unavailable())
 	}
 
+	if meta.WasDeleted(cr) { // There is no need to run isUpToDate if the resource is deleted
+		return managed.ExternalObservation{
+			ResourceExists: true,
+		}, nil
+	}
+
 	var upToDate bool
 	var diff string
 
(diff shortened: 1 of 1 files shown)


pkg/controller/database/rdsinstance/rdsinstance.go
```diff
@@ -161,13 +161,16 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 		cr.Status.SetConditions(xpv1.Unavailable())
 	}
 
-	var upToDate bool
-	var diff string
+	upToDate := true
+	diff := ""
```
MisterMX, line 165: Easier to read would be

```suggestion
	if meta.WasDeleted(cr) {
		return managed.ExternalObservation{}, nil
	}

	var upToDate bool
	var diff string
```

MisterMX (changes requested):

pkg/controller/database/rdsinstance/rdsinstance.go
wotolom: good point!

Just added `ResourceExists: true`, else the MR is deleted without sending a deletion-call due to never reach Delete().

MisterMX:
@wotolom can you rebase on the latest master to fix e2e tests?

MisterMX (approved):

PR #2018, opened 2024-03-22
fix(s3 bucket): Ignore AWS system tags

teeverr:
Ignore s3 bucket systemd tags(with prefix `aws:`) on observed resource side because these cannot be updated/deleted, otherwise it cause an error

Example: the controller takes control over existing s3 bucket which has systemd tag(s) and tries to reconcile tag set with desired state which doesn't include tags at all or regular tags only. As result the controller gets an error due it is impossible to delete system tag

I have:

- manual testing
- unit tests


MisterMX (approved):

PR #2023, opened 2024-03-28
fix: take into consideration the DesiredCount/RunningCount for resource readiness

dnlopes:
Currently, ECS Service resource is only taken into consideration the `.Status` field to decide on the readiness of the Crossplane resource.

I think this is an overlook, and Crossplane should also take into consideration the `DesiredCount/RunningCount`, since it is part of the input Spec. Returning _ready_ based only on the `.Status` might result in returning to the client a _ready_ state when in fact, no tasks are running yet.

Before wrapping up the PR with tests, I would like to get some feedback from you to undertstand if you agree with this change or not. 

I have:

Thanks.

### DIFF
--- a/pkg/controller/ecs/service/setup.go
+++ b/pkg/controller/ecs/service/setup.go
@@ -82,7 +82,11 @@ func postObserve(_ context.Context, cr *svcapitypes.Service, resp *svcsdk.Descri
 
 	switch aws.StringValue(resp.Services[0].Status) {
 	case "ACTIVE":
-		cr.SetConditions(xpv1.Available())
+		if resp.Services[0].DesiredCount != resp.Services[0].RunningCount {
+			cr.SetConditions(xpv1.Creating())
+		} else {
+			cr.SetConditions(xpv1.Available())
+		}
 	case "DRAINING":
 		cr.SetConditions(xpv1.Deleting())
 	case "INACTIVE":


pkg/controller/ecs/service/setup.go
```diff
@@ -80,6 +80,11 @@ func postObserve(_ context.Context, cr *svcapitypes.Service, resp *svcsdk.Descri
 		return obs, err
 	}
 
+	if resp.Services[0].DesiredCount != resp.Services[0].RunningCount {
+		obs.ResourceUpToDate = false
+		return obs, nil
+	}
```
MisterMX, line 86: This does not change the `READY` condition for the resource, but instead it triggers a rereconcilation which is not what we want here.

To change the ready state one needs to use `cr.SetConditions()`.

MisterMX (changes requested):

pkg/controller/ecs/service/setup.go
dnlopes: But, don't we want a rereconciliation here? What I'm advovating for is exactly that: we should not consider the resource ready if the `DesiredCount` != `RunningCount`.

I can add a `cr.SetConditions(Unavailable())`, but I followed the logic of the condition just before.

MisterMX: No, a reconcilation should only be scheduled explicitly if the (a) the reconcilation itself failed in case of an error or (b) we need a second run to finish the whole reconcilation process.

In this case the resource is just not ready yet. If we would always trigger a rereconcilation we would constantly spam AWS with request and may end up in request limit that would effect other resources as well. It is safe to return with `ResourceUpToDate = true` and set the condition to `Unavailable` or `Creating` and return later at a time that is chosen by the controller runtime.

dnlopes: Ok I think I got it. Thanks for the explanation. Already updated the PR accordingly.

pkg/controller/ecs/service/setup.go
```diff
@@ -80,6 +80,12 @@ func postObserve(_ context.Context, cr *svcapitypes.Service, resp *svcsdk.Descri
 		return obs, err
 	}
 
+	if resp.Services[0].DesiredCount != resp.Services[0].RunningCount {
+		obs.ResourceUpToDate = true
```
MisterMX, line 84: Is it necessary to set this? What if the tasks are not ready yet because of some misconfiguration? If we hardcode `obs.ResourceUpToDate = true` here users will never be able to update their configuration if there are no tasks running.

I think we should remove setting this here unless it is really necessary.

MisterMX (changes requested):

pkg/controller/ecs/service/setup.go
dnlopes: I don't think it's strictly necessary. I have update the PR accordingly.

pkg/controller/ecs/service/setup.go
```diff
@@ -80,6 +80,11 @@ func postObserve(_ context.Context, cr *svcapitypes.Service, resp *svcsdk.Descri
 		return obs, err
 	}
 
+	if resp.Services[0].DesiredCount != resp.Services[0].RunningCount {
+		cr.SetConditions(xpv1.Creating())
+		return obs, nil
+	}
+
```
MisterMX, line 87: Sorry, for being picky here, but I am not sure if we would want to return here already. What if the state is `INACTIVE` and the number of tasks does not match? We would never run into `obs.ResourceExists = false` in L96 probably preventing deletion of the resource?

I think best would be to put this `if` clause into the `case "ACTIVE"` block below since that is the only time were it is relevant to check if the number of desired tasks is running.

MisterMX (changes requested):

pkg/controller/ecs/service/setup.go
dnlopes: Hello @MisterMX , good catch.

Updated the PR accordingly.

MisterMX (approved):
LGTM. Thank you very much @dnlopes! I took the liberty to squash your commits into one before merging.

PR #2025, opened 2024-03-29
Remove `TagSpecification` from the EC2 instance spec

### LINKED ISSUE #2024: EC2 Instance Tags have two fields in the spec with different semantics
EC2 instance tags can currently be specified via `TagSpecification` and `Tags` ([link](https://github.com/crossplane-contrib/provider-aws/blob/master/apis/ec2/manualv1alpha1/instances_types.go#L273C2-L273C19)).
```
type InstanceParameters struct {
 	// Tags are used as identification helpers between AWS resources.
	// +optional
	Tags []Tag `json:"tags,omitempty"`

	// The tags to apply to the resources during launch. You can only tag instances
	// and volumes on launch. The specified tags are applied to all instances or
	// volumes that are created during launch. To tag a resource after it has been
	// created, see CreateTags (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateTags.html).
	// +immutable
	// +optional
	TagSpecifications []TagSpecification `json:"tagSpecifications,omitempty"`
}
```

However, only `TagSpecification` can be used for tags that must be specified during launch. The EC2 instance controller creates tag specified in `Tags` in a seperate `CreateTags` step after the `RunInstances` call.

Consolidating tags into a single struct `Tags` would simplify the Spec and make it easier to use. All tags specified in the spec should be applied when the instance is created to ensure that any IAM policies that require tagging are respected.

This would be a backwards incompatible change and any users using  `TagSpecification` would have to migrate to `Tags` or their tags will be removed on controller upgrade.

However this does make the instance creation a single API call. Currently, if the `RunInstances` call succeeds but the `CreateTags` call fails - the instance may be orphaned.

justinmir:

Instance parameters allows specifying tags in two methods: `TagSpecification` and `Tags`. These methods have different semantics - `TagSpecification` tags will be applied when the instance is created, `Tags` will be created in a separate `CreateTags` API call afterwards. All tags required to meet IAM policy must be specified in `TagSpecification`.

Fixes #2024 

I have:

This code has been tested on and deployed in our own fork used in production.

This continues to test the RunInstancesInput generation unit test that was updated to match passing in tags via the `Tags` field instead of `TagSpecification`.


MisterMX:
I'm very reluctant about introducing breaking changes to the API. Would it be possible to keep both fields but deprecate one and have both of them being merged together?

chris-bornmann:
If values attached to Tags are applied after creation, will this impact organizations with a set of required tags that must be present on creation?  My org, and I'm certain many others, requires certain tags on instances and volumes and CreateInstance calls are blocked without those tags specified.

justinmir:
Closing for now as I don't have the time to look at this.

> I'm very reluctant about introducing breaking changes to the API. Would it be possible to keep both fields but deprecate one and have both of them being merged together?

I think this is a great solution, keeping backwards compatibility makes sense until a new version of the spec is released.

> If values attached to Tags are applied after creation, will this impact organizations with a set of required tags that must be present on creation? My org, and I'm certain many others, requires certain tags on instances and volumes and CreateInstance calls are blocked without those tags specified.

This proposes using a single field that is applied when resources are created (instead of applied in a step after the fact), this still allows the same set of IAM policy that requires tags at the time resources are created. See line 825 in ec2/instance.go for the change from using `TagSpecifications` to `Tags`.

But for backwards compatibility its likely best this is changed to merge the two as an input to the RunInstancesInput.

PR #2026, opened 2024-03-29
ec2: Create security group with tags in the spec

justinmir:

Resource tags are used to ensure that only IAM roles with the right access can modify the resources. This ensures that the security groups will have the right tags from creation time. Without this change, IAM policy based on tags may fail to run subsequent steps of security group creation.

This change makes security groups created with the tags rather than waiting for the first observation / update to add tags.

I have:

Tested on our own internal production fork and staging environments.


MisterMX (approved):
LGTM. Thank you very much @justinmir!

PR #2027, opened 2024-03-29
ec2: include attributes in instance observation, fix reconciling tags

justinmir:

This ensures that when tags change, the EC2 instance reconciles the update. Previously the ec2 instance controller supported updating tags via `CreateTags` but did not perform reconciliation when tags changed as it was not included in the check for if the EC2 instance was up to date.

However, the way `Update` is currently implemented had the implicit assumption that the EC2 instance must be stopped for all updates, which is true for all the ModifyInstanceAttribute calls, however, tags can be applied at runtime. This changes the logic of the controller to (1) include the instance attributes in the instance observations, (2) only attempt to modify instance attributes if the observation does not match the spec.

I have:

This code has been tested on and deployed in our own fork used in production.

Unit tests are updated to include instance attributes.

### DIFF
--- a/apis/ec2/manualv1alpha1/instances_types.go
+++ b/apis/ec2/manualv1alpha1/instances_types.go
@@ -312,6 +312,8 @@ type InstanceObservation struct {
 	// +optional
 	CPUOptons *CPUOptionsRequest `json:"cpuOptions,omitempty"`
 	// +optional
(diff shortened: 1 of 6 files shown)


MisterMX (approved):
LGTM. Thank you very much @justinmir!

PR #2028, opened 2024-03-29
Reduce ec2 instance controller API calls for an observation

justinmir:

The EC2 instance controller synchronously makes several DescribeInstanceAttribut calls to the EC2 API and does not use the results. This slows down the reconcile time for a single observation.

This removes the API calls for attributes that are returned in `RunInstances`.

I have:

Tested in staging/production in our fork. See drastic decrease in reconcile time for the ec2 instance resource.
image

Related to #2029

### DIFF
--- a/pkg/controller/ec2/instance/controller.go
+++ b/pkg/controller/ec2/instance/controller.go
@@ -158,11 +158,7 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
 
 	for _, input := range []types.InstanceAttributeName{
 		types.InstanceAttributeNameDisableApiTermination,
-		types.InstanceAttributeNameEbsOptimized,
 		types.InstanceAttributeNameInstanceInitiatedShutdownBehavior,
-		types.InstanceAttributeNameInstanceType,
-		types.InstanceAttributeNameKernel,
-		types.InstanceAttributeNameRamdisk,
 		types.InstanceAttributeNameUserData,
 	} {
 		r, err := e.client.DescribeInstanceAttribute(ctx, &awsec2.DescribeInstanceAttributeInput{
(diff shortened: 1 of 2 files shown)


pkg/controller/ec2/instance/controller.go
```diff
@@ -178,26 +174,10 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
-		}
-
-		if r.KernelId != nil {
-			o.KernelId = r.KernelId
-		}
-
-		if r.RamdiskId != nil {
-			o.RamdiskId = r.RamdiskId
-		}
-
```
MisterMX, line 200: These values are used during `LateInitialize` to update the MRs spec. See https://github.com/crossplane-contrib/provider-aws/pull/2028/files#diff-a888525d6868082318c9307abc6c1d162e9647a348dd19b77f1c9695a18fc4ccR201. So these attributes cannot just be removed.

In order to prevent unnecessary AWS calls, we would need to check if the respective spec property needs late-initialization. If that is not the case, we might be able to skip it - if it is not needed for `isUpToDate`.

justinmir: I believe in the linked code (`LateInitializeInstance`) these values are actually pulled from the `RunInstances` result. My investigation below:

The function signature for `LateInitializeInstance` takes in two in:
-  `instance *types.Instance`, populated from the `DescribeInstance` call.
- `attributes *ec2.DescribeInstanceAttributeOutput`. The `attributes` field is what is populated by the calls I removed in this PR.

```
func LateInitializeInstance(in *manualv1alpha1.InstanceParameters, instance *types.Instance, attributes *ec2.DescribeInstanceAttributeOutput)
```

However, the attributes that are removed in my PR aren't actually pulled from `attributes` and instead are populated from the `DescribeInstance` call.

Snippet:
```
	in.EBSOptimized = pointer.LateInitialize(in.EBSOptimized, instance.EbsOptimized)
	in.KernelID = pointer.LateInitialize(in.KernelID, instance.KernelId)
	in.RAMDiskID = pointer.LateInitialize(in.RAMDiskID, instance.RamdiskId)
```

The only one that is not is `InstanceType` which this PR swaps over to use the `DescribeInstances` output.

MisterMX: Yeah that makes sense. I guess I mixed up those two.

MisterMX (approved):
LGTM. Thanks for the contribution and the clarification @justinmir!

Issue #2029, opened 2024-04-01
Improve scalability of the EC2/Route53 controllers


justinmir:
Reddit uses provider-aws to primarily manage EC2 and Route53 resources. In the future we may be adopting it for interacting with various AWS managed services. The provider-aws controllers currently manage on the order of several thousand resources each, ~3000+ EC2 instances, and ~6000+ route53 records.

We are currently running provider-aws version 0.46.

### What problem are you facing?
At this scale we run into issues with high queue depth in our controllers, even with 20+ active workers (provider-aws --max-reconcile-rate=20).

![image](https://github.com/crossplane-contrib/provider-aws/assets/8886628/d4e751e8-ed56-4c74-b96a-c3e55321ec12)
**Figure 1: Crossplane controller is unable to work through queue of reconcile requests even with 20 active workers at our scale.**

Reconcile times for EC2 instances typically take greater than one second at median and can take up to 6 seconds at p99.
![image](https://github.com/crossplane-contrib/provider-aws/assets/8886628/f179fd95-7172-40c0-b9b3-d6e0a57a3104)
**Figure 2: Reconcile time for `instance` resources**

Without any jitter, ResourceRecordSet resource observations can cause a backlog that can take up to an hour to resolve. AWS rate limits route53 API requests to 5 requests / second / account, which makes it extremely easy to hit the rate limit when performing observation on route53 resources. These queue depths exist even with poll intervals set to 30 minutes (up from 1 minute) in our fork.

![image](https://github.com/crossplane-contrib/provider-aws/assets/8886628/26cacd02-7c72-4fce-95c0-36e2c0000a5b)
**Figure 3: Resource record set controller is backlogged during instance observation.**

### How could Crossplane help solve your problem?

**Allow configuring per-resource poll interval / jitter**
Introduce jitter in resource record sets, introducing jitter smooths the request rate of the resource due to observations and minimizes the impact of the rate limit. We hard-code jitter in our crossplane provider-aws fork.

![image](https://github.com/crossplane-contrib/provider-aws/assets/8886628/db61ffbd-634e-45f8-8026-021c140b5d43)

**Reduce reconcile time for EC2 instance resources**
Reduce the time spent to perform an EC2 instance observation by: (1) reduce unnecessary API calls for duplicate data, (2) parallelize API calls where possible.

davimmt:
[...]

PR #2030, opened 2024-04-04
Allow to override options for specific controllers

max-melentyev:

We're facing long poll queues when using crossplane to manage thousands of EC2 instances,
and facing AWS API throttling while observing thousands of route53 records (#2029).
As an intermediate step we want to reduce poll interval for these resources and keep it short for other resources.

In further PRs I plan to make these resources to be monitored using EventBridge events 
and these controller to have even longer poll intervals.
Resources without EventBridge events will still have short poll intervals.

Options are parsed from env variables with `PROVIDER_AWS_` prefix, for example: `PROVIDER_AWS_ec2.instance.pollInterval=10m`.

I have:

- with new tests
- manually in dev setup with fmt.Printf :)


max-melentyev:
I'm happy to add more docs on this feature, but I didn't find any appropriate place for it other than source code comments. Please let me know if there's a better option.

cmd/provider/main.go
```diff
@@ -138,3 +147,18 @@ func UseISO8601() zap.Opts {
 		o.TimeEncoder = zapcore.ISO8601TimeEncoder
 	}
 }
+
+// Collects all env variables with the prefix OPTION_ENV_PREFIX and returns them as a map
+// with the prefix removed.
+func optionsOverridesFromEnv() map[string]string {
```
negz, line 153: Doesn't kingpin support natively loading flags from environment variables?

max-melentyev: I see it has binding of particular flag to an env variable. This will add ~320 new flags and it'll still require some code to collect these flags into overrides map.

negz: I'm not a maintainer of this provider anymore, so I'll defer to @MisterMX and @chlunde.

That said, I'm a bit concerned about:

* Adding a feature to only this provider to tune poll interval per resource type
* Introducing a new path to configure the provider (environment variables)

I think if this is a problem for this provider it's probably a problem for other providers. It would be ideal to agree on the problem, and a standard way to address it. Usually to do that we'd write a one-pager in the crossplane/crossplane repo, that provider owners can adopt.

@ulucinar is the TL of Upbound's extensions team (he owns tools like upjet that generate a lot of the other providers). He might be a good person to review a one-pager if you're interested in writing one.

MisterMX:
This looks like a very drastic change to me. And I honestly don't think that we are going merge that since it is very custom solution for a single provider which does not comply to the other providers.

I would prefer a more general approach that can be included in https://github.com/crossplane/crossplane-runtime. As @negz suggested it is probably good to start with a design proposal in https://github.com/crossplane/crossplane.

chlunde:
Also discussed here:
https://github.com/crossplane-contrib/provider-aws/pull/1625#issuecomment-1381225016

MisterMX:
Closing this for now as the goal is rather to find a generic solution that works across all Crossplane providers.

PR #2031, opened 2024-04-09
fix(database): ignore restoreFrom

wotolom:

for `RDSInstance.database`:

- ignore the parameter `restoreFrom` in the `isUpToDate()` check

this fixes an update-loop if the RDSInstance claim was created with `restoreFrom`. 

Same logic as in the `DBInstance.rds` controller: https://github.com/crossplane-contrib/provider-aws/blob/ddd940a4e790b6425efbf4141a30b480c6125671/pkg/controller/rds/dbinstance/setup.go#L495

I have:

manually


MisterMX (approved):

PR #2037, opened 2024-04-12
fix(dbinstance): in isUpToDate consider 0 and nil to be equal for iops/storageThroughput

wotolom:

AWS seems to be giving different responses for IOPS/StorageThroughput (either 0 or nil) depending on whether the instance was created as `gp2` or modified from another storageType (e.g. `gp3`) to `gp2`,

This changes their `isUptoDate()` checks to consider both 0 and nil to be equivalent.

Also I used the opportunity and added missing parameters to the built diff for the logs.

I have:

manually


MisterMX (approved):

PR #2039, opened 2024-04-15
Call route53.ListResourceRecordSets with MaxItems = 1

max-melentyev:
Otherwise AWS returns ALL routes with name >= requested one.

Fetching unnecessary records results in slow & inefficient reconciliations
when account has a lot or route53 resords.

I have:

- Manual integration testing

### DIFF
--- a/pkg/clients/resourcerecordset/resourcerecordset.go
+++ b/pkg/clients/resourcerecordset/resourcerecordset.go
@@ -71,9 +71,11 @@ func NewClient(cfg aws.Config) Client {
 // GetResourceRecordSet returns recordSet if present or err
 func GetResourceRecordSet(ctx context.Context, name string, params v1alpha1.ResourceRecordSetParameters, c Client) (*route53types.ResourceRecordSet, error) {
 	res, err := c.ListResourceRecordSets(ctx, &route53.ListResourceRecordSetsInput{
-		HostedZoneId:    params.ZoneID,
-		StartRecordName: &name,
-		StartRecordType: route53types.RRType(params.Type),
+		HostedZoneId:          params.ZoneID,
+		StartRecordName:       &name,
+		StartRecordType:       route53types.RRType(params.Type),
+		StartRecordIdentifier: params.SetIdentifier,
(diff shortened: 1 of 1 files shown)


pkg/clients/resourcerecordset/resourcerecordset.go
```diff
@@ -71,9 +71,11 @@ func NewClient(cfg aws.Config) Client {
...
 func GetResourceRecordSet(ctx context.Context, name string, params v1alpha1.ResourceRecordSetParameters, c Client) (*route53types.ResourceRecordSet, error) {
 	res, err := c.ListResourceRecordSets(ctx, &route53.ListResourceRecordSetsInput{
-		HostedZoneId:    params.ZoneID,
-		StartRecordName: &name,
-		StartRecordType: route53types.RRType(params.Type),
+		HostedZoneId:          params.ZoneID,
+		StartRecordName:       &name,
+		StartRecordType:       route53types.RRType(params.Type),
+		StartRecordIdentifier: params.SetIdentifier,
+		MaxItems:              aws.Int32(1),
```
MisterMX, line 78: I wonder if we need proving `maxItems` to AWS since we are providing type and name of the RRS. I am also unsure about `StartRecordIdentifier` which might cause problems since it is an optional field.

max-melentyev: The way ListResourceRecordSets work is that it returns records in ascending order starting from the first matching record. You can test it with
```
aws route53 list-resource-record-sets --start-record-name a --start-record-type A --hosted-zone-id {id}
```
This will return records with name >= 'a', and also include records of other type.

`StartRecordIdentifier` works the same was as `StartRecordName` and `StartRecordType` - it'll filter apply filters to find first matching record. It's already used later on line 98 (96) to find matching record so it won't cause any problems.

MisterMX: Makes sense. Thanks for the clarification.

MisterMX (approved):
LGTM. Thank you very much @max-melentyev!

PR #2047, opened 2024-04-23
fix(ec2): Call describe-instance API concurrently

max-melentyev:

Instance observation makes 4 AWS API calls. Running this calls concurrently reduces reconciliation time. Here is 0.99 quantile of reconciliation duration

image

I have:

with existing unit tests

### DIFF
--- a/pkg/controller/ec2/instance/controller.go
+++ b/pkg/controller/ec2/instance/controller.go
@@ -18,8 +18,6 @@ package instance
 
 import (
 	"context"
-	"sync"
-	"sync/atomic"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsec2 "github.com/aws/aws-sdk-go-v2/service/ec2"
@@ -33,6 +31,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -181,20 +180,18 @@ func (e *external) describeInstance(ctx context.Context, instanceId string) (
 	awsec2.DescribeInstanceAttributeOutput,
 	error,
 ) {
-	wg := sync.WaitGroup{}
+	eg := errgroup.Group{}
 
 	var describeOutput *awsec2.DescribeInstancesOutput
 	var describeError error
-	wg.Add(1)
-	go func() {
+	eg.Go(func() error {
 		describeOutput, describeError = e.client.DescribeInstances(ctx, &awsec2.DescribeInstancesInput{
 			InstanceIds: []string{instanceId},
 		})
-		wg.Done()
-	}()
+		return nil
+	})
 
 	attrs := awsec2.DescribeInstanceAttributeOutput{}
-	attrsErr := atomic.Pointer[error]{}
 	descAttr := func(attr types.InstanceAttributeName) (*awsec2.DescribeInstanceAttributeOutput, error) {
 		return e.client.DescribeInstanceAttribute(ctx, &awsec2.DescribeInstanceAttributeInput{
 			InstanceId: &instanceId,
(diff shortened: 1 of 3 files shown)


MisterMX:
Adding multi-threading generally increases the complexity of the code and makes it harder to debug. Since we have a bunch of other controllers (i.e. S3) where we make multiple, consecutive calls that cause no problem so far, I am very reluctant to merge this. 

Is there any issue that comes with the increased reconcile time for EC2 instances?

max-melentyev:
Having thousands of instances, this allowed us to reduce queue size and period of a single instance reconciliation from 10+ minutes down to several minutes.

pkg/controller/ec2/instance/controller.go
```diff
@@ -221,6 +176,86 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+	}()
+
+	attrs := awsec2.DescribeInstanceAttributeOutput{}
+	attrsErr := atomic.Pointer[error]{}
+	descAttr := func(attr types.InstanceAttributeName) (*awsec2.DescribeInstanceAttributeOutput, error) {
+		return e.client.DescribeInstanceAttribute(ctx, &awsec2.DescribeInstanceAttributeInput{
+			InstanceId: &instanceId,
+			Attribute:  attr,
+		})
+	}
```
MisterMX, line 203: I wonder if this could be done more elegantly by iterating over an array of attribute keys and start a Go routine in parallel. I would also prefer to use https://pkg.go.dev/golang.org/x/sync/errgroup instead of standard `WaitGroup`.

Something like this:

```go
func (e *external) describeInstanceAttributes(ctx context.Context, cr *svcapitypes.Instance) (*awsec2.DescribeInstanceAttributeOutput, error) {
	eg := errgroup.Group{}
	out := &awsec2.DescribeInstanceAttributeOutput{}
	describeAttribute := func(attr types.InstanceAttributeName, onSuccess func(res *awsec2.DescribeInstanceAttributeOutput)) {
		eg.Go(func() error {
			res, err := e.client.DescribeInstanceAttribute(ctx, &awsec2.DescribeInstanceAttributeInput{
				InstanceId: aws.String(meta.GetExternalName(cr)),
				Attribute: attr,
			})
			if err != nil {
				err = errors.Wrap(err, string(attr))
				return err
			}
			onSuccess(res)
			return nil
		})
	}

	describeAttribute(types.InstanceAttributeNameDisableApiTermination, func(res *awsec2.DescribeInstanceAttributeOutput) {
		out.DisableApiTermination = res.DisableApiTermination
	})
	describeAttribute(types.InstanceAttributeNameInstanceInitiatedShutdownBehavior, func(res *awsec2.DescribeInstanceAttributeOutput) {
		out.InstanceInitiatedShutdownBehavior = res.InstanceInitiatedShutdownBehavior
	})
	describeAttribute(types.InstanceAttributeNameUserData, func(res *awsec2.DescribeInstanceAttributeOutput) {
		out.UserData = res.UserData
	})
	
	err := eg.Wait()
	return out, err
}
```

MisterMX (changes requested):

pkg/controller/ec2/instance/controller.go
max-melentyev: Oh, cool. I didn't know about `errgroup`. It looks much better now.

pkg/controller/ec2/instance/controller.go
```diff
@@ -221,6 +175,81 @@ func (e *external) Observe(ctx context.Context, mgd resource.Managed) (managed.E
...
+
+	eg.Go(func() error {
+		if res, err := descAttr(types.InstanceAttributeNameDisableApiTermination); err != nil {
+			return errorutils.Wrap(err, "fetching DisableApiTermination")
+		} else {
+			attrs.DisableApiTermination = res.DisableApiTermination
+			return nil
+		}
+	})
```
MisterMX, line 209: This is still quite a lot of duplicate code. Can we do it like in https://github.com/crossplane-contrib/provider-aws/pull/2047/files#r1577643597?

Also, there is no need for `else` if the previous `if` block had a `return`.

max-melentyev: > This is still quite a lot of duplicate code.

This is just the way error handling works in go, all these `if err != nil`. `descAttr` is straightforward and has no side-effects now. Having error handling inside `descAttr` adds a side-effect, and creation of a goroutine inside a small helper method. This makes things more complicated. And it would have same or more LOC.

> Also, there is no need for else if the previous if block had a return.

`res` is not visible outside of `if ... else`.

MisterMX (approved):
LGTM. Thank you very much @max-melentyev!

PR #2064, opened 2024-05-24
feat: Do not set Status.AtProvider.LastUsedAt for IAM roles by default

max-melentyev:
Because it can call frequent unnecessary reconciliations of dependent resources.

`LastUsedAt` is updated frequently and triggers unnecessary reconciliations of dependent resources.

I presumed that this field is used rarely and reasonable default would be not to update it on every reconciliation, even though this will be a breaking change.

Please let me know if you think it's better to use `IGNORE_LAST_USED_AT` annotation instead.

I have:

new test case

### DIFF
--- a/apis/iam/v1beta1/role_types.go
+++ b/apis/iam/v1beta1/role_types.go
@@ -21,6 +21,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// Set this annotation to update Status.AtProvider.LastUsedAt.
+	// Note that this may trigger frequent updates and reconciliation of
+	// dependent resources.
+	TRACK_LAST_USED_AT = "roles.iam.crossplane.io/track-last-used-at"
+)
+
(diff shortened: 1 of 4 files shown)


MisterMX:
Can you give more information about the issue you are experiencing? Which reconcilations are triggered by an update of the roles status information?

max-melentyev:
We have a CRD and its controller manages roles used by the CRD. This controller [watches](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/builder/controller.go#L137) all its managed resources including roles to make sure these resources exist and have required specs. This makes controller-runtime to run controller every time watched resource changes. Roles are used often and this makes CRD controller run constantly even if nothing else happens.

MisterMX:
I'm not very confident with adding this exception for a very special use case. As long as there no other users having the same issue in their environment I would suggest to not merge this and propose to use a custom fork instead.

max-melentyev:
What if I change it to the "default to the current behaviour" and make new behaviour optional?

PR #2066, opened 2024-06-07
fix: add engine version for dbinstance update

akesser:
Currently, when changing the engine version of a rds db instance, the new version is not send to aws. This leads to a) the database does not get updated and b) an update loop, as the isUpToDate call always returns false.

Fixes #

I have:


haarchri:
add update functionality back after https://github.com/crossplane-contrib/provider-aws/pull/1765

PR #2116, opened 2024-11-04
fix(s3): dont set ExpiredObjectDeleteMarker in lifecycle policy if da…

kkendzia:
This PR solved the issue with a parameter collision in S3 Bucket Lifecycle Configurations.

According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html#lifecycle-config-conceptual-ex7 you cannot specify both `Days` / `Date` and `ExpiredObjectDeleteMarker`, resulting in a MalformedXML Error otherwise.

This change is adapted to the behaviour of terraform provider aws: https://github.com/hashicorp/terraform-provider-aws/blob/main/internal/service/s3/bucket_lifecycle_configuration.go#L664

When Date or Days is set, don't set ExpiredObjectDeleteMarker at all.

I have:

tested by @simwak 
* Built
* Tested with example MRs

### DIFF
--- a/pkg/controller/s3/bucket/lifecycleConfig.go
+++ b/pkg/controller/s3/bucket/lifecycleConfig.go
@@ -161,12 +161,18 @@ func GenerateLifecycleRules(in []v1beta1.LifecycleRule) []types.LifecycleRule { 
 			}
 		}
 		if local.Expiration != nil {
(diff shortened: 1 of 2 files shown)


MisterMX (approved):
LGTM. Thank you very much @kkendzia!

PR #2145, opened 2025-01-22
feat(transfer): add vpcendpoint dns entries to connection secrets

anagarlau:

Fixes #

I have:

### DIFF
--- a/pkg/controller/transfer/server/setup.go
+++ b/pkg/controller/transfer/server/setup.go
@@ -15,8 +15,11 @@ package server
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/aws/aws-sdk-go/service/ec2"
 	svcsdk "github.com/aws/aws-sdk-go/service/transfer"
+	svcsdkapitransfer "github.com/aws/aws-sdk-go/service/transfer/transferiface"
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/connection"
 	"github.com/crossplane/crossplane-runtime/pkg/controller"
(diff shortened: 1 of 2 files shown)


pkg/controller/transfer/server/setup.go
```diff
@@ -33,13 +36,32 @@ import (
...
+	wrapperOpts := []wrapperOption{
+		func(e *wrapperExternal) {
+			c := &custom{
+				client:            e.client,
+				kube:              e.kube,
+				external:          e.external,
+				vpcEndpointClient: e.vpcEndpointClient,
+			}
+
+			e.postObserve = c.postObserve
```
MisterMX, line 64: I think we can avoid a second GET on the providerconfig by implementing a custom connector that returns a custom external with the standard external nested in it. Something like:

```go
type customConnector struct {
	kube client.Client
	opts []option
}

func (c *connector) Connect(ctx context.Context, mg cpresource.Managed) (managed.ExternalClient, error) {
	// the code can either be copied from zz_controller.go or reused and the result casted (whatever feels better)

        cr, ok := mg.(*svcapitypes.Server)
	if !ok {
		return nil, errors.New(errUnexpectedObject)
	}
	sess, err := connectaws.GetConfigV1(ctx, c.kube, mg, cr.Spec.ForProvider.Region)
	if err != nil {
		return nil, errors.Wrap(err, errCreateSession)
	}

	ec2Client := c.newClientFn(*cfg)
	if vpcClient == nil {
		return nil, errors.New("failed to initialize EC2 client")
	}

	return &customExternal{
		ec2: ec2Client
		external: newExternal(c.kube, svcapi.New(sess), c.opts)
	}
}

type customExternal struct {
  *external
  ec2Client ec2.Client
}

func (e *customExternal) postObserve() {
  // endpoint observation + existing code
}
```

MisterMX (changes requested):

pkg/controller/transfer/server/setup.go
anagarlau: Technically this would be the best way to do it but given that the Connect method on the connector is already generated, this approach would require generated code to be removed.  I m not sure it can be skipped

PR #2146, opened 2025-01-22
fix(rdsinstance.database): do not send empty modify-calls

wotolom:

for RDSInstance.database:
- do not request Modify, when ModifyDBInstanceInput only includes `DBInstanceIdentifier` and the Modify-Flags(`AllowMajorVersionUpgrade`, `ApplyImmediately`)

When user only changed Tags, this could lead to getting the MR stuck, if `ApplyImmediately: false` was set.
AWS then blocks the request and the whole update-fails:

> "update failed: cannot modify RDS instance: api error InvalidParameterCombination: No modifications were requested"

This rarely happened, due to AWS curiously not rejecting empty Modify-Calls with `ApplyImmediately: true`.

I have:

manually

### DIFF
--- a/pkg/clients/database/rds.go
+++ b/pkg/clients/database/rds.go
@@ -887,6 +887,10 @@ func generateCloudWatchExportConfiguration(spec, current []string) *rdstypes.Clo
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


MisterMX (approved):

PR #2149, opened 2025-01-22
fix(rdsinstance.database): do not overwrite engineVersion if only major was set + add engineVersion to observe

wotolom:

for `RDSInstance.database`:
- remove lateinit overwrite of `engineVersion` when only major is set
  - this is no longer needed since we check in isUpToDate the `engineVersion` in detail (considering major and minor)
- now fill  - already existing status field... - with observed `engineVersion` in AWS
- change printcolumn to `.status.atProvider.engineVersion`

Fixes:

for setups where MRs were applied automatically (e.g. via Composition/XR) the `spec.forProvider.engineVersion` would constantly flicker due to this lateinit part, when user set only the major version (which is valid by AWS; and makes sense in combination with `autoMinorVersionUpgrade`)

I have:

manually


MisterMX (approved):

PR #2169, opened 2025-02-18
fix(ec2.instance): permit to create instance with no tags

akofman:

This PR fixes an issue in the EC2 instance controller where an empty tag list causes the `CreateTags` API call to fail, preventing the instance’s external name from being set. 

Currently, `Create` always attempts to call `CreateTags`, even when no tags are provided. However, AWS rejects `CreateTags` with an empty tag list, returning an `InvalidParameterValue`.
Since this error occurs before `SetExternalName` is called, the Crossplane reconciliation loop fails to detect that the instance was already created and keeps creating new instances in an infinite loop.

I have:

I have added a SuccessfulWithNoTags test to explicitly check that CreateTags is not called when no tags are provided.

### DIFF
--- a/pkg/controller/ec2/instance/controller.go
+++ b/pkg/controller/ec2/instance/controller.go
@@ -265,14 +265,16 @@ func (e *external) Create(ctx context.Context, mgd resource.Managed) (managed.Ex
 
 	instance := result.Instances[0]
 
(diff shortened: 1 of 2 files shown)


MisterMX (approved):
LGTM. Thank you very much @akofman!

PR #2182, opened 2025-03-11
fix(efs): Adjusted Filesystem Update

### LINKED ISSUE #2135: EFS FileSystem not updated
### What happened?

I create a FileSystem MR among other MR from a XR. The creation of all MRs succeed, and my XR is "Ready" and "Synced" (as well as the FileSystem).
When I update the parameters of my XR (FileSystem throughputMode, or FileSystem backup), my XR is ready (also the FileSystem MR) with updated parameter values but the external resource EFS FileSystem stays unchanged : throughputMode or/and backup parameters are not updated.
The FileSystem MR Status values are regularly well updated.
The external resource EFS FileSystem and its dependencies are well deleted when I delete my XR.

### How can we reproduce it?
Spec of the EFS FileSystem:
```bash
...
Spec:
  Deletion Policy:  Delete
  For Provider:
    Backup:            false
    Encrypted:         true
    Performance Mode:  generalPurpose
    Region:            eu-west-3
    Tags:
      Key:            Name
      Value:          default-testfab
    Throughput Mode:  bursting
  Management Policies:
    *
  Provider Config Ref:
    Name:  aws-provider
...
```

Crossplane version: 1.17.1
crossplane-contrib/provider-aws: 0.50.6 (and 0.44.1 on other cluster)

kuseler:

Fixes #2135

This change enables the update of Throughputmode and BackupPolicy. 

I have:

1. apply efs1.yml:

```yaml
kind: FileSystem
metadata:
  name: greattestingfilesystem
spec:
  deletionPolicy:  Delete
  forProvider:
    backup:            false
    encrypted:         false
    performanceMode:  generalPurpose
    region:            eu-west-3
    throughputMode:  elastic
  providerConfigRef:
    name: aws-community-provider
```

2. apply efs2.yml

```yaml
apiVersion: efs.aws.crossplane.io/v1alpha1
kind: FileSystem
metadata:
  name: greattestingfilesystem
spec:
  deletionPolicy:  Delete
  forProvider:
    backup:            true
    encrypted:         true
    performanceMode:  generalPurpose
    region:            eu-west-3
    throughputMode:  bursting
  providerConfigRef:
    name: aws-community-provider
```

3. confirm that the throughputmode and BackupPolicy have changed.

### DIFF
--- a/pkg/controller/efs/filesystem/setup.go
+++ b/pkg/controller/efs/filesystem/setup.go
@@ -4,7 +4,9 @@ import (
 	"context"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	svcsdk "github.com/aws/aws-sdk-go/service/efs"
+	svcsdkapi "github.com/aws/aws-sdk-go/service/efs/efsiface"
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/connection"
 	"github.com/crossplane/crossplane-runtime/pkg/controller"
@@ -12,6 +14,8 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/meta"
 	"github.com/crossplane/crossplane-runtime/pkg/reconciler/managed"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/pkg/errors"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	svcapitypes "github.com/crossplane-contrib/provider-aws/apis/efs/v1alpha1"
@@ -26,13 +30,14 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
 	name := managed.ControllerName(svcapitypes.FileSystemGroupKind)
 	opts := []option{
 		func(e *external) {
+			c := &custom{client: e.client, external: e}
 			e.isUpToDate = isUpToDate
 			e.preCreate = preCreate
 			e.postCreate = postCreate
 			e.preObserve = preObserve
 			e.preUpdate = preUpdate
 			e.preDelete = preDelete
-			e.postObserve = postObserve
+			e.postObserve = c.postObserve
 		},
 	}
 
@@ -67,11 +72,20 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
 		Complete(r)
 }
 
+type custom struct {
+	client   svcsdkapi.EFSAPI
+	external *external
+}
+
 func isUpToDate(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.DescribeFileSystemsOutput) (bool, string, error) {
 	for _, res := range obj.FileSystems {
 		if pointer.Int64Value(cr.Spec.ForProvider.ProvisionedThroughputInMibps) != int64(aws.Float64Value(res.ProvisionedThroughputInMibps)) {
 			return false, "", nil
 		}
+		if !ptr.Equal(cr.Spec.ForProvider.ThroughputMode, res.ThroughputMode) {
+			return false, "", nil
+		}
+
 	}
 	return true, "", nil
 }
@@ -79,11 +93,11 @@ func isUpToDate(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Descr
 func preObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.DescribeFileSystemsInput) error {
 	// Describe query doesn't allow both CreationToken and FileSystemId to be given.
 	obj.CreationToken = nil
-	obj.FileSystemId = pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr))
+	obj.FileSystemId = ptr.To(meta.GetExternalName(cr))
 	return nil
 }
 
-func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.DescribeFileSystemsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
+func (e *custom) postObserve(ctx context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.DescribeFileSystemsOutput, obs managed.ExternalObservation, err error) (managed.ExternalObservation, error) {
 	if err != nil {
 		return managed.ExternalObservation{}, err
 	}
@@ -92,6 +106,13 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
 	}
 	obs.ConnectionDetails = managed.ConnectionDetails{
 		svcapitypes.ResourceCredentialsSecretIDKey: []byte(meta.GetExternalName(cr)),
+	}
+
+	// BackupPolicy is managed separately from the FileSystem, so we update it here to avoid empty requests in preUpdate
+	if obs.ResourceExists {
+		if err := e.updateBackupPolicy(ctx, cr, obj); err != nil {
+			return managed.ExternalObservation{}, errors.Wrap(err, "cannot update backup policy")
+		}
 	}
 	return obs, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/controller/efs/filesystem/setup.go
```diff
@@ -67,11 +74,54 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
...
 		}
+		if !ptr.Equal(cr.Spec.ForProvider.ThroughputMode, res.ThroughputMode) {
+			return false, "", nil
+		}
+
+		err := e.cacheBackupPolicyHelper(res.FileSystemId)
+		if err != nil {
+			return false, "", err
+		}
+		if *cr.Spec.ForProvider.Backup != e.cache.backupPolicy {
```
MisterMX, line 122: Possible nil deref

pkg/controller/efs/filesystem/setup.go
```diff
@@ -67,11 +74,54 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
...
 		if pointer.Int64Value(cr.Spec.ForProvider.ProvisionedThroughputInMibps) != int64(aws.Float64Value(res.ProvisionedThroughputInMibps)) {
 			return false, "", nil
 		}
+		if !ptr.Equal(cr.Spec.ForProvider.ThroughputMode, res.ThroughputMode) {
+			return false, "", nil
+		}
+
+		err := e.cacheBackupPolicyHelper(res.FileSystemId)
+		if err != nil {
+			return false, "", err
```
MisterMX, line 120: We should always `errors.Wrap` with a custom message to easily identify where the error comes from.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -67,11 +74,54 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
...
+	if err != nil {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == "PolicyNotFound" {
+			e.cache.backupPolicy = false
+			return nil
+		}
+		return err
+	}
+
+	switch *policy.BackupPolicy.Status {
```
MisterMX, line 100: Possible nil deref

pkg/controller/efs/filesystem/setup.go
```diff
@@ -67,11 +74,54 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
...
+	// Get BackupPolicy
+	//default is no backup
+	policy, err := e.client.DescribeBackupPolicy(&svcsdk.DescribeBackupPolicyInput{FileSystemId: filesystemId})
+	if err != nil {
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) && awsErr.Code() == "PolicyNotFound" {
+			e.cache.backupPolicy = false
+			return nil
+		}
+		return err
```
MisterMX, line 97: Should be moved into an extra function like here: https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/controller/eks/addon/zz_conversions.go#L226

```suggestion
		return resource.Ignore(err, IsNotFound())
```

pkg/controller/efs/filesystem/setup.go
```diff
@@ -96,12 +146,41 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
...
+
+		if pointer.StringValue(obj.FileSystemId) == "" {
+			obj.FileSystemId = pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr))
+		}
+		_, err := e.client.PutBackupPolicyWithContext(ctx,
+			&svcsdk.PutBackupPolicyInput{
+				FileSystemId: obj.FileSystemId,
+				BackupPolicy: &svcsdk.BackupPolicy{Status: policy}})
+		if err != nil {
+			return err
```
MisterMX, line 172: missing `errors.Wrap`

pkg/controller/efs/filesystem/setup.go
```diff
@@ -96,12 +146,41 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
...
+			policy = aws.String("DISABLED")
+		}
+
+		if pointer.StringValue(obj.FileSystemId) == "" {
+			obj.FileSystemId = pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr))
+		}
+		_, err := e.client.PutBackupPolicyWithContext(ctx,
+			&svcsdk.PutBackupPolicyInput{
+				FileSystemId: obj.FileSystemId,
+				BackupPolicy: &svcsdk.BackupPolicy{Status: policy}})
```
MisterMX, line 170: For consistency the code formatting should be same as everywhere else: Newline before `}`.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -96,12 +146,41 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
...
-func preUpdate(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.UpdateFileSystemInput) error {
+func (e *custom) preUpdate(ctx context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.UpdateFileSystemInput) error {
 	obj.FileSystemId = pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr))
 	// Type of this field is *float64 but in practice, only integer values are allowed.
 	if cr.Spec.ForProvider.ProvisionedThroughputInMibps != nil {
 		obj.ProvisionedThroughputInMibps = aws.Float64(float64(pointer.Int64Value(cr.Spec.ForProvider.ProvisionedThroughputInMibps)))
 	}
+
+	if e.cache.backupPolicy != *cr.Spec.ForProvider.Backup {
+		var policy *string
```
MisterMX, line 157: Don't have to use a pointer here and stick with value. We can reference it with `&` later.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -96,12 +146,41 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
...
+			return err
+		}
+	}
+
+	/*	Backup is not a direct part of the FileSystem, but an externally managed BackupPolicy.
+		Changing only the backup policy will result in an empty request and thus an Error.
+		To avoid this, we return an error if the BackupPolicy is the only thing that has changed.
+	*/
+	if ptr.Equal(cr.Spec.ForProvider.ThroughputMode, obj.ThroughputMode) && pointer.Int64Value(cr.Spec.ForProvider.ProvisionedThroughputInMibps) == int64(aws.Float64Value(obj.ProvisionedThroughputInMibps)) {
+		return errors.New("throughputmode and provisionedthroughputinmibps are unchanged, aborting update")
```
MisterMX, line 181: Returning an error here will trigger a constant reschedule of the reconcile and make the resource show as unready - although there is actual not any error. If we really would want to prevent the update we would need to modify the `IsResourceUpToDate` result in `Observe`.

MisterMX (changes requested):

pkg/controller/efs/filesystem/setup.go
```diff
@@ -67,11 +73,25 @@ func SetupFileSystem(mgr ctrl.Manager, o controller.Options) error {
 		Complete(r)
 }
 
+type custom struct {
+	kube     client.Client
+	client   svcsdkapi.EFSAPI
+	external *external
+
+	cache struct {
+		backupPolicy bool
```
MisterMX, line 82: Boolean variables should always describe "yos-or-no" states. Better here would be `shouldUpdateBackupPolicy`.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -126,3 +152,60 @@ func postCreate(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.FileS
...
+		e.cache.backupPolicy = true
+	case "DISABLED", "DISABLING":
+		e.cache.backupPolicy = false
+	}
+	return nil
+}
+
+func (e *custom) updateBackupPolicy(ctx context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.DescribeFileSystemsOutput) error {
+	for _, res := range obj.FileSystems {
+		err := e.cacheBackupPolicy(res.FileSystemId)
```
MisterMX, line 176: This `cache` thingy is only used as a workaround to pass state information from `Observe` to `Create`, `Update` or `Delete` so you don't have to recompute it, i.e. difference in resource tags - because the controller framwork does not allow passing custom values in a `managed.ExternalObservation`.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -126,3 +152,60 @@ func postCreate(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.FileS
 	meta.SetExternalName(cr, pointer.StringValue(obj.FileSystemId))
 	return managed.ExternalCreation{}, nil
 }
+
+func (e *custom) cacheBackupPolicy(filesystemId *string) error {
```
MisterMX, line 156: Rather have a function `isBackupPolicyUpToDate` that returns a `(bool, error)`.

pkg/controller/efs/filesystem/setup.go
```diff
@@ -93,6 +113,12 @@ func postObserve(_ context.Context, cr *svcapitypes.FileSystem, obj *svcsdk.Desc
 	obs.ConnectionDetails = managed.ConnectionDetails{
 		svcapitypes.ResourceCredentialsSecretIDKey: []byte(meta.GetExternalName(cr)),
 	}
+
+	// BackupPolicy is managed separately from the FileSystem, so we update it here to avoid empty requests in preUpdate
+	updateBackupPolicyErr := e.updateBackupPolicy(ctx, cr, obj)
+	if updateBackupPolicyErr != nil {
+		return managed.ExternalObservation{}, updateBackupPolicyErr
+	}
```
MisterMX, line 121: I don't have much EFS insights but I think this section should only be executed if `obs.ResourceExists` is true?

pkg/controller/efs/filesystem/setup.go
```diff
@@ -2,9 +2,13 @@ package filesystem
...
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/pkg/errors"
+	"k8s.io/utils/ptr"
 
 	"github.com/aws/aws-sdk-go/aws"
 	svcsdk "github.com/aws/aws-sdk-go/service/efs"
+	svcsdkapi "github.com/aws/aws-sdk-go/service/efs/efsiface"
 	xpv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
 	"github.com/crossplane/crossplane-runtime/pkg/connection"
 	"github.com/crossplane/crossplane-runtime/pkg/controller"
```
MisterMX, line 14: The linter requires Go imports to be in a specific order (defined in `.golangci-lint.yml`) separated by empty lines:

* Go standard packages
* Third-party packages
* `github.com/crossplane-contrib/provider-aws/*`

MisterMX (changes requested):

MisterMX (changes requested):

MisterMX (approved):
LGTM. Thank you very much @kuseler!

PR #2187, opened 2025-03-18
fix(wafv2): WebACL isUpToDate bugs

teeverr:

Fixes #
 - an error in the first reconciliation loop while taking control over existing webacl with modified configuration.
 - controller doesn't detect diff in case if a parameter became to nill, e.g. removing of rules[] section doesn't trigger update on aws side

QoL improvements:
  - rules[].byteMatchStatement.searchString is not required to be base64 encoded anymore. Actually it is breaking change from the perspective of api(v1alpha1) but wafv2 webacl is not included in any releases yet. So I believe it is okay.

I have:

unse


MisterMX (approved):

PR #2206, opened 2025-04-09
fix(ec2): Partial revert of 138b8cf to fix issue crossplane-contrib#2196

### LINKED ISSUE #2196: AWS EC2 Route Tables with Tags, never gets Ready
### What happened?

After upgrading from version 0.49.2 to 0.52.0 we've detected that the EC2 RouteTable started updating non-stop.
We've been troubleshooting and debugging and it seems that the changes introduced in https://github.com/crossplane-contrib/provider-aws/commit/138b8cf25debbe75a98f5af919cb8e05b2850fd3 have changed the routetable.go code:
https://github.com/crossplane-contrib/provider-aws/blob/e02a4c2de22c21f188f5a514297b54e9bd7677cc/pkg/clients/ec2/routetable.go#L163

We have done some tests, and reverting back to `SortTagsV1Beta1` and also changing the 1st parameter to  `targetCopy.Tags`, we had success 
`SortTagsV1Beta1(targetCopy.Tags, in.Tags)`

### How can we reproduce it?

Create a RouteTable CR in version 0.49.0 with a list of unordered tags, upgrade to 0.52.0 and the result with show itself.

Crossplane version: 0.52.0
Cloud provider: AWS EKS 1.30

fernandoalexandre:

This PR reverts partially reverts the changes of 138b8cf to not detect changes when only the order of the tags change between what is returned by AWS and what is being submitted on the `RouteTable` CR.

The revert of this change will avoid detecting order-related differences because the `SortTagsV1Beta1(targetCopy.Tags, in.Tags)` function changes the order of `targetCopy.Tags` and `in.Tags` by reference which enables a proper comparison afterwards in:
`jsonPatch, err := jsonpatch.CreateJSONPatch(*currentParams, targetCopy)`

The `currentParams.Tags` will have the same tag order of `in.Tags` from the `LateInitializeRT(currentParams, &in)`.

Note: the `CompareTags` added in 138b8cf is effectively doing nothing as the result is not being used and nothing is changed by reference.

Fixes #2196

I have:

Added new unit test that validates the different order of the tags will still return an empty patch result (test fails with current code).

### DIFF
--- a/pkg/clients/ec2/routetable.go
+++ b/pkg/clients/ec2/routetable.go
@@ -160,7 +160,7 @@ func CreateRTPatch(in ec2types.RouteTable, target v1beta1.RouteTableParameters) 
 	targetCopy := target.DeepCopy()
 	currentParams := &v1beta1.RouteTableParameters{}
 
-	CompareTags(target.Tags, in.Tags)
+	SortTagsV1Beta1(targetCopy.Tags, in.Tags)
 
 	if !pointer.BoolValue(target.IgnoreRoutes) {
 		// Add the default route for fair comparison.
--- a/pkg/clients/ec2/routetable_test.go
+++ b/pkg/clients/ec2/routetable_test.go
@@ -12,11 +12,15 @@ import (
 )
 
 var (
-	rtVPC      = "some vpc"
-	otherRtVPC = "some other vpc"
-	rtID       = "some RT Id"
-	rtSubnetID = "some subnet"
-	rtOwner    = "some owner"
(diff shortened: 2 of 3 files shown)


anagarlau:
Please adjust your commit message(s) to match [git conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

fernandoalexandre:
Thank you for the feedback @anagarlau! commit message has been adapted

anagarlau:
LGTM. Thank you very much @fernandoalexandre

PR #2207, opened 2025-04-11
Remove default instance options that cause problems when using launch…

### LINKED ISSUE #1193: ec2-instance: imageId and launchtemplateId required even if launchtemplateName set
I am trying to spawn an Instance using a LaunchTemplate which has `imageId` set. I have set `launchTemplate.launchTemplateName` to reference said template. I expect that I only need to set `imageId` on the Instance CRD if I want to override the value  from the launchTemplate. Furthermore, based on the [documentation](https://doc.crds.dev/github.com/crossplane/provider-aws/ec2.aws.crossplane.io/Instance/v1alpha1@v0.24.1#spec-forProvider-launchTemplate), I expect that I must *not* also set `launchTemplate.launchTemplateId` since I have already set `launchTemplateName`

### What happened?

When trying to create the Instance Resource, I get two validation Errors, telling me that both `Instance.spec.forProvider.imageId` and `Instance.spec.forProvider.launchTemplate.launchTemplateId` are required fields

### How can we reproduce it?

The crd has the fields set as required ([here](https://github.com/crossplane/provider-aws/blob/4651608a57229248d92821af0eb03db9b80c7e6e/package/crds/ec2.aws.crossplane.io_instances.yaml#L500), and [here](https://github.com/crossplane/provider-aws/blob/4651608a57229248d92821af0eb03db9b80c7e6e/package/crds/ec2.aws.crossplane.io_instances.yaml#L948))

Crossplane version: 
- aws-provider v0.24.1

rayshoo:
Fixes #1193 
1. When using launch template, the ImageId set in launchTemplate is ignored when setting the imageId required by the default instance resource. Therefore, the imageId needs to be removed from crd.
2. **crossplane can only use one of launchTemplateId and launchTemplateName,** but both are required in crd. Therefore, we need to remove the required of these two in crd to use it.

I have:

# Example
```
---
kind: SecurityGroup
metadata:
  name: sample-cluster-sg
spec:
  deletionPolicy: Delete
  forProvider:
    region: ap-northeast-2
    vpcId: <change to Your VPC-ID>
    groupName: test-crossplane-security-group
    description: test-crossplane-security-group
    ingress:
    - ipProtocol: '-1'
      ipRanges:
      - cidrIp: 10.55.0.0/16
    egress:
    - ipProtocol: '-1'
      ipRanges:
      - cidrIp: 0.0.0.0/0
    tags:
    - key: managed-by
      value: crossplane
  providerConfigRef:
    name: default
---
kind: LaunchTemplate
metadata:
  name: test-crossplane-obj
spec:
  deletionPolicy: Delete
  forProvider:
    launchTemplateName: test-crossplane-obj
    launchTemplateData:
      imageID: ami-0d5bb3742db8fc264
      instanceType: m7i-flex.large
      # echo 'Hello crossplane! > hello.txt'
      userData: SGVsbG8gY3Jvc3NwbGFuZSEgPiBoZWxsby50eHQK
    region: ap-northeast-2
    tags:
    - key: managed-by
      value: crossplane
  providerConfigRef:
    name: default
---
kind: Instance
metadata:
  name: sample-instance
spec:
  deletionPolicy: Delete
  forProvider:
    region: ap-northeast-2
    securityGroupRefs:
    - name: sample-cluster-sg
    launchTemplate:
      launchTemplateName: test-crossplane-obj
    subnetId: subnet-02ccc64296fccbad7
  providerConfigRef:
    name: default
```

### DIFF
--- a/package/crds/ec2.aws.crossplane.io_instances.yaml
+++ b/package/crds/ec2.aws.crossplane.io_instances.yaml
@@ -604,9 +604,6 @@ spec:
 
                           Default: The default version for the launch template.
                         type: string
-                    required:
-                    - launchTemplateId
-                    - launchTemplateName
                     type: object
                   licenseSpecifications:
                     description: The Amazon Resource Name (ARN) of the license configuration
@@ -1221,7 +1218,6 @@ spec:
                     pattern: ^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$
                     type: string
                 required:
-                - imageId
                 - region
                 type: object
               managementPolicies:


package/crds/ec2.aws.crossplane.io_instances.yaml
```diff
@@ -604,9 +604,6 @@ spec:
 
                           Default: The default version for the launch template.
                         type: string
-                    required:
```
anagarlau, line 607: This approach does not work since CRDs are generated based on the API definitions in the `apis`-folder using sigs.k8s.io/controller-tools. 
Try re-running `make generate` and you'll notice that your changes will be overwritten. 
 
[In order to achieve the desired result, i.e. mark the field as optional, you would need to add the // +optional marker and the omitempty tag to the appropriate fields. Using the `omitempty`-tag ensures that these fields are also omitted from serialization if unset.](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required#)
 
For example, `launchTemplateId` and launchTemplateName in `apis/ec2/manualv1alpha1/common.go`, 638ff:

```
// LaunchTemplateSpecification defines the launch template to use.
// You must specify either the launch template ID or launch template
// name in the request, but not both.
type LaunchTemplateSpecification struct {
	// The ID of the launch template.
        // +optional
	LaunchTemplateID *string `json:"launchTemplateId,omitempty"`

	// The name of the launch template.
        // +optional
	LaunchTemplateName *string `json:"launchTemplateName,omitempty"`

	// The version number of the launch template.
	//
	// Default: The default version for the launch template.
	// +optional
	Version *string `json:"version,omitempty"`
}

```

Please consult https://book.kubebuilder.io/ for a short introduction to CRD generation.
More information on API Conventions can be found [here](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)

rayshoo: @anagarlau Thanks for the feedback! I'm using Windows and I'm having trouble reproducing the development environment, so I can't check the result of `make generate` :( I added optional to the variable that seems to manage the ImageId in item 1. If I'm wrong, please let me know. I'm not sure if region should also be optional. If someone needs it, please add it.

MisterMX (approved):
LGTM. Thank you very much @rayshoo!

Issue #2082, opened 2024-07-18
S3 bucket with loggingConfiguration never becomes READY


kubevalet:
### What happened?
Seems like this https://github.com/crossplane-contrib/provider-aws/issues/1382

```
{
  "atProvider": {
    "arn": "arn:aws:s3:::test-s3-log44-jl5lq-2x787"
  },
  "conditions": [
    {
      "lastTransitionTime": "2024-07-18T15:07:57Z",
      "reason": "Creating",
      "status": "False",
      "type": "Ready"
    },
    {
      "lastTransitionTime": "2024-07-18T15:07:57Z",
      "reason": "ReconcileSuccess",
      "status": "True",
      "type": "Synced"
    }
  ]
}
```

Makes this tech completely unusable in environments where security compliance is required e.g. SOC2

### How can we reproduce it?
Define a bucket with `loggingConfiguration` set ... here's my describe output (I already had to revert the change):
```
Spec:
  Deletion Policy:  Orphan
  For Provider:
    Location Constraint:  us-west-2
    Logging Configuration:
      Target Bucket:   log-bucket-dev-usw220240712151134571000000001
      Target Prefix:   test-s3-log44/
    Object Ownership:  BucketOwnerEnforced
    Payment Configuration:
      Payer:  BucketOwner
    Public Access Block Configuration:
      Block Public Acls:        true
      Block Public Policy:      true
      Ignore Public Acls:       true
      Restrict Public Buckets:  true
    Server Side Encryption Configuration:
      Rules:
        Apply Server Side Encryption By Default:
          Sse Algorithm:  AES256
    Versioning Configuration:
      Status:  Suspended
  Management Policies:
    *
  Provider Config Ref:
    Name:  provider-config-aws
```

### What environment did it happen in?
Crossplane version:  `1.15.1`
AWS provider `0.48.1`

kubevalet:
fwiw, tried to build this myself to fix the issue but can't get to build (on Ubuntu): 
```
build/run make -j4
==== building the cross container (this could take minutes the first time)
=== cross build image failed for build-bb8e9112/cross-amd64
18:12:55 [ .. ] docker build build-bb8e9112/cross-amd64
Dockerfile:63
--------------------
  62 |     # Node JS and chrome support
  63 | >>> RUN curl -fsSL https://deb.nodesource.com/setup_12.x | bash - && \
[...]

Issue #1760, opened 2023-05-23
Can not create S3 bucket faild beacuse ACL


Ehco1996:
### What happened?

Can not create s3 bucket in latest ver(0.40.0) with the yaml below

```
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  finalizers:
    - finalizer.managedresource.crossplane.io
  name: xxxx
spec:
  deletionPolicy: Delete
  forProvider:
    locationConstraint: us-west-2
    paymentConfiguration:
      payer: BucketOwner
    serverSideEncryptionConfiguration:
      rules:
        - applyServerSideEncryptionByDefault:
            sseAlgorithm: AES256
  providerConfigRef:
    name: xxx
```

### How can we reproduce it?

just create a bucket with that yaml and you will see this error

```
  conditions:
  - lastTransitionTime: "2023-05-23T02:39:38Z"
    reason: Creating
    status: "False"
    type: Ready
  - lastTransitionTime: "2023-05-23T02:39:52Z"
    message: 'observe failed: operation error S3: PutBucketAcl, https response error
      StatusCode: 400, RequestID: TZHYY9GBCV36BPRC, HostID: BkrtUQaMwKXFsViv+IN6m5aMgOZaUYyFkculK6Fauztb5SuRge05UUrecUl2CFapUMMc6FCkJ+I=,
      api error MissingSecurityHeader: Your request was missing a required header'
    reason: ReconcileError
    status: "False"
    type: Synced
```

### What environment did it happen in?
Crossplane version:  0.40.0

chlunde:
try

```patch
spec:
  forProvider:
+    objectOwnership: BucketOwnerEnforced
```

Ehco1996:
> ```diff
> BucketOwnerEnforced
> ```

it works  thanks

kelvinwijaya:
Hi @Ehco1996 .

Wanted to check if you managed to get the Bucket in available state, As for myself, even after changing the spec to  `objectOwnership: BucketOwnerEnforced`
The reconcile is ok however the resource Ready status is always false or not available 

  conditions:
  - lastTransitionTime: "2023-05-23T02:39:38Z"
    reason: Creating
    status: "False"
    type: Ready

Ehco1996:
> Hi @Ehco1996 .
> 
[...]

Issue #1723, opened 2023-04-12
Unable to create S3 buckets after the new S3 defaults


jwitrick:
### What happened?

Recently AWS updated their default AWS S3 bucket settings and since then I am unable to create S3 buckets due to ACL errors.

### How can we reproduce it?

Not sure if the settings are global or not. but in us-east-2 when i try to create an s3 bucket I get an error about ACL errors (Note: same code works in us-west-2).

Here is the example code i was using:
```
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: crosspln-buckettest-failed-acl
  annotations:
    # crossplane.io/external-name: crosspln-buckettest-failed-acl
spec:
  forProvider:
    acl: private
    locationConstraint: us-east-2
    publicAccessBlockConfiguration:
      blockPublicPolicy: false
      blockPublicAcls: true
      ignorePublicAcls: false
      restrictPublicBuckets: false
    objectLockEnabledForBucket: false
    serverSideEncryptionConfiguration:
      rules:
      - applyServerSideEncryptionByDefault:
          sseAlgorithm: AES256
    corsConfiguration:
      corsRules:
      - allowedMethods:
        - "GET"
        allowedOrigins:
        - "*"
        allowedHeaders:
        - "*"
```

The error seems to be when trying to put an acl on bucket. Here is a small bit from the AWS CloudTrail event:

```
"errorCode": "AccessControlListNotSupported",
"errorMessage": "The bucket does not allow ACLs",
```

### What environment did it happen in?
* Crossplane version: 1.11.1
* Provider-aws version: 0.38.0
* Kube version: 1.24
* Kube distro: AWS EKS

### Email from AWS:
We are reaching out to inform you that starting in April 2023 Amazon S3 will change the default security configuration for all new S3 buckets. For new buckets created after this date, S3 Block Public Access will be enabled, and S3 access control lists (ACLs) will be disabled.

[...]

Issue #1382, opened 2022-07-08
S3 Bucket reconciliation issues when using loggingConfiguration settings


pauloeliasjr1:
### What happened?
I am trying to set up a S3 bucket using the default Bucket resource on provider-aws, but I'm getting reconciliation issues while trying to set it up with a loggingConfiguration (to write access logs inside the same bucket). 

It sorts out that when you "enable" logging configuration for a bucket, [AWS automatically writes a bucket policy](https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html#grant-log-delivery-permissions-bucket-policy) to allow access from AWS logging service principal. This changes expected configuration from Crossplane's perspective and leads to several API calls on PutBucketPolicy and PutBucketLogging events on AWS due reconciliation. 

The resource itself stays stuck:

```
NAME                                  READY   SYNCED   AGE
bucket-loggingconfiguration-testing   False   True     11m
```
We also tried to write a custom bucket policy allowing this permission but to no avail. Other methods such as using a bucket ACL are not recommended by AWS and probably will lead to same situation.

Do you guys have any lead on that? This looks like a fully supported field on docs.

### How can we reproduce it?

Use this sample to deploy a new Bucket. It is the same sample available [here](https://github.com/crossplane-contrib/provider-aws/blob/master/examples/s3/bucket.yaml), but including `loggingConfiguration` parameters.

```yaml
apiVersion: s3.aws.crossplane.io/v1beta1
kind: Bucket
metadata:
  name: bucket-loggingconfiguration-testing
  annotations:
    # This will be the actual bucket name. It must be globally unique, so you
    # probably want to change it before trying to apply this example.
    crossplane.io/external-name: bucket-loggingconfiguration-testing
spec:
  deletionPolicy: Delete
  forProvider:
    acl: private
    locationConstraint: us-east-1
    paymentConfiguration:
      payer: BucketOwner
    loggingConfiguration:
      targetPrefix: bucket-loggingconfiguration-testing-
      targetBucket: bucket-loggingconfiguration-testing
    serverSideEncryptionConfiguration:
      rules:
        - applyServerSideEncryptionByDefault:
            sseAlgorithm: AES256
    versioningConfiguration:
      status: Enabled
  providerConfigRef:
    name: aws-provider
```

### What environment did it happen in?
* Crossplane version: v1.8.1
* aws-provider: provider-aws-controller:v0.27.1

PR #2214, opened 2025-04-23
fix(s3): bucket loggingConfig and objectOwnership

teeverr:

This pr includes fixes for:
  - observe function for loggingConfig permanently returns `NeedsUpdate` in case if `spec.forProvider.loggingConfiguration.targetGrants` is not specified. It happens because aws returns an empty array if target grants are not set, but the controller prepares(by func `GenerateAWSLogging`) configuration for comparison with nil. At the same time cmp.Equal in function `Observer` doesn't have option `cmpopts.EquateEmpty()` and detects diff every time. I fixed it by adding `EquateEmpty` and also a bit refactored(just in case) `GenerateAWSLogging`, now it returns an empty slice if `targetGrants` is not set.
  - the controller calls aws api to set ACLs if `spec.forProvider.objectOwnership` doesn't set explicitly to `BucketOwnerEnforced`, and it leads to errors if it is not set. But it is not a required parameter and by default it is `BucketOwnerEnforced` on aws side. So in this case aws returns an error and the managed resource becomes not ready and synced. Idealy backet acl handling should be refactored, but for now I fixed func `BucketHasACLsDisabled`.

Fixes: #2082 #1760 #1382 #1723

I have:


MisterMX (approved):

PR #2221, opened 2025-05-22
fix(rds): Only ignore weekday when comparing maintenanceWindow

### LINKED ISSUE #2009: Time parse issue in compareTimeRanges (maintenanceWindow)
Hi,

i guess i have discovered a bug in the comparison logic of the compareTimeRanges code for checking if the maintenanceWindow needs to be updated.
https://github.com/crossplane-contrib/provider-aws/blob/4b536aa882f6e4b4914c556bbfc4854ea5461a74/pkg/controller/rds/dbinstance/setup.go#L590-L616

### What happened?

The current implementation to check the timeRanges rely on the idea that the weekday in the string ("mon") is parsed into a time object. 
But according to the docs this is not the case
https://pkg.go.dev/time#Parse
> The day of the week is checked for syntax but it is otherwise ignored.

So its only compared to the hour and minute and not weekday. 

This results in a wrong comparison result to update the maintenanceWindow and thus could cause a skipped update if only the weekday of a maintenanceWindow would be changed.

### How can we reproduce it?

The following snippet will show the issue with a simplified function of the compareTimeRanges code.

https://go.dev/play/p/n2jRovBJBj5

Bug in the code

siddharth0801:

Fixes #2009: Time parse issue in compareTimeRanges (maintenanceWindow)

I have:

The following snippet will show the fix with a simplified function of the compareTimeRanges code.

https://go.dev/play/p/gNMLBbTgbtE 

### DIFF
--- a/pkg/controller/rds/dbinstance/setup.go
+++ b/pkg/controller/rds/dbinstance/setup.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/controller/rds/dbinstance/setup.go
```diff
@@ -656,16 +656,34 @@ func compareTimeRanges(format string, expectedWindow *string, actualWindow *stri
...
+		rightParts := strings.SplitN(rightSpans[i], ":", 2)
+
+		if len(leftParts) != 2 || len(rightParts) != 2 {
+			return false, errors.New("invalid time format, expected 'day:HH:MM'")
+		}
+
+		leftDay := strings.ToLower(strings.TrimSpace(leftParts[0]))
+		rightDay := strings.ToLower(strings.TrimSpace(rightParts[0]))
+
+		leftTime, err := time.Parse(format, leftSpans[i])
```
MisterMX, line 676: This way we implement the parsing functionality twice, one for each value. Can we move this into an extra code block (can be in this package) with a parser function and on for equality check?

Might also make sense to do the parsing with a regex which is a bit more stable?

MisterMX (changes requested):

pkg/controller/rds/dbinstance/setup.go
siddharth0801: Done. Moved parsing to a shared function and switched to regex-based parsing.

MisterMX (approved):
LGTM. Thank you very much @siddharth0801!

PR #2222, opened 2025-05-22
fix(secretsmanager): fix isUpToDate for tags only

### LINKED ISSUE #2205: Secretmanager does not update tags
### What happened?
Secretmanager seems to not trigger an update, when only the spec.forProvider.tags parameter is updated. 
Only through changing another field, we can trigger an update, which then also changes a tag in AWS.

### How can we reproduce it?
Create Secret with secretmanager group and only change tags value afterwards.

Crossplane version: 1.18.0
provider-aws: v0.50.6

anagarlau:

Fixes #2205

I have:

- manual tests with this branch's version 


MisterMX (approved):
LGTM. Thank you very much @anagarlau!

PR #2235, opened 2025-06-17
fix(ecs): issue with empty network configuration parameter

saiharsha-plivo:

When using host network mode users do not need to pass the network config , but the codes passes and empty object thus having issues creating a service with no network configuration , So when user is not passing any network configuration I removed if it is an empty object

Fixes #
This fixes https://github.com/crossplane-contrib/provider-aws/issues/2233

Added in the new examples I used to test the code , service are getting created without the network configuration now 

### DIFF
--- a/pkg/controller/ecs/service/setup.go
+++ b/pkg/controller/ecs/service/setup.go
@@ -537,7 +537,7 @@ func generateNetworkConfiguration(cr *svcapitypes.Service) *svcsdk.NetworkConfig
 	networkConfiguration := &svcsdk.NetworkConfiguration{}
 
 	if cr.Spec.ForProvider.NetworkConfiguration == nil {
-		return networkConfiguration
+		return nil
 	}
 
 	if cr.Spec.ForProvider.NetworkConfiguration.AWSvpcConfiguration != nil {
--- a/examples/ecs/service.yaml
+++ b/examples/ecs/service.yaml
@@ -37,3 +37,41 @@ spec:
       name: example
   providerConfigRef:
     name: example
+---
+apiVersion: ecs.aws.crossplane.io/v1alpha1
+kind: Service
+metadata:
+  name: nginx-test-ecs
+  annotations:
+    task-definition-network-mode: awsvpc
+spec:
+  forProvider:
+    cluster: arn:aws:ecs:us-east-1:123456789:cluster/crossplane-ecs-controller-test
+    launchType: EC2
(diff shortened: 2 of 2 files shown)


saiharsha-plivo:
@MisterMX , @anagarlau Can you review this once

pkg/controller/ecs/service/setup.go
```diff
@@ -533,11 +536,11 @@ func generateLoadBalancers(cr *svcapitypes.Service) []*svcsdk.LoadBalancer {
 
 // Helper func to generate SDK NetworkConfiguration types from the crossplane
 // types included in the Service api.
-func generateNetworkConfiguration(cr *svcapitypes.Service) *svcsdk.NetworkConfiguration {
+func generateNetworkConfiguration(cr *svcapitypes.Service) (*svcsdk.NetworkConfiguration, error) {
 	networkConfiguration := &svcsdk.NetworkConfiguration{}
 
 	if cr.Spec.ForProvider.NetworkConfiguration == nil {
-		return networkConfiguration
+		return nil, errors.New("network configuration is nil")
```
MisterMX, line 543: We should not misuse an error as an indicator here if we just want to not set a value.

Instead just return nil in this case so `obj.NetworkConfiguration` is just nil.

MisterMX (changes requested):

pkg/controller/ecs/service/setup.go
```diff
@@ -230,7 +230,10 @@ func preCreate(_ context.Context, cr *svcapitypes.Service, obj *svcsdk.CreateSer
 	obj.ClientToken = aws.String(string(cr.UID))
 	obj.Cluster = cr.Spec.ForProvider.Cluster
 	obj.LoadBalancers = generateLoadBalancers(cr)
-	obj.NetworkConfiguration = generateNetworkConfiguration(cr)
+	networkConfiguration := generateNetworkConfiguration(cr)
+	if networkConfiguration != nil {
+		obj.NetworkConfiguration = networkConfiguration
+	}
```
MisterMX, line 236: I believe this can stay as it is, right? Or is there another place where the value should be set?

```suggestion
	obj.NetworkConfiguration = generateNetworkConfiguration(cr)
```

MisterMX (changes requested):

MisterMX (approved):
LGTM. Thank you very much @saiharsha-plivo!

PR #2243, opened 2025-07-18
feat(userpool): add custom attributes when adding schema

kkendzia:
* fix(userpool): upToDate for schema was always false with at least one entry

Motivation:
Currently it is not possible to add custom attributes to userpools after initial creation. This PR changes the behaviour to enable users to add custom attributes later on by AddCustomAttributes API Endpoint. This also fixes the upToDate validation for SchemaAttributesTypes as this was not working properly.

For further discussion:
I don't think we have to check every field of SchemaAttributesTypes as customAttributes are immutable once created (also cannot be deleted). `mutable: true` means that the AppClient is capable of modifing this value. So a check for "Name" should be sufficient enough and we have to give further notice in documentation (and crd description) that the attributes are immutable.

Fixes #

I have:

### DIFF
--- a/pkg/controller/cognitoidentityprovider/userpool/setup.go
+++ b/pkg/controller/cognitoidentityprovider/userpool/setup.go
@@ -163,7 +163,7 @@ func (e *hooks) isUpToDate(_ context.Context, cr *svcapitypes.UserPool, resp *sv
 		!areEmailConfigurationEqual(spec.EmailConfiguration, pool.EmailConfiguration),
 		!areLambdaConfigEqual(spec.LambdaConfig, pool.LambdaConfig),
 		!arePoliciesEqual(spec.Policies, pool.Policies),
-		!e.areSchemaEqual(spec.Schema, pool.SchemaAttributes),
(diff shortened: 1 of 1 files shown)


pkg/controller/cognitoidentityprovider/userpool/setup.go
```diff
@@ -313,28 +316,43 @@ func arePoliciesEqual(spec *svcapitypes.UserPoolPolicyType, current *svcsdk.User
 	return true
 }
 
-func areSchemaEqual(spec []*svcapitypes.SchemaAttributeType, current []*svcsdk.SchemaAttributeType) bool {
+func (h *hooks) areSchemaEqual(spec []*svcapitypes.SchemaAttributeType, current []*svcsdk.SchemaAttributeType) bool { //nolint:gocyclo
```
ggkhrmv, line 319: Why do we need `(h *hooks)` here? No values of `hooks` are used by `areSchemaEqual`

kkendzia: Good catch. Removed it

PR #2250, opened 2025-07-24
feat(rds): add db-instance read replica

teeverr:
This PR adds params to create  rds `DBInstance` as a read replica.

I have:

This controller previously had zero test coverage. I've added tests for the new functionality. Existing features were tested manually for regressions.


kkendzia:
Thx for the Contribution! I'm unsure if we should follow this approach. `aws-sdk-go` supports a `DBInstanceReadReplica` type which we didn't include for now: https://github.com/crossplane-contrib/provider-aws/blob/master/apis/rds/generator-config.yaml#L99 - have you already tried it?

kkendzia:
Thanks for the explanation! I'm aware of ACK :) And yeah, I now generated the resources and, in fact, it lacks of independened methods as you stated. I just thought we don't have to reinvent the wheel and just use what the sdk provides us, but your solution fits way better.

kkendzia:
Thx! LGTM

PR #2276, opened 2026-01-24
feat(elasticache): auth token params and log delivery configuration

teeverr:
### Description

I added **one new parameter** and **two new `spec.forProvider` sections**:

- `spec.forProvider.authTokenSecretRef` — lets you provide your own auth token.
- `spec.forProvider.AuthTokenUpdateStrategy` — allows you to configure the auth token update strategy.
- `spec.forProvider.LogDeliveryConfiguration` — configures log delivery for **slow logs** and **engine logs**.

I have:

Unit tests, manually


pkg/controller/cache/replicationgroup/managed.go
```diff
@@ -381,3 +433,51 @@ func getVersion(version *string) (*string, error) {
 	}
 	return &versionOut, nil
 }
+
+// GetSecretValue retrieves the value of a secret key from a Kubernetes Secret.
+func getSecretPassword(ctx context.Context, kube client.Client, sks *xpv1.SecretKeySelector) (pw string, err error) {
+	secret := new(corev1.Secret)
+	secret.SetName(sks.Name) // TODO(teeverr): it is needed only for mock testing functions, will be overwritten by Get. didn't find a better way.
+	ref := sks.SecretReference
```
kkendzia: Maybe you can take a look at rds testing where they do something similar: https://github.com/crossplane-contrib/provider-aws/blob/fd3e357a4e3184e35d1870bf0b27df313f272cfb/pkg/clients/rds/common_test.go#L33 
I haven't taken a deeper look into it but other implementations don't need this workaround.

pkg/clients/elasticache/elasticache.go
```diff
@@ -158,6 +160,10 @@ func NewModifyReplicationGroupInput(g v1beta1.ReplicationGroupParameters, id str
 		SnapshotWindow:              g.SnapshotWindow,
 		SnapshottingClusterId:       g.SnapshottingClusterID,
 	}
+	if token != nil && *token != "" {
```
kkendzia: Please use ptr.Deref (pointer.StringValue) in those cases. We use it in other parts of the code, too. 

https://github.com/crossplane-contrib/provider-aws/blob/master/pkg/utils/pointer/pointer.go#L26

pkg/controller/cache/replicationgroup/managed.go
```diff
@@ -270,16 +317,21 @@ func (e *external) Update(ctx context.Context, mg resource.Managed) (managed.Ext
...
 		}
-		return managed.ExternalUpdate{}, nil
-
+		cr.Status.AtProvider.LastAppliedAuthTokenUpdateStrategy = cr.Spec.ForProvider.AuthTokenUpdateStrategy
 	}
-	err = e.updateTags(ctx, cr.Spec.ForProvider.Tags, rg.ARN)
-	return managed.ExternalUpdate{}, errorutils.Wrap(err, errUpdateReplicationGroupTags)
+	return managed.ExternalUpdate{ConnectionDetails: managed.ConnectionDetails{
+		xpv1.ResourceCredentialsSecretPasswordKey: []byte(e.cache.desiredAuthToken),
+	}}, nil
```
ridik-il, line 360: This rewrite of Update drops tag reconciliation. Previously the function fell through to e.updateTags(...) whenever ReplicationGroupNeedsUpdate returned no diff; now nothing calls updateTags — it has zero callers and is effectively dead code (the AddTagsToResource/RemoveTagsFromResource calls live only inside it, so they're unreachable from reconcile). The effect, for any ReplicationGroup that manages spec.forProvider.tags: Observe still computes tagsNeedUpdate independently (tags aren't part of rgDiff) and feeds it into ResourceUpToDate. So a tag change makes Observe report not-up-to-date → Update runs → the modify block is skipped (no rgDiff, no auth change) → returns without applying tags → next Observe reports not-up-to-date again. That's a permanent not-ready/requeue cycle with tags never converging. Resources that don't set tags are unaffected. Please restore the tag reconciliation in the success path — e.g. call e.updateTags(ctx, cr.Spec.ForProvider.Tags, rg.ARN) after the modify block — and add a TestUpdate case that asserts the tag API is exercised, since there's currently no coverage guarding this path.

pkg/controller/cache/replicationgroup/managed.go
```diff
@@ -177,13 +194,37 @@ func (e *external) Observe(ctx context.Context, mg resource.Managed) (managed.Ex
 	}
 
 	rgDiff := elasticache.ReplicationGroupNeedsUpdate(cr.Spec.ForProvider, rg, ccList)
+	// retrieve current pass from the connection secret
+	crPass, err := getCurrentPassword(ctx, e.kube, cr)
```
ridik-il: Observe calls getCurrentPassword(ctx, e.kube, cr) unconditionally, and that helper does SecretReference: *mg.GetWriteConnectionSecretToReference(). writeConnectionSecretToRef is optional (*SecretReference, returned raw by the generated getter), so it's nil for any ReplicationGroup that doesn't set it — and dereferencing it panics the reconcile loop. I confirmed this with a quick unit probe: invalid memory address or nil pointer dereference. Please guard against a nil ref, and ideally only look up the current/desired auth token when auth is actually in play (AuthEnabled or AuthTokenSecretRef set), so resources that use neither auth nor a connection secret are left untouched.

pkg/controller/cache/replicationgroup/managed.go
teeverr: Fixed

pkg/controller/cache/replicationgroup/managed.go
teeverr: Fixed

ridik-il (approved):
LGTM, thanks @teeverr

PR #2298, opened 2026-04-21
fix(docdb): omit AutoMinorVersionUpgrade from ModifyDBInstance requests

ldu112:

  The AWS DocumentDB API rejects the `AutoMinorVersionUpgrade` field in
  `ModifyDBInstance` calls. When this field was present in the resource
  spec, the controller would detect a mismatch in `isUpToDate`, trigger
  an update, include the field in the request, and AWS would reject it —
  putting the resource into a `Terminal` condition.

  Fixes #2274

  Changes:
  - **`isUpToDate`**: Remove `AutoMinorVersionUpgrade` from the comparison so changes to this field no longer trigger updates
  - **`lateInitialize`**: Remove `AutoMinorVersionUpgrade` since it cannot be updated via the API
  - **`preUpdate`**: Explicitly nil `AutoMinorVersionUpgrade` in `ModifyDBInstanceInput` so the auto-generated code cannot accidentally include it

  I have:

  Updated existing unit tests to reflect that `AutoMinorVersionUpgrade`
  differences are intentionally ignored (4 observe test cases renamed and
  `ResourceUpToDate` changed from `false` to `true`). Added a new
  `TestUpdate` case that asserts the field is absent from the
  `ModifyDBInstance` API call.

### DIFF
--- a/pkg/controller/docdb/dbinstance/setup.go
+++ b/pkg/controller/docdb/dbinstance/setup.go
@@ -129,8 +129,7 @@ func (e *hooks) isUpToDate(_ context.Context, cr *svcapitypes.DBInstance, resp *
 	instance := resp.DBInstances[0]
 
 	switch {
-	case pointer.BoolValue(cr.Spec.ForProvider.AutoMinorVersionUpgrade) != pointer.BoolValue(instance.AutoMinorVersionUpgrade),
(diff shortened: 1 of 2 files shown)


teeverr (approved):
LGTM, Thanks so much @ldu112!

PR #2299, opened 2026-05-07
ignore az in rds instance isUpToDate + expose it in status of managed…

### LINKED ISSUE #2300: Infinite drift loop on AvailabilityZone after Multi-AZ failover
### What happened?

After a Multi-AZ failover, the DBInstance controller enters an infinite reconciliation loop because isUpToDate
detects a diff on AvailabilityZone (spec holds the old AZ, AWS reports the new primary AZ). Since AZ is immutable
at the AWS API level and not part of ModifyDBInstanceInput, the resulting modify call is a no-op that briefly puts
the instance in modifying state, then the cycle repeats.

### How can we reproduce it?

Have `multiAz` set to true on a DBInstance, then change the DB Instance class or upgrade the engine version. You will see in the "Logs & Events" of the RDS Instance on AWS console events "Finished updating DB parameter group" caused by the continuous drift.

Image

Crossplane version: v1.20.4

stefnans:
RDS DBInstance resource

Fixes #2300

I have:

### DIFF
--- a/pkg/controller/rds/dbinstance/setup_test.go
+++ b/pkg/controller/rds/dbinstance/setup_test.go
@@ -526,6 +526,29 @@ func TestIsUpToDate(t *testing.T) {
 				err:      nil,
 			},
 		},
+		"AvailabilityZoneDiffIgnored": {
+			args: args{
+				cr: &svcapitypes.DBInstance{
+					Spec: svcapitypes.DBInstanceSpec{
+						ForProvider: svcapitypes.DBInstanceParameters{
+							AvailabilityZone: aws.String("eu-central-1b"),
+						},
+					},
+				},
+				out: &svcsdk.DescribeDBInstancesOutput{
+					DBInstances: []*svcsdk.DBInstance{
+						{
+							AvailabilityZone: aws.String("eu-central-1a"),
+						},
+					},
+				},
+				kube: test.NewMockClient(),
+			},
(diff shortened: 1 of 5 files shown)


pkg/controller/rds/dbinstance/setup_test.go
```diff
@@ -732,7 +778,10 @@ func TestPostObserve(t *testing.T) {
...
 			}
 			if diff := cmp.Diff(tc.want.statusAtProvider.DatabaseRole, cr.Status.AtProvider.DatabaseRole); diff != "" {
-				t.Errorf("r: -want, +got: \n%s", diff)
+				t.Errorf("DatabaseRole: -want, +got: \n%s", diff)
+			}
+			if diff := cmp.Diff(tc.want.statusAtProvider.AvailabilityZone, cr.Status.AtProvider.AvailabilityZone); diff != "" {
+				t.Errorf("AvailabilityZone: -want, +got: \n%s", diff)
 			}
 		})
 	}
```
teeverr, line 784: I propose changing `TestPostObserve` to use a generic `if` check for all values in `statusAtProvider`, instead of checking each field separately.

stefnans: Done thanks for the review!

teeverr (approved):
LGTM, thanks @stefnans !

PR #2301, opened 2026-05-11
fix(dbinstance): skip iops/storageThroughput diff in isUpToDate for g…

### LINKED ISSUE #2302: isUpToDate causes infinite reconciliation for gp3 volumes below allocatedStorage threshold
### What happened?

When a DBInstance uses storageType: gp3 with allocatedStorage below the engine-specific threshold (< 400GB for postgres/mysql/mariadb, < 200GB for oracle), setting custom iops or storageThroughput values causes an infinite reconciliation loop.

AWS rejects custom IOPS/storageThroughput for gp3 volumes below the threshold with the error: "You can't specify IOPS or storage throughput for engine postgres and a storage size less than 400". The provider already handles this in preCreate and preUpdate by stripping these fields via isStorageTypeGP3BelowAllocatedStorageThreshold (to avoid the AWS API error "You can't specify IOPS or storage throughput for engine postgres and a storage size less than 400"). However, isUpToDate still compares the desired values against the observed defaults, detects a diff every reconcile, and triggers an update — which is then a no-op because preUpdate strips the fields.

### How can we reproduce it?

1. Create a DBInstance with:
   - storageType: gp3
   - allocatedStorage: 20
   - engine: postgres
   - iops: 3200
   - storageThroughput: 150
2. Observe that the resource continuously triggers UpdatedExternalResource events
3. AWS always reports iops: 3000, storageThroughput: 125

Image

Crossplane version: 1.20.4

stefnans:
For gp3 volumes below the engine-specific allocatedStorage threshold (< 400GB for postgres/mysql/mariadb, < 200GB for oracle), AWS ignores custom iops/storageThroughput values and keeps defaults (3000/125).

The guard isStorageTypeGP3BelowAllocatedStorageThreshold already existed in preCreate and preUpdate to avoid AWS API errors, but was missing from isUpToDate, causing an infinite reconciliation loop.

Added a warning log when the user's desired values differ from observed but are being skipped due to the threshold, so the discrepancy is visible in provider logs.

Fixes #2302

I have:

Built a version of the provider and tested in the cluster with custom Iops and storageThroughput on a Postgres DB Instance with a low volume (allocatedStorage) that's below the threshold. No reconciliation drift.

### DIFF
--- a/pkg/controller/rds/dbinstance/setup.go
+++ b/pkg/controller/rds/dbinstance/setup.go
@@ -684,8 +684,20 @@ func (s *shared) isUpToDate(ctx context.Context, cr *svcapitypes.DBInstance, out
 	// Depending on whether the instance was created as gp2 or modified from another type (s.g. gp3) to gp2,
 	// AWS provides different responses for IOPS/StorageThroughput (either 0 or nil).
 	// Therefore, we consider both 0 and nil to be equivalent.
-	iopsChanged := !(pointer.Int64Value(cr.Spec.ForProvider.IOPS) == pointer.Int64Value(db.Iops))
-	storageThroughputChanged := !(pointer.Int64Value(cr.Spec.ForProvider.StorageThroughput) == pointer.Int64Value(db.StorageThroughput))
+	// For gp3 volumes below the engine-specific allocatedStorage threshold, AWS rejects custom
+	// iops/storageThroughput values. Return an error so the resource shows SYNCED=False.
+	iopsChanged := false
+	storageThroughputChanged := false
+	if isStorageTypeGP3BelowAllocatedStorageThreshold(cr) {
+		if pointer.Int64Value(cr.Spec.ForProvider.IOPS) != pointer.Int64Value(db.Iops) ||
+			pointer.Int64Value(cr.Spec.ForProvider.StorageThroughput) != pointer.Int64Value(db.StorageThroughput) {
+			return false, "", fmt.Errorf("cannot reconcile desired iops/storageThroughput: gp3 volumes below %dGB (engine: %s) use fixed defaults (3000 IOPS / 125 MB/s). Increase allocatedStorage to provision custom values",
+				gp3AllocatedStorageThreshold(cr), pointer.StringValue(cr.Spec.ForProvider.Engine))
+		}
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/controller/rds/dbinstance/setup.go
```diff
@@ -682,8 +682,21 @@ func (s *shared) isUpToDate(ctx context.Context, cr *svcapitypes.DBInstance, out
...
-	storageThroughputChanged := !(pointer.Int64Value(cr.Spec.ForProvider.StorageThroughput) == pointer.Int64Value(db.StorageThroughput))
+	// For gp3 volumes below the engine-specific allocatedStorage threshold, AWS ignores custom
+	// iops/storageThroughput values and keeps the defaults (3000/125). Skip the comparison to
+	// avoid an infinite reconciliation loop (the same guard already exists in preCreate/preUpdate).
+	iopsChanged := false
+	storageThroughputChanged := false
+	if isStorageTypeGP3BelowAllocatedStorageThreshold(cr) {
+		if pointer.Int64Value(cr.Spec.ForProvider.IOPS) != pointer.Int64Value(db.Iops) ||
+			pointer.Int64Value(cr.Spec.ForProvider.StorageThroughput) != pointer.Int64Value(db.StorageThroughput) {
+			log.Printf("WARNING: desired iops/storageThroughput ignored by AWS — gp3 volumes below %dGB (engine: %s) use fixed defaults (3000 IOPS / 125 MB/s). Increase allocatedStorage to provision custom values.",
```
teeverr: This "WARNING" is effectively a fake log level and should be removed.

The message will only be visible in debug mode anyway, and the issue will not be reflected at the MR resource level.

My suggestion is to call `ReconcileError()` when this condition is met. Yes, the AWS API silently ignores "wrong" parameters, so from the API perspective this is not an error (let's leave judging this design decision out of scope). However, from the Crossplane perspective, the controller is unable to reconcile the desired state. The most transparent way to communicate that is to set `SYNCED=False `with an appropriate error message — which is exactly what `ReconcileError()` in `crossplane-runtime` does.

teeverr:
@stefnans Please rebase your branch with the latest master.

pkg/controller/rds/dbinstance/setup.go
stefnans: Fixed

Screenshot 2026-05-28 124407

teeverr (approved):
@stefnans LGTM, thanks so much!

PR #2311, opened 2026-06-01
fix(rds/dbinstance): fix port and licenseModel handling in ModifyDBInstance

teeverr:
### Port
   The controller detected port changes as a diff but never sent the new value in the `ModifyDBInstance` request. Fixed by setting `DBPortNumber` from `spec.forProvider.port` in `preUpdate`.

   ### LicenseModel
   AWS rejects `ModifyDBInstance` requests on read replicas when `licenseModel` is included, even if unchanged. After `LateInitialize()` populates this field from observed state, it gets included in every
   subsequent modify call, causing all replica modifications to fail with:

   InvalidParameterValue: You can't change the license model on a Read Replica

   Fixed by excluding `licenseModel` from modify requests for read replicas (`SourceDBInstanceID != nil`).

I have:

Added `TestPreUpdate` covering both fixes.

### DIFF
--- a/pkg/controller/rds/dbinstance/setup.go
+++ b/pkg/controller/rds/dbinstance/setup.go
@@ -320,7 +320,7 @@ func (s *shared) updateConnectionDetails(ctx context.Context, cr *svcapitypes.DB
 	return details, nil
 }
 
-func (s *shared) preUpdate(ctx context.Context, cr *svcapitypes.DBInstance, obj *svcsdk.ModifyDBInstanceInput) (err error) {
+func (s *shared) preUpdate(ctx context.Context, cr *svcapitypes.DBInstance, obj *svcsdk.ModifyDBInstanceInput) (err error) { //nolint:gocyclo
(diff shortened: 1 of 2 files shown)


pkg/controller/rds/dbinstance/setup.go
```diff
@@ -362,6 +362,15 @@ func (s *shared) preUpdate(ctx context.Context, cr *svcapitypes.DBInstance, obj
 		obj.BackupRetentionPeriod = nil
 	}
 
+	obj.DBPortNumber = pointer.ToOrNilIfZeroValue(*cr.Spec.ForProvider.Port)
```
ridik-il: port is optional (*int64, omitempty), so this dereferences a nil pointer and panics when it's unset. Verified locally — calling preUpdate with Port == nil segfaults here. Both sides are *int64, so a direct assignment is nil-safe and also drops the redundant deref/re-pointer:
obj.DBPortNumber = cr.Spec.ForProvider.Port

teeverr: Fixed.

ridik-il (approved):
LGTM

PR #2321, opened 2026-07-23
fix(rds): sort VPCSecurityGroupIDs to prevent order-sensitive diffs

ChrisEllisSains:

This change affects the RDSInstance CR (database.aws.crossplane.io/v1beta1).

When AWS returns multiple VpcSecurityGroups from DescribeDBInstances, their order is arbitrary. The JSON patch comparison in CreatePatch is order-sensitive, causing false positives and re-reconciliation loops when the same security groups were returned in a different order.

This fix sorts VPCSecurityGroupIDs on both the current state (from AWS) and the desired spec before the patch comparison, making it order-insensitive. This prevents unnecessary ModifyDBInstance calls when 2 or more security groups are configured.

- Added test case VPCSecurityGroupIDsSameElementsDifferentOrder in TestIsUpToDate that verifies the same security group IDs in different order are correctly identified as equivalent
- This test would have failed before the fix and now passes
- Ran go test ./pkg/clients/database/... — all tests pass

I have:

- [x] Run make reviewable test to ensure this PR is ready for review.

### DIFF
--- a/pkg/clients/database/rds.go
+++ b/pkg/clients/database/rds.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ChrisEllisSains:
Hey @kkendzia @ggkhrmv when you get a chance would you mind to review this please?

We recently had an incident where an RDS instance was in `storage-full` state but crossplane couldn't update `allocatedStorage` as it was also trying to update `vpcSecurityGroupIds` at the same time so we were seeing:

```
Warning  CannotUpdateExternalResource  54m (x187 over 3h26m)   managed/rdsinstance.database.aws.crossplane.io  cannot modify RDS instance: InvalidDB
InstanceState: The specified database instance is currently in storage-full state. Please allocate more storage by modifying the DB instance.
```

The only fix was to manually adjust the storage through the console.  I believe this PR should fix this inconsistency.

PR #2323, opened 2026-07-27
fix(kms/key): skip key rotation calls for non-AWS_KMS origin keys

### LINKED ISSUE #2310: When a KMS Key resource has origin: AWS_CLOUDHSM or origin: EXTERNAL, the reconciler fails permanently
### What happened?
When a KMS Key resource has origin: AWS_CLOUDHSM or origin: EXTERNAL, the reconciler fails permanently with:

UnsupportedOperationException: <key-arn> origin is AWS_CLOUDHSM which is not valid for this operation
This happens because isUpToDate calls GetKeyRotationStatus unconditionally on every reconcile loop, and update calls EnableKeyRotation/DisableKeyRotation unconditionally. AWS does not support key rotation APIs for CloudHSM custom key store keys or externally-imported keys — the operation is only valid for AWS_KMS origin keys.

Fix

Guard both call sites with an origin check:

isUpToDate: skip GetKeyRotationStatus unless obj.KeyMetadata.Origin == "AWS_KMS" (origin is available directly from the DescribeKeyOutput already in scope)
update: skip EnableKeyRotation/DisableKeyRotation unless cr.Spec.ForProvider.Origin == "AWS_KMS" or is unset (origin is late-initialized from KeyMetadata.Origin after creation, so it is reliably populated by the time update runs; an empty value preserves existing AWS_KMS behaviour)

### How can we reproduce it?
Configure CloudHSM with custom KMS key

Crossplane version: 
Helm Charts v2.1.3

pujitha24:
Motivation:
KMS Key resources with origin AWS_CLOUDHSM or EXTERNAL fail to
reconcile permanently. AWS does not support the key-rotation APIs
(EnableKeyRotation, DisableKeyRotation, GetKeyRotationStatus) for
CloudHSM custom key store keys or externally-imported keys - only
AWS_KMS origin keys support rotation. The controller called these
APIs unconditionally on every reconcile, so Observe() (via isUpToDate)
returned UnsupportedOperationException every loop and the managed
resource never reached the Ready condition. This is a per-resource
reconcile failure, not a crash of the controller process, and it does
not affect AWS_KMS origin keys, which are unaffected before and after
this change.

Approach:
- isUpToDate: only call GetKeyRotationStatus when the live
  DescribeKey response reports Origin == AWS_KMS.
- update: only call EnableKeyRotation/DisableKeyRotation when
  spec.forProvider.origin is AWS_KMS or unset (unset preserves the
  previous AWS_KMS-only behaviour).
- lateInitialize: backfill spec.forProvider.origin from the live
  DescribeKey response, same as the existing Enabled field, so the
  update() guard is populated for Key resources imported into
  Crossplane rather than created through this controller (origin was
  previously only set on the Create() response path).

Validation:
go build ./pkg/controller/kms/...
go test ./pkg/controller/kms/... -v
All tests pass, including new table-driven tests in setup_test.go
that assert (via gomock strict expectations) that
GetKeyRotationStatus/EnableKeyRotationWithContext/
DisableKeyRotationWithContext are not called for AWS_CLOUDHSM and
EXTERNAL origin keys, are still called for AWS_KMS and unset origin
(regression check), and that lateInitialize backfills origin from the
DescribeKey response.

Fixes #2310

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/controller/kms/key/setup.go
+++ b/pkg/controller/kms/key/setup.go
@@ -132,19 +132,27 @@ func (u *updater) update(ctx context.Context, cr *svcapitypes.Key) (managed.Exte
 		return managed.ExternalUpdate{}, errorutils.Wrap(err, errUpdate)
 	}
 
-	if pointer.BoolValue(cr.Spec.ForProvider.EnableKeyRotation) {
-		// EnableKeyRotation
-		if _, err := u.client.EnableKeyRotationWithContext(ctx, &svcsdk.EnableKeyRotationInput{
-			KeyId: pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr)),
-		}); err != nil {
-			return managed.ExternalUpdate{}, errorutils.Wrap(err, errUpdate)
-		}
-	} else {
-		// DisableKeyRotation
-		if _, err := u.client.DisableKeyRotationWithContext(ctx, &svcsdk.DisableKeyRotationInput{
-			KeyId: pointer.ToOrNilIfZeroValue(meta.GetExternalName(cr)),
-		}); err != nil {
-			return managed.ExternalUpdate{}, errorutils.Wrap(err, errUpdate)
+	// Key rotation is only supported for AWS_KMS origin keys. AWS rejects
+	// EnableKeyRotation/DisableKeyRotation for AWS_CLOUDHSM and EXTERNAL
+	// origin keys, so skip the call for those. Origin is late-initialized
+	// from KeyMetadata.Origin after creation, so it is reliably populated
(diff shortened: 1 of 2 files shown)


pujitha24:
This is rebased, DCO-passing, and has no open change requests from a maintainer — happy to adjust anything that would make review easier.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #2326, opened 2026-08-10
fix(mq/broker): implement isUpToDate so spec changes are actually applied

### LINKED ISSUE #2093: [broker.mq] Update of RabbitMQ broker not possible
### What happened?
We want to update our AmazonMQ RabbitMQ brokers and have changed the engine version and the maintenance window accordingly but the provider-aws reconciler is still responding that the "External resource is up to date"

### How can we reproduce it?
- Setup a broker.mq with engineType RabbitMQ and an older version (e.g. 3.10.25)
- Wait till its running
- Update the engineVersion to `3.11.16` and maintenance window (to something in the next minutes)
- Check  the provider-aws logs and see "External resource is up to date"
- Check the AWS console for the changes of engineVersion and maintenance window and see nothing has changed

Crossplane version: 
xpkg.upbound.io/crossplane/crossplane:v1.15.2
xpkg.upbound.io/crossplane-contrib/provider-aws:v0.47.2

pujitha24:

The `Broker` (`broker.mq`) controller never overrode the generated `isUpToDate`
hook, so it always used the ACK-generated `alwaysUpToDate` stub
(`pkg/controller/mq/broker/zz_controller.go`), which unconditionally reports
`true`. This meant `Update` was effectively unreachable: changing
`engineVersion`, `hostInstanceType`, `maintenanceWindowStartTime`,
`authenticationStrategy`, or `logs` on the spec was silently ignored, and the
reconciler always logged "External resource is up to date" even though AWS
still had the old configuration.

This adds a real `isUpToDate` in `pkg/controller/mq/broker/setup.go`, wired up
the same way sibling ACK-generated controllers in this repo do it (e.g.
`mwaa/environment`, `glue/trigger`). It compares the spec against the observed
`DescribeBroker` response for exactly the fields that `GenerateUpdateBrokerRequest`
(`zz_conversions.go`) actually sends on Update: `AuthenticationStrategy`,
`AutoMinorVersionUpgrade`, `EngineVersion`, `HostInstanceType`, `Logs`, and
`MaintenanceWindowStartTime`.

Fields were deliberately excluded to avoid introducing a new, worse bug (a
permanent, unfixable "not up to date" diff / no-op update loop):
- `Configuration` and `LDAPServerMetadata` are not symmetrically returned by
  `DescribeBroker` into `Spec.ForProvider` (or are masked, e.g. the LDAP bind
  password), so comparing them would always show a diff.
- `DataReplicationMode` is mutable but not late-initialized, so an unset spec
  value could permanently diff against a non-nil AWS value that `Update` would
  never actually send.
- For `Logs` and `MaintenanceWindowStartTime`, each sub-field comparison is
  individually gated on the corresponding spec sub-field being non-nil,
  matching `GenerateUpdateBrokerRequest`'s own per-sub-field `if != nil`
  behavior, so partially-specified structs don't cause a permanent diff either.

Fixes #

I have:

Ran, from the repo root:

- `go build ./...` — passes.
- `go vet ./pkg/controller/mq/broker/...` — only a pre-existing warning in the
  generated `zz_conversions.go` (unkeyed `metav1.Time` struct literal),
  unrelated to this change and present before it too.
- `go test ./pkg/controller/mq/... ./pkg/clients/mq/...` — passes.

Added a new table-driven `TestIsUpToDate` in
`pkg/controller/mq/broker/setup_test.go` covering: no change, `engineVersion`
changed, `hostInstanceType` changed, `maintenanceWindowStartTime` changed, and
a partially-specified `maintenanceWindowStartTime` (only `dayOfWeek` set in
spec) correctly reporting up-to-date against an AWS response that also has
`timeOfDay`/`timeZone` set. Before this change `isUpToDate` didn't exist as a
standalone function (the generated stub was used directly), so these cases
could not even be expressed as a unit test — this failing-to-compile-then-passing
progression is the reproduction for this defect class (an always-true stub).

I did not have access to a live AWS account/broker to reproduce the original
end-to-end symptom described in the issue (updating a running RabbitMQ broker
and observing the reconciler log). The fix is scoped strictly to the fields
`GenerateUpdateBrokerRequest` actually sends on `Update`, so the mechanism by
which the bug occurred (missing `isUpToDate` override defaulting to
`alwaysUpToDate`) is directly addressed and verifiable by reading
`zz_controller.go`'s `newExternal()`/`alwaysUpToDate`.

Fixes #2093

### DIFF
--- a/pkg/controller/mq/broker/setup.go
+++ b/pkg/controller/mq/broker/setup.go
@@ -207,32 +207,53 @@ func isUpToDate(_ context.Context, cr *svcapitypes.Broker, obj *svcsdk.DescribeB
 	// GenerateUpdateBrokerRequest, an unset sub-field is never sent to AWS,
 	// so comparing it against the observed value could flag a permanent,
 	// unfixable diff.
-	var currentAudit, currentGeneral *bool
-	if obj.Logs != nil {
-		currentAudit, currentGeneral = obj.Logs.Audit, obj.Logs.General
+	if !logsUpToDate(p.Logs, obj.Logs) {
+		return false, "logs is not up to date", nil
 	}
-	if want := p.Logs; want != nil {
-		if (want.Audit != nil && pointer.BoolValue(want.Audit) != pointer.BoolValue(currentAudit)) ||
-			(want.General != nil && pointer.BoolValue(want.General) != pointer.BoolValue(currentGeneral)) {
-			return false, "logs is not up to date", nil
-		}
-	}
-
-	var currentDayOfWeek, currentTimeOfDay, currentTimeZone *string
-	if obj.MaintenanceWindowStartTime != nil {
-		currentDayOfWeek = obj.MaintenanceWindowStartTime.DayOfWeek
-		currentTimeOfDay = obj.MaintenanceWindowStartTime.TimeOfDay
-		currentTimeZone = obj.MaintenanceWindowStartTime.TimeZone
-	}
-	if want := p.MaintenanceWindowStartTime; want != nil {
-		if (want.DayOfWeek != nil && pointer.StringValue(want.DayOfWeek) != pointer.StringValue(currentDayOfWeek)) ||
-			(want.TimeOfDay != nil && pointer.StringValue(want.TimeOfDay) != pointer.StringValue(currentTimeOfDay)) ||
-			(want.TimeZone != nil && pointer.StringValue(want.TimeZone) != pointer.StringValue(currentTimeZone)) {
-			return false, "maintenanceWindowStartTime is not up to date", nil
-		}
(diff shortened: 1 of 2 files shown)


pujitha24:
Just checking in on this one — it's rebased and green, happy to make any changes that would help review go faster.

pujitha24:
The lint check was failing on gocyclo (complexity 19 in `isUpToDate`, limit 10). Split the Logs and MaintenanceWindowStartTime comparisons out into `logsUpToDate` and `maintenanceWindowUpToDate` helpers — same behavior, just under the threshold now. Tests and build pass locally.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

Issue #1108, opened 2022-01-31
Improve experience of configuring cloudfront distributions


samidbb:
### What happened?
Configuring cloudfront distribution seems to be a bit buggy. Some fields that are under `defaultCacheBehavior` are not required but once I introduced `cacheBehaviors `section then suddenly almost all fields become required and specifying `quantity` does not help on the experience.

Another issue is that the resource never goes into Ready state:
```
NAME                         READY   SYNCED   EXTERNAL-NAME
somedomain                   False   True     E1U92C17PI1M42
```
It seems like the provider has entered an infinite loop with reconciling the updates to the resource. This is maybe related to this issue https://github.com/crossplane/provider-aws/issues/894

### How can we reproduce it?
Here is the distribution that I was testing:

```
apiVersion: cloudfront.aws.crossplane.io/v1alpha1
kind: Distribution
metadata:
  name: somedomain
spec:
  providerConfigRef:
    name: my-namespace-aws
  forProvider:
    region: us-east-1
    distributionConfig:
      enabled: true
      comment: Example CloudFront Distribution 3
      priceClass: PriceClass_100
      origins:
        items:
          - domainName: app.my.cloud
            id: customOrigin
            customOriginConfig:
              httpPort: 80
              originProtocolPolicy: "https-only"
              originReadTimeout: 10
              originKeepaliveTimeout: 5
              httpSPort: 443
              originSSLProtocols:
                items:
                  - TLSv1.2
                quantity: 1
      cacheBehaviors:
        quantity: 1
        items:
        - pathPattern: /*/services-email-validation-api/*          
          smoothStreaming: false # Why is it required??
          targetOriginID: customOrigin
          compress: false
          viewerProtocolPolicy: https-only
          allowedMethods:        
            cachedMethods: # Why is it required??
              items:
              - HEAD
              - GET
              quantity: 2
            items:
            - GET 
            - HEAD 
            - OPTIONS
            - PUT
            - POST
            - PATCH
            - DELETE
            quantity: 7
          minTTL: 0
          maxTTL: 0
          defaultTTL: 0
          fieldLevelEncryptionID: "" # Why is this required ?
          forwardedValues:
            queryString: true 
            queryStringCacheKeys:  # Why is this required ?
              quantity: 0          # Why is this required ?              
            cookies:
              forward: all
            headers:
              items:
              - Accept
              - Accept-Charset
              - Accept-Encoding
              - Accept-Language
              - Authorization
              - Accept-Datetime
[...]

Issue #980, opened 2021-12-01
RDSInstance keep getting configuring-enhanced-monitoring state


kelvinwijaya:
### What happened?
 The resource are having more syncing than usual compared to the provider-aws version that i am using previously at 0.18. it keep getting configuring-enhanced-monitoring state to be updated again and again.

shown from my log:
```
test-mysqlinstance-knqk2-tsvkv        True    True     available                         mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    True     available                         mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    True     available                         mysql      5.7.33    14d
test-mysqlinstance-knqk2-tsvkv        True    False    available                         mysql      5.7.33    14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7.33    14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
test-postgresqlinstance-t9s4z-p4dd9   True    False    configuring-enhanced-monitoring   postgres   12.7      14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7       14d
test-mysqlinstance-knqk2-tsvkv        True    False    configuring-enhanced-monitoring   mysql      5.7.33    14d
test-mysqlinstance-knqk2-tsvkv        True    True     modifying                         mysql      5.7.33    14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
test-postgresqlinstance-t9s4z-p4dd9   True    True     available                         postgres   12.7      14d
```

event logs:
```
[...]

PR #2329, opened 2026-08-20
SNS Topic: Stop False Reconcile Loop on Plain Topics

enageshwari:
### Desc:

`GetTopicAttributes` omits keys for attributes that are not configured (e.g. no KMS key, no display name). Accessing a missing key in a Go map returns `""`, and the previous code passed `aws.String("")` — a non-nil pointer — to `pointer.LateInitialize`. This wrote an empty-string pointer into the spec on the first observe, marking `ResourceLateInitialized: true` every subsequent cycle and triggering a redundant spec patch on every reconcile indefinitely.

This is the same pattern behind #1108 and #980: absent optional attributes being treated as present defaults, creating infinite reconcile loops.

**What changed:**
- `LateInitializeTopicAttr` now guards each attribute behind a non-empty check — only real AWS-returned values are written into the spec.
- `lateInitStringAttr`: extracted helper for the repeated non-empty-check + nil-guard pattern; reduces cyclomatic complexity from 12 to 4.
- Returns `bool` (`modified`) for a precise late-init signal, removing the need for a full `reflect.DeepEqual` on the spec struct in the caller.
- `Observe` in `controller.go` uses the return value directly and drops the `reflect` import.
- One existing test expectation updated: `ResourceLateInitialized` is now correctly `false` when AWS returns no meaningful attributes.

Related: #1108, #980

I have:

New table-driven tests in `pkg/clients/sns/topic_test.go` (`TestLateInitializeTopicAttr`, 10 cases):

| Scenario | Validates |
|---|---|
| All attributes absent (empty map) | no write, `modified=false` — core fix |
| KMS key ARN present | copied correctly, `modified=true` |
| KMS key already set in spec | not overwritten, `modified=false` |
| DisplayName / DeliveryPolicy / Policy | same pattern for each |
| `FifoTopic: true` | copied correctly |
| `FifoTopic: false` or absent | no write (not a meaningful default) |
| FifoTopic already set | not overwritten |
| All fields populated in one pass | composite correctness |

```
ok  github.com/crossplane-contrib/provider-aws/pkg/clients/sns
ok  github.com/crossplane-contrib/provider-aws/pkg/controller/sns/topic
```

### DIFF
--- a/pkg/clients/sns/topic.go
+++ b/pkg/clients/sns/topic.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/utils/ptr"
 
 	"github.com/crossplane-contrib/provider-aws/apis/sns/v1beta1"
-	"github.com/crossplane-contrib/provider-aws/pkg/utils/pointer"
 	policyutils "github.com/crossplane-contrib/provider-aws/pkg/utils/policy"
 )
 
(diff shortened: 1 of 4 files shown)


enageshwari:
Hi, 

Please check below doc for further details about these reconciliation improvements and assist with approval workflow run. Thanks.

[reconciliation_improvements_provider_aws.pdf](https://github.com/user-attachments/files/31337174/reconciliation_improvements_provider_aws.pdf)

enageshwari:
Hi, 

Approval workflow passed with the improvements made in fork's master. See below, review and approve these changes to trigger the same run against upstream. Thanks. 

https://github.com/enageshwari/provider-aws/actions/runs/32434923189

enageshwari:
Hi @teeverr / @ridik-il / @kkendzia, 

Could you please help reviewing this PR when possible? Pls check the details about the changes and approval run info below. Thanks. 

**[PR #2329](https://github.com/crossplane-contrib/provider-aws/pull/2329) – SNS Topic LateInitialization Loop Fix**

- Problem: Empty attribute maps passed empty pointers ("") to LateInitialize, triggering infinite spec patches on every cycle (related to #1108, #980).
- Fix: Added non-empty check guards for AWS attributes; reduced cyclomatic complexity.

**Ref**

1. Project details: https://github.com/user-attachments/files/31337185/reconciliation_improvements_provider_aws.pdf
2. Approval run in fork: https://github.com/enageshwari/provider-aws/actions/runs/32434923189

cc: @christophrj @AndresNico @teeverr @ridik-il

PR #2331, opened 2026-09-01
fix(ec2/vpcendpoint): apply spec.forProvider.tags on create

### LINKED ISSUE #2330: VPCEndpoint - Tags are not configured
### What happened?
When settings up a VPCEndpoint it looks like the tags are not set.

### How can we reproduce it?
- Create a VPCEndpoint like this:

```
kind: VPCEndpoint
metadata:
  name: vpcendpoint
spec:
  deletionPolicy: Orphan
  forProvider:
    privateDNSEnabled: false
    region: eu-central-1
    securityGroupIds:
    - <some-securityGroupIds>
    serviceName: <some-serviceName>
    subnetIds:
    - <some-subnetIds>
    tags:
      Name: my-name
      Region: eu-central-1
      Stack: kubernetes
      cross-region: "true"
      service-region: eu-central-1
    vpcEndpointType: Interface
    vpcId: <some-vpcId>
  managementPolicies:
  - '*'
  providerConfigRef:
    name: aws-provider-config
```

see that neither Name nor tags are set

Crossplane version: 2.4.0
Provider-AWS version: 0.59.0

pujitha24:

Fixes #2330

`VPCEndpoint`'s `preCreate` hook never copied `spec.forProvider.tags`
(`map[string]string`) into the `TagSpecifications` field of the
`CreateVpcEndpoint` API request, so tags set on a `VPCEndpoint` resource
were silently dropped and never applied to the underlying AWS VPC
Endpoint.

This adds a `generateTagSpecifications` helper that converts
`spec.forProvider.tags` into a `vpc-endpoint` `TagSpecification` and
appends it (rather than overwrites) onto whatever `TagSpecifications`
was already populated from the separate, pre-existing
`spec.forProvider.tagSpecifications` field, so both fields are honored
if a user happens to set both.

Note: this only fixes tag application at creation time. This
controller has no tag-update handling at all today (no `isUpToDate`
tag comparison, no `preUpdate` tag diffing), so changing
`spec.forProvider.tags` after creation still won't be reconciled. That
is a separate, larger change and is left as a follow-up.

I have:

Ran `go build ./pkg/controller/ec2/vpcendpoint/...` and
`go test ./pkg/controller/ec2/vpcendpoint/... -v`; the full package
suite passes, including a new `TestPreCreate` with cases for no tags,
tags only, and tags alongside a pre-existing `TagSpecifications` entry
(verifying the merge/append behavior). I confirmed
`TestPreCreate/WithTags` fails against the pre-fix code and passes
against the post-fix code, i.e. it is a genuine regression test for
the reported bug. `gofmt -l` reports no issues on the changed files.

I could not run this repo's pinned `golangci-lint v1.63.4` locally
(only v2.12.2 is available in this environment, and its config format
is incompatible with the repo's v1-style `.golangci.yml`), nor
`make check-diff` / `make generate` — the change does not touch any
generated API types, so these should be a no-op, but I did not
execute them directly. `master`'s own CI (the `CI` workflow) is
currently green.

### DIFF
--- a/pkg/controller/ec2/vpcendpoint/setup.go
+++ b/pkg/controller/ec2/vpcendpoint/setup.go
@@ -2,6 +2,7 @@ package vpcendpoint
 
 import (
 	"context"
+	"sort"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


pujitha24:
This is rebased and all checks are green — let me know if there'\''s anything I can do to make review easier.

